Issue #6, opened 2016-02-05
rename OpenRDFException


jeenbroekstra:
Possible candidates are `RDFException` or `RDF4JException`. I personally prefer the first since it's shorter. OpenRDFException should remain as a deprecated class for backward compatibility.

ansell:
RDFException is probably simpler in the long term.

nikolavp:
I think that RDF4JException is better because it indicates that the this is an exception from the framework.

RDFException looks like an error in the RDF format when in fact it can be thrown from other places that are not RDF related.

ansell:
@nikolavp, they are good reasons to use RDF4JException. I am switching to preferring that.

jeenbroekstra:
Duly noted, and I agree with the both of you. Thanks for the feedback!

PR #34, opened 2016-03-29
OpenRDFException now (deprecated) superclass of RDF4JException

jeenbroekstra:
#6 Usage of OpenRDFException replaced throughout the code. 
Given that every concrete exception is a subtype of both RDF4JException _and_ OpenRDFException, old code can be easily migrated (minus the package rename, obviously).


jeenbroekstra:
Thanks for the comments @ansell . I didn't know Github would automatically pick those up when I opened the pull request. As you can see I switch the inheritance around, as per your feedback. I can't do anything about the package naming, but that's a general problem for any user who migrates.

ansell:
Even with the package rename it should make it slightly simpler to migrate. With this deprecation OpenRDFException can be removed in RDF4J-2.0 after the client base migrations are complete.

PR #47, opened 2016-05-04
Maven config for publishing snapshot/release build to OSSRH and Central

jeenbroekstra:


pom.xml
```diff
@@ -63,6 +57,67 @@
 
 	<profiles>
 		<profile>
+			<id>ossrh</id>
+			<distributionManagement>
```
ansell: This may also need the release repository information added.

ansell: An example from one of my other projects is:

	<distributionManagement>
		<snapshotRepository>
			<id>sonatype-nexus-snapshots</id>
			<name>Sonatype Nexus Snapshots</name>
			<url>https://oss.sonatype.org/content/repositories/snapshots/</url>
		</snapshotRepository>
		<repository>
			<id>sonatype-nexus-staging</id>
			<name>Nexus Release Repository</name>
			<url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
		</repository>
	</distributionManagement>

jeenbroekstra: @ansell are you sure that's necessary? The way I read the Sonatype guide you only need this if you do not use the nexus staging plugin.

ansell: If you want to use the basic mvn clean deploy without the release/nexus plugins you will need it, but I have never used the nexus plugin, so it may define all of that internally and you may be fine. :)

jeenbroekstra: Cool. Thanks for the tip anyway. We'll try it like this and see what happens. I kinda need to use the staging plugin because we are configuring this to be deployed and staged/promoted from Hudson, rather than from the command line.

Issue #13, opened 2016-02-11
Decide on version number for initial release


jeenbroekstra:
We need to make a decision on what version number to use for the initial RDF4J release. There are, roughly, three options:

1. continue where Sesame left off: Sesame 4.1.0 -> RDF4J 4.1.0. 
2. jump a minor version after the last Sesame release: Sesame 4.1.0 -> RDF4J 4.2.0.
3. reset. RDF4J 1.0.0.

Advantage of the first option is that it's a more 'gradual' transition. Potential downside is that it suggests it's not the first Eclipse RDF4J release. 

Advantage of the second option is that it is more clear that there may be compatibility problems between the last Sesame release and the first RDF4J release.

Advantage of the last option is that we get to start fresh. Downside is that it's not obvious how this release relates to existing Sesame releases.

No matter what we choose, we will always need to provide accompanying upgrade notes anyway.

ansell:
I would suggest going with 1.0 or 5.0 due to the package renames. I slightly prefer 1.0, as the Sesame-4 transition to RDF4J-1.0 will just be package names, but it is clear that they are not drop in substitutes.

ansell:
Would a compatibility package be useful if it could be created efficiently? Ie, a set of Sesame classes/interfaces that all override RDF4J classes/interfaces with the old package names, and simply deprecate themselves with constructors alone?

jeenbroekstra:
On 12/02/16 12:56, Peter Ansell wrote:
> Would a compatibility package be useful if it could be created efficiently? Ie,
> a set of Sesame classes/interfaces that all override RDF4J classes/interfaces
> with the old package names, and simply deprecate themselves with constructors alone?

That would be a quite useful, but I'm not sure how to go about creating this. 
You know of any tools we can use to generate such a package?

I can imagine it being doable by some clever scripting, but it sounds like quite 
a lot of work to get this done.

An alternative is that we offer a conversion script of some sort, that inspects 
user code and quickly replaces old package names with new ones. This is fairly 
trivial to do.

ansell:
In the distant past I remember doing it with the Eclipse IDE, but I can't seem to replicate that functionality with Eclipse Mars. The move/copy functions don't seem to do it and I am not sure what other tools I would have used.

A script may be useful and preferable, but compatibility packages would smooth over the transition period so that it wasn't a single switchover date/version for everyone.

jeenbroekstra:
I've created a new issue #14 for discussing transitional packages further.

[...]

PR #48, opened 2016-05-06
 #13 set version number for Java 8 release branch to 2.0M1-SNAPSHOT

jeenbroekstra:
The Java 8 release of RDF4J will be numbered 2.0. When we are ready to release a Java 7 backport we'll use 1.0 for that release branch. 

Unless someone has a better idea.


ansell:
That versioning works well IMO.

Issue #38, opened 2016-04-15
Support parallel Java 7-compatible release


jeenbroekstra:
We are currently focusing on the Sesame 4 code base as the launch point for RDF4J. However there are several core users who require to stay at Java 7 for a while longer. We should consider bringing over the Sesame 2.9 code base to RDF4J to live alongside the main branch, so we can do parallel releases for those users who wish to stick to Java 7.

jeenbroekstra:
I have sent off a question to Eclipse legal, to see if there is a way to fast-track bringing over the 2.9 codebase (given that they've already approved basically the same codebase, just a different version).

jeenbroekstra:
Sharon (Eclipse IP team) has let me know that this shouldn't be a problem from their POV, and that we can bring in the 2.9 codebase as a parallel branch when we're ready - they'll be able to fast-track through the approval process. 

So all we need to do is stabilize, bring out the release, then freeze development for a little while so we don't have a fast-moving target.

jeenbroekstra:
We are currently considering an alternative approach: instead of bringing over the 2.9 code base as a new "initial" contribution, we can also just branch off a parallel branch from the current master, and downgrade that branch to be Java 7-compatible again.

PR #137, opened 2016-05-18
Backport to Java 7 runtime compatibility

jeenbroekstra:
This PR addresses GitHub issue: #38 

- maven configuration enforces 1.7 compilation target and source/bytecode compliance
- all source modified to be 1.7 compatible (optionals and lambdas/streams removed) 

- [x] tests are included
- [x] all tests succeed


jeenbroekstra:
@pulquero I'd like your eyes on this PR, especially.

compliance/rio/src/test/java/org/eclipse/rdf4j/rio/helpers/RDFParserHelperTest.java
```diff
@@ -108,7 +108,7 @@ public final void testCreateLiteralLabelOnly()
 				valueFactory);
 
 		assertEquals(LABEL_TESTA, literal.getLabel());
-		assertFalse(literal.getLanguage().isPresent());
+		assertFalse(literal.getLanguage() != null);
```
ansell: If this were a long term branch (does it have an end of support date yet?), this pattern may be better suited to use assertNull, as the assertFalse(... != null) pattern isn't intuitive to read. However, it is functionally the same so really just a style issue. 

Not sure what Git will think about this change if/when you merge 1.0 into 2.0. Ie, will it change the code on 2.0 to use != null which will always be true for Optional's or will it conflict. May be simpler not to merge 1.0 to 2.0 and just recreate patches as necessary.

jeenbroekstra: I did it this way because that was the simplest change I could make, and one I could easily copy-paste to multiple places. You're right that it's not really intuitive, but I figure it's "good enough".

There's no fixed date (yet) on end-of-support for Java 7, but I treat this as "as long as there's a need and we can do it without it severely impacting our primary development". 

As for merging: 1.0.x is intended to be a parallel _backport_ of the 2.0 branch. This means we will _never_ merge 1.0 into 2.0, only the other way around (or rather, we merge the master branch into the 1.0 branch). Fixes or feature development branches should also never start by branching off from the 1.0.x branch, but always either from master or from 2.0.x. I plan to document this in a bit more detail in the contribution guidelines.

ansell: Just backporting will mean the asserts don't have to be read by humans as often so sounds good to me.

pulquero:
Doesn't compile for me:

> 
[INFO] --- maven-enforcer-plugin:1.3.1:enforce (enforce-bytecode-version) @ rdf4j-sail-solr ---
[INFO] Restricted to JDK 1.7 yet jdk.tools:jdk.tools:jar:1.6:system contains com/sun/codemodel/internal/ClassType.class targeted to JDK 1.8

solr-core-5.1.0 depends on hadoop-anotations-2.6.0 depends on jdktools-1.6

Will have to run with profile jdk1.7 or only enforce 1.7 and below.

jeenbroekstra:
@pulquero odd, doesn't happen for me - compilation goes off without a hitch here, and the enforcer plugin doesn't complain about jdk.tools (despite me actually being on JDK 8). 

How are you running compilation, exactly? 

> Will have to run with profile jdk1.7 or only enforce 1.7 and below.

Not sure what you mean with that.

pulquero:
Maven within eclipse running on jdk8.

Dunno if enforcer distinguishes between 1.7 and <=1.7.

jeenbroekstra:
I thought I had it configured not to use the enforcer plugin from m2e (I've put the enforcer plugin as an exception in the eclipse maven lifecycle plugin). 

In any case: I cannot reproduce the problem, either building in Eclipse, executing embedded maven from Eclipse or compiling from the command line. 

What the enforcer plugin does is set an upper limit on the bytecode of our code and of dependencies, to ensure it runs on jdk 1.7. So it's <= jdk 1.7.  

Can you try if tweaking the configuration to exclude dependencies with system scope helps? In the project root pom, find line 729:

`<maxJdkVersion>1.7</maxJdkVersion>`

underneath it, add the following:

`<ignoredScopes>system</ignoredScopes>`

See http://www.mojohaus.org/extra-enforcer-rules/enforceBytecodeVersion.html for more details on the enforcer rule we use. 

If that doesn't help: can you try resetting your local environment (cleaning out your local maven repo, etc), to see if that fixes things?

jeenbroekstra:
Actually, scratch that. I just had a look at the config we used for Sesame 2.9 and it had this:

```
								   <excludes>
								     <!-- system-provided jar that we have no influence over -->
								     <exclude>jdk.tools:jdk.tools:jar:1.6</exclude>
								   </excludes>
```
I'll update the PR to include this.

jeenbroekstra:
Done. Give it another go and let me know.

pulquero:
Built fine now, but I think the Model interface still has a default method.

pulquero:
FYI: this is still in the target branch:
https://github.com/eclipse/rdf4j/blob/releases/1.0.x/core/sail/rdbms/pom.xml

pulquero:
FYI: old package name in https://github.com/eclipse/rdf4j/blob/releases/1.0.x/core/http/server/src/main/webapp/WEB-INF/classes/logback.xml

jeenbroekstra:
Yeah, the rdbms stuff keeps popping up again (I think it comes from having to sync the old Sesame bitbucket stuff a couple more times). And thanks for finding the old package names in the log config.

Both problems need to be addressed on the 2.0 branch first, then merged to master, then finally to the 1.0 branch.

pulquero:
Just ported and built sesametools and ripple without problems.

Issue #29, opened 2016-03-16
FederationSail fails on SPARQL 1.1 query test subquery/sq14


jeenbroekstra:
The test evaluates the handling of a limit on a subselect in a larger CONSTRUCT query. Failure is specific to FederationSail as the same test succeeds on other store types.

jeenbroekstra:
@aschwarte10 is this something you could perhaps take a look at?

aschwarte10:
@jeenbroekstra: yes, I will try to look into this by Friday

jeenbroekstra:
There's no simple way to _only_ execute that specific test I'm afraid; this is due to how manifest-based test suites are generated (runtime). Either just run the complete `FederationSparqlTest` suite (with sq14 re-enabled), or create a local copy of that specific test case so you can quickly execute it without needing to run the entire suite.

aschwarte10:
@jeenbroekstra: I analysed the issue and also understood the cause.

The issue is that the optimizer evaluates the inner subquery (i.e. the SLICE) as the right argument of the join (i.e., in a nested loop fashion) while it needs to be evaluated at once - ideally first.

See below for the relevant excerpt of the query plan:

```
NaryJoin
                     StatementPattern
                        Var (name=person)
                        Var (name=_const_aba78b99_uri, value=http://xmlns.com/foaf/0.1/homepage, anonymous)
                        Var (name=homepage)
                     Slice ( limit=3 )
                        Projection
                           ProjectionElemList
                              ProjectionElem "person"
                              ProjectionElem "name"
                           Order
                              OrderElem (ASC)
                                 Var (name=name)
                              OwnedTupleExpr org.eclipse.rdf4j.sail.memory.MemoryStoreConnection@1e6bd367
                                 NaryJoin
                                    StatementPattern
                                       Var (name=person)
                                       Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
                                       Var (name=_const_e1df31e0_uri, value=http://xmlns.com/foaf/0.1/Person, anonymous)
                                    StatementPattern
                                       Var (name=person)
                                       Var (name=_const_23b7c3b6_uri, value=http://xmlns.com/foaf/0.1/name, anonymous)
                                       Var (name=name)
```

[...]

PR #147, opened 2016-05-20

aschwarte10:

as intermediate result.

The full fix (which will be done in another change) could be to use a hash
based join strategy.


aschwarte10:
Have to redo the submission - used the wrong email address as of the commit.

PR #148, opened 2016-05-20

aschwarte10:

as intermediate result.

The full fix (which will be done in another change) could be to use a hash
based join strategy.


jeenbroekstra:
I think you may have opened a PR against the wrong branch (master, rather than the 2.0 release branch).

jeenbroekstra:
The annoying thing about GitHub PRs is that you can't change the target of an existing PR. The only thing you can do to fix it is to close this PR, and then create a new PR with the correct target branch.

jeenbroekstra:
I see you created this branch off of master as well, instead of off the release branch. It's not really a big deal at this point because we're still in milestone development stage, but I am trying to be all biblical about good workflow habits :) 

In any case, this gives me an opportunity to try out how cherry-picking works (which I've always avoid before) so I'll just create a new branch and copy over your fix.  Then this PR can be closed and I'll create a new PR for the new branch.

jeenbroekstra:
New PR on correct branch (after cherry-picking your fix) now available: #149 .

PR #149, opened 2016-05-22

jeenbroekstra:
This PR addresses GitHub issue: #29. It replaces PR #148 which was accidentally created against the wrong branch. 

as intermediate result.

The full fix (which will be done in another change) could to use a hash
based join strategy.


jeenbroekstra:
@aschwarte10 if you're happy with this as-is, go ahead and merge. If anything is unclear about the tweaks to branches I did, let me know.

aschwarte10:
I think it is always good to evaluate subsets as early as possible, so from my side this change should be good. @pulquero had some comments about whether this is general enough.

See also the related pull request https://github.com/eclipse/rdf4j/pull/159 where a hash join is applied instead. We should decide whether we want both changes in or only this one, or only the other.

PR #159, opened 2016-05-23
#29 fix suqbquery evaluation in FederationSail (use hash join)

aschwarte10:

In certain cases the nested loop join approach is not applicable (e.g.,
if the right argument is a sub-select with ORDER BY or LIMIT, see sq14).

This change now applies a HashJoin strategy in such cases.

Note that the HashJoinIteration has to be extended such that the left
intermediate result iteration can be passed to it.

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/HashJoinIteration.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/HashJoinIteration.java
@@ -87,6 +87,20 @@ public class HashJoinIteration extends LookAheadIteration<BindingSet, QueryEvalu
 
 		this.leftJoin = leftJoin;
 	}
+	
+	public HashJoinIteration(EvaluationStrategy strategy, CloseableIteration<BindingSet, QueryEvaluationException> leftIter, 
+			Set<String> leftBindingNames, TupleExpr right, BindingSet bindings, boolean leftJoin)
+		throws QueryEvaluationException
+	{
+		this.leftIter = leftIter;
+		rightIter = strategy.evaluate(right, bindings);
+
+		Set<String> joinAttributeNames = leftBindingNames;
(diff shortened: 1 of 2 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/HashJoinIteration.java
```diff
@@ -87,6 +87,20 @@ public HashJoinIteration(EvaluationStrategy strategy, TupleExpr left, TupleExpr
 
 		this.leftJoin = leftJoin;
 	}
+
```
pulquero: It might be a good idea if the constructor above delegates to this new constructor with a this(...,strategy.evaluate(left, bindings),...).  You could also make it take CloseableIteration, Set<String> for both left and right to maintain a bit of consistency with the param types.

jeenbroekstra:
Looks good. @aschwarte10 have you ran the test suite, and does it greenline?

jeenbroekstra:
Does this supercede the previous PR? I notice you haven't merged that one in yet, just wondering what the status is.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/HashJoinIteration.java
aschwarte10: Ok, I will do this directly as another change on this branch.

aschwarte10:
@jeenbroekstra: yes, I did run all Federation related tests and they green line.

I wrote this change in a way that there are no conflicts with the previous PR. Note that this is the reason why the activation of sq14 is not part of this change.

From my point of view both changes can be merged, although the first one can be considered an improvement to the optimizer, while this one is the real fix.

jeenbroekstra:
Ok thanks Andreas, approved for my part, go ahead and merge (I already merged your other PR).

PR #162, opened 2016-05-25
Uses IRI class in all vocabularies

Tpt:
Uses IRI class in all vocabularies


jeenbroekstra:
Looks sensible, thanks. However, two things:

1. Can your refer an issue in our issue tracker that describes what you're fixing? 
2. This might be on the wrong branch - it looks like a bugfix (or code maintenance if you will) that should be part of the 2.0 release branch, rather than master (which is our feature development branch for the _next_ major/minor release).

Issue #171, opened 2016-05-27
Use IRI class in all vocabularies


Tpt:
The vocabularies APF, GEOF, GEO and LIST do not use IRI but the deprecated URI class.

As IRI is a subclass of URI this change should keep compatibility.

PR #172, opened 2016-05-27
Uses IRI class in all vocabularies

Tpt:
This PR addresses GitHub issue: #171 

It uses IRI instead of URI in all classes in the org.eclipse.rdf4j.model.vocabulary package


jeenbroekstra:
Looks good to me. I'll run quick smoke test and if nothing breaks, I'll merge. Thanks!

Issue #164, opened 2016-05-25
SPARQL engine/repository configuration to run in 'strict' or 'extended' mode


jeenbroekstra:
The SPARQL specification is extensible in nature. In particular, the behavior of boolean operators on datatyped literals is extensible: while the default specification has no support for all but a limited subset of XSD datatypes, implementors are free to extend behavior to include boolean comparison support - at least for those cases where in the strict minimal interpretation the comparison results in error.

There are several useful extensions that we can add, but some of these may trip up existing users that rely on the 'strict' behavior of the engine. To facilitate this, we should introduce a configuration option that sets the SPARQL engine to run in either 'strict' mode or in 'extended' mode.

jeenbroekstra:
I'm thinking about the design of this. We could do this on a per-query basis, or make it a simple extension to the Repository API (a basic setter method and in addition make it part of the config template so it can be configured when creating a repo from the Workbench/Console). Thoughts?

pulquero:
I think it should probably be a bit more generic than this else you could end up with a config zoo. To my mind this is all about being able to configure evaluation strategy being used. You should be able to set the class name/instance of the evaluation strategy supplier you want to use at the sail level. In workbench it is then a similar type option as async threshold or persist. The repo api seems the wrong place, how would it work for something like a sparqlrepo?

pulquero:
I've sketched out an impl @ https://github.com/pulquero/rdf4j/tree/issues/%23164-config-evalstrat

jeenbroekstra:
Basic mechanism now merged in. I'm keeping this open to track progress on actually making an extended evaluation strategy available as a config choice. Will do some work on that today.

PR #204, opened 2016-06-28
Initial design for pluggable SPARQL evaluation modes (strict/extended/...) 

pulquero:

- Support for pluggable evaluation modes in workbench.

- [] tests are included - tested in workbench
- [x] all tests succeed


jeenbroekstra:
Um, this was planned for milestone 2.1, so wrong branch I think.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailConfig.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.util.ModelException;
+import org.eclipse.rdf4j.model.util.Models;
+import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategyFactory;
+import org.eclipse.rdf4j.sail.config.AbstractSailImplConfig;
+import org.eclipse.rdf4j.sail.config.SailConfigException;
+
+import static org.eclipse.rdf4j.sail.base.config.BaseSailSchema.*;
+
+public abstract class BaseSailConfig extends AbstractSailImplConfig {
```
jeenbroekstra: Why a subclass, instead of adding the necessary properties to AbstractSailImplConfig directly?

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -135,10 +141,17 @@
 	 * @param resolver
 	 */
 	protected SailSourceConnection(AbstractSail sail, SailStore store, FederatedServiceResolver resolver) {
+		this(sail, store, new SimpleEvaluationStrategyFactory(resolver), resolver);
```
jeenbroekstra: I see what you're doing. If we go this route, it needs to be clear at the interface level what is expected. We possible need to add a setter-method to either Sail, or SailConnection, to indicate that EvaluationStrategyFactory is something that can be injected from outside.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -135,10 +141,17 @@
...
+	}
+
+	protected SailSourceConnection(AbstractSail sail, SailStore store,
+			EvaluationStrategyFactory evalStratFactory, FederatedServiceResolver resolver)
```
jeenbroekstra: Actually my comment above should be here :)

jeenbroekstra:
I'm not yet 100% convinced we need such a "heavy" structure (with a Factory class, etc) in place. An alternative would be to simply have (Simple)EvaluationStrategy accept a config parameter and adjust its behavior on the basis of that parameter. 

On the other hand I do see the appeal of a Factory, in terms of (future) flexibility.

pulquero:
The advantages I see are this:
- cleaner code in EvalStrat - use of polymorphism over if-else branching.
- config parameter only supports a limited number of possibilities. What about strict+spinx function, extended+spin functions, extended+spin functions+spinx functions?
- API consumers don't have to write so much boiler plate code and so can focus more on business logic, e.g. extending arithmetic operators to support currency datatypes would mean having to extend sail and sailconnection for each store required. With this change they would just have to extend evalstrat and plug it in to the stores.
- commoditizes evalstrat. Easy to share and re-use evalstraties within the user community.
- democratizies evalstrat. API consumers and API producers both use the same mechanism for extending evalstrat (rather than one having to do a lot of boiler plate and the other using some config param).




    On Wednesday, 29 June 2016, 3:26, Jeen Broekstra <notifications@github.com> wrote:
 

 I'm not yet 100% convinced we need such a "heavy" structure (with a Factory class, etc) in place. An alternative would be to simply have (Simple)EvaluationStrategy accept a config parameter and adjust its behavior on the basis of that parameter. On the other hand I do see the appeal of a Factory, in terms of (future) flexibility. —

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
pulquero: From my reading, EvaluationStrategy is something specific to the SailSourceConnection impl rather than the generic Sail API. It feels like there should be some layer between Sail and MemoryStore/NativeStore where EvaluationStrategy comes in.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailConfig.java
pulquero: Didn't want to containment AbstractSailImplConfig with something that might only be applicable to a particular subclass hierarchy of Sail impls.

pulquero:
I'm guessing there is no way of changing the target branch without closing and creating a new PR?

jeenbroekstra:
@pulquero  unfortunately not. Moreover you may also need to redo or rebase your changes - depending on whether or not you branched off from the master branch or not.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
jeenbroekstra: That will mean that only Sail implementations that rely on SailSourceConnection can actually make use of this - which is not the design intent. The overall design of the various APIs has always been that functionality such as SPARQL query evaluation is _provided_ to the SAIL implementation, and it can choose to reuse or override. These evaluation "modes" are core functionality that the query engine should provide for _any_ SAIL implementation out of the box.

jeenbroekstra: I will copy my line notes over to the new PR, we can continue discussion there.

Issue #187, opened 2016-06-15
Statement controller doesn't take into account max execution time


nikolavp:
Max execution time is ignored in StatementController. This affects https://openrdf.atlassian.net/browse/SES-2174 and the linked issue for remote update queries (going through HTTPRepositoryConnection). This is already done in repository controller so the implementation can be taken from there.

This affects 2.9.0, not sure if it is implemented for 4.0.0

jeenbroekstra:
Hm. It isn't documented that we support a timeout param on this operation, which is probably why it was overlooked. You're correct though that it should. I'll add the param and update the REST API documentation as well.

PR #205, opened 2016-06-29
update execution now takes timeout into account

jeenbroekstra:
This PR addresses GitHub issue: #187 

- timeout parameter now read and used by StatementsController
- added protocol test that verifies form-encoded post for update with timeout param is accepted

- [x] tests are included
- [x] all tests succeed

### DIFF
--- a/compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
+++ b/compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
@@ -17,18 +17,24 @@ import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Random;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.io.Charsets;
 import org.apache.http.HttpEntity;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.entity.ContentType;
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
+import org.apache.http.message.BasicNameValuePair;
 import org.eclipse.rdf4j.common.io.IOUtil;
 import org.eclipse.rdf4j.http.protocol.Protocol;
 import org.eclipse.rdf4j.model.ValueFactory;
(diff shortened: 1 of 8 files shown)


jeenbroekstra:
@nikolavp PR up, have a look.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/statements/StatementsController.java
```diff
@@ -246,13 +248,24 @@ private ModelAndView getSparqlUpdateResult(Repository repository, HttpServletReq
 			}
 		}
 
-		try {
+		final String timeout = request.getParameter(Protocol.TIMEOUT_PARAM_NAME);
```
nikolavp: This is the same code as the one in the RepositoryController. We should probably create a method in Protocol that contains the logic for parsing that parameter?

compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
```diff
@@ -197,6 +203,31 @@ public void testUpdateDirect_POST()
...
+		post.setEntity(entity);
+
+		CloseableHttpResponse response = httpclient.execute(post);
+
+		System.out.println("Update Form Post Status: " + response.getStatusLine());
+		int statusCode = response.getStatusLine().getStatusCode();
+		assertEquals(true, statusCode >= 200 && statusCode < 400);
+	}
+
+	/**
```
nikolavp: We should probably create an unit test that verifies that we are setting the update timeout parameter.

```java
@Test
	public void shouldUseTimeoutParameterForUpdateQueries() throws Exception {
		//prepare
		StatementsController controller = new StatementsController();

		final int maxExecution = 1;
		final MockHttpServletRequest request = new MockHttpServletRequest();
		request.setMethod(HttpMethod.POST.name());
		request.setContentType(Protocol.SPARQL_UPDATE_MIME_TYPE);
		request.addParameter(Protocol.TIMEOUT_PARAM_NAME, String.valueOf(maxExecution));
		final String updateString = "delete where { <monkey:pod> ?p ?o . }";
		request.setContent(updateString.getBytes(StandardCharsets.UTF_8));

		// prepare mocks
		final RepositoryConnection connection = Mockito.mock(RepositoryConnection.class);
		final Update updateMock = Mockito.mock(Update.class);
		Mockito.when(connection.prepareUpdate(QueryLanguage.SPARQL, updateString, null)).thenReturn(updateMock);

		// repository interceptor uses this attribute
		request.setAttribute("repositoryConnection", connection);

		//act
		controller.handleRequest(request, new MockHttpServletResponse());

		Mockito.verify(updateMock).setMaxExecutionTime(maxExecution);
	}
```

jeenbroekstra: I like this, good idea. I'll include it ASAP.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/statements/StatementsController.java
jeenbroekstra: The problem with that is that the protocol module currently has no dependency on javax.servlet, so it can not reference classes like HttpServletRequest. I can of course add the dependency but I'm always a bit hesitant about doing this unless it's necessary. 

However, there is a class ProtocolUtil in the server-spring module, which serves as a sort of go-between. I can probably add it there. I'll have a look.

compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
jeenbroekstra: This requires that we include spring-test (which contains ready-to-use Mock classes for HttpServletRequest etc) as a dependency. Not a problem, but I need to file a piggyback CQ for this ( https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11653 ). Given that it's already used by other Eclipse projects approval of this CQ should be swift.

jeenbroekstra: Drat. spring-test (4.2.1) requires servlet-api 3.0, while we're still relying on 2.5. We should probably upgrade, but that means that we need to drop support for tomcat 6. I have no problem with that but I think we should not just do this as a quick fix in an unrelated issue, just to get our test rig working :) I'll raise a new issue.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/statements/StatementsController.java
nikolavp: (y)

compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
nikolavp: We should be able to change the test but I think that future tests will be much cleaner if we have access to spring test. Didn't know that 2.5 was the servlet version we are using(I think that we bump that in GraphDB anyways).  I also didn't know that tomcat6 supports jdk8 in recent versions

jeenbroekstra: I fully agree. I'm all for upgrading, just didn't want to do it as part of this PR but rather as a separate fix. It's up now, as soon as it's merged I'll update this branch and we can include the test.

jeenbroekstra:
Test added, servlet api upgrade included, test now greenlines. Declaring victory.

Issue #215, opened 2016-07-01
RepositoryConnectionUtil.isHasStatementOptimized(member) disappeared from FederationJoinOptimizer$OwnerScanner.getSingleOwner()


pulquero:
It seems that in going from 2.9 to 4.0 this was removed from FederationJoinOptimizer and hence no longer present in rdf4j. The purpose of this was to support stores that have a very expensive hasStatement() operation. In this case, it is cheaper to give up then take the hit of multiple hasStatement() calls.

It was also used in EmptyPatternOptimizer.

pulquero:
Support for setting bloom filters for federation members has been added. If present, the bloom filter is then used where possible instead of hasStatement(). This allows for custom hasStatement logic, in particular, a constant true can simply be returned for federation members that have a very expensive hasStatement impl.

PR #221, opened 2016-07-04
Issue #215: federation performance regression

pulquero:
This PR addresses GitHub issue: #215 

- Fixes performance regression by replacing missing old code with bloom filter alternative.

- [x] all tests succeed


jeenbroekstra:
Looks fine. But in order to use this, shouldn't we provide an actual implementation of a bloom filter?

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/AbstractFederationConnection.java
```diff
@@ -434,10 +436,18 @@ private TupleExpr optimize(TupleExpr parsed, Dataset dataset, BindingSet binding
 		new QueryMultiJoinOptimizer().optimize(query, dataset, bindings);
 		// new FilterOptimizer().optimize(query, dataset, bindings);
 
-		new EmptyPatternOptimizer(members).optimize(query, dataset, bindings);
+		Map<Repository, RepositoryBloomFilter> bloomFilters = federation.getBloomFilters();
+		Map<RepositoryConnection, RepositoryBloomFilter> bloomFiltersByConn = new HashMap<>();
+		List<Repository> repoMembers = federation.getMembers();
+		for(int i=0; i<members.size(); i++) {
+			bloomFiltersByConn.put(members.get(i), bloomFilters.get(repoMembers.get(i)));
```
ansell: Does this need a null check on the result of repoMembers.get(i)?

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/FederationJoinOptimizer.java
```diff
@@ -370,6 +379,14 @@ else if (result != member) {
 			return result;
 		}
 
+		private boolean mayHaveStatement(RepositoryConnection conn, Resource subj, IRI pred, Value obj,
```
ansell: Can this method be reused as it seems to be implemented both here and in EmptyPatternOptimizer?

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/RepositoryBloomFilter.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.repository.RepositoryConnection;
+
+public interface RepositoryBloomFilter {
+
+	/**
+	 * Returns true if the repository may have such a statement or false if it definitely does not.
```
ansell: Needs javadoc to describe the parameters and preconditions.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/Federation.java
```diff
@@ -94,6 +100,18 @@ public void addMember(Repository member) {
...
 
+	List<Repository> getMembers() {
+		return Collections.unmodifiableList(members);
+	}
+
+	public void setBloomFilter(Repository member, RepositoryBloomFilter filter) {
+		bloomFilters.put(member, filter);
+	}
+
+	Map<Repository, RepositoryBloomFilter> getBloomFilters() {
```
ansell: Should this be a protected or private method?

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/Federation.java
```diff
@@ -94,6 +100,18 @@ public void addMember(Repository member) {
 		members.add(member);
 	}
 
+	List<Repository> getMembers() {
```
ansell: Should this be a protected or private method?

ansell: Ie, being package private (default) means it is accessible to AbstractFederationConnection but no subclasses of this non-final class unless they place themselves inside of this package.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/RepositoryBloomFilter.java
```diff
@@ -0,0 +1,37 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: Minor thing, but you're using the wrong copyright header. For new files, Aduna should not be mentioned. See the copyright header template in eclipse-settings/codetemplates.xml .

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/InaccurateRepositoryBloomFilter.java
```diff
@@ -0,0 +1,19 @@
+package org.eclipse.rdf4j.sail.federation.optimizers;
```
jeenbroekstra: Missing copyright header

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/AccurateRepositoryBloomFilter.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.rdf4j.sail.federation.optimizers;
```
jeenbroekstra: Missing copyright header

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/AbstractFederationConnection.java
```diff
@@ -434,10 +439,27 @@ private TupleExpr optimize(TupleExpr parsed, Dataset dataset, BindingSet binding
...
+		Map<RepositoryConnection, RepositoryBloomFilter> bloomFiltersByConn = new HashMap<>();
+		List<Repository> repoMembers = federation.getMembers();
+		for(int i=0; i<members.size(); i++) {
+			RepositoryBloomFilter bloomFilter = bloomFilters.get(repoMembers.get(i));
+			if (bloomFilter == null) {
+				bloomFilter = defaultBloomFilter;
+			}
+			bloomFiltersByConn.put(members.get(i), bloomFilter);
+		}
+		com.google.common.base.Function<RepositoryConnection, RepositoryBloomFilter> bloomFilterFunction = Functions.forMap(
```
jeenbroekstra: Given that we're Java 8 shop, shouldn't we use java.util.function.Function for this instead?

jeenbroekstra:
I can see these bloom filters potentially being useful for other Repository/Sail implementations as well (e.g. SPARQLRepository or HTTPRepository). Perhaps they should be moved into rdf4j-repository-api (in `oer.repository.optimizers` or `.filters` perhaps?)

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/AbstractFederationConnection.java
```diff
@@ -434,10 +439,27 @@ private TupleExpr optimize(TupleExpr parsed, Dataset dataset, BindingSet binding
 		new QueryMultiJoinOptimizer().optimize(query, dataset, bindings);
 		// new FilterOptimizer().optimize(query, dataset, bindings);
 
-		new EmptyPatternOptimizer(members).optimize(query, dataset, bindings);
+		// prepare bloom filters
+		RepositoryBloomFilter defaultBloomFilter = new AccurateRepositoryBloomFilter(includeInferred);
+		Map<Repository, RepositoryBloomFilter> bloomFilters = federation.getBloomFilters();
+		Map<RepositoryConnection, RepositoryBloomFilter> bloomFiltersByConn = new HashMap<>();
+		List<Repository> repoMembers = federation.getMembers();
+		for(int i=0; i<members.size(); i++) {
```
ansell: Is the index link between the List at federation.getMembers() and the members List variable here documented? It seems like it would be safer and simpler to manage them as a map then as two lists that need synchronisation to work here.

pulquero: I agree, but supporting the existing constructors makes it a bit awkward. It is not just the optimizers, but also the connection subclasses. This was less invasive. Ideally, I would want to add a new constructor which took a list of pair<Repository,RepositoryConnection> instead of list<RepoConn>. I can map repoconn -> pair<null,repoconn> to keep backwards compatibility (and I have its adjoint, the forgetful functor in the opposite direction). Every repo is then known (or not) for every repoconn so it is a straightforward lookup to get the filter (or anything else in the future).

pulquero: Are we happy as-is or should I go ahead with this change?

jeenbroekstra: Sorry mark, lost track of this for a bit - but your proposal for an approach with a constructor with a list of pairs sounds good, if that makes things more flexible and robust. If you do this, could you add a comment on issue #215 itself that quickly outlines the changes to the API (and whether or not they are backward compatible)? Just so I have this at hand when the time comes to prep release notes.

pulquero: It was even easier, I noticed there is a RepositoryConnection.getRepository() method which is exactly what was needed.

jeenbroekstra:
Looks good to me! @ansell if you're happy too give me a thumbs up or something, and I will merge. 

Note that this goes into master - so we have a bit of time to make further refinements if required.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/AbstractFederationConnection.java
```diff
@@ -434,10 +438,18 @@ private TupleExpr optimize(TupleExpr parsed, Dataset dataset, BindingSet binding
 		new QueryMultiJoinOptimizer().optimize(query, dataset, bindings);
 		// new FilterOptimizer().optimize(query, dataset, bindings);
 
-		new EmptyPatternOptimizer(members).optimize(query, dataset, bindings);
+		// prepare bloom filters
+		RepositoryBloomFilter defaultBloomFilter = new AccurateRepositoryBloomFilter(includeInferred);
+		Map<Repository, RepositoryBloomFilter> bloomFilters = federation.getBloomFilters();
+		java.util.function.Function<Repository, RepositoryBloomFilter> bloomFilterFunction = c -> Optional.ofNullable(
```
ansell: This can be simplified even further to bloomFilters.getOrDefault(c, defaultBloomFilter).

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/Federation.java
```diff
@@ -94,6 +100,18 @@ public void addMember(Repository member) {
 		members.add(member);
 	}
 
+	protected List<Repository> getMembers() {
+		return Collections.unmodifiableList(members);
+	}
+
+	public void setBloomFilter(Repository member, RepositoryBloomFilter filter) {
```
ansell: This needs javadoc to explain its assumptions, in particular, that not every Repository in getMembers needs to have this method called on it if the user doesn't need to override the default implementation.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/Federation.java
```diff
@@ -94,6 +100,18 @@ public void addMember(Repository member) {
 		members.add(member);
 	}
 
+	protected List<Repository> getMembers() {
```
ansell: This needs javadoc to explain the unmodifiable List return.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/Federation.java
```diff
@@ -94,6 +100,18 @@ public void addMember(Repository member) {
...
+		return Collections.unmodifiableList(members);
+	}
+
+	public void setBloomFilter(Repository member, RepositoryBloomFilter filter) {
+		bloomFilters.put(member, filter);
+	}
+
+	protected Map<Repository, RepositoryBloomFilter> getBloomFilters() {
```
ansell: This needs javadoc to explain the unmodifiable Map return.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EmptyPatternOptimizer.java
```diff
@@ -58,7 +70,8 @@ public void meet(StatementPattern node)
 		Value obj = node.getObjectVar().getValue();
 		Resource[] ctx = getContexts(node.getContextVar());
 		for (RepositoryConnection member : members) {
-			if (member.hasStatement(subj, pred, obj, true, ctx)) {
+			RepositoryBloomFilter bloomFilter = bloomFilters.apply(member.getRepository());
+			if (bloomFilter == null || bloomFilter.mayHaveStatement(member, subj, pred, obj, ctx)) {
```
ansell: This implementation detail, that a bloom filter is necessary for all repositories or else optimisation does not occur, should be documented in the class javadoc comment.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/FederationJoinOptimizer.java
```diff
@@ -356,7 +370,8 @@ else if (dataset != null) {
 				// fallback to using hasStatement()
 				// but hopefully we narrowed it down to results
 				for (RepositoryConnection member : results) {
-					if (member.hasStatement(subj, pred, obj, true, ctx)) {
+					RepositoryBloomFilter bloomFilter = bloomFilters.apply(member.getRepository());
+					if (bloomFilter == null || bloomFilter.mayHaveStatement(member, subj, pred, obj, ctx)) {
```
ansell: Not sure how to interpret this line. It seems to say that if the bloom filter is not available, this repository could be chosen, which doesn't seem to make sense.

ansell: Ie, I would have presumed the refactoring to go to the following:

if (bloomFilter != null && bloomFilter.mayHaveStatement(...))

ansell:
There are no new tests for this fairly complex pull request so even if the current tests all succeed, it doesn't provide any regression tests for the new functionality, which is fairly large considering that optimisation code is affected. I can't approve it without tests of each of the new branches.

I understand that BloomFilters are difficult to test in general. but at least each of the new branches could be tested, including the "bloomFilter == null" cases that may not be tested by existing code that gets an accurate bloom filter applied to it in the previous constructor versions.

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/filters/RepositoryBloomFilter.java
```diff
@@ -0,0 +1,37 @@
...
+	 * 
+	 * @param conn
+	 *        connection to the repository to check.
+	 * @param subj
+	 *        subject of the statement to check for (can be null).
+	 * @param pred
+	 *        predicate of the statement to check for (can be null).
+	 * @param obj
+	 *        object of the statement to check for (can be null).
+	 * @param ctxs
```
ansell: I dont believe that we support the case where ctxs is null in other interfaces. This should be an empty array or an array with a single null element for the all cases, and default cases respectively.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/Federation.java
ansell: Should this method support the removal case? I can't remember if HashMap.put(key, null) is supported, but in general the Map.put contract says that it can throw an NPE if value is null.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EmptyPatternOptimizer.java
pulquero: It is not needed for all. If I have a memory store and a MyDifferentStore, then only MyDifferentStore would need the bloom filter as the characteristics of memorystore.hasStatement() are generally good enough.

pulquero:
Or eliminate the branches through functional programming.

jeenbroekstra:
@pulquero what's the status of this PR? @ansell remarked that some additional regression tests for the new functionality are needed - are you in a position to add a few such tests?

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/AbstractFederationConnection.java
```diff
@@ -434,10 +437,18 @@ private TupleExpr optimize(TupleExpr parsed, Dataset dataset, BindingSet binding
 		new QueryMultiJoinOptimizer().optimize(query, dataset, bindings);
 		// new FilterOptimizer().optimize(query, dataset, bindings);
 
-		new EmptyPatternOptimizer(members).optimize(query, dataset, bindings);
+		// prepare bloom filters
+		RepositoryBloomFilter defaultBloomFilter = new AccurateRepositoryBloomFilter(includeInferred);
+		Map<Repository, RepositoryBloomFilter> bloomFilters = federation.getBloomFilters();
+		java.util.function.Function<Repository, RepositoryBloomFilter> bloomFilterFunction = c -> bloomFilters.getOrDefault(
```
jeenbroekstra: If you change this function to `<Function, Optional<RepositoryBloomFliter>>` (implementation would be `c -> Optional.ofNullable(bloomFilters.get...))` then in the places where you now use `MoreObjects` you can simply do `firstFilter.or(secondFilter)`.

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EmptyPatternOptimizer.java
```diff
@@ -58,7 +71,10 @@ public void meet(StatementPattern node)
 		Value obj = node.getObjectVar().getValue();
 		Resource[] ctx = getContexts(node.getContextVar());
 		for (RepositoryConnection member : members) {
-			if (member.hasStatement(subj, pred, obj, true, ctx)) {
+			RepositoryBloomFilter bloomFilter = MoreObjects.firstNonNull(
```
jeenbroekstra: As said above: if the output of the `bloomFilters` function is an `Optional<RepositoryBloomFilter>` you can replace this with:

    RepositoryBloomFilter bloomFilter = bloomFilters.apply(member.getRepository())
                                                                        .or(AccurateRepositoryBloomFilter.INCLUDE_INFERRED_INSTANCE);

jeenbroekstra: ...though I can see that in other places this may make the code a bit more complex. So not too hung up on changing this - if you prefer the current approach just leave it.

jeenbroekstra:
Running hudson build for this PR: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/22/

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EmptyPatternOptimizer.java
pulquero: Upon reflection, I think leave it as-is.

jeenbroekstra:
New (and old) tests all succeed, LGTM.

Issue #195, opened 2016-06-24
"Multiple statements for pattern" error when attempting to load RDF file w/ SPIN constraints


gwcox66:
I'm using Sesame OpenRDF Workbench:

> System Information
> Application Information
> Application Name    OpenRDF Workbench
> Version 4.1.2
> Runtime Information
> Operating System    Windows 8.1 6.3 (amd64)
> Java Runtime    Oracle Corporation Java HotSpot(TM) 64-Bit Server VM (1.8.0_91)
> Process User    Greg
> Memory
> Used    357 MB
> Maximum 3463 MB

My constraints, classes, properties, and example individuals are contained in one RDF file under TopBraid Composer Free Edition (attached).  I'm attempting to load into a repository of type "In Memory Store with SPIN+RDFS support".  Screen grab of full error page is also attached.

I've tried using the Sesame Open RDF workbench Modify/Add command to pull in the RDF file generated by TopBraid Free Edition, but I get an error that appears to be related to a blank node:

> javax.servlet.ServletException: org.openrdf.repository.RepositoryException: org.openrdf.repository.RepositoryException: org.openrdf.sail.SailException: org.openrdf.query.QueryEvaluationException: Multiple statements for pattern: _:node1alqlr2eix590 http://spinrdf.org/sp#where null

[SXXIComplianceCheck2.zip](https://github.com/eclipse/rdf4j/files/332392/SXXIComplianceCheck2.zip)
![image](https://cloud.githubusercontent.com/assets/20132242/16347819/59d8e84a-3a14-11e6-878e-099d154e955b.png)

pulquero:
I believe that error is due to that pattern resulting in more than one matching statement, I.e. multiple where clauses. I assume you are loading into an empty repository?

gwcox66:
Yes, I was loading into an empty repository, and because of this error, the repository remains empty.  No facts were loaded.

For your convenience, here's the SPARQL source code (from TopBraid Composer Free).  This should be the only SPIN constraint remaining in the RDF file...

```
# A proposal with no security classification (data item 005) is not valid
ASK WHERE {
    ?this a sxxicc:Pub7Proposal .
    OPTIONAL {
        ?this sxxicc:pub7ProposalHasDataItem ?dataItem .
        ?dataItem a sxxicc:Pub7DataItem005 .
        ?dataItem sxxicc:pub7DataItemHasStringValue ?securityClassificationString .
    } .
[...]

PR #223, opened 2016-07-05
Multiple context fix.

pulquero:
This PR addresses GitHub issue: #195 

- ...
- ...
- ...

- [x] tests are included
- [x] all tests succeed

Signed-off-by: Mark Hale <mark.hale@physics.org>

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/Statements.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/Statements.java
@@ -16,15 +16,19 @@ import org.eclipse.rdf4j.model.Literal;
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.model.Statement;
 import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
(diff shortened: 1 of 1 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/Statements.java
```diff
@@ -133,10 +145,13 @@ public static Statement single(Resource subj, IRI pred, Value obj, TripleSource
 				pred, obj);
 		try {
 			if (stmts.hasNext()) {
-				stmt = stmts.next();
-				if (stmts.hasNext()) {
-					throw new QueryEvaluationException(
+				stmt = createTripleStatement(stmts.next());
+				while (stmts.hasNext()) {
+					Statement nextStmt = createTripleStatement(stmts.next());
+					if (!nextStmt.equals(stmt)) {
```
jeenbroekstra: Wouldn't it be more efficient to adapt this equality-check rather than creating a new Statement object every time? createTripleStatement is likely to cause a lot of object churn. 

I remember that when we changed `Statement.equals()` to include the context field, we briefly talked about introducing a separate method `equalsIgnoreContext` to cover the old behavior. It seems that never made it through - but it might be a useful method to add to  the `org.eclipse.rdf4j.model.Statements` utility class.

jeenbroekstra:
I have some time to work on this - I'll put up an alternative PR based off of your proposed fix.

jeenbroekstra:
Opened alternative PR #225 with my proposed extensions to this fix.

PR #224, opened 2016-07-05
Initial design for pluggable SPARQL evaluation modes (strict/extended/...)

pulquero:

- ...
- ...
- ...

- [x] tested in workbench
- [x] all tests succeed


jeenbroekstra:
For reference, see discussion in previous PR #204 . Several good points about the design were made there.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -135,10 +141,17 @@
 	 * @param resolver
 	 */
 	protected SailSourceConnection(AbstractSail sail, SailStore store, FederatedServiceResolver resolver) {
+		this(sail, store, new SimpleEvaluationStrategyFactory(resolver), resolver);
+	}
+
+	protected SailSourceConnection(AbstractSail sail, SailStore store,
+			EvaluationStrategyFactory evalStratFactory, FederatedServiceResolver resolver)
```
jeenbroekstra: (in response to @pulquero 's comment on the old PR that he considers `EvaluationStrategy` specific to `SailSourceConnection`, and not generic for `Sail`). 

That will mean that only Sail implementations that rely on SailSourceConnection can actually make use of this - which is not the design intent. The overall design of the various APIs has always been that functionality such as SPARQL query evaluation is provided to any SAIL implementation. These evaluation "modes" are core functionality that the query engine should provide for any SAIL implementation out of the box.

pulquero: Should it be moved up to AbstractSail/AbstractSailConnection in that case? What you are describing sounds more like TripleSource - evaluation is provided for it. In the same way that there is a SailRepository for Repository, it appears to me you are missing something that plays the role of a TripleSourceSail/EvaluationStrategySail to complete the API journey:

>
Repository - query language
   |
   |- SailRepository - Repository delegates to Sail via QueryParser
   |
Sail - query algebra
   |
   |- ??? Sail delegates to TripleSource via EvaluationStrategy
   |
TripleSource - statements

At the moment, it is on a sail impl-by-impl basis, e.g. MemoryStore and NativeStore. Maybe AbstractSail is a good candidate?

jeenbroekstra: Possibly - it sounds like a good candidate, but I need to find some time to map this out. I feel a strong need to draw some diagrams :)

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailConfig.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.util.ModelException;
+import org.eclipse.rdf4j.model.util.Models;
+import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategyFactory;
+import org.eclipse.rdf4j.sail.config.AbstractSailImplConfig;
+import org.eclipse.rdf4j.sail.config.SailConfigException;
+
+import static org.eclipse.rdf4j.sail.base.config.BaseSailSchema.*;
+
+public abstract class BaseSailConfig extends AbstractSailImplConfig {
```
ansell: What is the reasoning behind creating this as a new abstract class and not adding it all to the abstract superclass?

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailConfig.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.util.ModelException;
+import org.eclipse.rdf4j.model.util.Models;
+import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategyFactory;
+import org.eclipse.rdf4j.sail.config.AbstractSailImplConfig;
+import org.eclipse.rdf4j.sail.config.SailConfigException;
+
+import static org.eclipse.rdf4j.sail.base.config.BaseSailSchema.*;
```
ansell: This seems to only be importing a single IRI constant for use here. Would it be simpler just to import that constant directly to avoid the ambiguous star import?

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailSchema.java
```diff
@@ -0,0 +1,29 @@
...
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.base.config;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+
+/**
+ * Defines constants for the BaseSail schema.
```
ansell: What is the reasoning behind not adding these to one of the existing config vocabularies such as SailConfigSchema?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
```diff
@@ -185,6 +189,23 @@ public long getSyncDelay() {
 	}
 
 	/**
+	 * @return Returns the {@link EvaluationStrategy}.
+	 */
+	public synchronized EvaluationStrategyFactory getEvaluationStrategyFactory() {
```
ansell: Will having this method synchronised instead of using volatile double-checked locking affect concurrent read thread performance on MemoryStore? Similar for the recently added FederatedServiceResolver methods below that could also use volatile double-checked locking instead.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/EvaluationStrategyFactory.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package org.eclipse.rdf4j.query.algebra.evaluation;
+
+import org.eclipse.rdf4j.query.Dataset;
+
+/**
+ * Factory for {@link EvaluationStrategy}s.
+ */
+public interface EvaluationStrategyFactory {
+	EvaluationStrategy createEvaluationStrategy(Dataset dataset, TripleSource tripleSource);
```
ansell: This needs javadoc. In particular, a reference to calling EvaluationStrategies.register as a recommendation to implementors before returning the EvaluationStrategy from this method if that is desired.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/SimpleEvaluationStrategy.java
```diff
@@ -154,7 +155,7 @@
  * @author David Huynh
  * @author Andreas Schwarte
  */
-public class SimpleEvaluationStrategy implements EvaluationStrategy, UUIDable {
+public class SimpleEvaluationStrategy implements EvaluationStrategy, FederatedServiceResolverClient, UUIDable {
```
ansell: Can this change be separated out into a separate pull request? It doesn't look like it is related to the rest of this one.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -135,10 +141,17 @@
 	 * @param resolver
 	 */
 	protected SailSourceConnection(AbstractSail sail, SailStore store, FederatedServiceResolver resolver) {
+		this(sail, store, new SimpleEvaluationStrategyFactory(resolver), resolver);
```
ansell: It is a red flag to me when the same variable is being passed through to two different locations from a constructor like this. 

Is the resolver actually being used here anymore and the fourth argument is just an historical artifact to keep the previous API?

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailConfig.java
pulquero: I wasn't sure about whether is was desirable to couple the base class with this feature. General consesus seems to be that it would be acceptable so I'll push it up.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailSchema.java
pulquero: I'll push this up too.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
pulquero: No, it is only used to configure the sail - one-call per lifetime.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/SimpleEvaluationStrategy.java
pulquero: This is used on line 173 in SailSourceConnection.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
pulquero: Historical artifact. Ideally, there would be no sight of FederatedServiceResolver here as the EvaluationStrategyFactory is responsible for manuafacturing fully configued EvaluationStrategies. But you are right, I can do a better job here.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/config/BaseSailConfig.java
pulquero: oh wait, that would mean sail-api would have to depend on queryalgebra-evaluation. Any objections to that?

PR #225, opened 2016-07-06
Context handling fix in SPIN 

jeenbroekstra:
This PR addresses GitHub issue: #195

- based on @pulquero 's provided fix for issue #195 .
- introduced `Statements.isSameTriple` utility method to compare statements disregarding context.
- reneamd `oer.query.algebra.evaluation.util.Statements` to `TripleSources` to avoid confusion with `oer.model.util.Statements`.
- added TODO to indicate we should look at closer integration of TripleSources code with existing util functionality in, e.g., `Models` and `RDFCollections`. 

- [x] tests are included
- [x] all tests succeed


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/TripleSources.java
```diff
@@ -0,0 +1,277 @@
...
+	{
+		Statement stmt;
+		CloseableIteration<? extends Statement, QueryEvaluationException> stmts = store.getStatements(subj,
+				pred, obj);
+		try {
+			if (stmts.hasNext()) {
+				stmt = stmts.next();
+				while (stmts.hasNext()) {
+					Statement nextStmt = stmts.next();
+					if (!org.eclipse.rdf4j.model.util.Statements.isSameTriple(stmt, nextStmt)) {
```
jeenbroekstra: This is the core of the fix as proposed by @pulquero . The main change is that instead of converting every statement to triple-only form, we simply do a different comparison.

pulquero: Looks fine but I  think we should still return a triple-only statement else we are selecting a context at random - it is better to null it out than expose it.

jeenbroekstra: I disagree, for several reasons:

1. Context is an intrinsic part of the Statement. Note also that for example a `RepositoryConnection.getStatements` operation that "disregards context" will still return statement objects with the context field filled in - so this way is more consistent with other parts of the API. 
2. it is a performance burden; you will have to create new Statement objects for every match, causing object churn. 

In a future enhancement we probably want to look at actually making the spin reasoner context-aware - which I think will be easier to do if we don't tamper with contexts too much at this point.

pulquero: You only need to create a new statement on return not every match. Happy with it as-is.

Issue #227, opened 2016-07-11
2.0.x: Wrong serialization of language tag in SPARQLConnection


QuarksToQuasars:
In org.openrdf.repository.sparql.SPARQLConnection#createDataBody, getLanguage() is called on the literal:
`				if (Literals.isLanguageLiteral(lit)) {
					qb.append("@");
					qb.append(lit.getLanguage());
				}` (c.f. https://github.com/eclipse/rdf4j/blob/releases/2.0.x/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java#L815)
However, getLanguage returns an Optional i.e. signature is ` Optional<String> org.openrdf.model.Literal.getLanguage()`. This results effectively in an invalid SPARQL insert serialization e.g. you get something like ```INSERT { <http://rdf4j.org/> <http://rdf4j.org/pred> "my literal"@Optional[en-US] .  } WHERE{}```

The following simple test fails with an UpdateExecutionException (i.e. most sparql repository implementations with rdf4j sparql parser will throw probably an MalformedQueryException internally):
```
@Test
    public void testSparqlConnectionAddingLanguageTags() throws IOException{
        java.net.URL documentUrl = new URL("http://www.w3.org/TR/2004/REC-rdf-syntax-grammar-20040210/example08.rdf");

        Repository rep= null;
        try{
            rep = new SPARQLRepository("http://127.0.0.1:10020/sparql");
            rep.initialize();
            try(RepositoryConnection con = rep.getConnection()){
                try(InputStream inputStream = documentUrl.openStream()){
                    RDFParser rdfParser = Rio.createParser(RDFFormat.RDFXML);
                    Model model = new LinkedHashModel();
                    rdfParser.setRDFHandler(new StatementCollector(model));
                    rdfParser.parse(inputStream, documentUrl.toString());
                    con.add(model);
                }
            }
        }finally{
            rep.shutDown();
        }
    }
```

jeenbroekstra:
Fixed. Fix should be available in next 2.0 nightly build (and of course the next milestone / release).

PR #230, opened 2016-07-12
#227 fixed use of getLanguage() return value

jeenbroekstra:
This PR addresses GitHub issue: #227

-    getLanguage() returns an Optional, if it has been verified to be non-empty, use getLanguage().get() (or getLanguage.orElse(null)) to retrieve the actual string value.

Make sure you've followed the Contributor Guidelines. In particular (please tick to indicate you've taken care of it):

- [x] tests are included
- [x]   all tests succeed

### DIFF
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -812,7 +812,7 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 
 				if (Literals.isLanguageLiteral(lit)) {
 					qb.append("@");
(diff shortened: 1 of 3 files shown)


core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -938,9 +938,9 @@ private void createBGP(StringBuilder qb, Resource subject, IRI predicate, Value
 				qb.append(SPARQLUtil.encodeString(lit.getLabel()));
 				qb.append("\"");
 
-				if (Literals.isLanguageLiteral(lit)) {
+				if (lit.getLanguage().isPresent()) {
```
jeenbroekstra: Arbitrary change this - both methods do exactly the same thing. We should perhaps look into deprecating the util method.

PR #233, opened 2016-07-13
#91 using secondary cache for maintenance

jeenbroekstra:
This PR addresses GitHub issue: #91

- primary cache in txn registry no longer has a timeout
- secondary cache with timeout verifies if entry is still actively used server-side. If not, entry is also removed from primary cache.
- added unit tests which verify expected behavior. 

- [x] tests are included
- [x] all tests succeed

### DIFF
--- a/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
+++ b/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
@@ -231,7 +231,7 @@ public enum ActiveTransactionRegistry {
 			throw new RepositoryException(
 					"transaction with id " + transactionId + " is no longer registered!");
 		}
-		bumpSecondaryCache(transactionId, entry);
+		updateSecondaryCache(transactionId, entry);
 
 		return entry.getConnection();
 	}
@@ -246,7 +246,7 @@ public enum ActiveTransactionRegistry {
 	public void returnTransactionConnection(UUID transactionId) {
 		final CacheEntry entry = primaryCache.getIfPresent(transactionId);
 		if (entry != null) {
-			bumpSecondaryCache(transactionId, entry);
+			updateSecondaryCache(transactionId, entry);
 			final ReentrantLock txnLock = entry.getLock();
 			if (txnLock.isHeldByCurrentThread()) {
 				txnLock.unlock();
@@ -256,14 +256,14 @@ public enum ActiveTransactionRegistry {
 
 	/**
 	 * Checks if the given transaction entry is still in the secondary cache (resetting its last access time
-	 * in the process) and if not reinsert it.
+	 * in the process) and if not reinserts it.
 	 * 
 	 * @param transactionId
 	 *        the id for the transaction to check
 	 * @param entry
 	 *        the cache entry to insert if necessary.
 	 */
-	private void bumpSecondaryCache(UUID transactionId, final CacheEntry entry) {
+	private void updateSecondaryCache(UUID transactionId, final CacheEntry entry) {
 		try {
 			secondaryCache.get(transactionId, new Callable<CacheEntry>() {
 
--- a/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
+++ b/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
@@ -59,8 +59,8 @@ public enum ActiveTransactionRegistry {
 
 	/**
 	 * The secondary cache does automatic cleanup of its entries based on the configured timeout. If an
-	 * expired entry no longer is no longer handed out to an actual server-side operation, it is considered
-	 * "orphaned" and discarded from the primary cache.
+	 * expired entry is no longer locked by any thread, it is considered "orphaned" and discarded from the
+	 * primary cache.
 	 */
 	private final Cache<UUID, CacheEntry> secondaryCache;
 
@@ -69,8 +69,6 @@ public enum ActiveTransactionRegistry {
 		private final RepositoryConnection connection;
 
 		private final ReentrantLock lock = new ReentrantLock();
-
-		private boolean handedOut;
 
 		public CacheEntry(RepositoryConnection connection) {
 			this.connection = connection;
@@ -88,23 +86,6 @@ public enum ActiveTransactionRegistry {
 		 */
 		public ReentrantLock getLock() {
 			return lock;
-		}
-
-		/**
-		 * Indicates if the entry is currently handed out to a thread
-		 * 
-		 * @return true iff the entry is handed out to a thread, false otherwise.
-		 */
-		public boolean isHandedOut() {
-			return handedOut;
-		}
-
-		/**
-		 * @param handedOut
-		 *        the handedOut to set
-		 */
-		public void setHandedOut(boolean handedOut) {
-			this.handedOut = handedOut;
 		}
 
 	}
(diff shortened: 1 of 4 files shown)


core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
```diff
@@ -94,30 +107,38 @@ private ActiveTransactionRegistry() {
...
-					try {
-						entry.getConnection().close();
+					final UUID transactionId = notification.getKey();
+					final CacheEntry entry = notification.getValue();
+					synchronized (primaryCache) {
+						if (!entry.getLock().isLocked()) {
+							// no operation active, we can decommission this entry
+							primaryCache.invalidate(transactionId);
+							logger.warn("deregistered expired transaction {}", transactionId);
+						}
```
jeenbroekstra: Note to self: as extra insurance we should re-insert the item in the secondary cache if it is still active.

ansell:
I am not familiar with the concept of a secondary cache or CacheBuilder, so not sure how to review this pull request myself.

jeenbroekstra:
I'll try and explain the design.

To keep transaction 'sessions' open over HTTP, we need to keep a connection open at server-side once a transaction has started on that connection. To enable this, we have a (primary) cache of connection objects, identified by their txn ID (a UUID). Whenever a new client-side request comes in, the server gets the txn id from the request, looks up the corresponding connection, and executes the new operation as part of the running transaction. If the incoming request is to end  the transaction, the connection can be discarded from the cache. The primary cache is a basic dictionary/hashmap, in effect. 

The problem is how to deal with dangling connections: if for whatever reason a client cannot finish the transaction, the server at some point should decide to abort and close the connection. This is handled by the timeout parameter. In the original design, the primary cache itself just had an expiry: it removed connections that hadn't been accessed for a while. The problem with this, however, is that sometimes the access timeout expires not because the client is unresponsive, but because the server itself is still busy with an operation (e.g. a delete operation on a large store). In such cases, the connection should not just be closed and ejected when the timeout occurs. 

The primary cache can not expire objects and then re-insert them in itself. This is because removal notification is a scheduled task and not immediate, so there is a potential period at which the connection is removed but not yet reinserted by the removal handler. This is why instead we add a secondary cache that keeps the timeout. On removal notice, it checks if the connection is still in use. If it isn't, it ejects it from the primary cache. If it is, it just reinserts it into itself (note that for the secondary cache this is no problem, because for request handling we always retrieve connections from the primary cache only).

nikolavp:
Can't we just reinsert the connection if it is currently locked and use a single cache instead?

jeenbroekstra:
@nikolavp, unfortunately not. The reason is that we need to do this by listening for a removal event, at which point we can indeed reinsert the object. However the removal event don't necessarily happen _immediately_ after removal. So there is likely to be some time between the object disappearing from the cache and the listener reinserting it. See http://stackoverflow.com/a/8919768/451941 for more details.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
```diff
@@ -216,12 +244,39 @@ public RepositoryConnection getTransactionConnection(UUID transactionId)
...
+	/**
+	 * Checks if the given transaction entry is still in the secondary cache (resetting its last access time
+	 * in the process) and if not reinsert it.
+	 * 
+	 * @param transactionId
+	 *        the id for the transaction to check
+	 * @param entry
+	 *        the cache entry to insert if necessary.
+	 */
+	private void bumpSecondaryCache(UUID transactionId, final CacheEntry entry) {
```
ansell: Naming this "ensureInSecondaryCache" may be easier to read than bump, unless it is a technical term.

jeenbroekstra: 'bump' is _totally_ a [technical term](http://www.urbandictionary.com/define.php?term=bump) :)

Seriously though, you're right, I'll  change it to 'updateSecondaryCache' .

core/http/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/transaction/TestActiveTransactionRegistry.java
```diff
@@ -0,0 +1,152 @@
...
+		Runnable r3 = new Runnable() {
+
+			@Override
+			public void run() {
+				try {
+					done.await();
+					registry.deregister(txnId1);
+				}
+				catch (InterruptedException e) {
+					// TODO Auto-generated catch block
```
ansell: If this exception isn't expected in practice (hopefully nothing is interrupting this thread during the test), this should be setup to fail the test, possibly through setting an AtomicBoolean variable and then checking for it at the end of the test.

core/http/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/transaction/TestActiveTransactionRegistry.java
```diff
@@ -0,0 +1,152 @@
...
+			}
+		};
+
+		Thread t1 = new Thread(r1, "r1");
+		Thread t2 = new Thread(r2, "r2");
+		Thread t3 = new Thread(r3, "r3");
+
+		t3.start();
+		t2.start();
+		t1.start();
```
ansell: I haven't used JUnit like this before for multithreaded testing. How does it know to keep the test running long enough for it to complete? Previously I have used join or wait to let them to complete and then test the exit conditions.

core/http/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/transaction/TestActiveTransactionRegistry.java
ansell: Ditto for all other try-catch in this test.

core/http/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/transaction/TestActiveTransactionRegistry.java
jeenbroekstra: Good point - I'm using a countdown latch to make all threads wait for each other, but I forgot to ensure the main thread doesn't exit. Will fix.

jeenbroekstra: Done. See below: main thread now waits for thread 3 (which, in turn, waits for all other threads to complete).

nikolavp:
LGTM +1

jeenbroekstra:
Before I merge I want to actually deploy a server and do an "in the wild" test to see that this fix actually solves #91

jeenbroekstra:
There is still a problem. While the above fix solves the exception we get at the client side, the transaction now seems to hang eternally upon the commit operation. Looking into it.

jeenbroekstra:
It appears that the cause is that the clear operation is converted to a call on RepositoryConnection.remove for all statements (which is very inefficient on large native stores). Two things to do:

- [ ]  fix TransactionController to use clear explicitly
- [ ]  optimize SailREpositoryConnection to use sailConnection.clear when subject, predicate and object are all null.

nikolavp:
I think a better place for this would be in RepositoryConnectionBase.remove? This way all even the memory store or other repository types would get the optimization.

jeenbroekstra:
@nikolavp we can't do this in RepositoryConnectionBase (or AbstractRepositoryConnection as it's known now), because at that level we don't know whether we are transferring to a Sail or not. In fact, the depenency is the other way around: the default implementation of AbstractRepositoryConnection.clear() calls AbstractRepositoryConnection.remove(). 

However, SailRepository is the correct level here since this is the point at which the Rep API operation is "translated" to a SAIL operation. Since the MemoryStore (and any other store) is a SAIL implementation, this will work for any RDF4J-compatible backend.

jeenbroekstra:
I have confirmed that with the above fixes the problem in workbench's clear operation no longer occurs.

Issue #239, opened 2016-07-16
Missing unit tests for spin:functions


jgrzebyta:
I am working on simple unit tests for spin functions.

pulquero:
Note there are some existing ones for some functions in rdf4j-sail-spin/org.eclipse.rdf4j.sail.spin.

jeenbroekstra:
Current merge has caused a test failure on hudson. I am re-opening this issue and reverting the merge.

jeenbroekstra:
See https://hudson.eclipse.org/rdf4j/job/rdf4j-2.0-verify/43/ for details on test failure.

jgrzebyta:
This issue was not finish yet. I will put new PR for that.

Issue #220, opened 2016-07-04
search query evaluator is not able to get query string from Literal


jgrzebyta:
Affected: rdf4j version _2.0M3_
Java: 1.8
Os: Ubuntu 16.04

In Lucene search query only direct _String_ is accepted, ie: `search:query "abc123"` works fine.
In case if the term is instance of Literal taken as result of other _atomic_ request it fails. For example:

```
prefix search: <http://www.openrdf.org/contrib/lucenesail#>
prefix y: <urn:raw:yeastract#>

select ?sub ?query (isLiteral(?query) as ?isLit) ?score where {
  [] y:Yeast_id ?query .
  ?sub search:matches [
  search:query ?query ;
  search:score ?score
  ] .
}
```

I have error:

```
Exception in thread "main" org.eclipse.rdf4j.query.QueryEvaluationException: Invalid Text Query: missing query string for Lucene query specification
    at org.eclipse.rdf4j.repository.sail.SailTupleQuery.evaluate(SailTupleQuery.java:59)
    at org.eclipse.rdf4j.repository.sail.SailTupleQuery.evaluate(SailTupleQuery.java:67)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:93)
    at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28)
    at sparql$process_sparql_query.invokeStatic(sparql.clj:92)
    at sparql$process_sparql_query.invoke(sparql.clj:83)
    at sparql$_main$fn__652.invoke(sparql.clj:68)
    at sparql$_main.invokeStatic(sparql.clj:67)
    at sparql$_main.doInvoke(sparql.clj:47)
    at clojure.lang.RestFn.applyTo(RestFn.java:137)
    at sparql.main(Unknown Source)
Caused by: org.eclipse.rdf4j.sail.SailException: Invalid Text Query: missing query string for Lucene query specification
    at org.eclipse.rdf4j.sail.lucene.QuerySpecBuilder.failOrWarn(QuerySpecBuilder.java:203)
[...]

PR #240, opened 2016-07-16
Added unit testing class for spin functions

jgrzebyta:
This PR addresses GitHub issue: #239 

- create class for testing different functions within SPARQL requests
- add SPARQL exmaple for **`apf:strSplit`** and **apf:concat**

- [x] tests are included
- [x] all tests succeed


pulquero:
This should probably be called ApfSailTest to match the naming convention used by SpifSailTest/SplSailTest and go in the sail-spin module.

jeenbroekstra:
I think there is some confusion here - these tests verify if the SPARQL engine accepts custom property functions. But this has not been implemented. Currently, custom property functions are _only_ available in the SPIN reasoner (which is a separate component from the SPARQL engine, although it does of course reuse some functionality). 

We'd need to extend the SPARQL engine's query parser to support this.

pulquero:
I took that as a mis-understanding of how they work - that they need the presence of the SpinSail, the memory sail alone isnt enough. In fact, you don't even need that. A simple sailwrapper that runs SpinMagicPropertyInterpreter() in evaluate() and wraps the evaluationstrategy in TupleFunctionEvaluationStrategy would be enough too.

jgrzebyta:
@pulquero and @jeenbroekstra Thanks a lot for your help.
I have relocated the unit tests into `sail-spin` module. Also added `SpinSail` as wrappef for the `MemoryStore` to activate the spin feature. The `SpinFunctionTest` might be used as the example of usage.

core/sail/spin/src/test/java/org/eclipse/rdf4j/spin/functions/SpinFunctionTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The property functionsunit tester. How does it work?</br>
+ * <ul>
+ * <li>The functions should be embedded into a SPARQL query.</li>
+ * <li>The query might be any type: graph, tuple or boolean.</li>
+ * <li>The file with sparql query should be written in <code>src/test/resources/</code> directory.</li>
+ * <li>The file pathway should be added into static FINCTIONS variable as presented.</li>
```
pulquero: spelling mistake.

core/sail/spin/src/test/resources/functions/apf/concat.sparql
```diff
@@ -0,0 +1,5 @@
+prefix apf: <http://jena.hpl.hp.com/ARQ/property#>
```
pulquero: Conventionally file extension is .rq for sparql queries. I think it is mentioned in the sparql federation sparql if you are interested. Is it possible to have .srx files to check the results against? Check out rdf4j/testsuites/sparql/src/main/resources/testcases-sparql-1.1-w3c/functions to see how it can be done.

pulquero:
Should SpifSailTest be done in the same style? Should SpinFunctionTest handle both spif and apf?

core/sail/spin/src/test/resources/functions/apf/concat.sparql
jgrzebyta: On 18 July 2016 at 23:10, pulquero <notifications@github.com> wrote:

> In core/sail/spin/src/test/resources/functions/apf/concat.sparql
> <https://github.com/eclipse/rdf4j/pull/240#discussion_r71240354>:
>
> > @@ -0,0 +1,5 @@
> > +prefix apf: <http://jena.hpl.hp.com/ARQ/property#>
>
> Conventionally file extension is .rq for sparql queries. I think it is
> mentioned in the sparql federation sparql if you are interested.
>
You are right with it. I changed. Sorry.


> Is it possible to have .srx files to check the results against? Check out
> rdf4j/testsuites/sparql/src/main/resources/testcases-sparql-1.1-w3c/functions
> to see how it can be done.
>
I will try to prepare. How are they compared with the results (I mean about
the pretty printing)?

pulquero: You should be able to use rdf4j-rio to load .srx as a list of bindings and then compare with the query results.

jeenbroekstra: You can also use `QueryResultIO.parseTuple()`, which is a static method giving you a `TupleQueryResult` object. You can then use `QueryResults.equals` or `QueryResults.isSubset` to compare your expected and actual results.

core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
```diff
@@ -72,7 +76,13 @@ public void tearDown()
 		}
 	}
 
+	/**
+	 * Probably that test is damaged because there is no spif.ttl file.
```
pulquero: spif.ttl is on the classpath via rdf4j-spin.

jgrzebyta:
@pulquero  So finally I have read SPIN ontology description and a lot of things I found clarified, eg `spinn:Function`. It seems there are much more issues I have expected in the beginning, eg syntax of function is different than syntax of property functions.

I will finish this PR with setting up of the unit tests and further I will solve API issues in #220.

jeenbroekstra:
I have only kept half an eye on this PR. I am currently operating under the assumption that this is a work in progress and not yet ready for merge. Is that correct or do you feel it's more or less complete?

jgrzebyta:
IMHO it is completed. I replaced `SpifSailTest` by tests run by `SpinFunctionTest`. I guess there is no need to rewrite all tests as the only those two base on comparison sparql request with the expected result.

core/sail/spin/src/test/java/org/eclipse/rdf4j/spin/functions/SpinFunctionTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+	public static final String SPARQL_RES_EXT = ".srx";
+
+	@Parameters
+	public static Collection<Object[]> loadSparqls()
+		throws Exception
+	{
+		List<Object[]> sparqls = new ArrayList<>();
+
+		List<Path> functions = scanTests(Paths.get(TESTS_ROOT));
```
ansell: This should be done using a manifest, (see the SPARQL/RDF compliance testsuites for examples) rather than a file scan.

core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
```diff
@@ -73,6 +77,7 @@ public void tearDown()
 	}
 
 	@Test
+	@Ignore
```
ansell: This method that uses a manifest should have been kept and extended rather than replaced.

jeenbroekstra:
I have reverted this PR for now. @jgrzebyta please have a look at the test failures (see comments on #239 ) and @ansell 's comments above. Once you have corrected the problems, please feel free to put up a new Pull  Request.

Issue #228, opened 2016-07-11
Send single Accept header with comma-separated values


jeenbroekstra:
(migrated from https://openrdf.atlassian.net/browse/SES-2383 ) 

RDF4J's SPARQL client currently sends multiple Accept headers with its query requests, to indicate which response formats it can process. For example:
```

 Accept: text/csv;q=0.8
 Accept: application/sparql-results+xml 
```

While this is [technically correct](http://i1.kym-cdn.com/photos/images/original/000/909/991/48c.jpg) (see [section 4 in RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) for details), it appears many SPARQL endpoint implementations have trouble processing request headers in this form. E.g. in the above example, they will choose to use text/csv (even if they do support application/sparql-results+xml).

To facilate better interoperability, we should instead compose a single accept header, with the supported formats as comma-separated values:

`Accept: text/csv;q=0.8, application/sparql-results+xml`

nikolavp:
+1 on this one. We are using wiremock and it also had problems with multiple accept headers - it only takes into account the first instance.

Although this is not a problem with sesame but with naive implementations out there it will also make the request more compact.

https://github.com/tomakehurst/wiremock/issues/334

PR #242, opened 2016-07-17
create single Accept header with comma-separated values for formats

jeenbroekstra:
This PR addresses GitHub issue: #228

- http query requests now send single, comma-separated Accept header 

- [ ] tests are included
- [x] all tests succeed

Signed-off-by: Jeen Broekstra <jeen.broekstra@gmail.com>

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
@@ -97,6 +97,8 @@ import org.eclipse.rdf4j.rio.helpers.ParseErrorLogger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Joiner;
+
 /**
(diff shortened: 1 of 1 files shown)


ansell:
I am not familiar with the Joiner class, but it looks good to me if that class does what join mechanisms do in other frameworks.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
```diff
@@ -988,10 +994,12 @@ private HttpResponse sendBooleanQueryViaHttp(HttpUriRequest method, Set<QueryRes
 					acceptParam += ";q=0." + qValue;
 				}
 
-				method.addHeader(ACCEPT_PARAM_NAME, acceptParam);
+				acceptValues.add(acceptParam);
 			}
 		}
 
+		method.addHeader(ACCEPT_PARAM_NAME, Joiner.on(", ").join(acceptValues));
```
nikolavp: Joiner is thread safe, we should probably extract that as a static variable in the class. Same for the joiner above.

Issue #163, opened 2016-05-25
Allows to compare literals with different date/time datatypes in SPARQL


Tpt:
It is not possible to compare literals with different date/time datatypes in SPARQL, e.g. "2005"^^xsd:gYear and "2015-01"^^xsd:gYearMonth.

A solution is to consider these literals as if they where xsd:dateTime and use xsd:dateTime comparison. So, for example, we would have "2005"^^xsd:gYear < "2015-01"^^xsd:gYearMonth = "true"^^xsd:boolean

SPARQL 1.1 specification does not talk about other date/time datatypes than xsd:dateTime so it would be compatible with the specification.

The comparison when a date/time is a prefix of an other like, "2005"^^xsd:gYear and "2005-01"^^xsd:gYearMonth, should be undefined.

jeenbroekstra:
Looks sensible, but I'll need to have a quick look at the spec again to be absolutely sure - there's things that trip one up in the SPARQL spec when it comes to datatypes :)

Tpt:
We should probably also have "2005"^^xsd:gYear != "2005-01"^^xsd:gYearMonth = "true"^^xsd:boolean and  "2005"^^xsd:gYear = "2005-01"^^xsd:gYearMonth = "false"^^xsd:boolean . The undefined behaviour is just for the <, >, <= and >= operators.

jeenbroekstra:
You can extend the behavior of <, >, <= and >= (because they currently return a type error when you use an unknown datatype),  but for = and !=  you have to be a little bit careful. The [definition of RDFTerm-equal in the SPARQL spec](https://www.w3.org/TR/sparql11-query/#func-RDFterm-equal) is _incredibly_ convoluted. But from memory and from what I can see in the spec your proposal is sound, so go ahead and put it up for testing.

ansell:
Personally I wouldn't expect 2005-01^^xsd:year to be automatically ordered before 2005^^xsd:year. In some ordering schemes the less specific cases may need to be ordered before the more specific cases.

Adding this behaviour would best be done with a flag that enables comparisons of different date types with the proprietary algorithm that this would implement, if it isn't already allowed for in the SPARQL-1.1 spec.

jeenbroekstra:
I agree. More generally I think there should a be config option to run queries in 'strict' mode or 'extended' mode. See #164

PR #247, opened 2016-07-18
Issue #164: configurable eval strategy

pulquero:

- ...
- ...
- ...

- [x] all tests succeed

Signed-off-by: Mark Hale <mark.hale@physics.org>


pulquero:
*sigh* yet another git disaster. Anyway, I managed to rescue it.

pulquero:
Previous comments and replies here: https://github.com/eclipse/rdf4j/pull/224

jeenbroekstra:
@pulquero is there any way I can help you prevent these git problems occurring? I'm assuming that the root cause is that you occasionally forget to check you're working on the correct branch and/or sync with the master repo?

pulquero:
I think I'm just cursed or maybe it is old age creeping up on me. I've established one thing, I prefer bitbucket to git. So, yes, I sometimes branch from the wrong branch or merge in the wrong branch. Maybe I should include the target branch in the branch name see if that helps. This time though, eclipse lost my author setting and I autocompleted with the wrong email address. That mixed up with merging in the wrong target branch at some stage maybe it difficult to sort out the history to fix ip validation. Anyway, reflog and cherry picking to the rescue.

jeenbroekstra:
Fair enough :) If it might help, we could adapt our workflow to use the source branch as part of the branch naming. Currently we prefix with 'issues/' only. We could adopt something more elaborate, e.g. 'hotfixes/' (for issues on the release branch) and 'features/' for any enhancements on the main dev branch. Anyway, getting offtopic here, let's continue this on the mailinglist.

jeenbroekstra:
There's a merge conflict with this PR, can you take a look at this @pulquero ?

Other than that: due to the long timeframe (and the multiple PRs) I kinda lost track where we are with this. I had a quick glance at the code and it looks fine to me. Do you consider this stable and ready for release as-is?

Also: given that we want to use this mechanism to disable/enable extended features such as extended datetime comparison (see #163 ), can you give some short pointers on what should happen in the PR for that issue to get it to work in extended mode but not in strict?

pulquero:
Yep, I consider it ready. So for #163, I imagine there would be an extended eval strategy subclass of simple eval strategy which overrides the relevant methods (I think most of them are private atm). There would then be an associated extended eval strat factory which can be set on a sail to use it. It can also be added to the create-*.xsl to allow it to be chosen in workbench.

jeenbroekstra:
Running hudson: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/28/

jeenbroekstra:
Hudson succeeds.

PR #256, opened 2016-07-20
Fix #255 Lower MAXIMUM_URL_LENGTH

### LINKED ISSUE #255: SPARQLRepository SparqlSession should have a configurable MAXIMUM_URL_LENGTH
This is another issue that appears to have been lost since the migration to Eclipse.

I've pasted the text of the original text of the issue below.  The issue still appears present in both the RDF4j 1.0 and 2.0 branches.

I've just tried upgrading one of our applications from Sesame 2.7.16 to Sesame 2.9.0 and have run into a problem caused by the change SES-2039. Jeen rightly predicted that this change might break backward compatability for some, as it turns out that the Stardog store (at least versions 3.1.4 and 4.0.0-4.0.3) has a default URL limit of 4084 characters (4096 if you also include the prefix string "GET " and the suffix " HTTP 1.1" from the HTTP requests request line).

Not being able to set this means that queries via SPARQLRepository that serialise to 4084 characters or more fail on the Stardog store.

I'm also pursuing the possibility of configuring that limit with Stardog, but it has occurred to me that making the limits on both sides configurable is probably a good idea; because results of GET requests are potentially cacheable by proxies etc.

So when you have the luxury of controlling both the client and the server (as we often do) it might desirable to set this value to the same high value on both, as it opens up the possibility of clients and servers to one day implement better result caching via etags etc...

It seems that Stardog use Netty internal, and inherit their default setting from here:

- http://netty.io/4.0/api/index.html?io/netty/handler/codec/http/HttpRequestDecoder.html

I'm not sure whether any other triple stores use Netty internally, but if they do they probably also inherit the default of 4083 characters for the URL length limit.  The popular Jetty web server also ships with a 4kb default for the first line of the request.

I think it might also be worth lowering the default in Sesame for this value to 4083, unless we know of any triple stores with an even lower limit?

RickMoynihan:
This PR addresses GitHub issue: #255 

Improves compatibility with SPARQL endpoints by lowering thresh-hold for when a GET request should become a POST

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
@@ -123,7 +123,7 @@ public class SparqlSession implements HttpClientDependent {
 	 * The threshold for URL length, beyond which we use the POST method based on the lowest common
 	 * denominator for various web servers
 	 */
-	public static final int MAXIMUM_URL_LENGTH = 8192;
+	public static final int MAXIMUM_URL_LENGTH = 4083;
 
 	final Logger logger = LoggerFactory.getLogger(this.getClass());
 


jeenbroekstra:
Thanks for submitting Rick, but shouldn't this be part of the 2.0 release branch instead? Otherwise it won't be included in the upcoming release but will have to wait for 2.1.... 

Also: the IP validation script yells that checks have failed. This is because of a bug in that script (it doesn't recognize noreply@github.com as a committer on merges), but it means that I can't verify that you yourself have signed the Eclipse CLA. Can you confirm that you did?

RickMoynihan:
Hi Jeen,

Yes I can confirm that I've signed the CLA.  I noticed the IP check failed too, but wasn't sure how things are setup.  I confess I made this change through the github UI, not through a local commit, so I suspect that might be the cause.  

I had a branch from sesame which 2.8.9 which has the same change, but that won't merge with RDF4j because of the rewritten histories.

Sorry about putting it on the wrong branch, it would be great if we could merge it so it ends up in 2.0 and 1.0, if that's still possible - though it might be some time till we can upgrade things from sesame to RDF4j.

Thanks again.

jeenbroekstra:
No worries. I'll just close this PR and redo the fix on the release branch. It's trivial enough.  

Tbh if we were strict about our workflow this should be a separate issue from #255  (which is about making the value configurable, rather than setting a lower default). But we'll cheat a bit :)

PR #257, opened 2016-07-21
SPARQL endpoint url length threshold lowered and made configurable

jeenbroekstra:
This PR addresses GitHub issue: #255

- default value of url length threshold set to 4083
- threshold now configurable via system property 'rdf4j.sparql.url.maxlength'

- [ ] tests are included
- [x] all tests succeed

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
@@ -120,10 +120,20 @@ public class SparqlSession implements HttpClientDependent {
 	protected static final Charset UTF8 = Charset.forName("UTF-8");
 
 	/**
-	 * The threshold for URL length, beyond which we use the POST method based on the lowest common
-	 * denominator for various web servers
+	 * The default threshold for URL length, beyond which we use the POST method. The default is based on the
+	 * lowest common denominator for various web servers.
 	 */
-	public static final int MAXIMUM_URL_LENGTH = 8192;
+	public static final int DEFAULT_MAXIMUM_URL_LENGTH = 4083;
+
+	/**
+	 * System property for configuration of URL length threshold: rdf4j.sparql.url.maxlength
+	 */
+	public static final String MAXIMUM_URL_LENGTH_PARAM = "rdf4j.sparql.url.maxlength";
+
+	/**
+	 * The default threshold for URL length, beyond which we use the POST method.
+	 */
+	private final int maximumUrlLength;
 
 	final Logger logger = LoggerFactory.getLogger(this.getClass());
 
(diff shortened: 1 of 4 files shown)


jeenbroekstra:
@RickMoynihan could use your feedback on this. Also added configuration option while I was at it.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
```diff
@@ -172,6 +188,24 @@ public SparqlSession(HttpClient client, ExecutorService executor) {
...
+			catch (NumberFormatException e) {
+				throw new RDF4JException("integer value expected for property " + MAXIMUM_URL_LENGTH_PARAM,
+						e)
+				{
+
+					private static final long serialVersionUID = 1L;
+				};
+			}
+		}
+		this.maximumUrlLength = maximumUrlLength;
```
ansell: Could add a check that it is a non-negative value here, just for sanity interpreting it later

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
```diff
@@ -172,6 +188,24 @@ public SparqlSession(HttpClient client, ExecutorService executor) {
...
+		if (propertyValue != null) {
+			try {
+				maximumUrlLength = Integer.parseInt(propertyValue);
+			}
+			catch (NumberFormatException e) {
+				throw new RDF4JException("integer value expected for property " + MAXIMUM_URL_LENGTH_PARAM,
+						e)
+				{
+
+					private static final long serialVersionUID = 1L;
```
ansell: This seems to be an issue for the RDF4JException class, not here.

jeenbroekstra: It needs to happen this way because RDF4JException is abstract: this code creates a concrete anonymous subclass (which of course needs its own serialization version number). Admittedly a bit cavalier to create an exception in this way but the problem is not in RDF4JException itself (unless you consider it being abstract a problem, of course :))

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
jeenbroekstra: We could also just say that a negative or 0 value means POST is always used.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
jeenbroekstra: The more I think on it, the more I like it: should we make RDF4JException concrete so we can just use it directly instead of always having to subclass? 

Given that it's a new class in RDF4J and not yet released, I can even get away with just fixing that as part of this PR :)

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
ansell: Either way works. Restricting it to non-negative would mean that 0 would be the only value where POST is always guaranteed to be used but negatives are also used as sentinels in various programs so they work too.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SparqlSession.java
ansell: Having it as an abstract class encourages good exception heirarchy design, but shows up in cases like this where we don't already have an "RDF4JConfigException" (or do we?)

jeenbroekstra: We don't have a generic one. There's RepositoryConfigException and SailConfigException. 

We could introduce a generic RDF4JConfigException, and make the above two subclasses of that. It wouldn't break existing code if we did this AFAICT.

ansell: I think combining those and any other config exception cases in the future under a single exception would be useful, and it shouldn't break anything if they are both direct sub-classes of RDF4JException right now.

RickMoynihan:
Yet to try it - but this looks good @jeenbroekstra - thanks for also making it configurable via a system property.

Issue #264, opened 2016-07-26
TurtleParser does not report line numbers / location in file


rwm:
The class `org.eclipse.rdf4j.rio.turtle.TurtleParser` does not report location info while parsing. Only during start of parsing (line number 1) and after processing a comment (e.g. line that starts with '#').

Instead of during processing comments, the `reportLocation()` method should be called after incrementing the line number  in the method `skipWSC()`.

I also suggest changing the visibility of the method `getLineNumber()` to protected (instead of private). This would allow subclasses to access line number information directly, instead of relying on reportLocation to do the job.

ansell:
Those changes sound good to me, they are proposed in pull request #264 for review.

rwm:
Upon review of the changes, I think I found another bug with the line counting in TurtleParser:
The method `processComment()` does on increase the lineNumber, even though it may/will read a line feed.
To fix this, the `lineNumber` should be incremented in `processComment()` after having read `0x0D`.

Maybe a unit test which verifies the correct line count after reading an example file will help..

ansell:
hi @rwm, I will look into processComment then create a regression test for it and add it to the pull request

ansell:
I added regression tests and a fix for line number counting within comments to the pull request

ansell:
This can be closed as the pull request has been merged.

PR #265, opened 2016-07-26
Issues/#264 turtle line numbers

ansell:
This PR addresses GitHub issue: #264

- TurtleParser does not report when lines tick over to the location listener
- TurtleParser does not increment line numbers for lines with comments
- TurtleParser does not allow subclasses, of which we have two internally and external users have more, to get the lineNumber information without using reportLocation that is fairly roundabout

- [x] tests are included
- [x] all tests succeed

### DIFF
--- a/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/SimpleParseLocationListener.java
+++ b/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/SimpleParseLocationListener.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright (c) 2015 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.eclipse.rdf4j.rio.helpers;
+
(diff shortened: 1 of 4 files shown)


core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/SimpleParseLocationListener.java
```diff
@@ -0,0 +1,44 @@
...
+package org.eclipse.rdf4j.rio.helpers;
+
+import org.eclipse.rdf4j.rio.ParseLocationListener;
+
+/**
+ * A simple implementation of {@link ParseLocationListener}.
+ * 
+ * @author Peter Ansell
+ */
+public class SimpleParseLocationListener implements ParseLocationListener {
```
jeenbroekstra: To what purpose would you expect to put this implementation, other than unit testing?

ansell: I have never used ParseLocationListener in the past and don't expect to do so in the future. 

The reasoning for introducing an implementation is simply that we have an interface that we do not plan to deprecate or remove, but we have no local implementation of it, particularly as it is this simple to implement and doesn't add any maintenance costs to the project to do so.

jeenbroekstra:
All looks good. You finished / can I merge?

ansell:
Yes, I am happy with it being merged.

Issue #280, opened 2016-08-09
spif:Test-dateFormat-1 test in module rdf4j-spin-sail fails when using the Italian locale


manuelfiorelli:
When compiling the 2.0.x branch, the compilation fails at the module `rdf4j-spin-sail` because of a test failure.

The code below is an actual test report.
```
Running org.eclipse.rdf4j.sail.spin.SpifSailTest
Tests run: 13, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 48.194 sec <<< FAILURE! - in org.eclipse.rdf4j.sail.spin.SpifSailTest
runTests(org.eclipse.rdf4j.sail.spin.SpifSailTest)  Time elapsed: 4.219 sec  <<< FAILURE!
java.lang.AssertionError: http://spinrdf.org/spif#Test-dateFormat-1 expected:<"February 13, 2008"^^<http://www.w3.org/2001/XMLSchema#string>> but was:<"febbraio 13, 2008"^^<http://www.w3.org/2001/XMLSchema#string>>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.eclipse.rdf4j.sail.spin.SpifSailTest.runTests(SpifSailTest.java:94)
```

The actual value contains the word "febbraio", which is the Italian word for February. 

The code below should be the failing test case.

```
spif:Test-dateFormat-1
  rdf:type spl:TestCase ;
  spl:testExpression [
      rdf:type spif:dateFormat ;
      arg:date "2008-02-13"^^xsd:date ;
      arg:pattern "MMMMMMMMM dd, yyyy" ;
    ] ;
  spl:testResult "febbraio 13, 2008"^^xsd:string ;
  rdfs:label "Test-date format"^^xsd:string ;
.
```

Indeed, it requires that the structured date `"2008-02-13"^^xsd:date` is formatted according to the template `"MMMMMMMMM dd, yyyy"`.

The implementation of `spif:dateFormat` seems to use the default locale, which explains the observed behavior.

If the implementation is right, then the tests should be updated to avoid the failure.

jeenbroekstra:
[...]

PR #284, opened 2016-08-10
workarounds in unit tests to account for different platform locales.

jeenbroekstra:
This PR addresses GitHub issue: #280

- SPIN unit tests now force English locale for test execution
- TriX and RDF/XML parser tests do not verify exact error message anymore.

- [x] tests are included
- [x] all tests succeed

### DIFF
--- a/core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
+++ b/core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
@@ -48,7 +48,11 @@ public class SpifSailTest {
 
 	private RepositoryConnection conn;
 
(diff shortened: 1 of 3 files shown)


core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
```diff
@@ -47,6 +48,8 @@
 
 	private RepositoryConnection conn;
 
+	private Locale locale;
```
ansell: If it is just named "locale" it looks like a test variable rather than just a storage position for something unrelated to the test itself. Naming this "oldLocale" or "previousLocale" may be easier to remember its purpose in future, but otherwise it all looks good.

Issue #262, opened 2016-07-25
Workbench does not support setting the empty prefix


manuelfiorelli:
Inside the Workbench, binding the empty prefix to a namespace causes the definition of a prefix named "namespaces".

This behavior may be determined by a bug in the server-side implementation the [REST API for setting the namespace for prefixes](http://rdf4j.org/doc/the-rdf4j-server-rest-api/#Set_the_namespace_for_a_specific_prefix)

To set the empty prefix, the relevant URL has this form: ``http://....../rdf4j-server/repositories/mem-rdf/namespaces/``

(Note that this URL differs from the one for bulk operations on namespaces, which does not end with a trailing slash)

The buggy behavior may be determined by the following piece of code:

https://github.com/eclipse/rdf4j/blob/releases/2.0.x/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java#L54

```
String pathInfoStr = request.getPathInfo();
String[] pathInfo = pathInfoStr.substring(1).split("/");
String prefix = pathInfo[pathInfo.length - 1];
```

Based on the documentation of [String.split(String)](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#split-java.lang.String-),  in case of a path such as ".../mem-rdf/namespaces/", the last element in the array would be "namespaces" (because trailing empty strings are ignored)

manuelfiorelli:
Hi @jeenbroekstra, I've noticed that you assigned this issue to yourself. In fact, I've developed a solution to this issue together with an associated test. However, I didn't make a PR because of the issue #280, which prevents me from compiling the whole project, and thus execute all the tests.

If you haven't already solved this issue by yourself, I could make a pull request by tomorrow.

jeenbroekstra:
@manuelfiorelli ok that'd be great, saves me some time :) 

I'll have a quick look at #280 as well to see if I can disable it for the time being.

manuelfiorelli:
Hi @jeenbroekstra I've pushed my proposed solution to a dedicated branch in my fork of the rdf4j project.

I can compile the whole project (after disabling the tests), and then the relevant tests for the issue pass.

(Also, I've manually checked that now the workbench works)

Do you think that this is enough to make a PR?

jeenbroekstra:
Sounds good to me - just put it up and we can take a look.

manuelfiorelli:
I've created the PR #285

jeenbroekstra:
[...]

PR #285, opened 2016-08-10
issue #262: add a test and fix the issue in the HTTP server with empt…

manuelfiorelli:
This PR addresses GitHub issue: #262 

- Use of String.lastIndexOf to find the index of the last slash "/" in the request path, in order to determine where the prefix starts

As the original code, the proposed modification accepts URLs that are not strictly conforming to the specified request format: e.g. any number of intermediate components are silently ignored.

- [x] tests are included
- [x] all tests succeed (in fact, because of a problem with locale-dependent tests, I was not able to run all tests, but I've executed the tests for the affected module)

### DIFF
--- a/compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
+++ b/compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
@@ -9,6 +9,7 @@ package org.eclipse.rdf4j.http.server;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.io.ByteArrayInputStream;
@@ -18,8 +19,11 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
@@ -37,7 +41,9 @@ import org.apache.http.impl.client.HttpClients;
 import org.apache.http.message.BasicNameValuePair;
 import org.eclipse.rdf4j.common.io.IOUtil;
 import org.eclipse.rdf4j.http.protocol.Protocol;
+import org.eclipse.rdf4j.model.Namespace;
 import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleNamespace;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.query.QueryLanguage;
 import org.eclipse.rdf4j.query.TupleQueryResult;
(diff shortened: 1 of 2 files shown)


compliance/http/src/test/java/org/eclipse/rdf4j/http/server/ProtocolTest.java
```diff
@@ -482,6 +516,62 @@ public void run() {
...
+			try {
+				int responseCode = conn.getResponseCode();
+				if (responseCode != HttpURLConnection.HTTP_OK) {
+					String response = "location " + location + " responded: " + conn.getResponseMessage()
+							+ " (" + responseCode + ")";
+					fail(response);
+				}
+
+				String responseText = IOUtil.readString(connIs);
+				String[] rows = responseText.split("\\r?\\n");
```
ansell: As we already have opencsv on the classpath at this point, inherited from queryresultio-text, it may be useful to rewrite the (very simple in this case) parsing code using opencsv, but that shouldn't stop this pull request from being accepted.

ansell: Looking at it further using 

```
mvn dependency:tree
```

You may need to add a runtime/test dependency to opencsv and/or queryresultio-text to get my suggestion to work as http-compliance only seems to link to queryresultio-binary, queryresultio-sparqljson and queryresultio-sparqlxml right now.

manuelfiorelli: Hi @ansell 

I decided to use CSV because it is relatively easily to parse, but in case I can use an existing parser I could opt to use another format, maybe the XML serialization, in order to avoid the introduction of another dependency.

I think that I could work on this modification tomorrow or the day after tomorrow.

ansell: There are no new RDF4J dependencies needed. You may just need to copy one, opencsv, to rdf4j-http-compliance that we already have in other modules. 

It shouldn't need to block this pull request though so don't feel you need to do it before getting this accepted.

manuelfiorelli: @ansell Just to be sure. Was your suggestion to replace my code for parsing the CSV response with the use of the opencsv library? At first, I thought that I should use the class `org.eclipse.rdf4j.query.resultio.text.csv.SPARQLResultsCSVParser`. However, now I think that direct use of opencsv could be better.

jeenbroekstra: Pretty sure that was what @ansell  meant, yes. If you want to do this go ahead, but as said don't feel obligated, it's icing on the cake but not exactly _necessary_ for such a simple use case.

jeenbroekstra:
Approved from me. @manuelfiorelli I will merge this PR. If you want to do further improvements wrt CSV parsing you can open a new PR for that.

manuelfiorelli:
@jeenbroekstra I don't know if I will make a PR for the "icing". However, I am interested to known what should I do. Should I create a new issue and branch?

jeenbroekstra:
No, that wouldn't be necessary. Essentially it's still part of the same issue, so even after this PR is merged, it's fine to keep working on the existing issue, and using the same branch. You will just need to put up a new PR for those further changes.

Btw I merged your PR because I wanted to make sure your original fix was included for the 2.0 release. If you work on the csv parsing bit like @ansell proposed there's always a chance your further fixes break something. By merging now I made sure that we can go ahead with release with the fix included, regardless of whether you improve the csv parsing.

Issue #286, opened 2016-08-11
Explore ways to add error handlers to queryresultio-api


ansell:
queryresultio-api doesn't currently offer the same error handling that rio-api does, which resulted in me adding three fixme notes to the SPARQL/XML parser in my last pull request: 

https://github.com/eclipse/rdf4j/commit/70fde6f5fd39bede881e8a2db0215c7d777f6d20#diff-433a3d547f906a56dc8d804323a08684R166

It would be ideal if this could be done without adding any new interfaces/etc. by sitting queryresultio-api on top of rio-api, but it may be necessary to add a new module, say "rdf4j-parser-api" to get this to work transparently without screwing with the way everything is currently working.

ansell:
Related to this is the lack of ways to configure the parsers like ParserConfig, so the error handling would likely be done in conjunction with adding parser configurability to queryresultio.

PR #282, opened 2016-08-10
issue#281 : Further use of ErrorHandler to prevent fatal errors emitted to console

ansell:
This PR addresses GitHub issue: #281 

Briefly describe the changes proposed in this PR:

- Ensure that all XMLReader instances have ErrorHandler objects attached to them to prevent fatal errors being emitted to the Console 
- Adds Locator to SimpleSaxParser so that error messages and exceptions for parsers using it can have line and column numbers attached to them.
- Adds regression test for TriXParser to ensure it doesn't start emitting error messages to the System.err or System.out in future

Make sure you've followed the [Contributor Guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md). In particular (please tick to indicate you've taken care of it):

- [x] RDF4J code formatting has been applied
- [x] tests are included
- [x] all tests succeed

Note that queryresultio doesn't have any formal infrastructure for reporting errors to users in a friendly fashion as rio does with ParseErrorListener, so only supporting fatal error exception rethrowing for SAXExceptions passing through the SPARQL/XML parser at this stage.

Signed-off-by: Peter Ansell <p_ansell@yahoo.com>

PR #287, opened 2016-08-11
Issues/#286 queryresultio error handling

ansell:
This PR addresses GitHub issue: #286 

- Adds capability to receive error messages from QueryResultParser's
- Resolves the FIXME notes that I added to the SPARQL/XML parser in pull request #282 so non-fatal SAX exceptions are not swallowed by a noop ErrorHandler
- Ports the fluent configuration style from RDFParser to QueryResultParser
- Fixes entity expansion limit security issue for SPARQL/XML in the same way as RDF/XML
- Fixes use of int instead of long for line/column numbers in QueryResultParseException, to match RDFParseException API

- [x] tests are included
- [x] all tests succeed

Even though this is a large set of changes, there are several API changes that would need to be done before the first 2.0 release or else wait until 3.0 due to not being able to make them backwards compatible, so targeting at 2.0 rather than 2.1.

This builds on the changes that were done to bring Tuple and Boolean parsing together in the past and hopefully makes it simpler in future to pull together the QueryResultParser and RDFParser interfaces in the future to reduce code duplication.

### DIFF
--- a/compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqlxml/SPARQLXMLParserCustomTest.java
+++ b/compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqlxml/SPARQLXMLParserCustomTest.java
@@ -1,0 +1,162 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.query.resultio.sparqlxml;
+
+import static org.junit.Assert.*;
+
+import org.eclipse.rdf4j.query.resultio.QueryResultIO;
+import org.eclipse.rdf4j.query.resultio.QueryResultParseException;
+import org.eclipse.rdf4j.query.resultio.QueryResultParser;
+import org.eclipse.rdf4j.query.resultio.TupleQueryResultFormat;
+import org.eclipse.rdf4j.query.resultio.helpers.QueryResultCollector;
+import org.eclipse.rdf4j.rio.ParserConfig;
+import org.eclipse.rdf4j.rio.helpers.ParseErrorCollector;
+import org.eclipse.rdf4j.rio.helpers.XMLParserSettings;
+import org.junit.Ignore;
+import org.junit.Test;
+
+/**
+ * Custom tests for SPARQL/XML Parser.
+ * 
+ * @author Michael Grove
+ * @author Peter Ansell
+ */
+public class SPARQLXMLParserCustomTest {
+
+    /**
+     * Test with the default ParserConfig settings. Ie, setParserConfig is not
+     * called.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testEntityExpansionDefaultSettings() throws Exception {
+        QueryResultCollector handler = new QueryResultCollector();
+        ParseErrorCollector errorCollector = new ParseErrorCollector();
+        QueryResultParser aParser = QueryResultIO.createTupleParser(TupleQueryResultFormat.SPARQL)
+                .setQueryResultHandler(handler).setParseErrorListener(errorCollector);
+
+        try {
+            // this should trigger a SAX parse exception that will blow up at
+            // the 64k entity limit rather than OOMing
+            aParser.parseQueryResult(this.getClass()
+                    .getResourceAsStream("/sparqlxml/bad-entity-expansion-limit.srx"));
+            fail("Parser did not throw an exception");
+        } catch (QueryResultParseException e) {
+            // assertTrue(e.getMessage().contains(
+            // "The parser has encountered more than \"64,000\" entity
+            // expansions in this document; this is the limit imposed by the
+            // "));
+        }
+        assertEquals(0, errorCollector.getWarnings().size());
+        assertEquals(0, errorCollector.getErrors().size());
+        assertEquals(1, errorCollector.getFatalErrors().size());
+    }
+
+    /**
+     * Test with unrelated ParserConfig settings
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testEntityExpansionUnrelatedSettings() throws Exception {
+        ParserConfig config = new ParserConfig();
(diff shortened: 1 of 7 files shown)


jeenbroekstra:
Looks like a good set of improvements but to be honest I'm a bit reluctant to include this in 2.0, for two reasons:

1. release is scheduled for the 17th, this is cutting it very close to introduce such a big new feature;
2. the original idea behind 2.0 was to stay "as compatible as possible" with Sesame 4. 

The second point is not etched in stone though. In terms of compatibility with user code, can you sketch some quick examples of what kind of user code is likely to need changes? The only thing I can spot is the change from int to long. If that is the only real incompatibility I can live with it as a late addition to 2.0. 

Note, by the way, that if we move this to the master branch, "waiting for a 3.0 release" will not take ages. You introduce this change, therefore the next release will be 3.0 - simple as that.

ansell:
The goal of the API changes are to make QueryResultParser as similar as possible to RDFParser to make integrating them into a single chain further along in the future easier.

These changes are similar to the bumps that RDFParser had between Sesame-2.8 and Sesame-4.0, but we didn't think about them at the same time then.

Sorry for dropping large changes just before a release, feel free to switch master to 3.0 and I will retarget this and a few other structural changes there rather than at the 2.0 release. Getting 2.0 out on time is the main goal and this would only really have the weekend to soak at this point so is at a higher risk than otherwise.

If this pull request as a whole is denied, I want to split out the SPARQL XML security part today into a new pull request (won't be configurable if I do that but I will make sure that it is always set on), as it is a basic denial of service protection that has been known for a long time and was just overlooked for the SPARQL XML parser for some reason.

On the new features, I think users will appreciate the fluent configuration model for QueryResultParser that they have had already for RDFParser, but again if you judge that it is too close then 3.0 is fine with me for that too.

The QueryResultParseException int/long change will be a very minor part of the migration for most people given Java's compile time support for int-to-long changes rather than making them a runtime error or compile warning like C does. It is part of the broader goal to make it's use consistent with ParseLocationListener and ParseErrorListener for a future integrated parser API that will reduce the code duplication that we have right now between rio and queryresultio.

None of these changes should surprise or confuse people who have used the Sesame-4.0 RDFParser model as they are simple copy-paste's between that and QueryResultParser (not ideal but I am aiming for removing that code duplication at some stage in the future)

If these changes are accepted for 2.0, then I can work on integrating more of the configuration support for the other queryresultio parsers in 2.1 without introducing any backwards incompatible changes as the infrastructure for that has been setup for that in this pull request.

jeenbroekstra:
What I'll do is download this locally, play with it a little bit to get a feel for how it affects legacy code. plus how stable this is. If that is sufficient I am OK with including this in 2.0.

jeenbroekstra:
From what I can tell, most changes are only likely to affect existing user code if those users have their own implementations of QueryResultParser. For use cases where they are simply _using_ a QueryResultParser, nothing needs changing (extending to a fluent API doesn't affect pre-existing use of void setter methods, for example), expect when reading line numbers from an exception (and there the change is trivial). 

All in all I'm fine with merging this in. 

Can you write one or two sentences for the [migration guide](http://rdf4j.org/doc/sesame-to-rdf4j-migration-guide/) to alert users on this minor api break? Should be a short paragraph somewhere in the 'Changes for Java programmers' section (1.1).

ansell:
The changes should also only affect QueryResultParser implementors if they do not inherit from AbstractQueryResultParser, as I didn't need to make any changes to queryresultio-text or binary to get this functional.

As part of the release notes, the security fix for SPARQL/XML results parser should be noted:

> The SPARQL/XML results parser now configures its internal XMLReader by default using XMLParserSettings.SECURE_PROCESSING to prevent entity expansion denial of service attacks. This should not be turned off under any normal circumstances for use in parsing untrusted data sources.

I can't remember how to login to edit rdf4j.org. There used to be a button to click for administration/login but I can't find it right now. The text for the migration guide is basically:

> QueryResultParseException constructors, getLineNumber and getColumnNumber now accept and return long instead of int to bring them inline with the changes that occurred for RDFParseException/ParseErrorListener/ParseLocationListener in Sesame-4.0
> Third-party implementors of QueryResultParser that do not inherit from AbstractQueryResultParser will notice some new methods such as setParseErrorListener and setParseLocationListener, that have been added from RDFParser to improve usability and configurability.

Examples of the fluent configuration for both QueryResultParser and RDFParser:

> Both RDFParser and QueryResultParser support a common fluent configuration style:

```
QueryResultCollector handler = new QueryResultCollector();
ParseErrorCollector errorCollector = new ParseErrorCollector();
QueryResultParser aParser = QueryResultIO.createTupleParser(TupleQueryResultFormat.SPARQL)
                                                         .setQueryResultHandler(handler)
                                                         .setParseErrorListener(errorCollector);
try {
    aParser.parseQueryResult(myInputStream);
} catch (QueryResultParseException e) {
    // log or rethrow QueryResultParseException
} finally {
    // Examine any parse errors that occurred before moving on
    System.out.println("There were " + errorCollector.getWarnings().size() + " warnings during parsing");
    System.out.println("There were " + errorCollector.getErrors().size() + " errors during parsing");
    System.out.println("There were " + errorCollector.getFatalErrors().size() + " fatal errors during parsing");
}
```

```
Model rdfStatements = new LinkedHashModel();
ParseErrorCollector errorCollector = new ParseErrorCollector();
RDFParser aParser = Rio.createParser(RDFFormat.TURTLE)
                                                .setRDFHandler(new StatementCollector(rdfStatements))
                                                .setParseErrorListener(errorCollector);
try {
    aParser.parse(myInputStream);
} catch (RDFParseException e) {
    // log or rethrow RDFParseException
} finally {
    // Examine any parse errors that occurred before moving on
    System.out.println("There were " + errorCollector.getWarnings().size() + " warnings during parsing");
    System.out.println("There were " + errorCollector.getErrors().size() + " errors during parsing");
    System.out.println("There were " + errorCollector.getFatalErrors().size() + " fatal errors during parsing");
}
```

A note about the configurability of XMLReader for the SPARQL/XML results parsers may be useful for the release notes and/or migration notes:

> SPARQL/XML results parsers now support custom XMLReader implementations to be configured in the same way as RDFXMLParsers using the following syntax:

```
QueryResultParser parser = QueryResultIO.createTupleParser(TupleQueryResultFormat.SPARQL)
                                                      .set(XMLParserSettings.CUSTOM_XML_READER, newInstanceOfXMLReader);
```

jeenbroekstra:
@ansell rdf4j.org runs on new software (Wordpress), so you will need a new login. Let me know which e-mail address you want to use and I will create an editor login for you.

jeenbroekstra:
Done. You should have received an e-mail with login details. If you have any trouble, shout out.

Issue #192, opened 2016-06-23
Lucene wrapped in SailRepository with multithreading gives error


jgrzebyta:
Affected: v 2.0-SNAPSHOT, 
Java: openjdk-8
os: Ubuntu 16.04

I use `LuceneSailRepository` wrapped in `SailsRepository`. `MemoryStore` with dump file is used as a storage.

I load data in multithreads managed by `Executors.newFixedThreadPool`pool. When I do `.invokeAll`
I got error about closed call when new request is done. It seems Lucene index in not thread safety.

The full stacktrace is:
```
23:29:47,677 [pool-3-thread-2] ERROR LuceneSailConnection: Rolling back org.apache.lucene.store.LockObtainFailedException: Lock obtain timed out: NativeFSLock@/tmp/rdf4j-loader-s5A6SCKMe.34724407296800170144873/lucenedir/write.lock
	at org.apache.lucene.store.Lock.obtain(Lock.java:89)
	at org.apache.lucene.index.IndexWriter.<init>(IndexWriter.java:780)
	at org.eclipse.rdf4j.sail.lucene.LuceneIndex.getIndexWriter(LuceneIndex.java:291)
	at org.eclipse.rdf4j.sail.lucene.LuceneIndex.invalidateReaders(LuceneIndex.java:581)
	at org.eclipse.rdf4j.sail.lucene.LuceneIndex.commit(LuceneIndex.java:651)
	at org.eclipse.rdf4j.sail.lucene.LuceneSailConnection.addRemoveStatements(LuceneSailConnection.java:252)
	at org.eclipse.rdf4j.sail.lucene.LuceneSailConnection.commit(LuceneSailConnection.java:217)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.commit(SailRepositoryConnection.java:169)
	at triple.loader$eval939$fn__940$fn__941.invoke(loader.clj:122)
	at triple.loader$eval939$fn__940.invoke(loader.clj:98)
	at clojure.lang.MultiFn.invoke(MultiFn.java:238)
	at triple.loader$eval935$fn__936.invoke(loader.clj:88)
	at clojure.lang.MultiFn.invoke(MultiFn.java:238)
	at sparql$load_multidata$fn__1094$fn__1095.invoke(sparql.clj:102)
	at clojure.lang.AFn.call(AFn.java:18)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

Issue #293, opened 2016-08-16
NPE with BackgroundTupleResult.handleClose in stacktrace


ansell:
The Jenkins server managed to find another NPE with BackgroundTupleResult.handleClose in the stacktrace. This is likely related to BackgroundTupleResult operating in its own thread and needing to be accessed by both cleanup threads and the originating thread.

https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/10/testReport/junit/org.eclipse.rdf4j.repository.sparql/SPARQLStoreConnectionTest/testQueryBaseURI_NONE_/

Similar issues in the past have been reported at:

https://openrdf.atlassian.net/browse/SES-1978

https://openrdf.atlassian.net/browse/SES-1039

https://bitbucket.org/openrdf/sesame/pull-requests/134/update-to-apache-httpcomponents-httpclient/diff#comment-293691

This was discovered solely on the Jenkins server while testing pull request #290 , but as it cannot be replicated locally and that pull request does not modify queryresultio code, it is not likely related to it specifically.

ansell:
Breaking down the stack trace gives the likely cause as being in EofSensorInputStream or ResponseEntityProxy because it allows multiple calls to close, which should be legitimate given the contract of java.io.Closeable.close that says that "If the stream is already closed then invoking this method has no effect."

First line is an NPE where the wrappedStream is accessed without a null-guard, although given it is an input parameter to the method, one could argue that the issue is with EofSensorInputStream for not checking before sending it up the chain.

```
java.lang.NullPointerException: null
	at org.apache.http.impl.execchain.ResponseEntityProxy.streamClosed(ResponseEntityProxy.java:140)
```

The next two calls, include methods that contain a finally to switch the wrappedStream to null (amongst other methods in the class, and the general contract, that sets it to null without synchronisation in any of the locations).

```
	at org.apache.http.conn.EofSensorInputStream.checkClose(EofSensorInputStream.java:228)
	at org.apache.http.conn.EofSensorInputStream.close(EofSensorInputStream.java:174)
```

There is an annotation on EofSensorInputStream that says "@NotThreadSafe", so it would be worthless to report a threading bug on such a class.

Hence, we need to do explicit synchronisation before every call down into Apache HTTPClient, given that they declare internally, and use broken null checking/setting patterns themselves.

ansell:
[...]

PR #294, opened 2016-08-17
Issue #293: Attempt to avoid nested NPE in BackgroundTupleResult

ansell:
This PR addresses GitHub issue: #293 

- Avoid NPE from Apache HTTPClient InputStream's that occur very intermittently being propagated through the stack. Not clear why the input streams are being closed twice, but the core reason is that HTTPClient is not attempting to be threadsafe and is using a set-to-null pattern that will always be vulnerable to issues without threadsafety wrappers
- Move almost all internal CloseableIteration implementations to subclasses of AbstractCloseableIteration to standardise the way closure occurs and attempt to prevent multiple calls to close getting anywhere near HTTPClient in future from our control
- Standardise on the handleClose pattern using try-finally to ensure all resource closures occur properly
- Add proactive calls to AbstractCloseableIteration.isClosed to exit earlier in cases where we have already called close on objects
- Adds active closure just before hasNext is about to return false or next is about to throw NoSuchElementException. This has been rebutted with anecdotal arguments in the past, but unless a minimal test case fails due to the changes then I will still be pushing for active closure on exhaustion. The main reason for pushing for it is that the behaviour looks correct. Ie, after-close behaviour is that hasNext returns false and next throws NoSuchElementException the same as just before the close-on-exhaust occurs.

- [x] tests are included
- [x] all tests succeed

The volatile closed variable requires double-checked locking to properly operate.

In particular, we are required by non-threadsafe Apache HTTPClient implementations to not call close multiple times in close succession.

Ie, they don't fully comply with the standard java.io.Closeable.close contract but try to release their obligation by declaring their implementations as "NotThreadSafe".

The underlying issue is possibly that the InputStream is having .close called on it directly from another location (thread/connection-pooling cleanup inside of HttpClient is the most likely place), but volatile variables require double-checked locking to operate safely, so there are no downsides to this change, even though I can't see how handleClose could be called twice given the use of AtomicBoolean.compareAndSet to isolate them, assuming there are no bugs with AtomicBoolean.compareAndSet...

The vast majority of classes with close methods have been audited and their implementations fixed so that all components of the close operations are attempted in a threadsafe manner as far as practical.


ansell:
All tests greenlined for me:

```
[INFO] RDF4J compliance tests ............................. SUCCESS [  0.012 s]
[INFO] RDF4J Model compliance test ........................ SUCCESS [  9.850 s]
[INFO] RDF4J Query Result IO compliance tests ............. SUCCESS [ 12.609 s]
[INFO] RDF4J Rio compliance tests ......................... SUCCESS [03:22 min]
[INFO] RDF4J HTTP server compliance tests ................. SUCCESS [ 14.865 s]
[INFO] RDF4J SAIL and Repository compliance test .......... SUCCESS [10:55 min]
[INFO] RDF4J SeRQL query parser compliance tests .......... SUCCESS [  5.420 s]
[INFO] RDF4J SPARQL query parser compliance tests ......... SUCCESS [02:32 min]
[INFO] RDF4J GeoSPARQL compliance tests ................... SUCCESS [  2.372 s]
[INFO] RDF4J BOM .......................................... SUCCESS [  0.005 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 24:35 min
[INFO] Finished at: 2016-08-17T11:27:50+10:00
[INFO] Final Memory: 51M/787M
[INFO] ------------------------------------------------------------------------
```

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/BackgroundGraphResult.java
```diff
@@ -87,19 +87,25 @@ public void remove()
...
+						}
+						finally {
+							in.close();
+						}
+					}
+					catch (IOException e) {
+						throw new QueryEvaluationException(e);
+					}
+					finally {
+						queue.close();
```
ansell: Is there an historical reason behind queue.close being called here but not in BackgroundTupleResult?

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/QueueCursor.java
```diff
@@ -29,11 +29,11 @@
...
 	private volatile boolean done;
 
-	private BlockingQueue<E> queue;
+	private final BlockingQueue<E> queue;
 
-	private E afterLast = createAfterLast();
+	private final E afterLast = createAfterLast();
 
-	private volatile Queue<Throwable> exceptions = new LinkedList<Throwable>();
+	private final Queue<Throwable> exceptions = new LinkedList<Throwable>();
```
ansell: This is technically a performance change, as it will help compiler avoid three volatile reference resolutions per getNextElement call since we were never reassigning this private variable anyway.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/QueueCursor.java
```diff
@@ -27,13 +28,13 @@
  */
 public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationException> {
 
-	private volatile boolean done;
+	private final AtomicBoolean done = new AtomicBoolean(false);
```
ansell: This wasn't being used in a strict sense as a hard limit with synchronized to perform once-only (consistent) actions (as evidenced by the use of queue.clear in a loop to get space for a sentinel), so switching to AtomicBoolean that has the ability to perform a lazySet that may be more useful for this scenario than a volatile variable.

jeenbroekstra:
I feel somewhat out of my depth here but it all looks very reasonable. I'll run this PR through Hudson, see what it says.

ansell:
I may yet split it up as it is growing further than originally planned. I am auditing the AbstractCloseableIteration subclasses for the threadsafety and resource closures in their handleClose methods to make sure they all use try-finally so some resource closures are not skipped as they can be currently, but hudson will help with some intermediate results.

ansell:
Only one of my local mvn clean install runs have failed, and only with one test, the interruptable order by queries test that had a different QueryEvaluationException triggering it. That condition is changed in:

https://github.com/eclipse/rdf4j/pull/294/commits/d8e8aeb0952b1a5d17e49cd507dfc8d91627f6d6

ansell:
Could you kill the Hudson job. It may have livelocked, as my local run on the commit just after it has  livelocked in NativeStoreConnectionTest so it may be occurring the same for RDFSchemaMemoryRepositoryConnectionTest:

https://github.com/eclipse/rdf4j/pull/294/commits/c2c55ef9fd252b9aadc3b83c206bc033ebfec039

At least that puts the likely regression before that commit though, as Hudson was run on:

https://github.com/eclipse/rdf4j/pull/294/commits/be9b5731b4dde7e87d8a406e0006ca5dd377909f

Can you trigger Hudson on the following commit (just before those two) to help get a regression range:

https://github.com/eclipse/rdf4j/pull/294/commits/5818cbbf21a02c2344603d8aa3c2a0b90b831893

ansell:
Thread dump for me after running the following commands:

```bash
$ ps aux | grep 'rdf4j' # To find PID
$ jsadebugd <PID>
```

then in another terminal:

```bash
$ jdb -connect sun.jvm.hotspot.jdi.SADebugServerAttachingConnector:debugServerName=localhost
Initializing jdb ...
>
```

Then using the "threads" command shows the following which seems to put the issue possibly in TimeLimitIteration so I will start there:

```
> threads
Group system:
  (java.lang.ref.Reference$ReferenceHandler)0x3 Reference Handler  cond. waiting
  (java.lang.ref.Finalizer$FinalizerThread)0x2  Finalizer          cond. waiting
  (java.lang.Thread)0x1                         Signal Dispatcher  running
Group main:
  (java.lang.Thread)0x4                         main               running
  (java.util.TimerThread)0x0                    TimeLimitIteration waiting in a monitor
```

core/util/src/main/java/org/eclipse/rdf4j/common/iteration/TimeLimitIteration.java
```diff
@@ -78,23 +79,41 @@ public void remove()
 	protected void handleClose()
 		throws X
 	{
-		interruptTask.cancel();
```
ansell: This seems to have something to do with it, ie, reordering handleClose and interruptTask.cancel may be causing issues. The livelock for me is inside of the infamous testOrderByQueriesAreInterruptable on line 2095 (the hasNext call and it is livelocked trying to complete TimeLimitIteration.interrupt, which is calling TImeLimitIteration.handleClose, eventually getting stuck at DelayedIteration.handleClose:

```
  [1] org.eclipse.rdf4j.common.iteration.DelayedIteration.handleClose (DelayedIteration.java:118), pc = 8
  [2] org.eclipse.rdf4j.query.algebra.evaluation.iterator.OrderIterator.handleClose (OrderIterator.java:232), pc = 1
  [3] org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close (AbstractCloseableIteration.java:49), pc = 13
  [4] org.eclipse.rdf4j.common.iteration.Iterations.closeCloseable (Iterations.java:121), pc = 11
  [5] org.eclipse.rdf4j.common.iteration.IterationWrapper.handleClose (IterationWrapper.java:99), pc = 8
  [6] org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close (AbstractCloseableIteration.java:49), pc = 13
  [7] org.eclipse.rdf4j.common.iteration.Iterations.closeCloseable (Iterations.java:121), pc = 11
  [8] org.eclipse.rdf4j.common.iteration.IterationWrapper.handleClose (IterationWrapper.java:99), pc = 8
  [9] org.eclipse.rdf4j.sail.base.SailClosingIteration.handleClose (SailClosingIteration.java:124), pc = 1
  [10] org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close (AbstractCloseableIteration.java:49), pc = 13
  [11] org.eclipse.rdf4j.common.iteration.Iterations.closeCloseable (Iterations.java:121), pc = 11
  [12] org.eclipse.rdf4j.common.iteration.IterationWrapper.handleClose (IterationWrapper.java:99), pc = 8
  [13] org.eclipse.rdf4j.sail.helpers.SailBaseIteration.handleClose (SailBaseIteration.java:62), pc = 1
  [14] org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close (AbstractCloseableIteration.java:49), pc = 13
  [15] org.eclipse.rdf4j.common.iteration.Iterations.closeCloseable (Iterations.java:121), pc = 11
  [16] org.eclipse.rdf4j.common.iteration.IterationWrapper.handleClose (IterationWrapper.java:99), pc = 8
  [17] org.eclipse.rdf4j.common.iteration.TimeLimitIteration.handleClose (TimeLimitIteration.java:83), pc = 1
  [18] org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close (AbstractCloseableIteration.java:49), pc = 13
  [19] org.eclipse.rdf4j.common.iteration.TimeLimitIteration.interrupt (TimeLimitIteration.java:119), pc = 16
  [20] org.eclipse.rdf4j.common.iteration.InterruptTask.run (InterruptTask.java:30), pc = 16
  [21] java.util.TimerThread.mainLoop (Timer.java:555), pc = 221
  [22] java.util.TimerThread.run (Timer.java:505), pc = 1
```

ansell:
The deadlocks seem to have been fixed by the following commit for me in my local testing:

https://github.com/eclipse/rdf4j/pull/294/commits/e6a3887a0b38a6288ecc14d587dae1488be97ef0

One cause may have been that more of the resource closures are now happening, and that invalidated the synchronisation model used in DelayedIteration that wasn't strictly necessary and was only tolerated with the relaxed resource closures and general lack of calls to super.handleClose(). It is worked around in that commit by doing the isClosed checks upfront before synchronisation occurs in hasNext and next and doing a single volatile resolution and a null-guard in the close rather than the full synchronisation that was occurring previously.

If synchronisation is actually needed in the close, and it isn't clear from looking at the other implementations that it would be, as they all close iterators in their handleClose without synchronisation, then some form of locking would be necessary. However, the locking couldn't be just simple reentrant, as the closures regularly occur from different threads to the threads calling to next and hasNext so they don't get classified as reentrant.

ansell:
Found the LuceneSailTest could deadlock in testMultithreadedAdd using this branch, although it doesn't always occur. Looking into how to fix it systematically other than the global Timeout. Cause seems to be that the exception is not propagated out of the runners to the test and it sits on await forever:

Stacktrace that was printed before it deadlocked was:
```
Exception in thread "Thread-1" org.eclipse.rdf4j.repository.RepositoryException: org.eclipse.rdf4j.sail.SailException: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.commit(SailRepositoryConnection.java:172)
	at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.conditionalCommit(AbstractRepositoryConnection.java:366)
	at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.add(AbstractRepositoryConnection.java:491)
	at org.eclipse.rdf4j.sail.lucene.AbstractLuceneSailTest$1.run(AbstractLuceneSailTest.java:1055)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.eclipse.rdf4j.sail.SailException: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.eclipse.rdf4j.sail.lucene.LuceneSailConnection.commit(LuceneSailConnection.java:237)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.commit(SailRepositoryConnection.java:169)
	... 4 more
Caused by: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.apache.lucene.index.IndexReader.ensureOpen(IndexReader.java:274)
	at org.apache.lucene.index.SegmentReader.getPostingsReader(SegmentReader.java:227)
	at org.apache.lucene.index.CodecReader.fields(CodecReader.java:102)
	at org.apache.lucene.index.LeafReader.terms(LeafReader.java:207)
	at org.apache.lucene.index.LeafReader.termDocsEnum(LeafReader.java:318)
	at org.eclipse.rdf4j.sail.lucene.LuceneIndex.addDocuments(LuceneIndex.java:461)
	at org.eclipse.rdf4j.sail.lucene.LuceneIndex.getDocuments(LuceneIndex.java:452)
	at org.eclipse.rdf4j.sail.lucene.LuceneIndex.getDocuments(LuceneIndex.java:338)
	at org.eclipse.rdf4j.sail.lucene.AbstractSearchIndex.addRemoveStatements(AbstractSearchIndex.java:237)
	at org.eclipse.rdf4j.sail.lucene.LuceneSailConnection.addRemoveStatements(LuceneSailConnection.java:251)
	at org.eclipse.rdf4j.sail.lucene.LuceneSailConnection.commit(LuceneSailConnection.java:217)
	... 5 more
```

ansell:
At the time it deadlocked the main thread was waiting at endLatch.await due to the endLatch.countDown inside of the runner not being in a finally block, will fix that issue, but not sure how to replicate the underlying exception yet:
```
  [1] sun.misc.Unsafe.park (native method)
  [2] java.util.concurrent.locks.LockSupport.park (LockSupport.java:175), pc = 14
  [3] java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt (AbstractQueuedSynchronizer.java:836), pc = 1
  [4] java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly (AbstractQueuedSynchronizer.java:997), pc = 72
  [5] java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly (AbstractQueuedSynchronizer.java:1,304), pc = 24
  [6] java.util.concurrent.CountDownLatch.await (CountDownLatch.java:231), pc = 5
  [7] org.eclipse.rdf4j.sail.lucene.AbstractLuceneSailTest.testMultithreadedAdd (AbstractLuceneSailTest.java:1,067), pc = 60
  [8] sun.reflect.NativeMethodAccessorImpl.invoke0 (native method)
  [9] sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62), pc = 100
  [10] sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43), pc = 6
  [11] java.lang.reflect.Method.invoke (Method.java:498), pc = 56
  [12] org.junit.runners.model.FrameworkMethod$1.runReflectiveCall (FrameworkMethod.java:50), pc = 15
  [13] org.junit.internal.runners.model.ReflectiveCallable.run (ReflectiveCallable.java:12), pc = 1
  [14] org.junit.runners.model.FrameworkMethod.invokeExplosively (FrameworkMethod.java:47), pc = 10
  [15] org.junit.internal.runners.statements.InvokeMethod.evaluate (InvokeMethod.java:17), pc = 12
  [16] org.junit.internal.runners.statements.RunBefores.evaluate (RunBefores.java:26), pc = 49
  [17] org.junit.internal.runners.statements.RunAfters.evaluate (RunAfters.java:27), pc = 12
  [18] org.junit.runners.ParentRunner.runLeaf (ParentRunner.java:325), pc = 17
  [19] org.junit.runners.BlockJUnit4ClassRunner.runChild (BlockJUnit4ClassRunner.java:78), pc = 30
  [20] org.junit.runners.BlockJUnit4ClassRunner.runChild (BlockJUnit4ClassRunner.java:57), pc = 6
  [21] org.junit.runners.ParentRunner$3.run (ParentRunner.java:290), pc = 12
  [22] org.junit.runners.ParentRunner$1.schedule (ParentRunner.java:71), pc = 1
  [23] org.junit.runners.ParentRunner.runChildren (ParentRunner.java:288), pc = 44
  [24] org.junit.runners.ParentRunner.access$000 (ParentRunner.java:58), pc = 2
  [25] org.junit.runners.ParentRunner$2.evaluate (ParentRunner.java:268), pc = 8
  [26] org.junit.runners.ParentRunner.run (ParentRunner.java:363), pc = 20
  [27] org.apache.maven.surefire.junit4.JUnit4Provider.execute (JUnit4Provider.java:264), pc = 116
  [28] org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet (JUnit4Provider.java:153), pc = 70
  [29] org.apache.maven.surefire.junit4.JUnit4Provider.invoke (JUnit4Provider.java:124), pc = 158
  [30] org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader (ForkedBooter.java:200), pc = 23
  [31] org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess (ForkedBooter.java:153), pc = 14
  [32] org.apache.maven.surefire.booter.ForkedBooter.main (ForkedBooter.java:103), pc = 183
```

core/util/src/main/java/org/eclipse/rdf4j/common/iteration/IterationWrapper.java
```diff
@@ -53,20 +55,28 @@ public IterationWrapper(Iteration<? extends E, ? extends X> iter) {
...
+		}
+		return result;
 	}
 
 	/**
 	 * Returns the next element from the wrapped Iteration.
 	 * 
 	 * @throws java.util.NoSuchElementException
-	 *         If all elements have been returned.
-	 * @throws IllegalStateException
```
ansell: This part of the contract was contradictory to CloseableIteration, and was picked up by automated testsuite once it was actually implemented here. Hence, it is changed to be consistent with CloseableIteration.

jeenbroekstra: ITYM `Iteration` (which is where `next()` is defined), but good call.

jeenbroekstra:
looks good Peter. Hudson is reporting a test failure on the current branch though, see https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/12/ .

It's `LuceneSailTest.testMultiThreadedAdd` that fails. Not sure if that is caused by your changes or if it is an intermittent thing. I'm taking a closer look.

jeenbroekstra:
I have now merged the 2.0 release branch back into master, which includes fix #192 - hopefully this fixes the intermittent error you reported as issue #301. Can you update your PR?

core/util/src/main/java/org/eclipse/rdf4j/common/iteration/ConvertingIteration.java
```diff
@@ -56,10 +58,18 @@ protected abstract T convert(S sourceObject)
...
 	public boolean hasNext()
 		throws X
 	{
-		return iter.hasNext();
+		if (isClosed()) {
+			return false;
+		}
+		boolean result = iter.hasNext();
+		if (!result) {
+			close();
```
jeenbroekstra: I seem to remember that I proposed something very similar earlier (in Sesame) and there were some negative responses to this approach, so we abandoned it. See https://bitbucket.org/openrdf/sesame/pull-requests/550/ses-2323-added-autoclose-of-exhausted/diff

ansell: I may be wrong (been working on this for a while), but I think there are already self-closing cases in the code before this pull request. 

I also don't accept the general premise that we are hurting anything, especially users, in general by halting iterations either when hasNext will be returning false or just before next() would throw a NoSuchElementException. If a CloseableIteration knows that it can't immediately deliver a result, it should either block on hasNext or not respond to the close request as MutableTupleQueryResult does. If it doesn't want to follow the general contract that if hasNext ever returns false it is all over it is free to do so but we shouldn't encourage resource leaks just to cater for some situations where the programmer doesn't know their codebase well enough to know where to start looking when they accidentally exhause iterations and actually get a stack trace to help them start debugging it.

I don't accept the specific case that close on exhaustion can create deadlock bugs due only to the closure and would need specific examples to convince me of that one.

Relying on a complex finalizer mechanism instead of cleaning up after oneself at the time when the contract says we are free to do so is troublesome and will require more memory (while things leak onto the heap before getting finalised later) and time (both developer and running time) determining the source of bugs. Having said that, users are free to maintain a finalizer mechanism to pick up the odd case where this doesn't succeed, but it isn't a clean solution to resource management. Deadlocks are much more likely to happen in finalizers, due to having to get monitors on other threads, compared to when the cleanup is done within an objects own code where the immediate monitor is already guaranteed by the JVM to be held and any necessary locks are in place and don't need to be renegotiated.

I can imagine more cases where the source of the intermittent bug behind this pull request (that of HttpClient internals being modified outside of the handleClose method), that would be exacerbated by relying on lazy/queue-driven/thread-driven finalisation compared to active early closing.

FWIW, the only failures I have noticed with any of these changes are the Lucene failures that are not replicable and not unique to this so cannot be directly related to these changes. All of our other tests have been rock-solid throughout the development of this pull request, based on running full builds after ever 1-3 commits on my local machine given the complexity as a whole of the pull request.

If someone wanted to avoid the autoclose behaviour where they were using a subclass of AbstractCloseableIteration in their own code (say one of the query results iterators) but didn't want it to close just because it got to the end of its iteration, all they need to do is override the handleClose method to do nothing and the calls to close will silently short-circuit after going into handleClose the first time and then be avoided everytime after that. 

This pull request is just about being consistent internally and following the stated contract for CloseableIteration internally to avoid as many resource leaks and intermittent bugs in the future as possible.

I don't mind having a discussion about this, but fear of the unknown or anecdotal cases isn't a reason not to engineer it well IMO. The lack of consistency in the handleClose pattern (given that all of these iteration classes are used and stacked all over the API) triggered a few other changes attached to this, but they can be removed if someone can show they cause a bug in a minimal test case that can't be fixed by more cleanup.

jeenbroekstra: I agree. At the time I thought the autoclose behavior was undocumented - which is why I accepted the criticism and abandoned the fix. But I later discovered that in fact it _was_ documented (and also implemented in several places, just not consistently). But I never got around to picking this up again.

So I'm good with this - it certainly makes everything more consistent. If we later find a good motivation for changing this we have a solid basis.

ansell:
@jeenbroekstra Could you run this through Hudson again? It may still barf on the Lucene exception, but it shouldn't deadlock this time, and at best it runs through the full testsuite to get a second verification apart from my personal machine.

I haven't finished auditing yet (there are a lot! of instances of Iteration/CloseableIteration in the code base), but its would be useful to try Hudson again as the last times failed due to unrelated errors.

jeenbroekstra:
Hudson build running now: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/14/

jeenbroekstra:
Hudson build succeeded.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/federation/JoinExecutorBase.java
```diff
@@ -24,20 +25,29 @@
  */
 public abstract class JoinExecutorBase<T> extends LookAheadIteration<T, QueryEvaluationException> {
 
+	/**
+	 * @deprecated No replacement, don't use static shared int variables.
+	 */
 	protected static int NEXT_JOIN_ID = 1;
```
ansell: If this variable is actually used, it should be a static final AtomicInteger rather than a static mutable int. There were no references to it, but still deprecating it rather than changing or removing it as protected variables are presumed to be part of the public API until we clarify the procedure for these cases.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/federation/JoinExecutorBase.java
```diff
@@ -24,20 +25,29 @@
...
+	protected final CloseableIteration<T, QueryEvaluationException> leftIter;
 
-	protected CloseableIteration<T, QueryEvaluationException> rightIter;
+	protected volatile CloseableIteration<T, QueryEvaluationException> rightIter;
 
-	protected volatile boolean closed;
+	/**
+	 * @deprecated Use {@link AbstractCloseableIteration#isClosed()} instead.
+	 */
+	protected volatile boolean closed = false;
```
ansell: This is still set to true in the close method incase third-party code is relying on it, but is not accessed anymore so could be removed in future.

ansell:
This is ready for review. The entire patch has been squashed down to a single commit rather than the 50+ original commits and has been rebased on top of the current master so it contains all patches that made it onto master recently.

jeenbroekstra:
Due to house move I may not have time to look at this until next week somewhere - but I had a quick look through it, and overall I think you've done an awesome job here. Will try and find time for more thorough review ASAP.

ansell:
I also sent a patch upstream to isolate the original NPE that caused this. It will be available in the next apache httpclient release, 4.5.3. 

https://issues.apache.org/jira/browse/HTTPCLIENT-1767

The changes in this pull request still need to be applied locally to ensure that all closures are given the chance to complete fully and as expected in our multi-threaded environment.

jeenbroekstra:
Looks as if a recent merge caused a conflict, can you check?

ansell:
I rebased again to fix the conflict. All tests still greenlining for me locally.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/RepositoryInterceptor.java
```diff
@@ -51,18 +53,18 @@
...
-	private RepositoryConnection repositoryCon;
+	private volatile RepositoryConnection repositoryCon;
 
 	/*---------*
 	 * Methods *
 	 *---------*/
 
 	public void setRepositoryManager(RepositoryManager repMan) {
-		repositoryManager = repMan;
+		repositoryManager = Objects.requireNonNull(repMan, "Repository manager was null");
```
jeenbroekstra: Nice, had no idea you could do this in a single line.

jeenbroekstra:
One minor thing: does issue #293 still adequately sum up what you've done here? It seems to me you've done a lot more than just fix a potential NPE.

ansell:
I will add more content to 293 to describe the reasoning for the audit.

jeenbroekstra:
Hudson verify running at https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/25/

jeenbroekstra:
Build success. Merging.

PR #297, opened 2016-08-22
Fixes #296 by closing the embedded iteration.

### LINKED ISSUE #296: CrossProductIteration doesn't close embedded iteration
The problem manifests as hanging queries with federation of the type:

select * {
    ?s ?p ?o 
    service <http://localhost/repositories/repo1> {
        ?x ?y ?z
    }
}

avataar:
This PR addresses GitHub issue: #296 

- Closes the embedded iteration in handleClose().

- [ ] RDF4J code formatting has been applied
- [ ] tests are included
- [ ] all tests succeed

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/CrossProductIteration.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/CrossProductIteration.java
@@ -7,14 +7,14 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.iterator;
 
-import java.util.Iterator;
-import java.util.List;
-
 import org.eclipse.rdf4j.common.iteration.CloseableIteration;
 import org.eclipse.rdf4j.common.iteration.LookAheadIteration;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.QueryEvaluationException;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryBindingSet;
+
+import java.util.Iterator;
+import java.util.List;
 
 /**
  * Iteration which forms the cross product of a list of materialized input bindings with each result obtained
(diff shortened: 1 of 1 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/CrossProductIteration.java
```diff
@@ -67,4 +67,12 @@ protected BindingSet getNextElement()
 
 		return null;
 	}
+
+	@Override
+	protected void handleClose()
+		throws QueryEvaluationException
+	{
+		super.handleClose();
```
ansell: The two actions here need to be in a try-finally configuration to ensure both actions occur when exceptions are thrown from super.handleClose.

jeenbroekstra:
@avataar are you available to fix the PR with @ansell 's suggestion?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/CrossProductIteration.java
avataar: That's fine but if we do that we should do it everywhere. Almost all implementations of handleClose() in iterations do the same as I did (some first close the wrapped iterations and then call super.handleClose()). Shouldn't we rather have a more generic way to do this consistently?

jeenbroekstra: @avataar thanks for applying the fix, and you're quite right that it should be done everywhere. @ansell  is in fact doing this as part of Pull Request #294 - which has turned into a general audit and standardization of the Iteration hierarchy.

jeenbroekstra:
Approved

jeenbroekstra:
@avataar thanks for applying the fix, and you're quite right that it should be done everywhere. @ansell is in fact doing this as part of Pull Request #294 - which has turned into a general audit and standardization of the Iteration hierarchy.

PR #307, opened 2016-08-31
Fix #239, second try

### LINKED ISSUE #239: Missing unit tests for spin:functions

jgrzebyta:
This PR addresses GitHub issue: #239 

- Second try to fix issue #239 
- Replacement of PR eclipse/rdf4j#240

- [x] tests are included
- [x] all tests succeed

    
More information at eclipse/rdf4j/pull/240
Signed-off-by:Jacek Grzebyta <grzebyta.dev@gmail.com>

### DIFF
--- a/core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
+++ b/core/sail/spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpifSailTest.java
@@ -1,12 +1,15 @@
-/*******************************************************************************
+/**
+ * *****************************************************************************
  * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Distribution License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/org/documents/edl-v10.php.
- *******************************************************************************/
+ ******************************************************************************
+ */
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
I notice that this PR only contains a single changed file, while the original was a lot larger. I think the problem is that I reverted the entire merge, and git doesn't see that you now want to re-merge your changes. 

The fix is to "revert the revert". The commit that reverted the merge was 6acf75b . What you need to do on your branch is revert that commit, then push. That should update your PR to include all the changes you made before I reverted the merge.  

Command: `git revert 6acf75b`

See https://www.kernel.org/pub/software/scm/git/docs/howto/revert-a-faulty-merge.txt for more details.

jgrzebyta:
That was intentional. @ansell suggested that using manifest-base approach is better that the file scanning and I found that was true indeed. Java resource finding (that is what I have used) is not suitable if the software is built with changing the application context   and that was shown in the error message.

Unfortunately it is solved only in some additional libraries, eg.: Spring Framework with its ``ResourceLoader`` interface but not in the Java core. 

Thus this PR should be treated as a replacement of eclipse/rdf4j#240 PR.

jeenbroekstra:
I see, thanks. In that case I'll let hudson do a verify run on this PR and once that succeeds I'll merge!

jeenbroekstra:
Hudson running: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/16/

Issue #301, opened 2016-08-30
LuceneSailConnection is not threadsafe


ansell:
The LuceneSail sends the same LuceneIndex to each LuceneSailConnection created from it. Each of the connections are implemented to call LuceneIndex.endReading when they are closed, which fails sporadically with the following (or similar) stacktrace:

```
Exception in thread "Thread-1" java.lang.AssertionError: org.eclipse.rdf4j.repository.RepositoryException: org.eclipse.rdf4j.sail.SailException: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.eclipse.rdf4j.sail.lucene.AbstractLuceneSailTest$1.run(AbstractLuceneSailTest.java:1069)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.eclipse.rdf4j.repository.RepositoryException: org.eclipse.rdf4j.sail.SailException: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.commit(SailRepositoryConnection.java:172)
	at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.conditionalCommit(AbstractRepositoryConnection.java:366)
	at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.add(AbstractRepositoryConnection.java:491)
	at org.eclipse.rdf4j.sail.lucene.AbstractLuceneSailTest$1.run(AbstractLuceneSailTest.java:1063)
	... 1 more
Caused by: org.eclipse.rdf4j.sail.SailException: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.eclipse.rdf4j.sail.lucene.LuceneSailConnection.commit(LuceneSailConnection.java:237)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.commit(SailRepositoryConnection.java:169)
	... 4 more
Caused by: org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
	at org.apache.lucene.index.IndexReader.ensureOpen(IndexReader.java:274)
	at org.apache.lucene.index.SegmentReader.getPostingsReader(SegmentReader.java:227)
	at org.apache.lucene.index.CodecReader.fields(CodecReader.java:102)
	at org.apache.lucene.index.LeafReader.terms(LeafReader.java:207)
[...]

PR #322, opened 2016-09-05
issue#301 : Work on tightening the contract for lucene sail components

ansell:
This PR addresses GitHub issue: #301 

- Adds more checks and sychronisation to avoid calling into Lucene API at unexpected times

- [x] tests are included
- [x] all tests succeed

Lock more Lucene methods. The internal Lucene API isn't stable for long term threaded access so external synchronisation is necessary.

Add checks on calls after closure

With these changes I have not been able to replicate the issue after several (30+) invocations of the testsuite. Previously I was getting the error about 1 in every 10 runs through LuceneSailTest.

### DIFF
--- a/core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/AbstractReaderMonitor.java
+++ b/core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/AbstractReaderMonitor.java
@@ -33,14 +33,22 @@ public abstract class AbstractReaderMonitor {
 		this.index = index;
 	}
 
-	public int getReadingCount() {
+	public final int getReadingCount() {
 		return readingCount.get();
 	}
 
 	/**
 	 * 
 	 */
-	public void beginReading() {
+	public final synchronized void beginReading() {
+		if (closed.get()) {
+			throw new IllegalStateException("Cannot begin reading as we have been closed.");
+		}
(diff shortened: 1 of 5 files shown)


core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/AbstractReaderMonitor.java
```diff
@@ -79,10 +88,10 @@ public boolean closeWhenPossible()
 		return closed.get();
 	}
 
-	public void close()
+	public final void close()
 		throws IOException
 	{
-		if (!closed.getAndSet(true)) {
+		if (closed.compareAndSet(false, true)) {
```
ansell: This line is an actual bug fix, as getAndSet doesn't have a similar contract to the useful compareAndSet method, but on its own it didn't fix the issue.

jeenbroekstra:
Great stuff, thanks Peter! I will merge and then cherrypick to have it on the 2.0 branch.

jeenbroekstra:
I was a bit over-eager in merging this - should have run verification first and also get @pulquero's feedback. But it appears no harm was done and all tests succeed.

ansell:
Hopefully the tests stay stable ;) Thats the trouble with an intermittent bug like this.

pulquero:
fwiw, looks fine to me.

PR #323, opened 2016-09-05
Sync of Releases/2.0.x to master

jeenbroekstra:
Putting up this PR to see how GitHub deals with cherrypicking. Somewhat surprisingly it shows @ansell's fixes in the diff, despite the fact that both branch have these fixes already.


jeenbroekstra:
@ansell you have any idea why this would happen? I sort of assume that it doesn't see the merge-commit and the cherry-pick commit are "the same thing". Any dangers you see in merging this? Anything I could be doing differently in handling merging and cherry-picking that you know of?

ansell:
I haven't had enough experience cherry-picking to know what issues there will be merging after cherry-picking the other way. Hopefully none but we won't know for sure until you hit the merge button. I have had some experience with Git intelligently knowing that a commit has already been done and avoiding committing it again, but I can't remember the exact circumstances.

Issue #266, opened 2016-07-27
Improving our git workflow


jeenbroekstra:
Our current git workflow is outlined in the [contributing guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md). 

Summarized, it's this:

1. each enhancement or bug fix gets its own GitHub issue.
2. each enhancement or bug fix gets its own branch, named `issues/#<issue>-<keywords>`
3. issues for bug fixes or small (backward compatible) improvements are branched off from the latest release branch (currently `releases/2.0.x`).
4. issues for larger improvements are branched off from master branch. 
5. every issue branch, once complete, is merged back into the _same_ branch from which it originally came. 

I have noticed that several people, in particular new developers, struggle with picking the correct branch to start fixing things and/or with picking the right target branch to merge their fixes back into. 

Partly this could be caused by unclear explanation in our docs, or by inattentive reading on the part of the contributor, but I feel that perhaps our workflow itself is not as simple and straightforward as it could be. 

In particular, I think that people struggle to remember step 5 (perhaps they simply forgot where their branch originated after a while), and this is not something that is trivially simply to see in most git tools. 

So, open question: does anybody have any suggestions on how to improve our workflow and/or our workflow documentation? Any ideas to make it easier, both for new contributors _and_ for existing contributors and full committers, are most welcome.

ansell:
It is much more difficult than the previous workflow where we generally had a single minor version that was active and a single minor version that was only given major bug fixes. The introduction of the master branch and the lack of any releases yet to give a baseline for what makes a change significant has made it a bit confusing about where things should be targeted.

In the case of issue#264 today, I opened its pull request against master because it isn't a critical fix (data loss/stability/etc.) and I didn't want to interrupt the upcoming review of the first 2.0/1.0 releases. 

Having a clear roadmap for which versions are upcoming along with expected maintenance dates for major/minor version combinations, with the name of the current git branch to start on would be useful. Then when you want to lock off a branch from features, such as when 2.0.0 gets released, you would change the roadmap to indicate that even small features get put on master.

I have also found it a little confusing that the two release branches have had -M1/-M2 releases and then are bumped back to the same snapshot build as previously. Bumping back to 2.0-SNAPSHOT rather than 2.0-M3-SNAPSHOT, for example, makes it difficult to easily visualise which milestone we are currently up to on each branch

jeenbroekstra:
[...]

PR #325, opened 2016-09-06
Update to readme and contributor guidelines

jeenbroekstra:
GitHub issue: #266 

Added a short version of the guidelines to the readme. Guidelines updated to use simplified branching model, and added info on release cycle.

Please review for correctness. I'm not 100% sure about the organization of this either: if you have suggestions for changes to make it all easier to digest by all means go ahead.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -6,13 +6,12 @@ Before you dive in, here are some things you need to know.
 **Table of Contents**  
 
 - [Legal stuff](#legal-stuff)
+- [Creating your contribution](#creating-your-contribution)
 - [Workflow](#workflow) 
  - [Releases](#releases)
   - [Release branches](#release-branches)
   - [A note on the 1.0.x release branch and Java 7 compatibility](#a-note-on-the-10x-release-branch-and-java-7-compatibility)
 - [Code formatting](#code-formatting)
-- [Creating your contribution](#creating-your-contribution)
-- [Submitting your changes](#submitting-your-changes)
 	
 ## Legal stuff
 
@@ -34,6 +33,31 @@ To sign the Eclipse CLA you need to:
 
 Every commit you make in your patch or pull request MUST be "signed off".
 You do this by adding the `-s` flag when you make the commit(s).
+
+## Creating your contribution
+
+Once the legalities are out of the way you can dig in. Here's how:
+
+1. Create an issue in the [RDF4J GitHub issue tracker](https://github.com/eclipse/rdf4j/issues) that describes your improvement, new feature, or bug fix. Alternatively, comment on an existing issue to indicate you're keen to help solve it.
+2. Fork the repository on GitHub.
+3. Create a new branch for your changes starting from the `master` branch. See [Workflow](#workflow) for details.
+4. Make your changes. Apply [RDF4J code formatting guidelines](#code-formatting)
+5. Make sure you include tests.
+6. Make sure the test suite passes after your changes.
+7. Commit your changes into the branch. Use meaningful commit messages. Reference the issue number in the commit message (for example "issue #276: added null check").
+8. **Sign off** every commit you do (as explained in the [legal stuff](#legal-stuff).
+9. Optionally squash your commits (not necessary, but if you want to clean your commit history a bit, _this_ is the point to do it).
+10. Push your changes to your branch in your forked repository.
+11. If your contribution is complete, use GitHub to submit a pull request (PR)
+	for your contribution back to `master` in the central RDF4J repository.
+	Once you have submitted your PR, do not use your branch for any other
+	development (unless asked to do so by the reviewers of your PR). 
+
+Once you've put up a PR, we will review your contribution, possibly make some
+suggestions for improvements, and once everything is complete it will be merged
+into the `master` branch, to be included in the next minor or major release. If
+your contribution is a bug fix in an existing release, we will also schedule it
+for inclusion in a patch release.
 
 ## Workflow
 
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.md
```diff
@@ -28,40 +29,87 @@ To sign the Eclipse CLA you need to:
...
+
+## Workflow
+
+The short-short version for contributors: work from the `master` branch. 
 
-RDF4J feature development takes place against the `master` branch. In addition,
-each minor release has its own release branch (e.g. `releases/1.0.x`, and
-`releases/2.0.x`), on which hotfixes/patches are done.
+The more complete version: RDF4J uses a git branching model where feature
+development takes place on the `master` branch. This is where all development
```
ansell: This might be able to clarified from "on the master branch" to "on branches from the master branch"

.github/CONTRIBUTING.md
```diff
@@ -28,40 +29,87 @@ To sign the Eclipse CLA you need to:
...
+time to complete release review, we use a feature cutoff date. This date is
+typically three weeks before the release date. The feature cutoff date does not
+mean that each feature needs to be complete, but it does mean that we ask every
+contributor to make a commitment to have their feature complete and stable in
+time for final review and release. If a contributor cannot make that
+commitment, the feature is not included in the release and is simply
+"postponed" to the next release cycle.
 
-Since `releases/1.0.x` is to be kept compatible with Java 7, it is inevitable that it will occassionally become unstable when merging the `releases/2.0.x` branch into it: the build process verifies that the branch is Java 7-compatible, so it will fail with a compilation error when this happens. These compilation failures can be fixed by branching a hotfix branch directly directly off the `releases/1.0.x` branch.
+We occassionally also schedule *patch releases*, which only include (backward
```
ansell: spelling, "occasionally"

.github/CONTRIBUTING.md
```diff
@@ -28,40 +29,87 @@ To sign the Eclipse CLA you need to:
...
+compatible) bug fixes. These releases are typically planned on short notice and
+can be done without much in the way of formal review.
+
+#### Release branches
+
+A few days before release of a minor or major version, we create a *release
+branch* from the current head of the master branch. The purpose of this release
+branch is to perform last-minute tweaks and little fixes. Once we are ready to
+release, the head of the release branch is tagged with the version number, and
```
ansell: Replacing "head", which is technically correct, with "latest commit" may make this slightly more fluent.

ansell:
Looks fairly good, just a few comments inline.

jeenbroekstra:
I merged this using GitHub's "squash and commit" option, instead of the regular merge commit. The advantage of this is that it makes cherrypicking easier, and we get a simpler history view. The downside is that we lose the original commit history (and my local git client gets confused about local branches and whether or not they are merged or not).

ansell:
Losing the commit history is cleaner to browse, but hopefully it won't result in more local merge conflicts. 

I also notice that the Signed-Off tag isn't on the squashed/merged commit. Is it possible to have it there (hopefully without having users type it out, which is prone to spelling mistakes)? It may be a little confusing having users always use -s but then not have the signed-off by (which Eclipse seems to require) in the final squashed commit.

I also use -S (gpg signing) which is also wiped by squashing, but that isn't as big an issue in the long term given the other inbuilt safeguards that Git has.

Both -s and -S could be preserved if we change the instructions to include "git rebase -i master" and "git commit --amend -as" or "git commit --amend -asS" for them to use locally after the pull request was approved but before merging. However, it is difficult to explain the rebase process, and would be much simpler to use the Github squash if it can preserve "-s" that Eclipse requires.

jeenbroekstra:
NB I am using squashing for merges, _not_ rebasing. They're different things!

jeenbroekstra:
> Losing the commit history is cleaner to browse, but hopefully it won't result in more local merge conflicts. 

I don't think it will - unless people do local changes on their own master branch directly (but in that case you'd get merge conflicts no matter what merging technique we use).

> I also notice that the Signed-Off tag isn't on the squashed/merged commit. Is it possible to have it there (hopefully without having users type it out, which is prone to spelling mistakes)? It may be a little confusing having users always use -s but then not have the signed-off by (which Eclipse seems to require) in the final squashed commit.

Well we don't get the signoff in the 'normal' merge commit either. I don't think I can easily configure this. However, I should point out that the way I'm aiming for this to work is that the commit message includes not just the issue number, but also the Pull Request number. This actually happens automagically if I use the GitHub squash-and-merge - and this gives you an automatic link back to the original PR, which still includes the full commit history.

ansell:
I know they are technically separate, but "rebase -i" is how you would practically do squashing locally, and rebasing onto the current tip/head is the simplest way to rebase locally.

ansell:
Is the full original commit history still visible if someone deletes the branch or their repository? I haven't kept up with the way that the github pull request branches work to know the specifics of that. 

Even though the merge doesn't have signed-off it is still present on the original commits if we don't use squash and merge.

jeenbroekstra:
> Is the full original commit history still visible if someone deletes the branch or their repository? I haven't kept up with the way that the github pull request branches work to know the specifics of that

Ah good point - I was looking at this particular PR, but wasn't considering PRs from forked repos. Hm, probably not.

Perhaps I should be a little less keen to use squashing everywhere. My main motivation was to have easier cherrypicking. But to be fair using Sourcetree, at least, I've found that selecting multiple commits for cherrypicking is still quite straightforward.

jeenbroekstra:
Shall we continue this discussion in issue #266 ?

PR #564, opened 2016-09-08
Issue #266: simplified Pull Request template

jeenbroekstra:
Got rid of the redundant link to the guidelines (GitHub shows it right at the top itself) and the tick boxes for formatting and tests, etc (nobody really uses them anyway it seems).


ansell:
The tasks were useful to me, but they weren't consistently used by everyone to confirm whether they had actually run mvn verify or equivalent locally. Streamling in general is useful though.

Issue #319, opened 2016-09-03
Add "distinct" query parameter to rdf4j sparql endpoint


pulquero:
Issuing a construct query to an RDF4J repository can return a serialization e.g. rdf+xml, that includes dups. While it is fine not to dedup the direct result of a construct query, arguably its serialization should be to save space/bandwidth. Probably, RIO should be responsible for doing this, either the framework itself or the individual graph writers.

jeenbroekstra:
The problem is that de-duplicating means keeping results in memory, which breaks the streaming character of the writers and make them less scalable - not to mention that the extra verification step of course lowers writer performance. The philosophy behind the writers has always been to be as lean as possible - and barring corner cases usually the duplication is not so much that you appreciably save bandwidth (especially when using gzip compression). 

However, if you want to filter out duplicates, there are several options you can currently already use: the simplest one is to use `QueryResults.distinctResults(GraphQueryResult)`, before sending the result to the Rio writer. Or you can first read the result into a `Model` object: since a `Model` is a `Set`, it automagically removes duplicate statements. 

So instead of making this the responsibility of the writer itself I'd suggest we document the available options for de-duplication, and if there are cases where we need something extra we introduce utility functions to allow users to optionally filter out duplicates.

pulquero:
I was thinking maybe there should be a RIO setting to enable dedup (off by default). It would be a natural compliment to the pretty_print option.

I would also like to see a "distinct" query parameter option added to the sesame server to enable it. As a user, I don't want to have to write/hack code just to get a canonical RDF+XML output from the server.

jeenbroekstra:
If the scenario you have in mind is an option in the RDF4J Server / Workbench, then that sounds very useful. But this can be easily implemented without adding a Rio setting. 

I'm reluctant to add a Rio setting, mainly because the current options are, in my view, good enough to solve most cases and adding more options just makes the API more cluttered. The situation is different with pretty-printing, where there is no viable other approach; it _has_ to be implemented inside the writer.

PR #567, opened 2016-09-11
Issues/#319 distinct query param

pulquero:
This PR addresses GitHub issue: #319  .

* Added distinct query parameter.


core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/RepositoryController.java
```diff
@@ -363,4 +378,18 @@ private IRI createURIOrNull(Repository repository, String graphURI) {
 			return null;
 		return repository.getValueFactory().createIRI(graphURI);
 	}
+
+	private static QueryResult<?> distinct(QueryResult<?> qr) {
```
jeenbroekstra: Might be better to reuse `QueryResults.distinctResults` here. If necessary extend to accept `QueryResult<?> qr` as a param.

pulquero: Done.

PR #569, opened 2016-09-14
Improve the performance and create less garbage in the turtle parser

nikolavp:
This pull request will hopefully make the turtle parser a little bit faster and prevent a lot of memory allocations when parsing reasonably turtle files. The pull request contains two major changes:
* Reusing the string builder instead of creating a new one - this is good for memory usage and also prevents resizing throughout the parsing
* Character.toChars is creating a char array which is not actually needed. We can easily just reuse the code from there and append the result into the string builder


nikolavp:
Hmm I thought that I have already signed the CLA?

PR #571, opened 2016-09-14
Issue #572: Improve the performance and create less garbage in the turtle parser

nikolavp:
This pull request will hopefully make the turtle parser a little bit faster and prevent a lot of memory allocations when parsing reasonably turtle files. The pull request contains two major changes:

Reusing the string builder instead of creating a new one - this is good for memory usage and also prevents resizing throughout the parsing
Character.toChars is creating a char array which is not actually needed. We can easily just reuse the code from there and append the result into the string builder

Issue that describes how we found about the problem https://github.com/eclipse/rdf4j/issues/572


core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
```diff
@@ -0,0 +1,30 @@
...
+ * @author Nikola Petrov nikola.petrov@ontotext.com
+ */
+class Chars {
+	/**
+	 * Appends the characters from codepoint into the string builder. This is the same
+	 * as Character#toChars but prevents the additional char array garbage.
+	 * @param dst the destination in which to append the characters
+	 * @param codePoint the codepoint to be appended
+	 */
+	public static void append(StringBuilder dst, int codePoint) {
```
pulquero: Personally, I would have just stuck this method in TurtleParser rather than having a new class for it. I'm not sure whether there will be any license issue with taking code from the jdk. I guess it depends on which? openjdk, harmony, that gnu one, etc.

pulquero:
You should probably create an associated issue for this to be accepted.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
```diff
@@ -0,0 +1,30 @@
+package org.eclipse.rdf4j.rio.turtle;
```
ansell: This file is missing the license header

jeenbroekstra: Copyright/license header is missing

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
```diff
@@ -0,0 +1,30 @@
...
+class Chars {
+	/**
+	 * Appends the characters from codepoint into the string builder. This is the same
+	 * as Character#toChars but prevents the additional char array garbage.
+	 * @param dst the destination in which to append the characters
+	 * @param codePoint the codepoint to be appended
+	 */
+	public static void append(StringBuilder dst, int codePoint) {
+		/*
+		 * This code is copy pasted from Character#toChars but doesn't allocate new arrays.
```
ansell: It may be more accurate to say, "doesn't allocate new arrays for BMP codepoints" or similar.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
```diff
@@ -0,0 +1,30 @@
...
+		 */
+		if (Character.isBmpCodePoint(codePoint)) {
+			dst.append((char)codePoint);
+		} else if (Character.isValidCodePoint(codePoint)) {
+			char[] result = new char[2];
+			// NB: Code copied Character#toSurrogates
+			dst.append(Character.highSurrogate(codePoint));
+			dst.append(Character.lowSurrogate(codePoint));
+		} else {
+			throw new IllegalArgumentException();
```
ansell: Having the codepoint that was invalid in the exception message here may help in future if this gets triggered from some strange place.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -70,6 +70,7 @@
 	protected Value object;
 
 	private int lineNumber = 1;
+	private StringBuilder parsingBuilder = new StringBuilder();
```
ansell: This should be declared as final to ensure that it is recognised as a permanent placeholder that can't be accidentally reassigned.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
jeenbroekstra: Yeah, this seems so specific that I'd prefer it as a protected/private method in TurtleParser, rather than a public utility. Also agree that we need to be certain this can be included, license-wise.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -1318,4 +1319,9 @@ protected void throwEOFException() throws RDFParseException {
 	protected int getLineNumber() {
 		return lineNumber;
 	}
+
+	public StringBuilder getBuilder() {
```
jeenbroekstra: This should probably be private or at least protected

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -304,7 +305,7 @@ protected void parsePrefixID() throws IOException, RDFParseException, RDFHandler
 				throwEOFException();
 			}
 
-			prefixID.append(Character.toChars(c));
+			Chars.append(prefixID, c);
```
ansell: prefixID could be a private final member variable in addition to the other StringBuilder.

jeenbroekstra:
This new github review system will take some getting used to. Better than it was though :)

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -1069,8 +1070,8 @@ protected BNode parseNodeID() throws IOException, RDFParseException {
 			reportError("Expected a letter, found '" + (char) c + "'", BasicParserSettings.PRESERVE_BNODE_IDS);
 		}
 
-		StringBuilder name = new StringBuilder(32);
-		name.append(Character.toChars(c));
```
ansell: The other use of name.append below could also be modified, as its casting straight to a char may not be correct in all cases (I can't remember now if there are only BMP characters allowed for blank node label characters, but safer to reuse the Chars.append method instead of a risky cast.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
ansell: This should be a protected method so it isn't exposed to users, just to subclasses as necessary.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
nikolavp: The snippet was taken from OpenJDK, I guess there is no problem with it as long as the result is open source again. 

I put the class as package private and created a dedicated class so we can update it easier in the future(if there is a need). But OK will move it in TurtleParser.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
nikolavp: True, thought that the maven build process is adding one for me. I will remove the file anyways.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
nikolavp: Oh uh. Missed that one as it was written before me. Will update it.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
nikolavp: Done :)

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
nikolavp: Yes, I guess the IDE extracted with a public modifier for some reason.

nikolavp:
Please check the new changes. Also linked a small summary of how we found about the memory usage being a problem.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Chars.java
jeenbroekstra: Unfortunately, it is not quite that simple. OpenJDK source code is licensed under GPL, which means we can't reuse it in our own code (unless we then distribute the entire RDF4J framework under GPL license as well - which is not something we want to do :)).

Given that this is such a small code snippet though, and a really quite basic algorithm, I'll doublecheck with the Eclipse legal team if we can get away with this somehow.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -1320,8 +1320,30 @@ protected int getLineNumber() {
...
+
+	/**
+	 * Appends the characters from codepoint into the string builder. This is the same
+	 * as Character#toChars but prevents the additional char array garbage for BMP codepoints.
+	 * @param dst the destination in which to append the characters
+	 * @param codePoint the codepoint to be appended
+	 */
+	private static void appendCodepoint(StringBuilder dst, int codePoint) {
+		/*
+		 * This code is copy pasted from Character#toChars but doesn't allocate new arrays.
```
jeenbroekstra: On the license issue: given that this is _such_ a basic/simple bit of code, I think that we can simply remove this specific comment to avoid confusion later. Other than that the PR (with the latest changes) looks good to me!

jeenbroekstra:
Good stuff, thanks Nikola. Apart from the licensing issue (which may turn out to be no issue at all given it's only ~10 lines of code) this now looks fine to me.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
ansell: I will submit a follow up patch to this to reuse your method for prefixID here. It should be okay as it is a distinct branch in parseDirective (but the call to prefixID.toString below would need to be above the call to parseURI that would otherwise clear its StringBuilder instance.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -1318,4 +1319,31 @@ protected void throwEOFException() throws RDFParseException {
...
```
jeenbroekstra: Someone at Eclipse pointed out that you really only copy 3 lines of code and that is so trivial that it is unlikely to cause a problem. So I'm fine with just including this. However, to avoid future confusion/questions, can you remove this comment about it being "copy pasted"?

nikolavp:
Removed the comment as per the comment from pulquero so it avoids confusion.

jeenbroekstra:
Running hudson build on this PR now: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/18/ 

If nothing breaks this can be merged.

PR #578, opened 2016-09-23
Issue #577  : added logback logger for elasticsearch tests

### LINKED ISSUE #577: RDF4J Elastic Search shows only INFO level logs
I use the tests quite intensively and need have more detailed logs than INFO. I will replace *slf4j-simple* implementation to *log4j2* one but for tests only.

jgrzebyta:
This PR addresses GitHub issue: #577  .

*  Implement log4j2 implementation:
 * add log4j2.yml config file
 * add required dependencies: log4j:2.x, YAML files handler


jeenbroekstra:
Given that we use logback almost exclusively everywhere else in the framework, would it be too much to ask to change this to use logback as well, instead of log4j?

core/sail/fts/elasticsearch/pom.xml
```diff
@@ -15,6 +15,8 @@
 
 	<properties>
 		<elasticsearch.version>1.7.0</elasticsearch.version>
+                <log4j.version>2.6.2</log4j.version>
+                <fasterxml.jackson.version>2.8.3</fasterxml.jackson.version>
```
jeenbroekstra: Can't upgrade a 3rd party dependency without filing a CQ and getting legal approval first.

core/sail/fts/elasticsearch/pom.xml
```diff
@@ -47,11 +49,35 @@
 			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
-		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-simple</artifactId>
-			<version>${slf4j.version}</version>
-			<scope>test</scope>
-		</dependency>
+                <dependency>
```
jeenbroekstra: I'd prefer instead switching to logback, and including a logback-text.xml config file in src/test/resources, like it's done in other parts of the code base. Makes things more consistent, and also avoids adding new dependencies (and having to bump the jackson version).

core/sail/fts/elasticsearch/pom.xml
jeenbroekstra: Also: the jackson version is declared at the project root pom level, it should not be set here.

jeenbroekstra:
Or even simpler: stick with using SFL4J SimpleLogger, and just configure it to enable debug-level logging. See http://www.slf4j.org/api/org/slf4j/impl/SimpleLogger.html for details.

jgrzebyta:
Done. I just added `simplelogger.properties` file with defined logger.

core/sail/fts/elasticsearch/pom.xml
```diff
@@ -1,57 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>rdf4j-lucene-testsuite</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
```
jeenbroekstra: Both logback deps should be scoped to `test` - this to avoid them being included in the production build.

jgrzebyta: All done

jeenbroekstra:
Running hudson to verify PR: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/23/

Issue #97, opened 2016-05-11
Add a ModelBuilder


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-2354 )

Programmatically creating RDF triples is currently a bit of a juggle with ValueFactories and collections of Statements - not to mention a lot of boilerplate is involved in creating individual IRIs, BNodes and Literals.

A ModelBuilder class that provide convenience methods for quickly adding triples would be a useful but non-invasive addition to the model utilities package. Something along these lines:

```

Model m = new ModelBuilder().withNamespace("ex", "http://example.org/")
                                     .subject("ex:p1")
                                     .add("ex:phone", "555-0142")
                                     .add(RDF.TYPE, FOAF.PERSON).build();

```

jeenbroekstra:
```
Namespace n = new Namespace("http://www.example.org/");
IRI foo = n.resource("foo"); // http://www.example.org/foo
```

PR #581, opened 2016-09-27
Issue #97: fluent ModelBuilder

jeenbroekstra:
This PR addresses GitHub issue: #97 .

* Introduced a new ModelBuilder class that allows fluent creation of RDF Models.

Early draft for feedback purposes. Tests to follow.

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
@@ -54,7 +54,7 @@ import org.eclipse.rdf4j.model.vocabulary.XMLSchema;
  *    builder.defaultGraph().subject("ex:graph1").add(RDF.TYPE, "ex:Graph");
  *    
  *    // return the Model object
- *    Model m = builder.builder();
+ *    Model m = builder.build();
  * </code>
  * </pre>
  * 
@@ -150,7 +150,7 @@ public class ModelBuilder {
 	}
 
 	/**
-	 * Set the current graph in which to add new statements to the supplied named graph. This method resets
+	 * Set the current graph in which to add new statements to the supplied named graph. This method clears
 	 * the current subject.
 	 * 
 	 * @param namedGraph
@@ -164,7 +164,7 @@ public class ModelBuilder {
 	}
 
 	/**
-	 * Set the current graph in which to add new statements to the default graph. This method resets the
+	 * Set the current graph in which to add new statements to the default graph. This method clears the
 	 * current subject.
 	 * 
 	 * @return this {@link ModelBuilder}
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
@@ -1,0 +1,357 @@
+/**
+ * Copyright (c) 2015 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
```diff
@@ -0,0 +1,357 @@
...
+				model.setNamespace(ns);
+				return vf.createIRI(ns.getName(),
+						prefixedNameOrIRI.substring(prefixedNameOrIRI.indexOf(':') + 1));
+			}
+		}
+		return vf.createIRI(prefixedNameOrIRI);
+	}
+
+	private Namespace[] getDefaultNamespaces() {
+		return new Namespace[] { RDF.NS, RDFS.NS, OWL.NS, XMLSchema.NS, DC.NS, FOAF.NS, SKOS.NS };
```
ansell: Although there has been extensive discussion about the prefix for DCTERMS (we currently use dcterms but others prefer either dc or dct), it should be represented in the list here as there are many additions that don't map onto the original DC Elements 1.1 vocabulary.

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
```diff
@@ -0,0 +1,357 @@
...
+import org.eclipse.rdf4j.model.vocabulary.DC;
+import org.eclipse.rdf4j.model.vocabulary.FOAF;
+import org.eclipse.rdf4j.model.vocabulary.OWL;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.model.vocabulary.SKOS;
+import org.eclipse.rdf4j.model.vocabulary.XMLSchema;
+
+/**
+ * Builder to facilitate easy creation of new RDF {@link Model} objects.
```
ansell: Could you add a clarification that any methods which return ModelBuilder will always return an immutable reference to the current object. This is how it is implemented, but if it is part of the contract, then it may prevent people presuming that they need to reassign the reference to their ModelBuilder object if they use it in a for-loop.

jeenbroekstra:
ip-validation fails because I accidentally did a merge commit before I had properly reconfigured my git settings (I had to do a factory reset of my laptop yesterday...). 

I will cheerfully ignore it this time and just go ahead and merge when ready (the only alternative is a rebase and I really hate those).

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelBuilder.java
```diff
@@ -0,0 +1,360 @@
...
+		return add(mapToIRI(predicate), object);
+	}
+
+	/**
+	 * Return the created {@link Model}
+	 * 
+	 * @return the {@link Model}
+	 */
+	public Model build() {
+		return model;
```
jeenbroekstra: Question: I currently just return the original `model` object. This means that if further actions are invoked on the `ModelBuilder` _after_ the `Model` has been retrieved from the builder, those actions will influence that `Model`. Do we consider that acceptable/useful, or should we instead return a copy?

ansell: I was okay with that when I reviewed it, based on the Javadoc for the constructor where they would submit the model. If you wanted another option, a "buildImmutable()" or similar version that returns an unmodifiable copy of the Model may be useful, but the usual model for Java collections is to be mutable by default. 

In addition, if they submit a Model, we have already been modifying it before this point, so we would need to take the copy in the constructor rather than here.

PR #582, opened 2016-09-27
Fix problem with logger

jgrzebyta:
This PR addresses GitHub issue: #578  .

I found that logger's level in the config file should be declared in different way:

```
<logger name="my.superclass.full.of.errors.Foo" level="debug" />
```
* Change logger's definition according to documentation
* Change logging layout


jgrzebyta:
I found solution for my problem:

Loggers in the abstract classes should be `private static` to be registered otherwise the logger from subclasse handles logs.

Issue #299, opened 2016-08-26
Lower q value for RDF+XML format


RickMoynihan:
I've just ran into an issue where `CONSTRUCT` queries that constructed a predicate of the URI form: `http://blah.com/blah/(tailing-element-in-parens)` were failing.

It turns out that the above is a valid URI (though it contains reserved characters - this is legal).  This works fine for all serialisations except RDF+XML which fails.  Some more digging and it turns out that it only fails for URI's of this form in predicates; because of this [documented limitation of RDF+XML](https://www.w3.org/TR/REC-rdf-syntax/#section-Serialising).

I realise we currently use an algorithm to determine these qValues for `CONSTRUCTS` preferring Turtle the most, and then weighting it based upon whether we need named graphs etc; however given the limitations of RDF+XML I think we should also be penalising it by giving it a significantly lower qValue, ideally something like `;q=0.1`.

Are there any reasons not to do this?

ansell:
I don't see any reasons not to do it. Definitely need to keep application/rdf+xml on the list for compatibility, but a low q value is a good idea IMO. I wouldn't lower it further than 0.5 at first though. Even 0.5 should ensure with a properly implemented content negotiation process that you never get it if you say you can handle turtle with no q value or q=1.0. If we go lower than 0.5, it seems like we have issues processing/roundtripping it, as is the case for text/csv for Tuple results (per its ill designed spec), for instance. In this case we don't have round tripping issues, there are just issues generating it from triple stores in some cases.

jeenbroekstra:
Agreed with Peter on all points: no reason not to lower it, but 0.1 might be a bit harsh - 0.5 seems good, and gives you sufficient "wiggling room" on both ends of the scale to later fit in other formats with higher (or lower) preference. 

Of course an rigorous alternative (if you _really_ wish to avoid ever getting back RDF/XML) is to remove the RDF/XML from your runtime classpath; then the client library will not send that mime type in the Accept header at all. Of course this is at the cost of not being able to read the format if you're communicating with a service that has no alternatives (though in that case, changing the q-value won't make a difference either :)).

PR #586, opened 2016-09-29
Issue #299 added downgrading of RDF/XML q-value for accept headers

jeenbroekstra:
This PR addresses GitHub issue: #299  .

* q-value for RDF/XML is downgraded by 0.4 (so best-case scenario is it becomes 0.6)
* added failsafe to completely skip a format if the q-value goes below 0.1.

### DIFF
--- a/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/RDFFormat.java
+++ b/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/RDFFormat.java
@@ -264,14 +264,22 @@ public class RDFFormat extends FileFormat {
 				qValue -= 1;
 			}
 
-			for (String mimeType : format.getMIMETypes()) {
-				String acceptParam = mimeType;
+			if (RDFXML.equals(format)) {
+				// We explicitly dislike RDF/XML as it has limitations in what it can serialize. See #299.
+				qValue -= 4;
+			}
 
-				if (qValue < 10) {
-					acceptParam += ";q=0." + qValue;
+			// if the qValue did not go negative, we add this format to the accept params.
+			if (qValue > 0) { 
+				for (String mimeType : format.getMIMETypes()) {
+					String acceptParam = mimeType;
+
+					if (qValue < 10) {
+						acceptParam += ";q=0." + qValue;
+					}
+
+					acceptParams.add(acceptParam);
 				}
-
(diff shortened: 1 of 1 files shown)


core/rio/api/src/main/java/org/eclipse/rdf4j/rio/RDFFormat.java
```diff
@@ -264,14 +264,22 @@
...
+			if (RDFXML.equals(format)) {
+				// We explicitly dislike RDF/XML as it has limitations in what it can serialize. See #299.
+				qValue -= 4;
+			}
 
-				if (qValue < 10) {
-					acceptParam += ";q=0." + qValue;
-				}
+			// if the qValue did not go negative, we add this format to the accept params.
+			if (qValue > 0) {
```
ansell: I would prefer to have a minimum value rather than excluding them. If there are issues on the other side with low q-value formats being selected then I don't think it will be fixed by eliminating some that users know are on their classpath.

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/RDFFormat.java
```diff
@@ -264,14 +264,22 @@
 				qValue -= 1;
 			}
 
-			for (String mimeType : format.getMIMETypes()) {
-				String acceptParam = mimeType;
+			if (RDFXML.equals(format)) {
+				// We explicitly dislike RDF/XML as it has limitations in what it can serialize. See #299.
+				qValue -= 4;
```
ansell: It is possible with this change that RDF/XML gets excluded if exclusions are active below, as it doesn't support contexts, so -5, and it may not be the preferred format, so -2, which with -4 here, seems to add up to -11 which would make it go negative.

The goal of the issue wasn't to exclude RDF/XML, as it is far too popular for that, just to lower its priority.

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/RDFFormat.java
jeenbroekstra: I think I agree. `qValue = max(1, qValue)` or something along those lines?

ansell: That should work.

PR #590, opened 2016-09-29
Use repositories over TLS.

josephw:
Use an HTTPS URL for Restlet, so code isn't fetched over an insecure connection. Switch to their current preferred TLD of .com rather than .org (they redirect .org to .com over HTTP, but not HTTPS).

Adjust the Eclipse link at the same time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
 
 	<organization>
 		<name>Eclipse Foundation</name>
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
Hudson verify running at https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/27/

Issue #544, opened 2016-09-07
handle new logback based logging in RDF4J Console


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-568)

The aduna application base switched from log4j to logback for logging. The Console needs to be modified accordingly, most notably the "set log" command.

ansell:
Is this still an issue? I thought that we had switched to SLF4J-API for libraries and Logback for applications (console/workbench) already but I haven't verified that.

josephw:
The `set log` command [suggested by `PrintHelp`](https://github.com/eclipse/rdf4j/blob/e42f4eebc55bd8ac2245cd4883b2f4704bfa4880/core/console/src/main/java/org/eclipse/rdf4j/console/PrintHelp.java#L74) doesn't work ("`unknown parameter: log`") since [SES-750](https://github.com/eclipse/rdf4j/commit/3349b45474d531cd1b483ca5fd5da2f2ed9270f1#diff-de6b8752185c42821ad34923dbeec7cbL1656).

PR #591, opened 2016-09-29
Issue #544: reimplement 'set log' for Logback

josephw:
This PR addresses GitHub issue: #544.

* Add tests for 'set log'
* Bring back the previous JDK logging implementation and adjust for Logback.

### DIFF
--- a/core/console/src/main/java/org/eclipse/rdf4j/console/SetParameters.java
+++ b/core/console/src/main/java/org/eclipse/rdf4j/console/SetParameters.java
@@ -7,10 +7,31 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.console;
 
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.slf4j.LoggerFactory;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+
 /**
  * @author dale
  */
 public class SetParameters implements Command {
+
+	private static final Map<String, Level> LOG_LEVELS;
+
+	static {
+		Map<String, Level> logLevels = new LinkedHashMap<String, Level>();
+		logLevels.put("none", Level.OFF);
+		logLevels.put("error", Level.ERROR);
+		logLevels.put("warning", Level.WARN);
+		logLevels.put("info", Level.INFO);
+		logLevels.put("debug", Level.DEBUG);
+		LOG_LEVELS = Collections.unmodifiableMap(logLevels);
+	}
 
 	private final ConsoleIO consoleIO;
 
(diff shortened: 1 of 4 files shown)


core/console/src/test/java/org/eclipse/rdf4j/console/SetParametersTest.java
```diff
@@ -0,0 +1,90 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: Wrong header for new files. Should just read "Copyright (c) 2016 Eclipse RDF4J contributors."

Which reminds me that I need to update the code template to have the correct year.

core/console/src/main/java/org/eclipse/rdf4j/console/SetParameters.java
```diff
@@ -65,6 +90,29 @@ else if ("queryprefix".equalsIgnoreCase(key)) {
...
+	private void setLog(final String value) {
+		// Assume Logback
+		Logger logbackRootLogger = (Logger)LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
+
+		if (value == null) {
+			Level currentLevel = logbackRootLogger.getLevel();
+
+			String levelString = LOG_LEVELS.inverse().getOrDefault(currentLevel, currentLevel.levelStr);
+
+			consoleIO.writeln("log: " + levelString);
```
ansell: In the case where they successfully set the logger below, the log level should be written out also?

josephw: The other `set`s don't seem to echo the new value; worth making this one a special case because it so directly affects output?

jeenbroekstra:
After merging this, the master branch has become unstable. See discussion on rdf4j-dev. Unless we can find the culprit and do a quick fix, we may need to revert this merge.

josephw:
I can reproduce the failure by upgrading `maven-surefire-plugin` to 2.19.1 and running the `core/console` tests with `-Dsurefire.runOrder=random`. When failure happens, the new test is run first. That suggests that one of the tests (the new one?) isn't setting up or tearing down its mocking correctly - at the very least, it's leaving the logging in a visibly different state.

If it's only a problem with the tests, consider `git rm`-ing the test I've added.

josephw:
@jeenbroekstra PR #611 includes a fix that works for me.

jeenbroekstra:
@josephw we have those - sort of. I can manually initiate a branch build on a PR from our [Hudson server](https://hudson.eclipse.org/rdf4j), and I usually do that for every PR, but in this case I just forgot and got trigger-happy with the merge button. Btw if you become a full committer to the project you will have access to Hudson yourself :)

PR #160, opened 2016-05-25
Issue #163: [Query Algebra] Allows to compare date times with different datatypes

Tpt:
[Query Algebra] Allows to compare date times with different datatypes

e.g. "2005"^^sd:gYear < "2015-01"^^xsd:gYearMonth

Adds also some unit tests

This behaviour seems to me more in the spirit of the SPARQL 1.1
specification that considers only the xsd:dateTime datatype

PR #595, opened 2016-10-04
Issue #164: extended query eval strategy

jeenbroekstra:
This PR addresses GitHub issue: #164 .

* `SimpleEvaluationStrategy` renamed to `StrictEvaluationStrategy`
* new class `ExtendedEvaluationStrategy` does virtual property evaluation and extended datetime literal comparison
* includes changes supplied by @Tpt in PR #160 
* added config option to Workbench for memory and native stores. 

This is an early PR, not quite ready for merging as it needs further testing and probably also some additional tweaks to complete the feature set. Nevertheless, feedback welcome.


jeenbroekstra:
Hudson build started: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/29/

jeenbroekstra:
Hudson build succeeds. However, for some reason when I run `mvn verify` on my local machine, testing hangs somewhere in `CustomGraphQueryInferencerNativeRepositoryConnectionTest`. I have never seen this before. It may be a quirk in my local setup - can someone run a clean `mvn verify` on this branch from their local machine and let me know if they see the same problem?

jeenbroekstra:
Can I get some eyeballs on this? @Tpt , @pulquero, @ansell, @nikolavp ?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
```diff
@@ -123,9 +129,48 @@ private static boolean valuesEqual(Value leftVal, Value rightVal) {
...
+	 * @param strict
+	 *        boolean indicating whether comparison should use strict (minimally-conforming) SPARQL 1.1
+	 *        operator behavior, or extended behavior.
+	 * @return {@code true} if execution of the supplied operator on the supplied arguments succeeds,
+	 *         {@code false} otherwise.
+	 * @throws ValueExprEvaluationException
+	 *         if a type error occurred.
+	 */
+	public static boolean compareLiterals(Literal leftLit, Literal rightLit, CompareOp operator,
+			boolean strict)
```
ansell: I am not sure whether a boolean for strict->extended will be maintainable in the long term. Ie, will everyone want all of the extended behaviours or will they want to customise which ones/categories they want to allow. May not have to answer that question right now though.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/limited/iterator/LimitedSizeZeroLengthPathIteration.java
```diff
@@ -35,7 +35,7 @@
 	 * @param contextVar
 	 * @param bindings
 	 */
-	public LimitedSizeZeroLengthPathIteration(SimpleEvaluationStrategy evaluationStrategyImpl, Var subjectVar,
```
ansell: What obstacles are there to having this take EvaluationStrategy instead of a specific implementation?

jeenbroekstra: None, AFAIK. I'll take a closer look.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
jeenbroekstra: If we start making finer distinctions we can replace the boolean with something like an enum, I guess. But I prefer not to over-engineer at this point - it seems unlikely to me that we will very quickly implement fine-grained picking-and-choosing of extended behaviors. For one thing it will make testing complex, and for another, it puts a burden on third-party Sail implementors. So all for two flavors for now: either strict, or extended, where extended is just whatever we feel is useful.

pulquero: I generally dislike the use of a boolean flag in this context too, but since there is a lot of common code between the two flavours, it seems reasonable to me. To avoid it, you would probably have to turn that method into a class with fine-grained methods that could be overridden to extend the comparison behaviour. Or perhaps have the method return something like an Optional<Boolean>, notPresent being return when there is no commonDataType. A caller could then check for this to decide whether to delegate the comparison to some compareExtendedLiterals() method, e.g. caller code could look something like: return coalesce(compareStrictLiterals(...), compareExtendedLiterals(...)

pulquero:
Looks like how I would imagine it would look.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/limited/iterator/LimitedSizeZeroLengthPathIteration.java
jeenbroekstra: Fixed, see update.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/SimpleEvaluationStrategyTest.java
```diff
@@ -31,7 +31,7 @@
 	public void setUp()
```
jeenbroekstra: Good call, done.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
jeenbroekstra: Yes, exactly, a design with more fine-grained control might be better, but it would mean significant refactoring and basically rewriting QueryEvaluationUtil - I don't think we're at the point where we need to do that, yet.

PR #611, opened 2016-10-10
Issue #544: Restore same log level.

josephw:
This PR addresses the build break from merging #591.

Restore the same log level from the start of the test, rather
than assuming that it's DEBUG.

If I run the tests using a more recent Surefire and `-Dsurefire.runOrder=random`, the tests fail before this change, and pass afterwards.

### DIFF
--- a/core/console/src/test/java/org/eclipse/rdf4j/console/SetParametersTest.java
+++ b/core/console/src/test/java/org/eclipse/rdf4j/console/SetParametersTest.java
@@ -12,6 +12,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.mockito.InjectMocks;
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
Running hudson verify on this PR: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/32/

jeenbroekstra:
Can you elaborate on what exactly you think the problem was? Because I find it odd that simply fixing the log level setting would influence the test outcomes this way.

jeenbroekstra:
Hudson build succeeded, ready to merge IMHO.  Thanks @josephw !

ansell:
The proposed fix here indicates that the log level setting code is still broken. Ie, the fact that only a test was changed indicates that there is still a bug in the actual code that was recently merged in relation to the log level setting.

jeenbroekstra:
That is true, which is why I am keen to hear more details on _what_ exactly caused the issue in the first place (and why this fix resolves it). I merged immediately because I wanted the branch stable, but it'd be good to discuss this a bit further.

ansell:
I am working through the SetParameters code to see if I can improve it, particularly to not rely on null as a sentinel.

ansell:
The Federate command class has a switch on log.isDebugEnabled() at line 77 that affects control flow. I.e., the ROOT log level that is being changed by SetParametersTest has side effects on Federate/FederateTest. This behaviour doesn't look to be documented (we don't say that if a user sets log level to debug that federation will be not performed). 

This may have been the first time that this code has actually been run through the testsuite. Code coverage tools would have picked this up as it is a fairly glaring mistake.

I am working on a pull request to patch it.

ansell:
@jeenbroekstra Debugging this included updating Mockito to version 2.2.0 that has some extra debugging features, and bumping the Logback version to the latest to verify it wasn't related. 

Do you want me to open an issue to get legal approval to use Mockito 2.2.0 and Logback 1.1.7 and include it in a pull request or back out those version bumps?

jeenbroekstra:
Neither of those versions are existing in Orbit yet, so they haven't been pre-approved. This close to release I'd prefer not to file CQs that might hold up release review. Can you back out of those bumps for the immediate fix and log issues to update these libraries for a later release?

josephw:
> The Federate command class has a switch on log.isDebugEnabled()

Thanks, @ansell - that explains what I saw, and why this change to the test has any effect. I didn't have time to dig in, so I appreciate learning what's going on.

Issue #609, opened 2016-10-07
Provide constants for the SKOS-XL vocabulary


manuelfiorelli:
Package `org.eclipse.rdf4j.model.vocabulary` should be extended with a utility class for the [SKOS eXtension for Labels (SKOS-XL)](https://www.w3.org/TR/skos-reference/#xl) vocabulary. It is worth mentioning that the aforementioned package already contains a utility class for the SKOS vocabulary. Therefore, the addition of another class for a standard extension (supporting reified labels) should be acceptable.

ansell:
This sounds useful given it has been standardised as part of the W3C SKOS standard.

manuelfiorelli:
I've already written the utility class. If you like, I can make a pull request in a day or two.

PR #612, opened 2016-10-10
issue #609: provide vocabulary constants for SKOS-XL

manuelfiorelli:
This PR addresses GitHub issue: #609 

* Add a vocabulary class for the namespace SKOS-XL

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SKOSXL.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SKOSXL.java
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) 2016 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.eclipse.rdf4j.model.vocabulary;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Namespace;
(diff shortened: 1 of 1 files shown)


manuelfiorelli:
I forgot to add a sign-off header. What should I do?

jeenbroekstra:
Given that this is only in your private fork sofar, you should be able to amend your commits with the sign-off message. See http://stackoverflow.com/questions/13043357/git-sign-off-previous-commits .

manuelfiorelli:
Hi @jeenbroekstra thank you for the support. Although I speculated about the use of `--amend`, I was afraid of messing things up (by the way, I hesitated before adding `--force` to the push command). That said, it seems that the operation was successful.

core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SKOSXL.java
```diff
@@ -0,0 +1,104 @@
...
+	 * @see <a href="https://www.w3.org/TR/skos-reference/#xl-Label">The skosxl:Label Class</a>
+	 */
+	public static final IRI LABEL;
+
+	/* literal form */
+
+	/**
+	 * The skosxl:literalForm property.
+	 * 
+	 * @see <a href="https://www.w3.org/TR/skos-reference/#xl-Label">The skosxl:Label Class</a>
```
jeenbroekstra: Needs to point to #xl-literalform instead (I assume).

jeenbroekstra: Ah, sorry, I saw this and assumed it was a copy-paste error (because it's the same links as the javadoc entry above). But I now see both features are documented in the same section, so it's all fine as-is then.

PR #615, opened 2016-10-11
issue #614 : Console Federate command broken by log level debug

### LINKED ISSUE #614: Console Federate command broken by log level debug
The Console Federate command is broken if log level is set to debug due to the incorrect use of log.isDebugEnabled for control flow.

Debugging this involved updating Mockito to 2.2.0 and Logback to 1.1.7, but those changes don't seem to be related to the solution so they can be backed out if legal can't approve by the time of the next planned release.

ansell:
This PR addresses GitHub issue: #614 .

* The Console Federate command was using log level for control flow, resulting in federation failing to occur if the log level was debug or finer.

Debugging this involved updating Mockito to 2.2.0 and Logback to 1.1.7, but those changes don't seem to be related to the solution so they have been backed out due to the close release time.

Debugging also involved refactoring SetParametersTest to reuse AbstractCommandTest as designed, and included refactoring to use MockitoRule, @Mock and @InjectMocks to verify they were not related and can be maintained in future.

There is now a known baseline log level of debug setup for both FederateTest and SetParametersTest, based on their need to consistently verify console IO messages that may be affected by the log level and should function properly even at the debug log level. This is reset after each test so the global base logging level will still be able to be manipulated through a configuration file as necessary for other tests.

Fixes #614

### DIFF
--- a/core/console/src/main/java/org/eclipse/rdf4j/console/Federate.java
+++ b/core/console/src/main/java/org/eclipse/rdf4j/console/Federate.java
@@ -77,34 +77,32 @@ public class Federate implements Command {
 		if (LOGGER.isDebugEnabled()) {
 			logCallDetails(distinct, readonly, fedID, memberIDs);
 		}
-		else {
-			RepositoryManager manager = state.getManager();
-			try {
-				if (manager.hasRepositoryConfig(fedID)) {
-					cio.writeError(fedID + " already exists.");
-				}
-				else if (validateMembers(manager, readonly, memberIDs)) {
-					String description = cio.readln("Federation Description (optional):");
-					RepositoryManagerFederator rmf = new RepositoryManagerFederator(manager);
-					rmf.addFed(fedID, description, memberIDs, readonly, distinct);
-					cio.writeln("Federation created.");
-				}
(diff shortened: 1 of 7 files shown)


ansell:
I am running the full testsuite locally to verify that the changes didn't break anything else but that is unlikely given that the version bumps have been backed out now.

core/console/src/main/java/org/eclipse/rdf4j/console/Federate.java
```diff
@@ -77,35 +77,33 @@ private void federate(boolean distinct, boolean readonly, String fedID, Deque<St
 		if (LOGGER.isDebugEnabled()) {
```
ansell: This is where the bug occurred, this should never have had an else clause linked to it, as that linked it to control flow.

ansell:
All tests are passing for me after these changes, where the FederateTest was failing for me previously.

core/console/src/main/java/org/eclipse/rdf4j/console/Federate.java
jeenbroekstra: Well spotted!

PR #618, opened 2016-10-13
Fixes #617: Tabs in update with federate query are not parsed correctly.

### LINKED ISSUE #617: Updates with a federated service in the WHERE clause are not parsed correctly if tabs are present before and inside the SERVICE block
Observed: syntactically valid federated query cannot be parsed by the remote endpoint. Minimal update to reproduce (note the leading blanks are tabs):
```
INSERT
{
	?a ?b ?c .
} WHERE {
	SERVICE <http://localhost>
		{
			?a ?b ?c .
		}
    }

```
Cause: Tab characters in updates are miscounted (treated as taking more than 1 characters) and thus the SERVICE part of the WHERE clause gets cut at the wrong offsets and a syntactically invalid query is sent to the remote endpoint.

avataar:
This PR addresses GitHub issue: #617 .

The fix addresses the issue by setting the tab size to 1 when parsing updates (copied the same code that is applied to parsing queries). I also edited the generated SyntaxTreeBuilder.java instead of rerunning jjtree/javacc on sparql.jjt (although I verified that works).

Also updated package names in sparql.jjt to org.eclipse.rdf4j as they seem to have been forgotten.


jeenbroekstra:
Running hudson verify: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/34/

jeenbroekstra:
Hudson verify succeeded.

Issue #576, opened 2016-09-19
Allow store-specific custom parameters to be injected via Repository/SAIL API


abrokenjester:
The Repository and SAIL APIs should be extended to allow for custom (store-specific) parameters to be injected.

jeenbroekstra:
More generally, we need a generalized parameter mechanism instead of (or rather, next to) the current store-specific getters and setters (setForceSync, setIterationCacheSyncThreshold, setDatadir, etc etc).

PR #624, opened 2016-10-20
#164 critical pre-release hotfix in native and memory store strategy selection

jeenbroekstra:
This PR addresses GitHub issue: #164 .

* hotfix/workaround for issue with configuration of evaluation strategy in memory and native stores.
* both stores now use the provided Factory to create an EvaluationStrategy.
* EvaluationStrategy interface extended to allow passing through of `iterationCacheSyncThreshold` param.

We may want to revisit the design later, as it all is a bit tightly coupled. However, this is a hotfix and unless somehow has strong objections I'd like this merged in so we can proceed with release.

@pulquero, @ansell , your feedback in particular is appreciated.

### DIFF
--- a/compliance/store/src/test/java/org/eclipse/rdf4j/repository/sail/memory/MemoryEvaluationStrategyTest.java
+++ b/compliance/store/src/test/java/org/eclipse/rdf4j/repository/sail/memory/MemoryEvaluationStrategyTest.java
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) 2016 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.eclipse.rdf4j.repository.sail.memory;
+
+import org.eclipse.rdf4j.repository.EvaluationStrategyTest;
+import org.eclipse.rdf4j.sail.base.config.BaseSailConfig;
+import org.eclipse.rdf4j.sail.memory.config.MemoryStoreConfig;
+
+
+/**
+ * @author jeen
+ *
+ */
+public class MemoryEvaluationStrategyTest extends EvaluationStrategyTest {
+
+	@Override
+	protected BaseSailConfig getBaseSailConfig() {
+		return new MemoryStoreConfig(false);
+	}
+
+}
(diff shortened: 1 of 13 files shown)


jeenbroekstra:
Hudson verify: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/38/

jeenbroekstra:
Hudson detected a compilation failure. Fix coming up.

jeenbroekstra:
New hudson job running: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/39/

jeenbroekstra:
Build succeeds.

jeenbroekstra:
Somewhat tricky to provide a solid regression test for this as programmatic configuration of extended evaluation strategy is difficult to access. I'll cobble something together after release based on the ad-hoc test code I wrote to debug the issue but I'm guessing we will want to provide a few convenience methods later on to make this a bit easier.

pulquero:
In sailsourceconnection the method is correctly implemented. It is overriden in the subclasses as the sails have the sync parameter which only made sense with strict. To support sync in the general case we probably need to add setparameter/attribute methods to the factory interface or a sub interface then sail.setsync just delegates to factory.setparameter. Unfortunately I have a flight to catch so can't do anything to the respect.

jeenbroekstra:
No worries, and thanks for your input. 

I'm not sure why you say the parameter only makes sense with the strict strategy - it seems just as relevant for the TupleFunctionStrategy. What it controls is the behavior of internal iterations during query evaluation (in particular when sorting or grouping, which internally cache results) - it sets a threshold at which the evaluation strategy will start syncing cached results to disk to avoid out of memory errors. 

So I think there's a definite case to make this a general parameter (I'm all for getting a better name by the way: `iterationCacheSyncThreshold` is rather a mouthful). But refactoring for this is is out of scope for this particular patch - I want to get a release out today. 

I'd like to schedule issue #576, which is about a more general design for setting Sail parameters in a generic manner, for the next minor release. We can take this discussion up as part of that issue.

jeenbroekstra:
What I'll try and do in the short term is remove the overloaded create method from the Factory class, and instead introduce a setter for that particular parameter. That changes existing behavior less.

jeenbroekstra:
Cleaned up setup a bit, and added a simple regression test. 

Hudson verify running now: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/40/ . If this greenlines I will merge and go ahead with release.

Issue #630, opened 2016-10-26
IllegalMonitorStateException in Transaction handling over HTTP


jeenbroekstra:
During transactions executed over HTTP an exception can occur when commit is invoked.

Stacktrace:
```
org.eclipse.rdf4j.http.server.ServerHTTPException: Transaction handling error: null
        at org.eclipse.rdf4j.http.server.repository.transaction.TransactionController.processModific
ationOperation(TransactionController.java:311)
        at org.eclipse.rdf4j.http.server.repository.transaction.TransactionController.handleRequestI
nternal(TransactionController.java:194)
        at org.springframework.web.servlet.mvc.AbstractController.handleRequest(AbstractController.j
ava:147)
        at org.springframework.web.servlet.mvc.SimpleControllerHandlerAdapter.handle(SimpleControlle
rHandlerAdapter.java:50)
        at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:959)
        at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:893)
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:970
)
        at org.springframework.web.servlet.FrameworkServlet.doPut(FrameworkServlet.java:883)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:651)

....

Caused by: java.lang.IllegalMonitorStateException: null
        at java.util.concurrent.locks.ReentrantLock$Sync.tryRelease(ReentrantLock.java:151)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.release(AbstractQueuedSynchronizer.java:1261)
        at java.util.concurrent.locks.ReentrantLock.unlock(ReentrantLock.java:457)
        at org.eclipse.rdf4j.sail.nativerdf.NativeSailStore$NativeSailSink.close(NativeSailStore.java:341)
        at org.eclipse.rdf4j.sail.nativerdf.SailSourceModel.dataset(SailSourceModel.java:397)
        at org.eclipse.rdf4j.sail.nativerdf.SailSourceModel.iterator(SailSourceModel.java:299)
        at org.eclipse.rdf4j.sail.nativerdf.MemoryOverflowModel.iterator(MemoryOverflowModel.java:128)
        at org.eclipse.rdf4j.sail.base.SailSourceBranch.flush(SailSourceBranch.java:536)
[...]

PR #631, opened 2016-10-26
Issue #630 simplified concurrent locking code

jeenbroekstra:
This PR addresses GitHub issue: #630 .

* simplified locking code in NativeSailStore and MemorySailStore: eliminated use of separate boolean value to track locking state 

This is not yet ready for merging, as I need to run further tests to confirm that this is consistent and that it indeed addresses the observed issue in #630. Nevertheless, reviews welcome.

### DIFF
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
@@ -384,8 +384,6 @@ class MemorySailStore implements SailStore {
 
 		private volatile Set<StatementPattern> observations;
 
-		private volatile boolean txnLock;
-
 		private boolean requireCleanup;
 
 		public MemorySailSink(boolean explicit, boolean serializable)
@@ -409,7 +407,7 @@ class MemorySailStore implements SailStore {
 			else {
 				sb.append("inferred ");
 			}
-			if (txnLock) {
+			if (txnLockManager.isHeldByCurrentThread()) {
 				sb.append("snapshot ").append(nextSnapshot);
 			}
 			else {
@@ -458,7 +456,7 @@ class MemorySailStore implements SailStore {
 		public synchronized void flush()
 			throws SailException
 		{
-			if (txnLock) {
+			if (txnLockManager.isHeldByCurrentThread()) {
 				currentSnapshot = Math.max(currentSnapshot, nextSnapshot);
 				if (requireCleanup) {
 					scheduleSnapshotCleanup();
@@ -469,9 +467,7 @@ class MemorySailStore implements SailStore {
 		@Override
 		public void close() {
 			try {
-				boolean toCloseTxnLock = txnLock;
-				txnLock = false;
-				if (toCloseTxnLock) {
+				if (txnLockManager.isHeldByCurrentThread()) {
 					txnLockManager.unlock();
 				}
 			}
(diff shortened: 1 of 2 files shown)


core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
```diff
@@ -569,13 +565,12 @@ public synchronized void deprecate(Resource subj, IRI pred, Value obj, Resource
 			}
 		}
 
-		private void acquireExclusiveTransactionLock()
+		private synchronized void acquireExclusiveTransactionLock()
```
ansell: Private methods are rarely synchronized, as it is usually done on public/protected/package methods which then already have the monitor lock when calling into private methods. Is there a particular reason why this one needs to be synchronized or should we synchronize all of the methods that call it?

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeSailStore.java
```diff
@@ -435,24 +428,14 @@ public void deprecate(Resource subj, IRI pred, Value obj, Resource ctx)
...
-						tripleStore.startTransaction();
-					}
-					nextTxnLockAcquired = txnLockAcquired = true;
+					txnLockManager.lock();
+					tripleStore.startTransaction();
 				}
 				catch (IOException e) {
 					throw new SailException(e);
 				}
-				finally {
```
ansell: The lock needs to be released if any exception occurs. Simplest way of doing this is a local boolean (see the "allGood" booleans introduced in #293) that are switched to true after tripleStore.startTransaction returns successfully within the try, to stop the lock being immediately released here. That way we don't need to catch all exceptions, but we can safely reset the state to unlocked if any exceptions are thrown.

jeenbroekstra: Good point, will fix.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
jeenbroekstra: I synchronized this to bring it in line with the same method in NativeSailStore. You're right though that we should probably look at the calling public methods. I'll have a closer look and adapt where necessary.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
```diff
@@ -569,13 +565,12 @@ public synchronized void deprecate(Resource subj, IRI pred, Value obj, Resource
...
 			throws SailException
 		{
-			if (!txnLock) {
+			if (!txnLockManager.isHeldByCurrentThread()) {
```
ansell: Ok

ansell:
With the changes it LGTM

jeenbroekstra:
Thanks Peter. I will let Hudson loose now, and will also run some local tests to see that any of this actually helps with the exception reported in #630.

jeenbroekstra:
Hudson verify: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/41/

jeenbroekstra:
Hudson build failing on 12 tests in (mostly) memorystore, so there's something I overlooked. Investigating.

jeenbroekstra:
On a positive note the fixes do seem to take care of the exception in NativeStore that I reported. If all else fails I may simply roll back the changes to the MemoryStore (I did not observe the original bug in MemoryStore) and only leave the fix in place for NativeStore.

PR #633, opened 2016-10-28
Issue: #630 TransactionController threading 

jeenbroekstra:
This PR addresses GitHub issue: #630 .

* new class Transaction ensures single-threaded access to RepositoryConnection: it encapsulates both a RepositoryConnection and a dedicated Thread (provided by an ExecutorService)
* Drastically simplified concurrency handling: ActiveTransactionRegistry now keeps tracks of Transactions instead of RepositoryConnections - no longer any need for locking access to the RepositoryConnection

All integration/compliance tests succeed, and I have manually verified that this fix also fixes the IllegalMonitorStateException described in #630.

### DIFF
--- a/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
+++ b/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
@@ -17,6 +17,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.rdf4j.IsolationLevel;
 import org.eclipse.rdf4j.model.IRI;
@@ -56,6 +57,8 @@ import org.eclipse.rdf4j.rio.helpers.BasicParserSettings;
  */
 class Transaction {
 
+	private final AtomicBoolean isClosed = new AtomicBoolean(false);
+
 	private final UUID id;
 
 	private final Repository rep;
@@ -93,8 +96,6 @@ class Transaction {
 		return id;
 	}
 
-
-
 	/**
 	 * Start the transaction.
 	 * 
@@ -130,7 +131,7 @@ class Transaction {
 			txnConnection.rollback();
 			return true;
 		});
-	
+
 		futures.add(result);
 		result.get();
 	}
@@ -146,9 +147,9 @@ class Transaction {
 			txnConnection.commit();
 			return true;
 		});
-	
+
 		futures.add(result);
-	
+
 		result.get();
 	}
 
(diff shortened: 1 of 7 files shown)


core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -0,0 +1,477 @@
...
+			update.execute();
+			return true;
+		});
+
+		futures.add(result);
+		result.get();
+	}
+
+	boolean hasActiveOperations() {
+		for (Future future : futures) {
```
jeenbroekstra: Now that I look at it - this may need to be synchronized on `futures` - and all methods that create/add a new `Future` should also synchronize on it.

jeenbroekstra:
Hudson verify PR: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/42/

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -0,0 +1,516 @@
...
+
+	private final AtomicBoolean isClosed = new AtomicBoolean(false);
+
+	private final UUID id;
+
+	private final Repository rep;
+
+	private final RepositoryConnection txnConnection;
+
+	private final ExecutorService executor = Executors.newSingleThreadExecutor();
```
ansell: The javadoc for this indicates that it will attempt to keep a single thread alive, but if it dies, it will happily create a new single thread to process further operations:

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newSingleThreadExecutor()

That doesn't quite fit the use case here that strictly requires a single thread for the monitor behind the locking, as far as I can tell.

jeenbroekstra: I thought so too at first. But all of our tasks are (implicitly) executed as a FutureTask, which handles all computational exceptions by wrapping them in an ExecutionException. If I understand correctly this prevents the thread from terminating. See also this note from the [Javadoc on ThreadPoolExecutor](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#afterExecute-java.lang.Runnable-java.lang.Throwable):

> Note: When actions are enclosed in tasks (such as FutureTask) either explicitly or via methods such as submit, these task objects catch and maintain computational exceptions, and **so they do not cause abrupt termination**.... 

(emphasis mine)

jeenbroekstra: In other words: I can not think of a single scenario in which the thread would actually die and need replacing.

ansell: So things we do should not kill the thread used by the executor. The only other cases to think about would be whether the thread could be killed by other system elements, which are in some cases unrecoverable from (OOM for example) so we don't need to cater for them.

jeenbroekstra:
Hudson build succeeded. I'll leave this PR open over the weekend so you and other can review and test further if you wish. If nothing new comes up over the weekend I'll consider this ready for merging.

PR #651, opened 2016-11-16
Fix #67: fixed auto detect and baseURI-as-context update in Workbench

### LINKED ISSUE #67: File format autodetect does not work in 'Add RDF' screen
(Migrated from https://openrdf.atlassian.net/browse/SES-2185 )

When uploading a file through the "Add RDF" screen, the (autodetect) option is supposed to determine the correct format and select the right parser. However, this does not work. In the current system, for any format other than RDF/XML, file upload with autodetect results in an error "Content is not allowed in prolog. [line 1, column 1] "

 Only after explicitly selecting the correct format from the dropdown does file upload work.

jeenbroekstra:
This PR addresses GitHub issue: #67 .

* recompiled all typescripts to include correct project root - this fixes the autodetect issue
* tweaked add.ts to fix problem with context field being uneditable if "use base URI" is not checked


jeenbroekstra:
Just ran another manual test and for some reason the bug reappeared after I thought I had fixed it. Grmbl. I will take a closer look tomorrow.

jeenbroekstra:
Tweaked. Should work now (tested in Firefox). If you can, give it a spin locally and see if nothing obvious breaks.

PR #653, opened 2016-11-16
Fix #652: Keep BOM updated so they are valid dependencies

### LINKED ISSUE #652: BOM is out of date
It lists several dependencies that no longer exists, like `rdf4j-sail-lucene4` and `rdf4j-sail-rdbms`.

Some of the dependencies should be `<type>war</type>`

Some are missing, e.g. `rdf4j-lucene-testsuite`.

This causes problems when depending on the bom and using the Maven site plugin. When building `mvn package site` of Apache Commons RDF 0.3.0 I get stack traces like:

    [WARNING] Unable to create Maven project for
    org.eclipse.rdf4j:rdf4j-sail-lucene-test:pom:2.1.1 from repository.
    org.apache.maven.project.ProjectBuildingException: Error resolving
    project artifact: Failure to find
    org.eclipse.rdf4j:rdf4j-sail-lucene-test:pom:2.1.1 in
    https://repo.maven.apache.org/maven2 was cached in the local repository,
    resolution will not be reattempted until the update interval of central
    has elapsed or updates are forced for project
    org.eclipse.rdf4j:rdf4j-sail-lucene-test:pom:2.1.1

stain:
This PR addresses GitHub issue: #652.

* specify <type>war</type> were needed (thus the `<dependencies>` could also be used outside `<dependencyManagement>`
* remove now missing rdf4j-sail-lucene4 etc.
* Reordered some dependencies to be closer to file system alhabetical

This avoids problems with maven site plugin when using the BOM, which would otherwise log various errors from the dependency plugin.

Signed-off-by: Stian Soiland-Reyes <stain@apache.org>

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -43,6 +43,7 @@
 				<groupId>org.eclipse.rdf4j</groupId>
 				<artifactId>rdf4j-http-server</artifactId>
 				<version>${project.version}</version>
+				<type>war</type>
 			</dependency>
 			<dependency>
 				<groupId>org.eclipse.rdf4j</groupId>
@@ -53,6 +54,7 @@
 				<groupId>org.eclipse.rdf4j</groupId>
 				<artifactId>rdf4j-http-workbench</artifactId>
 				<version>${project.version}</version>
+				<type>war</type>
 			</dependency>
 			<dependency>
 				<groupId>org.eclipse.rdf4j</groupId>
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
Nice! Thanks for this. I wonder if we can do some automated verification of the BOM so we catch these kinds of things in the future. It's so easy to forget to update this when we tweak the modules.

stain:
What I did to verify was to just change it to <packaging>jar and unwrapped
the <dependenchManagement>, then attempt mvn install.

This of course could not identify any missing modules and is probably
tricky to do automatically.

The plugin
http://maven.apache.org/plugins/maven-dependency-plugin/analyze-dep-mgt-mojo.html
might help, but would only detect if there was a version mismatch.

PR #664, opened 2016-11-23
Fix #216: include W3C DAWG SPARQL 1.0 tests

### LINKED ISSUE #216: Re-Include ManifestTest and testcases-dawg
This set of testcases was in the 4.0/master branch of sesame in sparql-testsuite but have now gone from rdf4j.

jeenbroekstra:
This PR addresses GitHub issue: #216 .

* Re-include SPARQL 1.0 W3C-approved query and syntax tests

Currently, two tests fail. I need to look closer but we may have discovered two genuine regressions.


testsuites/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/manifest/SPARQL10ManifestTest.java
```diff
@@ -0,0 +1,183 @@
...
+		throws IOException, RepositoryException, RDFParseException, RDFHandlerException
+	{
+		if (baseURI == null) {
+			baseURI = url.toExternalForm();
+		}
+
+		InputStream in = url.openStream();
+
+		try {
+			OpenRDFUtil.verifyContextNotNull(contexts);
```
ansell: This was probably here in the past, but generally is the first line in a method in other places in the codebase.

testsuites/sparql/src/main/resources/testcases-sparql-1.0-w3c/data-r2/LICENSE
```diff
@@ -0,0 +1,21 @@
+All tests materials are licensed under the W3C Software Notice and License, as follows:
+____________________________________
+W3C� SOFTWARE NOTICE AND LICENSE
```
ansell: This may just be an issue with Github, but there may be an undisplayable character here, likely UTF-8 versus latin1 or something like that but may want to look at it.

jeenbroekstra:
One of the two failing tests is caused by a problem in the literal hashcode implementation. See issue #665.

pulquero:
What is the blocker currently for this? I think I need this so I can PR further tests I have based on this, and associated fixes!

jeenbroekstra:
We have one failing test, still, which needs to be fixed. I haven't yet found the time to look at the issue in detail. It's possible the test is out of date with respect to SPARQL 1.1, but it's of course also possible this is an actual bug in the query engine.

jeenbroekstra:
Failed tests: 
  ```
W3CApprovedSPARQL10QueryTest>SPARQLQueryTest.runTest:224->SPARQLQueryTest.compareTupleQueryResults:373 
============ date-2 =======================
Unexpected bindings: 
v="2006-08-23Z"^^<http://www.w3.org/2001/XMLSchema#date> x=http://example/d2 
v="2006-08-23+00:00"^^<http://www.w3.org/2001/XMLSchema#date> x=http://example/d3 
===========================================
```

pulquero:
Is this a regression against sesame 2.8 or a test that was previously disabled?

jeenbroekstra:
The former. I think it may have been introduced by some enhancements we did to support date-related datatypes more extensively. I'll take a closer look at the failing test right now.

jeenbroekstra:
There were two more failing tests after syncing with master. Both dealt with the handling of decimal values ending with a point (but no fraction number) - which is no longer legal in SPARQL 1.1. 

As far as I have been able to determine, the date-related test is also incorrect or at least incompatible with SPARQL 1.1 semantics: it assumes that two date values, one of which has a timezone, and one of which doesn't, can be determined to be (un)equal. But if I read https://www.w3.org/TR/xmlschema-2/#dateTime correctly, that is not the case, and RDF4J actually behaves correctly here. 

All three failing tests have therefore been put on the ignore list. AFAIC this is ready for merge (as soon as Travis finishes building).

Issue #665, opened 2016-11-23
SimpleLiteral hashCode should ignore case for language tags


nikolavp:
The contract for equals and hashcode is now broken for the following literals

```java
final SimpleValueFactory vf = SimpleValueFactory.getInstance();
Literal v1 = vf.createLiteral("test", "EN");
Literal v2 = vf.createLiteral("test", "en");
v1.equals(v2) // true
v1.hashCode() == v2.hashCode() // false
```
this is caused because equals is using ignore case but the hashcode doesn't. Please lowercase the literal 's language in the setLanguage method.

ansell:
+1, since we moved to the new hashCode implementation including the language field, it needs to take it into account equivalently for hashCode. The alternatives if the performance hit to using equalsIgnoreCase for hashCode are caching the calculated hashCode or taking language out of the hashCode again.

Lowercasing the literal's language has been discussed in the past, but my recollection was there were complaints about it.

jeenbroekstra:
This bug is the exact cause of SPARQL 1.0 test case "lang-03" failing (see PR #664). If only we had reintegrated those test cases sooner :)

jeenbroekstra:
This is related to #449. That issue was originally postponed because we weren't updating to Java 7 just yet. Now that we're a Java 8 club, we should probably have a closer look at it.

jeenbroekstra:
As a further point of interest: the request to remove lower-casing all language tags actually came from Vladimir Alexiev of Ontotext. See https://openrdf.atlassian.net/browse/SES-1999 . Also relevant: #478 (which is the migrated issue SES-1659).

jeenbroekstra:
The immediate fix for the broken hashcode impl is clear though: we need to either ignore case for language tags in both hashCode and equals, or in neither. I'm in favor of ignoring case in both - that way they behave consistently regardless of the actual case that we store (whether it's all lowercase or follows BCP 47 recommendations). `String.toLowerCase` is sufficiently fast that I don't consider this a significant performance regression (though I'll try and run a benchmark to make sure).

jeenbroekstra:
Possible tweak to increase performance is to have a private field that stores the lower-cased tag, and use that in equals and hashCode. I'll run some tests.

ansell:
Sorry for confusion in my comment above. Sesame-4.0 removed lowercasing of language tags by default, but it is still available as a parser setting.

[...]

Issue #449, opened 2016-09-07
Support BCP47 for language tags


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-1826)

The RDF-1.1 draft currently includes a normative reference to BCP47 as the specification for language tags used in language literals. BCP obsoletes both RFC4646 and RFC3066. RDF-1.0 had a reference to the earlier RFC3066 however all RFC3066 and RFC4646 codes are grandfathered into BCP47 [1]

A new implementation of LanguageHandler needs to be created, and enabled when RDF-1.1 profile/mode is active.

[1] https://tools.ietf.org/html/bcp47#section-2.2.8

PR #666, opened 2016-11-23
issue #449 : Add support for BCP47, and use by default

ansell:
This PR addresses GitHub issue: #449 to support BCP47.

* Adds implementation of BCP47 based on Java-7+ Locale
* Switches BasicParserSettings to use BCP47 by default, per our reliance on RDF-1.1

RFC3066 is now only used if it is specifically setup in parser settings, or referenced directly.

The use of BCP47 is required for RDF-1.1, where RFC3066 was required for RDF-1.0. Since we have changed to RDF-1.1, we need to change the default language tag handler to match.

I still want to add some regression tests related to issue #665 and other language tag case-sensitivity issues to verify and preserve the new BCP47 behaviour unless we change the regression test, but the core implementation of BCP47 ```new Locale.Builder().setLanguageTag(languageTag).build().toLanguageTag()```, is done.

### DIFF
--- a/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BasicParserSettings.java
+++ b/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BasicParserSettings.java
@@ -242,18 +242,18 @@ public class BasicParserSettings {
 			_DEFAULT_PREFIX);
 
 	static {
-		List<DatatypeHandler> defaultDatatypeHandlers = new ArrayList<DatatypeHandler>(5);
+		List<DatatypeHandler> defaultDatatypeHandlers = new ArrayList<>(5);
 		try {
 			DatatypeHandlerRegistry registry = DatatypeHandlerRegistry.getInstance();
-			for (String nextHandler : Arrays.asList(DatatypeHandler.XMLSCHEMA, DatatypeHandler.RDFDATATYPES,
+			for (String nextDatatype : Arrays.asList(DatatypeHandler.XMLSCHEMA, DatatypeHandler.RDFDATATYPES,
 					DatatypeHandler.DBPEDIA, DatatypeHandler.VIRTUOSOGEOMETRY, DatatypeHandler.GEOSPARQL))
 			{
-				Optional<DatatypeHandler> nextdt = registry.get(nextHandler);
-				if (nextdt.isPresent()) {
-					defaultDatatypeHandlers.add(nextdt.get());
+				Optional<DatatypeHandler> nextDatatypeHandler = registry.get(nextDatatype);
+				if (nextDatatypeHandler.isPresent()) {
+					defaultDatatypeHandlers.add(nextDatatypeHandler.get());
 				}
 				else {
-					log.warn("Could not find DatatypeHandler : {}", nextHandler);
+					log.warn("Could not find DatatypeHandler : {}", nextDatatype);
 				}
 			}
 		}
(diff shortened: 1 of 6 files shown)


core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BasicParserSettings.java
```diff
@@ -270,7 +270,7 @@
 		List<LanguageHandler> defaultLanguageHandlers = new ArrayList<LanguageHandler>(1);
 		try {
 			LanguageHandlerRegistry registry = LanguageHandlerRegistry.getInstance();
-			for (String nextHandler : Arrays.asList(LanguageHandler.RFC3066)) {
+			for (String nextHandler : Arrays.asList(LanguageHandler.BCP47)) {
```
jeenbroekstra: Why is this is a for-loop? There is only ever a single value, no?

core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java
```diff
@@ -0,0 +1,101 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: wrong copyright header.

core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * Default constructor.
+	 */
+	public BCP47LanguageHandler() {
+	}
+
+	@Override
+	public boolean isRecognizedLanguage(String languageTag) {
+		if (languageTag == null) {
+			throw new NullPointerException("Language tag cannot be null");
```
jeenbroekstra: Not critical really, but we use `Objects.requireNonNull` in many places now:

    Objects.requireNonNull(languageTag, "Language tag cannot be null");

core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java
```diff
@@ -0,0 +1,101 @@
...
+	{
+		if (languageTag == null) {
+			throw new NullPointerException("Language tag cannot be null");
+		}
+
+		try {
+			Locale asBCP47 = parseAsBCP47(languageTag);
+			
+			if (literalValue == null) {
+				throw new NullPointerException("Literal value cannot be null");
```
jeenbroekstra: Can be replace by `Objects.requireNonNull` as well, and shouldn't this check be done _before_ trying to parse the language tag locale?

core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java
ansell: I will look for other copyright headers to see if I can fix. That header comes from RFC3066LanguageHandler which I copied to create BCP47LanguageHandler

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BasicParserSettings.java
ansell: I can remove the loop, likely was copied/pasted from the datatype handler loop.

core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java
ansell: I left this check until after the language tag check to match the previous behaviour. Not likely an issue to do it up-front though. Will change RFC3066 to match.

core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java
ansell: I can't find any other version of the copyright header in the files I am skimming through. Do you have an example of the correct header to use? (Will require a bulk replacement as this one is used extensively)

jeenbroekstra: My remark was under the assumption that this is a new file (and not code that was already present in Sesame). The copyright line that says "Eclipse RDF4J contributors, Aduna, and others" is _only_ used for content that was already present in Sesame. For everything new, we use copyright header that is present in the current template (and you can also quickly copy-paste it from the [code formatting section in the contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#code-formatting), if needed):

```
 * Copyright (c) 2016 Eclipse RDF4J contributors.
```

As you can see the difference is really minor.

ansell: I was not completely sure about changing it, given it started as a straight copy-paste of the RFC3066 file, with minor changes, and hence to me it seems like it should inherit the header from that file. However, I have changed it, can revert if necessary.

PR #669, opened 2016-11-24
Fix #648 : Log failure to rollback to help with debugging

### LINKED ISSUE #648: Improvement: Create logger warning if rollback is called without an active transaction
Use case:

I had a sail-memory repository and was wondering why my changes where not rolled back although I called the rollback method. I later found out that I forgot to start the transaction with begin().

I would propose to add a logger warning if someone tries to rollback changes without an active transaction as this would make debugging a lot easier.

I would propose to make the changes in AbstractSailConnection:453

if (txnActive) {
 //do internal commit
}else{
 logger.warn("Cannot rollback transaction because transaction is not active", new Throwable());
}

ansell:
This PR addresses GitHub issue: #648 .

* Adds logging when rollback fails because transaction is not active

### DIFF
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
@@ -888,10 +888,6 @@ public abstract class AbstractSailConnection implements SailConnection {
 					logger.warn("Forced closing of unclosed iteration that was created in:",
 							debugEnabled ? creatorTrace : null);
 				}
-				else {
-					logger.warn("Forced closing of unclosed iteration, current stacktrace:",
-							debugEnabled ? new Throwable() : null);
-				}
 				ci.close();
 			}
 			catch (SailException e) {
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
@@ -221,7 +221,8 @@ public abstract class AbstractSailConnection implements SailConnection {
 					forceCloseActiveOperations();
 
 					if (txnActive) {
-						logger.warn("Rolling back transaction due to connection close", new Throwable());
+						logger.warn("Rolling back transaction due to connection close",
+								debugEnabled ? new Throwable() : null);
 						try {
 							// Use internal method to avoid deadlock: the public
 							// rollback method will try to obtain a connection lock
@@ -461,7 +462,7 @@ public abstract class AbstractSailConnection implements SailConnection {
 				}
 				else {
 					logger.warn("Cannot rollback transaction on connection because transaction is not active",
-							new Throwable());
+							debugEnabled ? new Throwable() : null);
 				}
 			}
 			finally {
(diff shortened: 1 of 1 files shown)


core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -459,6 +459,10 @@ public final void rollback()
 						txnPrepared = false;
 					}
 				}
+				else {
+					logger.warn("Cannot rollback transaction on connection because transaction is not active",
+							new Throwable());
```
jeenbroekstra: What's the `new Throwable()` for?

jeenbroekstra: Never mind, figured it out: I I guess this is to include a stack trace in the warning that the user can use to trace where he called rollback?

ansell: It was in the suggested patch by @elky87. It will generate a stack trace to show where the message came from. Should I wrap it in a system property the same way as we do the other debug stacktraces?

jeenbroekstra: Hm, yeah, perhaps a good idea. You can just read the private boolean `debugEnabled` by the way, so no need to read the system prop yourself:

`logger.warn("Cannot rollback transaction on connection because transaction is not active", debugEnabled ? new Throwable() : null);`

ansell: Ie, should it be similar to the following pattern for specifying whether to show it or not:

https://github.com/eclipse/rdf4j/blob/a283e758fb79e7b62e94b669547e6e64c64106cf/core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/Properties.java#L19

jeenbroekstra: Looks like that may need a bit of revision (for one thing the system property should be renamed), as well as documentation.

ansell: The current system property that is used for AbstractSailConnection.debugEnabled is "org.eclipse.rdf4j.repository.debug", as setup in AbstractSail:

https://github.com/eclipse/rdf4j/blob/a283e758fb79e7b62e94b669547e6e64c64106cf/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSail.java#L66

Agree that we should migrate the lock debugging property, "info.aduna.concurrent.locks.trackLocks", but probably useful to grandfather the existing property name for a few versions.

jeenbroekstra:
Travis fails because the log length of the tests exceeds 4MB (which they take to mean the same error is raised over and over). There's an awful lot of warning messages of this form:

    00:26:55.170 [Time-limited test] WARN  Forced closing of unclosed iteration, current stacktrace:

It looks as if several of our compliance tests are not up to scratch in properly closing iterations.

ansell:
I can remove that branch, after looking through the code again. Switching org.eclipse.rdf4j.repository.debug on will do what it does, but return stacktraces attached to each case.

PR #670, opened 2016-11-24
Fix #668: reverted hashCode implementation to conform to API spec

### LINKED ISSUE #668: Revert change to Literal hashCode implementation
In #649 we included a literal's language tag and datatype in generation of the hashcode.

However, there are several problems with this:

1. It breaks the RDF4J API specification: it is expicitly documented on the `Literal` interface that the hashcode is calculated using the literal's label only. 
2. It introduces an unnecessary performance penalty which may impact performance of `Statement` collections (any `Set` or `List` as well as `Model`), as well as the `MemoryStore`.
3. The current implementation breaks the Java hashcode contract because it generates different hashcodes for two Literals which are considered equal (that is, with uppercase or lowercase language tags). See issue #665 for extended discussion.

To fix, we will revert the changes to the hashcode implementation, and go back to a label-only implemention. This will resolve all three above problems at once.

jeenbroekstra:
This PR addresses GitHub issue: #668 .

* SimpleLiteral.hashCode now uses label only again (cf. Literal API spec)
* updated unit tests to remove incorrect assertions and add test on different-case language tags

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
@@ -169,18 +169,10 @@ public class SimpleLiteral implements Literal {
 		return false;
 	}
 
-	// overrides Object.hashCode(), implements hashCode()
+	// overrides Object.hashCode(), implements Literal.hashCode()
 	@Override
 	public int hashCode() {
-		int hashCode = label.hashCode();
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
Travis is being slow today. But regardless, before I merge I'd like some additional ontotext eyeballs on this as well, just to make sure we're all on the same page. @nikolavp , @damyan-ognyanov , let me know if the solution as proposed at the end of #665 is workable for you.

PR #674, opened 2016-11-28
Fix #673 : Add utility for BCP47 normalisation

### LINKED ISSUE #673: Utlity function for BCP47 language tag normalisation
It would be useful to have a simple utility function to perform the steps necessary for BCP47 language tag normalisation:

```java
new Locale.Builder().setLanguageTag(tag).build().toLanguageTag()
```

ansell:
This PR addresses GitHub issue: #673 .

* Adds a utility function for the steps required for BCP47 language tag normalisation
* Converts BCP47LanguageHandler to use the utility function

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
@@ -10,6 +10,7 @@ package org.eclipse.rdf4j.model.util;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.Date;
+import java.util.IllformedLocaleException;
 import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
@@ -553,7 +554,8 @@ public class Literals {
 	public static boolean canCreateLiteral(Object object) {
 		if (object == null) {
 			// Cannot create a literal from a null
-			// Avoid throwing a NullPointerException here to enable universal usage
+			// Avoid throwing a NullPointerException here to enable universal
+			// usage
 			// of this method
 			return false;
 		}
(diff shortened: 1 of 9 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
```diff
@@ -580,6 +582,22 @@ public static boolean isLanguageLiteral(Literal literal) {
...
+	 * Normalises the given <a href="https://tools.ietf.org/html/bcp47">BCP47</a> language tag according to
+	 * the rules defined by the JDK in the {@link Locale} API.
+	 * 
+	 * @param languageTag
+	 *        An unnormalised, valid, language tag
+	 * @return A normalised version of the given language tag
+	 * @throws IllformedLocaleException
+	 *         If the given language tag is illformed according to the rules specified in BCP47.
+	 */
+	public static String normaliseBCP47Tag(String languageTag)
```
jeenbroekstra: Given that BCP47 is the "format to follow" according to the RDF 1.1 spec, could we simplify the method name to just `normaliseTag`?

ansell: I would prefer not to remove BCP47 without adding Language in its place, if that makes sense. 

On another note, I have just noticed that I used the Australian English spelling of normalise, instead of the US English that we use in the rest of the API, normalize, so I will need to change that to stay consistent.

normalizeLanguageTag?

jeenbroekstra: Just realized that giving a 👍 probably doesn't notify you, so: your proposal sounds good to me.

ansell:
I changed the name to normalizeLanguageTag, and also added deprecation tags for some ancient RFC3066 code that we have sitting around that wasn't being maintained. We will now rely on the JDK to maintain the language tag registry, through the Locale APIs.

jeenbroekstra:
Hm, looking at that older code: I am ashamed to admit that I had no idea we even have a `LanguageTag` class (or indeed a whole package with related utilities). It might be worth looking into that a bit further and see if it's useful to update it to work with BCP47.

PR #677, opened 2016-11-30
Fix #676: revert changes to hashcode of AbstractBindingSet

### LINKED ISSUE #676: Regression in hashCode vs equals for AbstractBindingSet
In 2.0.2 hashCode() and equals() were consistent for AbstractBindingSet (and thus all of its descendants) but in 2.0.3 hashCode() suddenly began to be sensitive to the order bindings are added, while equals() stayed agnostic to the order. As this is a Set I believe both hashCode() and equals() shouldn't care about the order and it's a good idea to keep them consistent either way.

This change was introduced by https://github.com/eclipse/rdf4j/pull/650.

jeenbroekstra:
This PR addresses GitHub issue: #676 .

* reverted changed hashcode impl back to be API-conformant 
* add regression tests for all BindingSet implementations

### DIFF
--- a/core/query/src/test/java/org/eclipse/rdf4j/query/BindingSetTest.java
+++ b/core/query/src/test/java/org/eclipse/rdf4j/query/BindingSetTest.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.query;
+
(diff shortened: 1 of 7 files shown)


core/query/src/test/java/org/eclipse/rdf4j/query/BindingSetTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+		assertEquals(bs1.hashCode(), bs2.hashCode());
+	}
+
+	/**
+	 * Creates two equal, but differently ordered, BindingSet objects.
+	 * 
+	 * @return an array of two equal but differently ordered BindingSets.
+	 */
+	protected abstract T[] createEqualBindingSets();
```
ansell: The javadoc for this explains the situation, but it may be useful (for future expansion of this test to other cases) to change the name of this method to createTwoEqualReorderedBindingSets or something shorter than indicates the reordering is the core test being undertaken.

PR #683, opened 2016-12-02
Fix #680: concurrent requests over http IllegalMonitorException

### LINKED ISSUE #680: RDF4J Server java.util.concurrent.ExecutionException
I'm calling the RDF4J server from two different executions of a program that is updating some data on the server. The problem point is a clean-the-old before adding the new data. Both executions are at this code in my client at the same time. 

More discussion and stack traces from RDF4J versions 2.0.1 and 2.1.3 are shown at https://groups.google.com/forum/#!topic/rdf4j-users/qtqso7TFnWM

jeenbroekstra:
This PR addresses GitHub issue: #680 .

* RepositoryInterceptor in RDF4J Server no longer keeps a shared RepositoryConnection, instead it creates a new, pre-configured connection whenever needed. Callers are responsible for closing the connection when done.
* Native store internal locking simplified 

These fixes have been tested in combination with the test case provided in #680, and it seems to resolve the issue. In addition all existing tests greenline.

I will look into adding the provided test case to our suite as a regression test.

### DIFF
--- a/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
+++ b/core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
@@ -84,13 +84,10 @@ public class NamespaceController extends AbstractController {
 	private ModelAndView getExportNamespaceResult(HttpServletRequest request, String prefix)
 		throws ServerHTTPException, ClientHTTPException
 	{
-		try {
+		try (RepositoryConnection repositoryCon = RepositoryInterceptor.getRepositoryConnection(request)) {
 			String namespace = null;
 
-			RepositoryConnection repositoryCon = RepositoryInterceptor.getRepositoryConnection(request);
-			synchronized (repositoryCon) {
-				namespace = repositoryCon.getNamespace(prefix);
-			}
+			namespace = repositoryCon.getNamespace(prefix);
 
 			if (namespace == null) {
 				throw new ClientHTTPException(SC_NOT_FOUND, "Undefined prefix: " + prefix);
@@ -117,11 +114,8 @@ public class NamespaceController extends AbstractController {
 		}
 		// FIXME: perform some sanity checks on the namespace string
 
-		try {
-			RepositoryConnection repositoryCon = RepositoryInterceptor.getRepositoryConnection(request);
-			synchronized (repositoryCon) {
-				repositoryCon.setNamespace(prefix, namespace);
-			}
+		try (RepositoryConnection repositoryCon = RepositoryInterceptor.getRepositoryConnection(request)) {
+			repositoryCon.setNamespace(prefix, namespace);
 		}
 		catch (RepositoryException e) {
 			throw new ServerHTTPException("Repository error: " + e.getMessage(), e);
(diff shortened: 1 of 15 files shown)


core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
```diff
@@ -84,13 +84,10 @@ else if ("DELETE".equals(reqMethod)) {
 	private ModelAndView getExportNamespaceResult(HttpServletRequest request, String prefix)
 		throws ServerHTTPException, ClientHTTPException
 	{
-		try {
+		try (RepositoryConnection repositoryCon = RepositoryInterceptor.getRepositoryConnection(request)) {
 			String namespace = null;
```
ansell: namespace = null is redundant now that the synchronized section has been removed

jeenbroekstra:
FWIW I have the feeling I keep falling into the same trap, over and over. The fix I'm doing in NativeStore is similar to something I've tried before, and, on the face of it, reasonable. When testing locally everything succeeds. But for some reason it breaks the build on Travis.

jeenbroekstra:
Also not 100% sure about the changes to RepositoryInterceptor yet. Hold off on deep review for now, as it may not be necessary for #680, so I might  revert it.

jeenbroekstra:
I had a closer look, and I think the failing test (RDFStore.testMultiThreadedAccess) is actually a flawed test - it checks multithreaded access on the same SailConnection, which is something we don't guarantee to work.

ansell:
Yes, I agree, the test is flawed by the shared connection. 

Its use of "sleep" instead of a CountDownLatch or similar to ensure synchronicity at the join points is another flaw.

testsuites/store/src/main/java/org/eclipse/rdf4j/sail/RDFStoreTest.java
```diff
@@ -718,6 +726,7 @@ public void testQueryBindings()
 	}
 
 	@Test
+	@Ignore("test is fundamentelly flawed as it uses multithreaded access on the same connection")
```
jeenbroekstra: An alternative to ignoring the test is fixing it, by putting the loop that checks if a txn is active (and the `con.begin()` call) inside a synchronized block that locks on the connection. That would probably fix the test failing, but on the other hand I'm not sure what this test is even testing for - we simply do not guarantee multi-threaded access on a single connection. Thoughts?

ansell: I would change the focus to open N new connections for the N threads being tested and create join points using CountDownLatch's between the stages to verify the overall sequence is somewhat replicable and not completely random (no point having a test that isn't replicable to some degree, even if it is trying to test concurrency).

jeenbroekstra: The thing is: we have other test suites that specifically focus on this. Both `SailConcurrencyTest` and `SailIsolationLevelTest` do extensive tests using multiple concurrent connections. So I don't think this particular test adds much value.

jeenbroekstra: Nevertheless I'll have a quick look at a rewrite along the lines you suggest.

jeenbroekstra: Yeah, nah. The test seems to be specifically about the behavior of things like the `close()` method when using the same connection in multiple threads. Let's leave it as ignored, or even just remove it.

ansell: I won't be sad to see the back of this test. It caused a lot of impossible to replicate errors in the past.

We do need a concurrent test suite for stores that support concurrent access in the future, but it needs to be replicable and solely test things that the API actually guarantees.

jeenbroekstra:
I decided to leave the changes to RepositoryInterceptor in, as I think the new design is more robust and easier to maintain. Any further reviews are appreciated.

Issue #684, opened 2016-12-02
Improved handling of thread closure in Background query results [RDF4J Server]


rune66:
I'm running rdf4j in a Tomcat 8.5 container. From another web app I use the rdf4j-api to access the repositories hosted in rdf4j. I don't know weather it's a problem or not, but when stopping the web app I get a ton of "failed to stop it"-exceptions that seems to originate from rdf4j.

Is this recognizable to anyone, and how to avoid it?

WARNING: The web application [REST] appears to have started a thread named [pool-28-thread-1] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 sun.misc.Unsafe.park(Native Method)
 java.util.concurrent.locks.LockSupport.park(Unknown Source)
 java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(Unknown Source)
 java.util.concurrent.ArrayBlockingQueue.put(Unknown Source)
 org.eclipse.rdf4j.http.client.QueueCursor.put(QueueCursor.java:80)
 org.eclipse.rdf4j.http.client.BackgroundTupleResult.handleSolution(BackgroundTupleResult.java:139)
 org.eclipse.rdf4j.query.resultio.sparqlxml.SPARQLResultsSAXParser.endTag(SPARQLResultsSAXParser.java:187)
 org.eclipse.rdf4j.common.xml.SimpleSAXParser$SimpleSAXDefaultHandler.endElement(SimpleSAXParser.java:337)
 com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.endElement(Unknown Source)
 com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanEndElement(Unknown Source)
 com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(Unknown Source)
 com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(Unknown Source)
 com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.next(Unknown Source)
 com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanDocument(Unknown Source)
 com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(Unknown Source)
 com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(Unknown Source)
[...]

PR #687, opened 2016-12-03
issue #684 : More reliably clean up for InterruptedException in BackgroundTupleResult

ansell:
This PR addresses GitHub issue: #684 .

* Cleans up more reliably in BackgroundTupleResult when InterruptedException occurs
* Removes internal synchronisation in QueueCursor for exceptions so it now uses ConcurrentLinkedQueue instead of LinkedList.

All tests pass for me after the changes.

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/QueueCursor.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/QueueCursor.java
@@ -9,11 +9,12 @@ package org.eclipse.rdf4j.http.client;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.rdf4j.RDF4JException;
@@ -34,7 +35,7 @@ public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationExcepti
 
 	private final E afterLast = createAfterLast();
 
-	private final Queue<Throwable> exceptions = new LinkedList<Throwable>();
+	private final Queue<Throwable> exceptions = new ConcurrentLinkedQueue<Throwable>();
 
 	/**
 	 * Creates an <tt>QueueCursor</tt> with the given (fixed) capacity and default access policy.
@@ -65,9 +66,7 @@ public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationExcepti
 	 * QueryEvaluationException or RuntimeException it will be wrapped in a QueryEvaluationException.
 	 */
 	public void toss(Exception exception) {
-		synchronized (exceptions) {
-			exceptions.add(exception);
-		}
+		exceptions.add(exception);
 	}
 
 	/**
@@ -76,8 +75,21 @@ public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationExcepti
 	public void put(E item)
 		throws InterruptedException
 	{
-		if (!done.get()) {
-			queue.put(item);
+		try {
+			while (!isClosed() && !done.get() && !Thread.currentThread().isInterrupted()
+					&& !queue.offer(item, 1, TimeUnit.SECONDS))
+			{
+				// No body, just iterating regularly through the loop conditions to respond to state changes without a full busy-wait loop
+			}
+			// Proactively close if interruption didn't propagate an exception to the catch clause below
+			if (done.get() || Thread.currentThread().isInterrupted()) {
+				close();
+			}
+		}
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			close();
+			throw e;
 		}
 	}
 
@@ -87,21 +99,23 @@ public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationExcepti
 	public void done() {
 		// Lazily set here, and then come back in handleClose and use set if necessary
 		done.lazySet(true);
-		try {
-			queue.add(afterLast);
-		}
-		catch (IllegalStateException e) {
-			// no thread is waiting on this queue anyway
+		boolean offer = queue.offer(afterLast);
+		if (!offer) {
+			// TODO: Log inability to add sentinel at debug level
+			// The sentinel is forced onto the queue during the close method
 		}
 	}
 
 	/**
-	 * Returns the next item in the queue or throws an exception.
+	 * Returns the next item in the queue, which may be <tt>null</tt>, or throws an exception.
 	 */
 	@Override
 	public E getNextElement()
 		throws QueryEvaluationException
 	{
+		if (isClosed()) {
+			return null;
+		}
 		try {
 			checkException();
 			E take;
@@ -119,10 +133,12 @@ public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationExcepti
 				done(); // put afterLast back for others
 				return null;
 			}
+			checkException();
 			return take;
 		}
 		catch (InterruptedException e) {
 			checkException();
+			close();
 			throw new QueryEvaluationException(e);
 		}
 	}
@@ -147,37 +163,36 @@ public class QueueCursor<E> extends LookAheadIteration<E, QueryEvaluationExcepti
 	public void checkException()
 		throws QueryEvaluationException
 	{
-		synchronized (exceptions) {
-			if (!exceptions.isEmpty()) {
-				try {
-					throw exceptions.remove();
-				}
-				catch (RDF4JException e) {
-					if (e instanceof QueryEvaluationException) {
-						List<StackTraceElement> stack = new ArrayList<StackTraceElement>();
-						stack.addAll(Arrays.asList(e.getStackTrace()));
-						StackTraceElement[] thisStack = new Throwable().getStackTrace();
-						stack.addAll(Arrays.asList(thisStack).subList(1, thisStack.length));
-						e.setStackTrace(stack.toArray(new StackTraceElement[stack.size()]));
-						throw e;
-					}
-					else {
-						throw new QueryEvaluationException(e);
-					}
-				}
-				catch (RuntimeException e) {
-					// any RuntimeException that is not an OpenRDFException should be
-					// reported as-is
+		if (!exceptions.isEmpty()) {
+			try {
+				close();
+				throw exceptions.remove();
+			}
+			catch (RDF4JException e) {
+				if (e instanceof QueryEvaluationException) {
 					List<StackTraceElement> stack = new ArrayList<StackTraceElement>();
 					stack.addAll(Arrays.asList(e.getStackTrace()));
 					StackTraceElement[] thisStack = new Throwable().getStackTrace();
-					stack.addAll(Arrays.asList(thisStack));
+					stack.addAll(Arrays.asList(thisStack).subList(1, thisStack.length));
 					e.setStackTrace(stack.toArray(new StackTraceElement[stack.size()]));
 					throw e;
 				}
-				catch (Throwable e) {
+				else {
 					throw new QueryEvaluationException(e);
 				}
+			}
+			catch (RuntimeException e) {
+				// any RuntimeException that is not an RDF4JException should be
+				// reported as-is
+				List<StackTraceElement> stack = new ArrayList<StackTraceElement>();
+				stack.addAll(Arrays.asList(e.getStackTrace()));
+				StackTraceElement[] thisStack = new Throwable().getStackTrace();
+				stack.addAll(Arrays.asList(thisStack));
+				e.setStackTrace(stack.toArray(new StackTraceElement[stack.size()]));
+				throw e;
+			}
+			catch (Throwable e) {
+				throw new QueryEvaluationException(e);
 			}
 		}
 	}
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SesameClientImpl.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SesameClientImpl.java
@@ -7,8 +7,10 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.http.client;
 
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.utils.HttpClientUtils;
@@ -16,6 +18,8 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
 import org.eclipse.rdf4j.http.client.util.HttpClientBuilders;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 /**
  * Uses {@link HttpClient} to manage HTTP connections.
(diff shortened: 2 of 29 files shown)


core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SesameClientImpl.java
```diff
@@ -107,25 +125,26 @@ public synchronized SesameSession createSesameSession(String serverURL) {
...
+	 * No-op
+	 * 
+	 * @deprecated Create a new instance instead of trying to reactivate an old instance.
 	 */
-	public synchronized void initialize() {
-		if (executor == null) {
-			executor = Executors.newCachedThreadPool();
-		}
+	@Deprecated
+	public void initialize() {
```
jeenbroekstra: What's the consequence of this change for code that still _does_ rely on reinitialization? I didn't check but I assume you changed our own code to not do this (if it ever did)?

core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPRepository.java
```diff
@@ -111,49 +113,76 @@ public HTTPRepository(final String repositoryURL) {
...
-		if (client == null) {
-			client = dependentClient = new SesameClientImpl();
+	@Override
+	public SesameClient getSesameClient() {
+		SesameClient result = client;
+		if (result == null) {
+			synchronized (this) {
+				result = client;
+				if (result == null) {
+					result = client = dependentClient = new SesameClientImpl();
```
jeenbroekstra: I remember protesting that this `foo = bar = new Object()` was ugly when James first introduced it. Now you go and exacerbate it :)

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SesameClientImpl.java
ansell: There was no instance of this method being called, according to Eclipse Call Heirarchy. The code that was there is easy enough to reinstate if necessary. If it turns out that people do need to reinitialise these, instead of replacing them with new copies, then I will need to change some of the other parts of the code also. 

I am not happy with the pull request yet, but it is great to get comments on it so far. In particular, the sample code referenced the .evaluate method that I haven't inspected closely yet.

core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPRepository.java
ansell: It has proven performance advantages by reducing volatile reads on the variable reference. I agree though, Java should have a better mechanism for caching volatile reads to avoid it.

ansell: The dependentClient stuff can't be helped though, that is how we know if we should be closing it ourselves or leaving it open then it goes out of scope.

ansell:
The only tools I used were the Eclipse Call Hierarchy, and the Eclipse "find implementations of this method" functionality.

core/http/client/src/test/java/org/eclipse/rdf4j/http/client/BackgroundGraphResultHangTest.java
```diff
@@ -63,11 +63,9 @@ public void testBGRHang()
 				new ByteArrayInputStream(data.getBytes(Charset.forName("UTF-8"))), Charset.forName("UTF-8"),
 				"http://base.org");
 
+		thrown.expect(QueryEvaluationException.class);
```
ansell: I am now a little suspicious of this change and its side effects in terms of which thread sees the exception now. I may reverse some of the changes around this if I can prove that it is breaking something, but it may actually be useful behaviour rather than waiting until .hasNext is called to throw the wrapped parse exception.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SesameClientImpl.java
jeenbroekstra: The reason I ask is: can we get away with adding this fix as part of a patch release (SemVer-wise)?

ansell:
I will look over it again today and check, haven't been back to it in a while so will likely see something.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/QueueCursor.java
```diff
@@ -76,8 +75,21 @@ public void toss(Exception exception) {
 	public void put(E item)
 		throws InterruptedException
 	{
-		if (!done.get()) {
-			queue.put(item);
+		try {
+			while (!isClosed() && !done.get() && !Thread.currentThread().isInterrupted()
```
ansell: After spending the last two weeks wrangling BlockingQueue uses I have really been turned off queue.put because of its blocking behaviour and the likelihood that thread interrupts are either silent because nothing was blocking when they occurred, or a thread somewhere silently swallowed them without resetting the interrupt status. I have changed this to be a 1-second wait loop, that will still respond during the 1 second offer immediately, but will check for closure/interrupts every 1 second before going back to the queue to offer the item again if it wasn't accepted the first time.

This is at the core of the stacktrace in #684, but on its own it was not going to ensure closure occurred, so the other changes are still necessary to get either the done or closed variables (and hopefully the thread interruption status) reliably propagated to the background thread operating here.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -113,13 +117,12 @@ UUID getID() {
...
-			result = executor.submit(() -> {
-				txnConnection.begin(level);
-				return true;
-			});
+		Future<Boolean> result = executor.submit(() -> {
+			txnConnection.begin(level);
+			return true;
+		});
 
+		synchronized (futures) {
```
ansell: I moved all of the executor.submit commands outside of synchronized(futures) as the order of submit isn't guaranteed to be the order that the executor will actually pick them in, and the JVM gives no guarantees about the order of releasing/locking monitors for synchronized blocks. I can revert this if it was intentionally done for another reason.

jeenbroekstra: This was done because threads need concurrent access to `futures` to determine if operations are currently active. See `hasActiveOperations`. The order in which the Future objects get added to the list is not really relevant, but this is done in a synchronized manner because `ArrayList` is not thread-safe. 

I think that this should  indeed be reverted.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -134,13 +137,12 @@ void begin(IsolationLevel level)
...
-			result = executor.submit(() -> {
-				txnConnection.rollback();
-				return true;
-			});
+		Future<Boolean> result = executor.submit(() -> {
+			txnConnection.rollback();
+			return true;
+		});
 
+		synchronized (futures) {
```
jeenbroekstra: Like I mentioned above, this needs to be rolled back I think. It is now possible that `hasActiveOperations` is called concurrently with this operation and returns `false`, despite the rollback already being scheduled for execution.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -57,7 +60,7 @@
  * 
  * @author Jeen Broekstra
  */
-class Transaction {
+class Transaction implements AutoCloseable {
```
jeenbroekstra: This shouldn't be necessary. The `ActiveTransactionRegistry` is responsible for closing the `Transaction` object when done.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
ansell: I will add some locking to ensure that there is concurrent access from the thread executor. Putting the .execute instruction inside of synchronized does nothing for thread-safety purposes, so I will add a semaphore to the executes to ensure they are complete before the next one starts.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
ansell: Adding it helps with compiler and static analysis tools, and has no downsides as far as I know.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
ansell: I will work through this with either semaphores or putting the Future.get inside of the synchronized block. It was already possible that hasActiveOperations was being called concurrently, as the executor doesn't retain the monitor from this thread.

jeenbroekstra: Just to be clear the problem is not that `hasActiveOperations` can be called concurrently (that is by design), but that it inspects the list of Futures. It needs to lock access to that list to avoid concurrent threads modifying it _while it is looping over it_.

ansell: I have added a Semaphore to make task executions mutually exclusive of the hasActiveOperations method

I have readded the synchronisation for the futures back around the ExecutorService.submit method to make ExecutorService.submit and futures.add atomic from the perspective of hasActiveOperations.

I am a little concerned that hasActiveOperations and the other transaction methods do not take into account the closed status to abort early if the Transaction has already closed up shop. Ie, they currently need to obtain the monitor (increasing the likelihood of deadlocks) before being given an ExecutionException to indicate that nothing more can occur, once ExecutorService.shutdown has been called in the close method. I have added two AtomicBooleans to track the start and end of the closed method, but they are not linked to hasActiveOperations or ActiveTransactionRegistry yet.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -64,18 +66,54 @@
...
+	 * The {@link ExecutorService} that performs all of the operations related to this Transaction.
+	 */
 	private final ExecutorService executor = Executors.newSingleThreadExecutor(
 			new ThreadFactoryBuilder().setNameFormat("rdf4j-transaction-%d").build());
 
-	private final List<Future<?>> futures = new ArrayList<>();
+	/**
+	 * An append-only {@link Queue} that tracks the state of all of the futures involved in this Transaction.
+	 */
+	private final Queue<Future<?>> futures = new LinkedBlockingQueue<>();
```
jeenbroekstra: My experience with BlockingQueues is limited. Why is a Queue better than a simple ArrayList in this case?

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -64,18 +66,54 @@
...
+
+	/**
+	 * This {@link Semaphore} is used to ensure that only a single task is executed by the ExecutorService,
+	 * and that these tasks are not run while {@link #hasActiveOperations()} is determining whether there are
+	 * outstanding tasks open for this Transaction.
+	 */
+	private final Semaphore semaphore = new Semaphore(1);
```
jeenbroekstra: "only a single task is executed by the ExecutorService"... there is no need to ensure this externally, this is guaranteed by the fact that we use a `singlethreadExecutorService`. The second part makes sense though, and I think this clarifies for me why you felt there was an issue with the approach with synchronized blocks: it didn't guarantee anything about the state of the individual future objects changing _while iterating_ - is that correct?

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -117,14 +155,10 @@ UUID getID() {
 	void begin(IsolationLevel level)
 		throws InterruptedException, ExecutionException
 	{
-		Future<Boolean> result = executor.submit(() -> {
+		Future<Boolean> result = submit(() -> {
```
jeenbroekstra: Regardless of anything else, having a centralized `sumbit` method that can take care of all the book-keeping makes sense. 👍

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
ansell: Sorry, I was going to remove this but forgot. It was done to replace the requirement that results.add be single threaded, but we do actually want add to be atomic with respect to submit, so ArrayList is simpler. LinkedBlockingQueue is the lowest overhead concurrent queue implementation for append-only situations.

Do we want to support removal from results to allow the GC to collect memory if needed, rather than buffering everything in this list? If so, LinkedBlockingDeque, which can efficiently remove from the head of the queue without resizing/moving arrays) may be useful. It could be efficiently done in hasActiveOperations without much overhead. I wasn't previously aware that we had a place in the code that buffered everything (ie, tuple and query results) for a transaction while it was alive, even after the operations were complete.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
ansell: Yes. My assumption, based on looking through code lines related to #684 was that high accuracy on the hasActiveOperations method is required to make sure the Transaction is not closed prematurely, which may have in the past caused BackgroundGraphResult threads to be isolated by broken close methods throughout the codebase wrt exceptions being thrown while close was proceeding.

The use of an internal single threaded ExecutorService gives ordering guarantees only if the user had already given operations in order (which they can't given the asynchronous closing of transactions in ActiveTransactionRegistry). These changes make both the threads inside of the ExecutorService and the ActiveTransactionRegistry get (almost) atomic snapshots of the state, so when the close blocks on the semaphore it also blocks calls to hasActiveOperations until it is ready to return. If hasActiveOperations begins and gets both the lock (so new operations cannot be submitted), and the semaphore, (so existing operations are either blocked on starting, or have only just finished), then it receives as accurate a state as possible while it is iterating. The only case then is a safe case with respect to cache removal, that of "hasActiveOperations" returning true because a Callable has just released the semaphore but the ExecutorService has not yet set isDone to true.

However, that may be entirely irrelevant and the semaphore can be easily removed and hasActiveOperations would then be independent again of the state of running tasks.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
jeenbroekstra: I think I never bothered because I assumed that 1) the Transaction lifetime isn't very long and 2) the Future objects aren't particularly big. But of course in cases where the operation is a query, futures can become big (because they keep an internal reference to the query result), so it might indeed be worth it cleaning them up.

core/http/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
jeenbroekstra: I still haven't quite gotten my head around the issue. Let me talk my way through this: 

In the previous setup (synchronized blocks, no semaphore), when a new operation was invoked on the transaction, this locked (via `synchronized`) on the futures list. An operation was scheduled in the executor and its future added to the list, after which the lock is released. `hasActiveOperations` itself also locks on `futures` access. Which means that while `hasActiveOperations` is running, no other threads can add new operations to the transaction. Correct sofar? 

Potential problem 1: `hasActiveOperations` loops over `Future` objects that may concurrently change their state to `done`. It iterates until it finds at least one future that _isn't_ done. So the question is: is there a problem when a future gets adapted during looping? As far I can tell, there isn't: if it checks whether Future `x` is done and it gets a `false` (and immediately after that `x` is completed), it just continues iterating. Worst case is that it mistakenly returns `true` because it believes `x` is still active, while in reality it's already completed. So a false positive, which is no problem from a cleanup perspective.

Potential problem 2: `hasActiveOperations` loops and finds no active operations, however concurrently another thread is waiting to obtain a lock to schedule a new operation and add it to the futures list. `hasActiveOperations` returns `false`, the registry closes and removes the transaction, despite the fact that a new operation was in fact received. This sounds dire, but I think it can't actually happen: every incoming transaction operation first obtains the Transaction object from the registry. So it either obtains the Transaction object (which means the timer is reset, so it won't activate the registry cleanup for that object and therefore won't invoke `hasActiveOperations`), or the `Transaction` has already been cleaned up, and is therefore no longer accessible from the cache, resulting in an _expected_ timeout error. 

Sorry to ramble. Do you see any flaws in my thinking? I'd be happy to add the semaphore stuff in, but I am slightly wary of over-complicating the locking logic without good cause, so I want to make sure I understand _why_ this is necessary :)

ansell: I was concerned about problem 2 and invalidation for sparsely accessed transactions and hasActiveOperations returning false because. However, it shouldn't be an issue, based on the cache revalidating just before each operation.

I have removed semaphores and switched back to List. The only practical change now to hasActiveOperations is that .remove is called until we get to one where .isDone returns false (given the single threaded FIFO queue ExecutorService, we don't need to look further than that, based on the atomic submit+add which is now neatly packaged up in the submit helper method. The other changes are all to make the close method more reliable, and add the isClosed and isComplete accessors for future use.

ansell:
I also simplified the awaitTermination logic in a commit after you approved the changes (removing it and its variables entirely) by calling shutdown while we have the lock to prevent future threads which were waiting on the lock from being able to successfully submit new tasks. This was implemented by adding a new helper method, submitAndShutdown.

jeenbroekstra:
LGTM.

ansell:
Travis passed: https://travis-ci.org/eclipse/rdf4j/builds/185652830

This patch doesn't have the elastic search test fixes so it may not pass on Hudson.

PR #690, opened 2016-12-07
Fix #689: bug in ZeroLengthPathIteration + cost estimate improvement in planner

### LINKED ISSUE #689: Zero or more path query is broken with some datasets
We've just upgraded our software from using Sesame 2.7.15 to rdf4j 2.1.2 and found that property paths (https://www.w3.org/TR/sparql11-property-paths/#path-language) that use * do not work with some datasets. I had trouble building up a dataset from scratch to reproduce the issue (it seemed to work on smaller datasets) so instead took the larger dataset that we did have that was showing the issue and have attached that (after scrambling some IP out of it). There are screen captures showing the steps to reproduce the issue and the dataset and queries.

We have been able to work around the issue by re-writing the query using a UNION clause, but have now found another query that doesn't seem to work, even after doing that. At this point we can't upgrade as we are not confident that the queries will continue to work correctly as the underlying dataset changes.

[rdf4j-bug.zip](https://github.com/eclipse/rdf4j/files/635373/rdf4j-bug.zip)

jeenbroekstra:
This PR addresses GitHub issue: #689 .

* QueryJoinOptimizer incorrectly reorganized joins involving ArbitraryLengthPaths due to too-low estimate of cost.
* cost estimate now based on store estimate of single statement pattern involving the ALP's subject and object, and an unknown predicate, times 2 (we estimate an ALP's path length will be at least 2). 
* Tested manually that the bug is specifically caused by the optimizer, not by the execution engine: if the optimizer is completely disabled, the engine does compute the correct query result.

### DIFF
--- a/core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ZeroLengthPathIterationTest.java
+++ b/core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ZeroLengthPathIterationTest.java
@@ -1,0 +1,89 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.query.algebra.evaluation.iterator;
+
+import static org.junit.Assert.assertTrue;
+
+import org.eclipse.rdf4j.common.iteration.CloseableIteration;
+import org.eclipse.rdf4j.common.iteration.CloseableIteratorIteration;
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.LinkedHashModel;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.query.BindingSet;
+import org.eclipse.rdf4j.query.QueryEvaluationException;
+import org.eclipse.rdf4j.query.algebra.Var;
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Unfortunately this is problematic to write a regression test for as the bug is only apparent for sufficiently large datasets (in which case store statistics for individual statement patterns become significant enough to influence join order).

ansell:
Adding a short test to https://github.com/eclipse/rdf4j/blob/master/testsuites/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ComplexSPARQLQueryTest.java#L1469 for /* patterns may still be useful (unless it is tested by another testsuite already) even if it doesn't trigger the bug.

Given the fix is adjusting a heuristically defined parameter, there may be other cases that aren't tested, but if you could programmatically generate a file that triggers it it may be useful.

What effect does turning off optimisation for property paths have?

jeenbroekstra:
Good point re: the tests, I'll have a closer look. 

Turning off optimization simply leaves the original join order intact. The reordered join tree evaluates the ALP too early, which leaves several of its variables unbound. To be fair: this should still work regardless of how inefficient it is, so there may be a deeper cause that I haven't yet pinpointed.

jeenbroekstra:
Found the root cause, a bug in ZeroLengthPathIteration (see latest commit). I'll work on some regression tests. I'll leave the statistics improvement in place as well.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ZeroLengthPathIterationTest.java
```diff
@@ -0,0 +1,89 @@
...
+			}
+		};
+		evaluator = new StrictEvaluationStrategy(ts, null);
+	}
+
+	@Test
+	public void testRetainInputBindings() {
+
+		MapBindingSet bindings = new MapBindingSet();
+		bindings.addBinding("a", RDF.FIRST);
```
ansell: Just to check I understand this, the regression test is that "a", which isn't otherwise involved in the zero length path, is still included in the bindings along with "x" and "y"

jeenbroekstra: Exactly. The problem was that ZeroLengthPathIteration, in a certain code path (namely when none of its involved variables have a binding), simply create a completely new BindingSet from scratch, omitting all the input bindings - which means it would effectively erase bindings insert by a VALUES clause. 

I'll add some comments to the test to make this clearer.

jeenbroekstra: Btw this also explains why the problem was related to the statistics / dataset size: on smaller datasets the ZLP evaluation is pushed further down, resulting in its subject variable being given a fixed value - so it followed a different code path and didn't erase any bindings.

PR #692, opened 2016-12-09
Fix #685: sparql decimal/integer parsing

### LINKED ISSUE #685: SPARQL parser fails to parse integer object value immediately followed by full stop
Here is sparql query which RDF4J fails to parse properly:
```
select ?Concept where { 
  ?Concept a 1. 
  ?Concept2 a 1.
} LIMIT 100
```
Probably this is a bug in RDF4J parser, but I'm not quite sure.
Tested with RDF4J workbench v2.1.2.

See also discussion at https://github.com/RubenVerborgh/SPARQL.js/pull/33

jeenbroekstra:
This PR addresses GitHub issue: #685 .

* grammar DECIMAL production updated to conform to SPARQL 1.1 Recommendation
* BGP ending with integer object value followed by a full-stop now parsed correctly
* added regression test


core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/JJTSyntaxTreeBuilderState.java
```diff
@@ -1,137 +1,123 @@
-/*******************************************************************************
```
ansell: Can javacc be modified to add the license header?

jeenbroekstra: Hm, good catch, I'll have a look.

PR #694, opened 2016-12-13
Fix #693: sailsourcemodel concurrency

### LINKED ISSUE #693: Native Store SailSourceModel concurrency bug
This is a followup of issue #680 . The Native Store internally uses a MemoryOverflowModel to cope with large transaction sizes by caching data to disk. This in turn uses a SailSourceModel to physically store and retrieve the cached data. SailSourceModel does not properly synchronize its public access methods, which means that when concurrent txns try to access it, it can lead to intermittent exceptions on manipulation or retrieval of data. Example stacktrace:

```
Exception in thread "Thread-2" java.lang.NullPointerException
	at org.eclipse.rdf4j.sail.nativerdf.SailSourceModel.dataset(SailSourceModel.java:397)
	at org.eclipse.rdf4j.sail.nativerdf.SailSourceModel.contains(SailSourceModel.java:227)
	at org.eclipse.rdf4j.sail.nativerdf.MemoryOverflowModel.contains(MemoryOverflowModel.java:111)
	at org.eclipse.rdf4j.model.impl.FilteredModel.contains(FilteredModel.java:147)
	at org.eclipse.rdf4j.model.impl.AbstractModel.isEmpty(AbstractModel.java:45)
	at org.eclipse.rdf4j.sail.base.SailDatasetImpl.union(SailDatasetImpl.java:336)
	at org.eclipse.rdf4j.sail.base.SailDatasetImpl.getStatements(SailDatasetImpl.java:305)
	at org.eclipse.rdf4j.sail.base.DelegatingSailDataset.getStatements(DelegatingSailDataset.java:74)
	at org.eclipse.rdf4j.sail.base.DelegatingSailDataset.getStatements(DelegatingSailDataset.java:74)
	at org.eclipse.rdf4j.sail.base.SailSourceConnection.remove(SailSourceConnection.java:720)
	at org.eclipse.rdf4j.sail.base.SailSourceConnection.removeStatement(SailSourceConnection.java:551)
	at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.removeStatements(AbstractSailConnection.java:491)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.removeWithoutCommit(SailRepositoryConnection.java:382)
	at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.remove(AbstractRepositoryConnection.java:587)
```

jeenbroekstra:
This PR addresses GitHub issue: #693 .

* made all public access methods on SailSourceModel synchronized
* added a compliance test that verifies large concurrent transactions do not cause errors or inconsistencies

Note that due to the nature of the problem, the added test may fail/succeed intermittently rather than consistently. I have, when running locally, been able to reproduce the error every two runs of the test, on average, and get consistent succes after applying the synchronization fixes.

### DIFF
--- a/compliance/store/src/test/resources/logback-test.xml
+++ b/compliance/store/src/test/resources/logback-test.xml
@@ -8,7 +8,7 @@
 	</appender>
 	
 	<root>
-		<level value="warn" />
+		<level value="trace" />
 		<appender-ref ref="STDOUT" />
 	</root>
 </configuration>
--- a/compliance/store/src/test/resources/logback-test.xml
+++ b/compliance/store/src/test/resources/logback-test.xml
@@ -8,7 +8,7 @@
 	</appender>
 	
 	<root>
-		<level value="trace" />
+		<level value="warn" />
 		<appender-ref ref="STDOUT" />
 	</root>
 </configuration>
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
@@ -26,6 +26,8 @@ import org.eclipse.rdf4j.model.impl.TreeModelFactory;
 import org.eclipse.rdf4j.query.algebra.StatementPattern;
 import org.eclipse.rdf4j.query.algebra.Var;
 import org.eclipse.rdf4j.sail.SailException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * An {@link SailSource} that keeps a delta of its state from a backing {@link SailSource}.
@@ -33,6 +35,8 @@ import org.eclipse.rdf4j.sail.SailException;
  * @author James Leigh
  */
 class SailSourceBranch implements SailSource {
+
+	private static final Logger logger = LoggerFactory.getLogger(SailSourceBranch.class);
 
 	/**
 	 * Used to prevent changes to this object's field from multiple threads.
@@ -270,10 +274,14 @@ class SailSourceBranch implements SailSource {
 			semaphore.lock();
 			if (!changes.isEmpty()) {
 				if (prepared == null) {
+					logger.trace("creating new prepared");
 					prepare();
 				}
+				logger.trace("flushing prepared {}", prepared);
 				flush(prepared);
+				logger.trace("flushing step 2");
 				prepared.flush();
+				logger.trace("prep flushed");
 				try {
 					if (prepared != serializable) {
 						prepared.close();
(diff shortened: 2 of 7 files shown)


jeenbroekstra:
Folks, I need some fresh eyes on this problem. See also discussion on issue #693 and #680. It appears that the fixes I did in this PR do not quite solve the problem. I've spent most of the day trying to debug the issue but I'm not really getting anywhere. It's definitely a concurrency issue as it happens intermittently (and, rather nastily, never when I step through with a debugger, so I'm reduced to output trace logging). From what I can tell, in some cases, one of the concurrent threads does not fully sync its pending changes to the master store. It looks as if one of the SailSourceModels created by one of the transactions somehow just gets "lost". I have verified that statements are added to it as normal, but for some reason those changes are never finalized and merged into the master store. 

I'm running out of ideas here. Would really appreciate some help. @ansell , @nikolavp , @damyan-ognyanov , @aschwarte10 , if you have some time, have a look please, and tell me what I'm missing.

jeenbroekstra:
As an aside: to debug this problem locally (rather than having to work with a remote server) I rolled back the changes I made in this PR. _With_ those fixes the problem does not occur when running locally, but it does still happen when running remote. So I figured the underlying cause is still there, and the synchronization I added merely influences the timing a bit. Easier to debug when running local, so...

jeenbroekstra:
Core of the issue seems to be that `SailSourceBranch` obtains references to all outstanding changes (from concurrent threads) and tries to sync/merge those changes. Part of that process is the flushing and then closing of the Model objects that hold the approved changes, but if the Model uses a NativeSailStore for disk storage, it may be impossible to flush/closes as it internally uses a ReentrantLock, which may currently be held by a different thread. 

In previous versions this gave us an IllegalMonitorStateException when it tried to unlock when it had no "right to". We fixed that in #680, but now the issue seems to be that instead of trying to sync and faling, it now simply exits silently - and the changes therefore get dropped.

jeenbroekstra:
Investigating a change that is either brilliantly simple or simply stupid: I'm looking at removing the ReentrantLock from NativeSailStore and instead just relying on basic synchronization to ensure serialized writing access to the triplestore. Stay tuned.

ansell:
In the synchronized code, what is stopping other threads from accessing/mutating the stores concurrently? Ie, not all of the methods are synchronized, but it isn't clear whether there is blocking code to make concurrent threads wait.

jeenbroekstra:
@ansell. Nothing. This is merely my first attempt to simplify things. You gotta wonder about the fact that with all the locking removed, suddenly all our tests (including those with concurrent modifications) succeed. But I'm looking at hardening it further with a new, simpler ReentrantLock strategy. Stay tuned.

jeenbroekstra:
@ansell added locking. Have another look if you can.

jeenbroekstra:
I've ran all unit and integration tests multiple times, without failures. Travis also succeeds, and I have also re-ran the user-submitted test code supplied by @rknowles9 in #693, several times. No errors, and expected outcomes (two contexts with the expected number of statements).

jeenbroekstra:
Unless I hear reactions to the contrary, I will merge this in tomorrow and publish it as part of the 2.1.4 patch release.

Issue #682, opened 2016-12-01
Clean up compliance tests to avoid resource leaks


jeenbroekstra:
Current compliance test implementations often fail to properly close connections or shut down repositories, especially if an error occurs during testing. This produces a lot of unnecessary error output on the build log, and in addition may interfere with subsequent tests. 

To fix, the compliance tests should make rigorous use of `try-with-resources`, ensuring RepositoryConnections and QueryResults are always properly closed even when exceptions occur.

ansell:
I am going to work on this, found some new stacktraces (mixed into all of the previous stacktraces) in the debugging logs from the new close operations added to Background(Graph/Tuple)Result.run and want to make sure things will still work as expected.

In particular, I want to remove e.printStackTrace from places where we are expecting or allowing an exception to occur. They are possibly masking other cases where tests may have started silently failing and we didn't notice them in the jumble.

jeenbroekstra:
Go for it @ansell . I can't assign this to you (still need to ask about allowing contributors to be assignee) but it's all yours.

PR #701, opened 2016-12-22
Issue #682 Cleanup test resource closure

ansell:
This PR addresses GitHub issue: #682 .

* Close resources reliably in tests, using try-with-resources for most cases
* Turns on debugging for tests, to highlight cases where iterations are being silently closed by the automatic closer. Only known case where the automatic closer is still kicking in is testOrderByQueriesAreInterruptable, although the relevant code blocks still look like they should work as desired, and they have been adjusted slightly in this pull request to suit.
* Closes iterations in ParallelJoinCursor if they are not added to the queue successfully, which was highlighted by adding the debugging
* Adds Autocloseable to SailConnection. No practical effect on the API, as the close method already existed. This just enables identification by static analysis of cases where SailConnection objects are going out of scope without being closed.


ansell:
I am investigating why ParallelJoinCursor is having so many issues on Travis, causing the log file to grow too large and Travis to kill the tests, but not reliably able to replicate it locally yet.

ansell:
Fixed the logging issue by disabling it in RepositoryConnectionTest. Given that the close methods are all executed correctly in debug mode, I am imagining that there isn't any actual issue shown by the AbstractSailConnection.forceCloseActiveOperations method kicking in and closing before the iteration gets to closing itself when it is interrupted by TimerTask/TimeLimitIteration. The design of AbstractSailConnection.close, which goes straight to forceCloseActiveOperations may have a factor, but it doesn't appear to be broken at this stage. The Travis builds are not showing the issue anymore on FederationConnectionTest where it was occurring and ParallelJoinCursor seems to be stable now.

This is therefore ready for review.

PR #705, opened 2017-01-01
Fix #703: service clause bind clause varname fix + unit test

### LINKED ISSUE #703: Vars in statement patterns in service are not added to service vars
```sparql
       select * {
               service <http://localhost:7200/repositories/test> {
               ?s <urn:p> ?o. bind (?o + 1*?foo as ?opp) . bind(5-?opp as ?next) . values ?foo {7 8 9}
               {select (NOW() as ?zoo){}}
               }
       }
```
here the variales *o* and *s* are not added. This is related to the change from https://github.com/eclipse/rdf4j/issues/646. It seems that for some reason the statement pattern is attached to the extension element and it needs to be visited recursively otherwise the vars from the statement pattern are not added.

jeenbroekstra:
This PR addresses GitHub issue: #703 .

* `Service.computeServiceVars` now also visits all descendents of `Extension` objects.
* added unit test, greenlines after fix.
* update test case javadoc


jeenbroekstra:
@nikolavp  please review

nikolavp:
LGTM

PR #715, opened 2017-01-12
Fix #713: Support for late evaluation of geo distance search

### LINKED ISSUE #713: Support late evaluation of geo distance search
This is a continuation of issues/#220. The changes made to support late evaluation of full text search queries should also be made to geo distance search. This means all the search inputs are not required to be known upfront and can be determined as part of sparql evaluation.

pulquero:
This PR addresses GitHub issue: #713  .

This is a PR for work that was started as part of #220.

### DIFF
--- a/core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/DistanceQuerySpec.java
+++ b/core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/DistanceQuerySpec.java
@@ -11,61 +11,98 @@ import java.util.List;
 
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Literal;
+import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.query.algebra.EmptySet;
 import org.eclipse.rdf4j.query.algebra.Extension;
 import org.eclipse.rdf4j.query.algebra.ExtensionElem;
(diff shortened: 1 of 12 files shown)


core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/DistanceQuerySpec.java
```diff
@@ -11,63 +11,100 @@
...
+		this.distanceFunction = null;
+		this.distanceExpr = null;
+		this.filter = null;
 	}
 
 	public Literal getFrom() {
 		return from;
 	}
 
-	public void setUnits(IRI units) {
```
jeenbroekstra: Is this a class that you would expect user code to access directly? Or is this an internal functionality thing only? The reason I ask is that if this class is part of the public API, then this is an API-breaking change, which we can only do in a major release.

pulquero: It is internal functionality, but as it is a setter I can easily add it back in.

ansell: I like final instance variables as a pattern, but I agree that this would break the API and doesn't seem necessary for the other changes to be useful.

jeenbroekstra:
Fixed issue numbers. The actual issue that this corresponds to is #713, not #714 (which is a PR), nor #716 .

PR #720, opened 2017-01-13
Fix #719: build configuration tweaks

### LINKED ISSUE #719: Tweak project build (memory) settings
Travis is currently unreliable, killing builds every so often. The hunch is that this is related to the amount of allocated memory, we should try and tweak the config to explicitly set a higher heap size for both compilation and test execution.

I will also look into bumping the versions of some of our basic maven plugins, like the compiler, failsafe, and surefire plugins. We're on ancient versions for some of this and while it currently works, I gotta believe newer versions will be more efficient and stable. 

And while we're at it, we should remove the MaxPermSize setting from our config since that is no longer used in Java 8.

jeenbroekstra:
This PR addresses GitHub issue: #719 .

* maven failsafe and surefire plugins both use increased heap space for forked jvms 
* obsolete MaxPermSize parameter removed
* bumped versions of various core build / test plugins (compiler, jar, clean, failsafe, surefire). 
* travis environment tweaked to set default maven heap size to 1G

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ branches:
   only:
   - master
   - /^releases.*$/
+env:
+  - MAVEN_OPTS=-Xmx1024M
--- a/pom.xml
+++ b/pom.xml
@@ -659,7 +659,7 @@
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-jar-plugin</artifactId>
-					<version>3.0.2</version>
+					<version>2.3.1</version>
 				</plugin>
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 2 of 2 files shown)


ansell:
Well it failed with a different error this time ;) Changes look good though.

jeenbroekstra:
I really hate build servers. all of them.

jeenbroekstra:
Also doing a build with Hudson: https://hudson.eclipse.org/rdf4j/job/rdf4j-verify-pr/48/

jeenbroekstra:
At first glance the problem appears related to how the test suite jars are packaged (it fails on a missing MANIFEST.MF file somewhere). I'll see about rolling back that change first. We might get lucky.

jeenbroekstra:
Ok that just made my weekend :)

PR #724, opened 2017-01-17
Fix #721: remove RDF4J-specific params from SPARQL Protocol comms and refactor

### LINKED ISSUE #721: SPARQLRepository sends non-standard parameters
SPARQLRepository, when sending requests to a SPARQL endpoint, sends non-standard request parameters, including `queryLn`. 

While this is not a problem for most endpoint implementations (they simply ignore them), a few known endpoint implementations (such as MarkLogic) choke on this. It would be good to remove these parameters from SPARQLRepository requests.

jeenbroekstra:
This PR addresses GitHub issue: #721 .

* SparqlSession (renamed to SPARQLProtocolSession) no longer adds RDF4J-specific request parameters to updates and queries.
* Refactored httpclient classes and interfaces for clearer indication of purpose, and getting rid of legacy names such as "SesameClient". 
* added deprecated classes, interfaces, and methods to keep code backward-compatible. 
* Several context-related compliance tests disabled for SPARQLRepository since it turns out they relied on RDF4J-specific request params (so these tests would fail if executed against a non-RDF4J SPARQL endpoint).


jeenbroekstra:
As it's rather a significant refactor I'd like some opinions on it, especially from @nikolavp , @damyan-ognyanov , @ansell , if possible.

ansell:
I like the new names. I was always getting confused with the previous names what role each class had, but the new names are completely clear IMO. 

However, I am not sure what practical effect the change will have on backwards compatibility and don't have time to review it right now. I have been meaning to suggest that we try adding japicmp to programmatically understand API changes but haven't got around to it yet.

The only place I have used it so far is JSONLD-Java, but it seems to work okay there to generate reports about API changes:

https://github.com/jsonld-java/jsonld-java/blob/f713c7e53e57c173da532ab66facc1e793bdc1d0/pom.xml#L328

jeenbroekstra:
I ran japicmp reporting on this PR (comparing to 2.1.6 API), and it reports no source incompatibilities. I also went over some of the fields and methods reported as a "removed" again and have verified that they are still in place (in superclasses or super-interfaces), and can still be accessed without changing client code. As far as I'm concerned this is ready to be merged.

PR #726, opened 2017-01-17
Fix #678: removed hashcode implementation details from javadoc

### LINKED ISSUE #678: Remove implementation details from Javadoc for Literal.hashCode and BindingSet.hashCode
Followup of #676 and #668 . The Javadoc in the interfaces for Literal and BindingSet contains too much implementation detail. This should be removed as it can falsely lead users to assume the implementation will never change.

jeenbroekstra:
This PR addresses GitHub issue: #678 .

* javadoc for model objects (and `BindingSet`) cleaned up: implementation details of hashcode algorithms removed, added override annotations, reformatted where necessary.

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/BNode.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/BNode.java
@@ -32,12 +32,14 @@ public interface BNode extends Resource {
 	 * @return <tt>true</tt> if the other object is an instance of {@link BNode} and their IDs are equal,
 	 *         <tt>false</tt> otherwise.
 	 */
+	@Override
 	public boolean equals(Object o);
 
 	/**
-	 * The hash code of a blank node.
+	 * Returns the hash code value of the blank node.
 	 * 
 	 * @return A hash code for the blank node.
 	 */
+	@Override
 	public int hashCode();
 }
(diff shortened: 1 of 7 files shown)


ansell:
I disagree with the removal on the principle that we don't track or convert the types of objects in most of our APIs (MemValueFactory and similar excepted), so we need hashcodes to be compatible.

jeenbroekstra:
@ansell I'm not sure I follow. The rationale behind removing the implementation detail was that it falsely leads users to believe that the hashcode will never change between releases.  Are you suggesting that we should in fact make that commitment? Can you elaborate on why you think this is necessary?

ansell:
Users will virtually always be using Value objects from different ValueFactory's, both internally and from third-parties. If they add these Value objects to a HashSet or a LinkedHashModel, they want to be sure that they will obtain consistent results. Having the hashCode contract stated explicitly (whatever its contract turns out to be) is the only way to ensure that core promise of the Java Collections API (and out Model API) in the long term across different implementations. This partly relates to the removal of ValueFactory when we changed from Graph to Model, but users will expect to be able to use the hash-based Java Collections API in the same way.

ansell:
It is possible that they may change, but the only possible stable point at which to change the hashcode contract would be during a major release IMO.

PR #731, opened 2017-01-18
Fix #181: federation scoping of joins from different graphs

### LINKED ISSUE #181: Federation does not correctly scope joins from different graphs
The following test does not pass:

``` java
        SailRepository repo1 = new SailRepository(new MemoryStore());
        repo1.initialize();
        SailRepository repo2 = new SailRepository(new MemoryStore());
        repo2.initialize();
        Federation fed = new Federation();
        fed.addMember(repo1);
        fed.addMember(repo2);
        SailRepository repoFed = new SailRepository(fed);
        repoFed.initialize();
        RepositoryConnection conn = repo1.getConnection();
        conn.add(getClass().getResource("/testcases-dawg/data-r2/algebra/var-scope-join-1.ttl"), null, null, conn.getValueFactory().createURI("http://example/g1"));
        conn.close();
        conn = repo2.getConnection();
        conn.add(getClass().getResource("/testcases-dawg/data-r2/algebra/var-scope-join-1.ttl"), null, null, conn.getValueFactory().createURI("http://example/g2"));
        conn.close();
        String query = "PREFIX : <http://example/> SELECT * { graph :g1 {?X  :name 'paul'} { graph :g2 {?Y :name 'george' . OPTIONAL { ?X :email ?Z } } } }";
        conn = repoFed.getConnection();
        TupleQuery tq = conn.prepareTupleQuery(QueryLanguage.SPARQL, query);
        TupleQueryResult tqr = tq.evaluate();
        boolean hasResults = tqr.hasNext();
        conn.close();
        assertFalse(hasResults);
```

but the equivalent does for a MemoryStore.

Possible fix is to replace the use of TupleExprs.containsProjection(rightArg) with

>    TupleExprs.containsProjection(expr) || (expr instanceof OwnedTupleExpr && ((OwnedTupleExpr)expr).getQuery() != null)

as expr when executed directly by the owner is also essentially a sub-query/projection.

pulquero:
This PR addresses GitHub issue: #181  .

* Better naming to Federation compliance tests.
* Test case for #181.
* Fix for #181.

### DIFF
--- a/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/algebra/OwnedTupleExpr.java
+++ b/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/algebra/OwnedTupleExpr.java
@@ -55,6 +55,10 @@ public class OwnedTupleExpr extends UnaryTupleOperator {
 		this.variables = bindings;
 	}
 
+	public TupleQuery getQuery() {
+		return query;
+	}
+
 	public CloseableIteration<BindingSet, QueryEvaluationException> evaluate(Dataset dataset,
 			BindingSet bindings)
 		throws QueryEvaluationException
--- a/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/algebra/OwnedTupleExpr.java
+++ b/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/algebra/OwnedTupleExpr.java
@@ -55,8 +55,8 @@ public class OwnedTupleExpr extends UnaryTupleOperator {
 		this.variables = bindings;
 	}
 
-	public TupleQuery getQuery() {
-		return query;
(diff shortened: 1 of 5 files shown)


core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/algebra/OwnedTupleExpr.java
```diff
@@ -55,6 +55,10 @@ public void prepare(QueryLanguage queryLn, String qry, Map<String, String> bindi
 		this.variables = bindings;
 	}
 
+	public TupleQuery getQuery() {
```
jeenbroekstra: I'm a bit wary of publicly exposing too many internals. The only thing I see you do with it is retrieving it to check it's not null. Perhaps a boolean hasQuery method would be better?

Issue #732, opened 2017-01-18
Add japicmp maven plugin to check API changes


ansell:
The japicmp maven plugin can be used to programmatically generate reports about API changes. This is useful even before we define the Public API, so we can get a feel for how likely refactorings are to effect users:

https://siom79.github.io/japicmp/MavenPlugin.html

My previous use has been in JSONLD-Java where it works fairly well so far:

https://github.com/jsonld-java/jsonld-java/blob/f713c7e53e57c173da532ab66facc1e793bdc1d0/pom.xml#L328

ansell:
Relates to #619 and #724

ansell:
I will add some initially to the api modules that are definitely on the Public API but not others which we are not sure about per #619. They will only generate reports so far, so they won't fail the build if incompatible changes are found, but that is an option in future.

jeenbroekstra:
Sounds good. Can you point us to an example of what one of these reports looks like?

jeenbroekstra:
Never mind, found an example on the plugin page itself: https://siom79.github.io/japicmp/Examples.html . Looks pretty straightforward?

ansell:
Yes, they are very straightforward as long as you turn on the "onlyModified" setting. They are generated on build into the target/japicmp/ directory as HTML files. The files are basically empty if no interfaces have changed, and populated with all of the relevant classes/methods/fields if any were added/removed/changed.

The reports differentiate between source and binary compatible. "Binary incompatible changes are marked with (!) while source incompatible changes are marked with (*)." I don't think we expect our code to be binary compatible (based on past behaviour), but we do want it to be source compatible as far as practical.

The default is also to look at protected/public, which I think is relevant for us in some cases with our Abstract implementations that we intend third-parties to reuse. We could change it to "public-only" but it wouldn't be as useful.

We can also customise it to only particular packages or classes, which will help for some of the embedded "impl" packages inside of api modules.

The code also shows up serialisation issues, which we have some of between 2.0 and 2.2-SNAPSHOT.

It is a little annoying in our large scale case that the reports are not yet aggregated to the reactor parent, but it should still be useful for us to review pull requests.

If you run "mvn clean install -DskipTests" on the pull request #733 code you can get a fairly quick idea of what some of the reports look like.

Issue #619, opened 2016-10-14
Declare our Public API for use with semantic versioning


jeenbroekstra:
We are aiming to apply [semantic versioning](http://semver.org/), and to a large degree already follow its principles. However, a requirement of SemVer is that we have explicitly specified what our public API is.  

Clearly, not every public class/method in the RDF4J code base is considered part of the public API - some of these are "internal", and are declared public for technical reasons rather than a statement that every user can/should use them and rely on their behavior to remain unchanged for minor/patch versions.

A good starting point is that all "...-api" modules and all public interfaces are considered part of the public API. But we'll need to go further than that - I assume at least a large part of the abstract implementations and a lot of utility code and _some_ public classes should be considered part of the public API as well.

Two main questions:
1. what's the best way to document/specify what we consider the RDF4J public API?
2. which parts of the code base should be part of it?

ansell:
As you say the -api modules are a given for consideration, although the ".impl" or ".util" packages within them could be off limits if they are only intended to be used by other RDF4J modules.

The repository-sail module is a substantially reused link between the 

There is a clear service driven access service to the specific parser modules within rio along with extensive configuration options that should make it possible to leave the individual parser implementations and entire modules off the public API. However, given that they are reused and overridden by third-parties, we should still take them into account and try hard to deprecate signatures in them rather than remove signatures.

The service driven creation and configuration for the sail and repository APIs are not as broadly used and almost all calls into them are done through constructors for the various Sail/Repository implementations. So it may require some modification to the service driven approach for those to make them a little simpler and more broadly used before pushing those constructors to the internal API.

In terms of programmatic verification of the API before releases, one recent tool that we could look at is: 

https://github.com/siom79/japicmp

Another is the animal sniffer maven plugin:

http://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/

It may be useful to put removals to a broader public API off until a major version bump such as 3.0 and work with deprecations solely before then.

jeenbroekstra:
I've done several attempts to build a draft spec of what is considered "in" and "out" of our public API, but I must admit I'm struggling to make any rules that stop short of enumerating every single package. 

[...]

PR #733, opened 2017-01-19
issue #732 : Add support for japicmp to core modules

ansell:
This PR addresses GitHub issue: #732 .

* Add japicmp maven plugin to core modules to start gathering knowledge about the scope of our likely public API for #619
* Reenables animal sniffer with the java-8 signatures. This will be useful with Java-9 betas being used more regularly for experimentation.
* Removes some unused managed dependencies

The japicmp inclusions may not be all of the eventual public API modules, but it forms a likely base. The japicmp plugin does not yet have aggregation support, so each of the individual modules that will make up the public API need to have the plugin added to them manually.

Notably, none of the rio or queryresultio modules need to be in the public API IMO, as they are all accessible as services by all of our core APIs solely using rio-api and queryresultio-api.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,6 +163,7 @@
 		<httpcore.version>4.4.4</httpcore.version>
 		<jackson.version>2.8.2</jackson.version>
 		<jsonldjava.version>0.8.3</jsonldjava.version>
+		<last.japicmp.compare.version>2.0</last.japicmp.compare.version>
 	</properties>
 
 	<dependencyManagement>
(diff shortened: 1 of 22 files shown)


jeenbroekstra:
Interesting. I just checked this out locally and ran a build, and the reports definitely look useful. However, somewhat surprisingly it seems to compare the current snapshot against the 2.0 release (rather than against 2.1). Is this something that we can configure somehow?

pom.xml
```diff
@@ -163,6 +163,7 @@
 		<httpcore.version>4.4.4</httpcore.version>
 		<jackson.version>2.8.2</jackson.version>
 		<jsonldjava.version>0.8.3</jsonldjava.version>
+		<last.japicmp.compare.version>2.0</last.japicmp.compare.version>
```
ansell: @jeenbroekstra This is where the compare version is defined by us.

jeenbroekstra:
Happy to merge this in. Do you have any thoughts on how we can best make the generated reports available? Travis doesn't really have any support for accessing generated reports, as far as I know, so we should either tinker with Hudson a bit, or just rely on occasionally checking this locally.

ansell:
Until there is a single aggregated report available, it would require a manual trigger to collect up the files. Tracking issue for that is:

https://github.com/siom79/japicmp/issues/156

Once there is a single aggregated report available in the plugin we could more easily check them locally, and possibly figure out a way of putting the reports on the interwebs.

Once we are happy with the way it works, we could also add a "fail" mode to the release profile or a new "check" profile to make the maven build fail rather than having to manually check the results.

jeenbroekstra:
Works for me. Shall I merge?

ansell:
Yes, merging is okay with me.

Issue #738, opened 2017-01-26
Pull based parsers for RDF & Tuple formats


RickMoynihan:
The current parsers in RDF4j all follow a push based model, where they control the execution flow entirely - firing events at a statement handler.

This makes it awkward to support certain use cases which are better suited to a pull based parser.  For example currently if you wanted to display a paginated list of statements/tuples without consuming the whole stream, you'd have to run the parser on a separate thread and block the parser with something like a `java.util.concurrent.BlockingQueue` to communicate results to your display; then when a page is displayed stop the parser by signaling your handler to raise an exception.

Another more general use case is in supporting a functional/lazy programming style where the amount of data to be processed is decided by a `take`that occurs outside of the parse code.  Grafter (a Clojure wrapper for RDF processing over sesame) takes this approach and has to jump through some of [the hoops mentioned above](https://github.com/Swirrl/grafter/blob/master/src/rdf-common/grafter/rdf/io.clj#L534) to allow an RDF stream to be viewed as a lazy sequence.  I suspect also that trying to do things like this with Java 8 streams will result in the same kinds of work around.

It would be nice if all the sesame parsers, for both RDF and Tuple formats could support these use cases.  One approach would be to provide a pull based interface to all of the parsers.  Another might be to provide a method for the current push based parsers to terminate the parse as either a success or a failure early.

Obviously also the pull based interface could theoretically be built on top of the push based implementation or vice versa.

ansell:
The entire implementation of the current parsers is based around the handler methodology. 

Blocking as you say can already be done by blocking inside of handleStatement/etc using a Blocking Queue. The lazy approach is simply a BlockingQueue limited to size 1.

Pagination is easily supported using counting handlers along with aborting inside of handleStatement, without changing the internal implementation of the parsers.

ansell:
I was wrong about pagination being easy though, because of blank nodes. However, blank nodes make all RDF decomposition difficult, so nothing new there though.

ansell:
PR #741 has some enhancements related to this issue. In particular, you will be able to use the QueryResults.parseGraphBackground and QueryResultIO.parseTupleBackground methods, along with the QueryResults.limitResults methods to achieve most of what you expect, without having to modify the way any of the parsers internally work. 

This is done using BlockingQueue.put and/or BlockingQueue.offer(e, time, unit) (where the offer is used to check for interruptions before offering the item again without returning from the handlers handle method.

[...]

PR #741, opened 2017-01-27
Fix #740: New background results API

### LINKED ISSUE #740: Standardise methodolgoy for reusing BackgroundGraphResult and BackgroundTupleResult
We need a standard method for reusing both for BackgroundGraphResult and BackgroundTupleResult so they can be easily reused by third-parties who wish to simulate a pull-based parser.

This refers to issue #738

ansell:
This PR addresses GitHub issue: #740 .

* Move BackgroundGraphResult and BackgroundTupleResult (and QueueCursor) to more accessible locations
* Add new API calls that will use the background methods (QueryResults.parseGraphBackground and QueryResultIO.parseTupleBackground) with a small fixed length BlockingQueue to simulate a pull based parser. Parser will now block (ie, wait) on their respective handle methods until the queue has the capacity to allow it to go further.

This pull request addresses part of #738, in that the internal push-based parsing now has an easy to use push-back mechanism to rate-limit and (almost) short-circuit the parsing based on the consumer. In reality, the capacity of the BlockingQueue used imposes a limit on the amount of short-circuiting in play. In this iteration, the capacity is fixed at 1, but QueueCursor has options for various other limits that users can experiment with themselves if 1 doesn't work for them. In the future we can extend the API to customise the capacity or change the default.

In terms of paging that was referred to in #738, the functionality already exists in the form of QueryResults.limitResults, which can be chained with the result of QueryResults.parseGraphBackground or QueryResultIO.parseTupleBackground.

The reasoning for the slightly odd placement of the helper methods relates to the current sub-optimal module layout, where rdf4j-rio-api is a dependency of rdf4j-query and rdf4j-query is a dependency of rdf4j-queryresultio-api.

In the QueueCursor case, rdf4j-query is a much better module in terms of dependencies compared to rdf4j-http-client, which should make it slightly more accessible for reuse.

In this iteration, ForkJoinPool.commonPool() is used to enable users to configure the maximum number of concurrent threads across calls to the parseGraphBackground/parseTupleBackground methods without having to add a dedicated ExecutorService to rdf4j-query.

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.LinkedBlockingQueue;
 
 import javax.xml.datatype.XMLGregorianCalendar;
@@ -220,7 +221,7 @@ public class QueryResults extends Iterations {
 				new QueueCursor<>(new LinkedBlockingQueue<>(1)), parser, in, format.getCharset(), baseURI);
 		boolean allGood = false;
 		try {
-			new Thread(result).start();
+			ForkJoinPool.commonPool().submit(result);
 			allGood = true;
 		}
 		finally {
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
@@ -73,10 +73,14 @@ public class QueryResults extends Iterations {
 		throws QueryEvaluationException
 	{
 		Statement singleResult = null;
-		if (result.hasNext()) {
-			singleResult = result.next();
+		try {
+			if (result.hasNext()) {
+				singleResult = result.next();
+			}
 		}
-		result.close();
+		finally {
+			result.close();
+		}
 		return singleResult;
 	}
 
@@ -90,10 +94,14 @@ public class QueryResults extends Iterations {
 		throws QueryEvaluationException
 	{
 		BindingSet singleResult = null;
-		if (result.hasNext()) {
-			singleResult = result.next();
+		try {
+			if (result.hasNext()) {
+				singleResult = result.next();
+			}
 		}
-		result.close();
+		finally {
+			result.close();
+		}
 		return singleResult;
 	}
 
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
@@ -201,8 +201,9 @@ public class QueryResults extends Iterations {
 	/**
 	 * Parses an RDF document and returns it as a GraphQueryResult object, with parsing done on a separate
 	 * thread in the background.<br>
-	 * IMPORTANT: As this method will spawn a new thread in the background, it is vitally important that it be
-	 * closed consistently when it is no longer required, to prevent resource leaks.
+	 * IMPORTANT: As this method will spawn a new thread in the background, it is vitally important that the
+	 * resulting GraphQueryResult be closed consistently when it is no longer required, to prevent resource
+	 * leaks.
 	 * 
 	 * @param in
 	 *        The {@link InputStream} containing the RDF document.
@@ -210,7 +211,8 @@ public class QueryResults extends Iterations {
 	 *        The base URI for the RDF document.
 	 * @param format
 	 *        The {@link RDFFormat} of the RDF document.
-	 * @return A {@link GraphQueryResult} that parses in the background.
+	 * @return A {@link GraphQueryResult} that parses in the background, and must be closed to prevent
+	 *         resource leaks.
 	 */
 	public static GraphQueryResult parseGraphBackground(InputStream in, String baseURI, RDFFormat format)
 		throws UnsupportedRDFormatException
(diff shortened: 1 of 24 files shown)


ansell:
I am getting the following error while testing, but it is not clear whether it is a regression here or a preexisting condition:

```  SPARQLStoreConnectionTest>RepositoryConnectionTest.testGetStatementsMalformedTypedLiteral:1077 » TestTimedOut
```

ansell:
Travis seems to have passed, so possibly a pre-existing or intermittent error. It may be hard to pin down a regression range in that case given that much of SPARQLConnectionTest is ignored right now and this is the first time I have noticed it failing.

ansell:
I was only able to replicate the error once, and given it is a timeout error, its possible it was just a slow operation and unrelated to this pull request.

core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
```diff
@@ -178,6 +199,40 @@ public static GraphQueryResult limitResults(GraphQueryResult queryResult, long l
 	}
 
 	/**
+	 * Parses an RDF document and returns it as a GraphQueryResult object, with parsing done on a separate
+	 * thread in the background.<br>
+	 * IMPORTANT: As this method will spawn a new thread in the background, it is vitally important that it be
```
jeenbroekstra: Editorial: change "...that it be closed..." to "...that the resulting GraphQueryResult be closed...". Also perhaps this warning should be moved (or repeated) in `@return` javadoc

core/queryresultio/api/src/main/java/org/eclipse/rdf4j/query/resultio/QueryResultIO.java
```diff
@@ -316,24 +320,77 @@ public static TupleQueryResult parseTuple(InputStream in, QueryResultFormat form
...
 		UnsupportedQueryResultFormatException
 	{
-		TupleQueryResultParser parser = createTupleParser(format);
+		return parseTupleInternal(in, format, false);
+	}
+
+	/**
+	 * Parses a query result document and returns it as a TupleQueryResult object, with parsing done on a
+	 * separate thread in the background.<br>
+	 * IMPORTANT: As this method will spawn a new thread in the background, it is vitally important that it be
```
jeenbroekstra: Same as before: change "it" to "the TupleQueryResult" and move/copy this warning to the `@return` javadoc.

core/query/src/main/java/org/eclipse/rdf4j/query/impl/BackgroundGraphResult.java
```diff
@@ -0,0 +1,244 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.query.impl;
```
jeenbroekstra: Why is this class in `oer.query.impl`, but its companion in `oer.query.resultio.helpers`? Shouldn't they both be in the `helpers` package?

core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
```diff
@@ -73,10 +82,14 @@ public static Statement singleResult(GraphQueryResult result)
...
 	{
 		Statement singleResult = null;
-		if (result.hasNext()) {
-			singleResult = result.next();
+		try {
+			if (result.hasNext()) {
+				singleResult = result.next();
+			}
+		}
+		finally {
```
jeenbroekstra: Well spotted.

core/query/src/main/java/org/eclipse/rdf4j/query/impl/BackgroundGraphResult.java
ansell: They can't both be in the same module because the module heirarchy rio-api -> query -> queryresultio-api doesn't allow it. Ie, to do so I would need to relocate this from core/query to core/queryresultio/api where it would be the only class dealing with Statement's, as all of the others in queryresultio-api deal with BindingSet.

I can create a new helpers package in core/query/ if you prefer, but the impl class contains all other similar classes in core/query.

This relates to my comment about wanting to merge queryresultio-api into rio-api in a future major version, to remove differences like this. To do so we would need to cut the tie from queryresultio-api to query, possibly by pushing some classes into query from queryresultio-api.

jeenbroekstra: Ah, that makes sense, thanks.

jeenbroekstra:
@ansell from what I've seen this is backward compatible and can therefore be included in the 2.2 release. I haven't yet ran the new japicmp plugin on this to verify though (currently doing exactly that for a different PR, in fact). Are you confident that this is ready to merge?

ansell:
Yes, I added regression tests against all of the RDF and Tuple parsers that verify the background parsing works well with each of them. We have already been using BackgroundQueryResult and BackgroundTupleResult for the server for years, and I worked on them last year to improve their reliability in a previous pull request. The only test I saw fail during testing was the single instance of a test timeout I mentioned above, which on its own isn't indicative of a bug.

I haven't run japicmp directly on this PR, but I will do that now to double check I haven't missed something.

ansell:
japicmp isn't showing any binary or source incompatibility warnings for me, particularly on http-client which previously contained the three classes that were moved and replaced with deprecated shells.

jeenbroekstra:
Sweet. I'll merge once Travis checks in.

PR #746, opened 2017-02-01
Fix #643: class names shortened and number of tests trimmed a bit

### LINKED ISSUE #643: New RDFS reasoner with better performance on stable schemas
Create a new reasoner that:
 - Uses a fixed point rule approach on the Tbox to create an inference map
 - Uses the map to infer new triples during data insert

jeenbroekstra:
This PR addresses GitHub issue: #643 .

* `ForwardChainingSchemaCachingRDFSInferencer` renamed to `SchemaCachingRDFSInferencer`: both shorter and more accurate 
* removed inactive testsuite for combination of spin and new inferencer


jeenbroekstra:
@hmottestad I know you had a preference to have the "ForwardChaining..." bit in there, but there are other/better ways to inform our users about this new reasoner, and it was also not quite accurate (given that 'forward chaining' is something that applies to rule-based inferencers, and this new reasoner is not, strictly speaking, rule-based).

compliance/store/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeStoreInferencingTestForwardChainingSchemaCachingRDFSInferencer.java
```diff
@@ -1,35 +0,0 @@
...
-import java.io.IOException;
-
-import org.eclipse.rdf4j.sail.InferencingTest;
-import org.eclipse.rdf4j.sail.NotifyingSail;
-import org.eclipse.rdf4j.sail.Sail;
-import org.eclipse.rdf4j.sail.inferencer.fc.ForwardChainingSchemaCachingRDFSInferencer;
-import org.junit.Rule;
-import org.junit.rules.TemporaryFolder;
-
-public class NativeStoreInferencingTestForwardChainingSchemaCachingRDFSInferencer extends InferencingTest {
```
ansell: This would never have been run as its classname doesn't end in Test or TestCase

hmottestad:
Forward chaining is accurate. It first forward chaines the tbox into lookup tables, and then uses that to forward chain inferences on each inserted triple. 

The difference between this forward chaining and the one done by a rule engine is that I cache parts of the "chain" so that that part doesn't need to forward chained again each time data is added. 

Also, with forward chaining all the inferred triples are generated at insertion time. Which is the case for this reasoner.

hmottestad:
Anyhow, it is a very long and unwieldy name.

hmottestad:
Can you change it to "RdfsInferencer"

hmottestad:
or RDFSReasoner

jeenbroekstra:
The problem with that is that it doesn't distinguish it sufficiently from the existing (forward chaining) inferencer. 

I'm also out of time - release is today and I need to get this merged. I suspect that once we have things stable and tested we'll do a refactor of this stuff anyway, hopefully we can fit that in for the next major release.

Issue #751, opened 2017-02-05
Wrong rdf4j version in benchmark pom


hmottestad:
I've been running some of the benchmarks again and found out that some of them are running much faster than anticipated. Did a double check with assert enabled and found out that there is an issue with the reasoner not detecting that the schema has been updated.

I'll keep digging.

To reproduce run `java -jar -ea target/rdf4j-benchmark.jar -wi 1 -i 1 -f 1 -gc` in the benchmarks folder.

hmottestad:
Renamed the issue

hmottestad:
The benchmark pom file has the wrong rdf4j version. It wasn't updated after the release.

Also, the rename of ForwardChainingSchemaCachingRdfsInferencer to SchemaCachingRdfsInferencer hasn't been applied to the benchmark classes.

PR #752, opened 2017-02-05
#751 updated benchmark pom and applied renaming

hmottestad:
This PR addresses GitHub issue: #751 .

* Fixed version number in the benchmark pom
* Applied the renaming of ForwardChainingSchemaCachingRDFSInferencer to the benchmarking classes

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.eclipse.rdf4j</groupId>
         <artifactId>rdf4j</artifactId>
-        <version>2.2-SNAPSHOT</version>
+        <version>2.3-SNAPSHOT</version>
     </parent>
 
     <artifactId>rdf4j-benchmark</artifactId>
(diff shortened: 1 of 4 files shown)


benchmark/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.eclipse.rdf4j</groupId>
         <artifactId>rdf4j</artifactId>
-        <version>2.2-SNAPSHOT</version>
+        <version>2.3-SNAPSHOT</version>
```
jeenbroekstra: This should have been automatically done by maven. Is the benchmark module not registered as a module in the root pom?

ansell: It looks that way. Not a big issue that it wasn't released in 2.2 though.

hmottestad:
Fixed the root pom

jeenbroekstra:
I just realized that it may have been deliberate to keep the benchmark module out of the root pom, to avoid the benchmark tests automatically executing. I haven't verified that this actually happens (it may not as the benchmarks are run using a shell script I think, rather than a maven task) but we should doublecheck before merging.

Issue #641, opened 2016-11-02
Allow interactive stacking of SAILs in Console/Workbench


jeenbroekstra:
Currently, both Console and Workbench require a template that specifies a full SAIL stack - a user picks one template that describes which store + which reasoners to use, and then fills in some final variables. 

However, this is cumbersome: we support several different stacked SAILs (Lucene indexing, RDFS reasoning, SPIN, custom inference, etc), all of which can be employed in different configurations on top of various base persistence SAILs. To offer every possibly combination as a fixed template is inflexible and not a good user experience (we end up with a very long dropdown in Workbench, for example). It also is hard to extend for third-party SAIL developers. 

To fix, we should look into supporting a mechanism in Console/Workbench that allows users to construct the SAIL Stack interactively (a set of dialogs/screens that allows the user to pick each element of the stack, one a time).

jgrzebyta:
I thing also that it is worth to simplify API (or add new factories for more complex layers). Ref #739.

PR #758, opened 2017-02-13
Fix #739: spin lucene

### LINKED ISSUE #739: Complex API of initiation sail+spin+lucene
Related to #220  and #584 

Example of SailRepository with Lucene and SpinSail initiation showed in class `org.eclipse.rdf4j.sail.lucene.AbstractLuceneSailSpinTest.setUp` is complex.

I guess it should be simplified and after that it is worth to add layout file for new store in `rdf4j-workbench`.

Also Lucene index is not built auto-magically. It might be built after commit (?).

### LINKED ISSUE #848: Make LuceneIndex internally more flexible
I suggest to add getter/setter for `org.apache.lucene.search.similarities.Similarity`.

jgrzebyta:
This PR addresses GitHub issue: #739, #771 and #848 .

* Create separate rdf4j-lucene-spin for LuceneSpinSail wrapper
* Support Lucene Index within SPIN sail
* Move tests reletad to #220 to new module
* Subtask of Ref #641
* Found workaround for issue #771 
* Fix #848  - Add getter/setter for `Similarity` class instance


jgrzebyta:
New implementation of Connection should follow `LuceneSailConnection` with `LuceneSailBuffer` internally.

jeenbroekstra:
Everyone: given that my involvement with the project is currently a lot less than it was, I have less time to do proper PR reviews, so I'm relying on your explicit input. Please leave comments or reviews on any pull requests, indicating if you are happy for it to be merged or not.

I will not merge any pull request unless I see a comment/review from at least one other reviewer that he/she is ok with a merge.

jgrzebyta:
IMHO that is ready to merge. Please, can anyone review it?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+					else if (op instanceof LuceneSailBuffer.ClearContextOperation) {
+						// clear context
+						clearContexts(((ClearContextOperation)op).getContexts());
+					}
+					else if (op instanceof LuceneSailBuffer.ClearOperation) {
+						logger.debug("clearing index...");
+						luceneIndex.clear();
+					}
+					else
+						throw new RuntimeException(
```
hmottestad: Are you throwing an exception here and then catching it in the catch all later?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+								"Cannot interpret operation " + op + " of type " + op.getClass().getName());
+					i.remove();
+				}
+			}
+			catch (Exception e) {
```
hmottestad: This is a big catch all.

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+				logger.error("Committing operations in lucenesail, encountered exception " + e
+						+ ". Only some operations were stored, " + buffer.operations().size()
+						+ " operations are discarded. Lucene Index is now corrupt.", e);
```
hmottestad: Since the index is corrupt, should it be accessible through the Sail anymore? Would it make sense for other operations on the Sail to throw exceptions now rather than return corrupt data?

Also, is there no way to roll back the operations on the lucene index?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+			luceneIndex.begin();
+		}
+		catch (IOException e) {
+			throw new SailException(e);
+		}
+	}
+
+	@Override
+	public void commit()
+		throws SailException
```
hmottestad: Also, I noticed that there are no locks within this method, even though you are making potentially multiple commits to the underlying lucene index without having an open transaction with the index. Would this mean that the Sail is at most READ_UNCOMMITTED? Locking could bring that up to READ_COMMITTED, but I'm not sure if you could get it to SNAPSHOT_ISOLATION or above.

hmottestad:
Hi @jgrzebyta. I've left some comments as part of an initial review. I'm not very familiar with this code, so I'm not sure if all the comments are that relevant.

@ansell You previously pointed me in the right direction for testing my rdfs sail, do you think the test coverage here is good enough or does this sail need to be integrated into more tests?

core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
```diff
@@ -387,6 +388,14 @@ public void initialize()
...
 	}
 
+        /**
+         * The method is relocated to {@link SearchIndexUtils#createSearchIndex(java.util.Properties) }.
+         * @param parameters
+         * @return
+         * @throws Exception
+         * @deprecated
+         */
+        @Deprecated
```
ansell: Is the body of this method able to be replaced with a call to SearchIndexUtils.createSearchIndex to reduce double coding?

core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/util/SearchIndexUtils.java
```diff
@@ -0,0 +1,35 @@
...
+import org.eclipse.rdf4j.sail.lucene.SearchIndex;
+
+/**
+ *
+ * @author githib.com/jgrzebyta
+ */
+public class SearchIndexUtils {
+
+    /**
+     * The method is relocated to
```
ansell: Fix Javadoc

core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/util/SearchIndexUtils.java
```diff
@@ -0,0 +1,35 @@
...
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.lucene.util;
+
+import java.util.Properties;
+import static org.eclipse.rdf4j.sail.lucene.LuceneSail.DEFAULT_INDEX_CLASS;
+import static org.eclipse.rdf4j.sail.lucene.LuceneSail.INDEX_CLASS_KEY;
+import org.eclipse.rdf4j.sail.lucene.SearchIndex;
+
+/**
+ *
```
ansell: Add a basic description

core/sail/fts/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/util/SearchIndexUtils.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author githib.com/jgrzebyta
```
ansell: We tend to use full names only, rather than URLs and/or email addresses for @author tags

core/sail/fts/lucene-spin/pom.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>rdf4j-sail-fts</artifactId>
+        <groupId>org.eclipse.rdf4j</groupId>
+        <version>2.3-SNAPSHOT</version>
+    </parent>
+
+
```
ansell: Remove extra empty lines in this file to reduce spaces to one line

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSail.java
```diff
@@ -0,0 +1,101 @@
...
+	private final Logger log = LoggerFactory.getLogger(LuceneSpinSail.class);
+
+	private SearchIndex si;
+
+	private Properties parameters = new Properties();
+
+	public LuceneSpinSail() {
+	}
+
+	public LuceneSpinSail(NotifyingSail baseSail) {
```
ansell: The initialize method presumes (no instanceof check) that baseSail is a direct instance of SpinSail. Is it better to enforce that in this constructor?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSail.java
```diff
@@ -0,0 +1,101 @@
...
+			throw new SailException(ex);
+		}
+
+		super.initialize();
+	}
+
+	@Override
+	public void setDataDir(File dataDir) {
+		parameters.setProperty(LuceneSail.LUCENE_DIR_KEY, dataDir.getAbsolutePath() + "/index");
+		getBaseSail().setDataDir(dataDir);
```
ansell: I am not familiar with the semantics of this method, but what is the rationale for not calling super.setDataDir?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSail.java
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * @return @throws SailException
+	 */
+	@Override
+	public NotifyingSailConnection getConnection()
+		throws SailException
+	{
+
+		NotifyingSailConnection connection = super.getConnection();
+		if (si == null) {
```
ansell: The connection created above is not closed if this check fails and exits before returning the connection object to the user.

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSail.java
```diff
@@ -0,0 +1,101 @@
...
+			throw new SailException("Index is not created");
+		}
+		return new LuceneSpinSailConnection(connection, si);
```
ansell: This call needs to be wrapped in a try-finally to ensure that connection is closed if the LuceneSpinSailConnection is not able to be created successfully.

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+public class LuceneSpinSailConnection extends NotifyingSailConnectionWrapper {
+
+	private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+	private final SearchIndex luceneIndex;
+
+	/**
+	 * the buffer that collects operations
+	 */
+	final private LuceneSailBuffer buffer = new LuceneSailBuffer();
```
ansell: The convention for RDF4J is to label fields as "private final" rather than "final private"

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+	/**
+	 * The listener that listens to the underlying connection. It is disabled during clearContext operations.
+	 */
+	protected final SailConnectionListener connectionListener = new SailConnectionListener() {
+
+		@Override
+		public void statementAdded(Statement statement) {
+			// we only consider statements that contain literals
+			if (statement.getObject() instanceof Literal) {
+				if (statement == null)
```
ansell: +1, this method needs to be reformatted and calls moved around to fit with RDF4J conventions

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+		throws SailException
+	{
+		super.addStatement(arg0, arg1, arg2, arg3);
+	}
+
+	@Override
+	public void close()
+		throws SailException
+	{
+		if (closed.compareAndSet(false, true)) {
```
ansell: @hmottestad All connections must support multiple threads calling close simultaneously and only let one thread through to the actual close operations. This pattern is directly sourced from other code in RDF4J and is known to be thread-safe per the contract for AtomicBoolean.compareAndSet

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+				super.close();
+			}
+			finally {
+				try {
+					luceneIndex.endReading();
+				}
+				catch (IOException e) {
+					logger.warn("could not close IndexReader or IndexSearcher " + e, e);
+				}
+				// remember if you were closed before, some sloppy programmers
```
ansell: This comment is irrelevant per the AtomicBoolean check above, and it isn't sloppy programming, there are background resource leak fix mechanisms that legitimately call close without knowing if it has been called before.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -48,89 +53,96 @@
...
-			try {
-				if (repository != null) {
-					repository.shutDown();
-				}
-			}
-			finally {
-				if (searchIndex != null) {
-					searchIndex.shutDown();
-				}
+		if (connection != null) {
```
ansell: The try finally blocks above should not have been replaced with this unsafe version. The reasoning behind them is to isolate cross-test failures to more efficiently identify failing tests.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -48,89 +53,96 @@
...
  */
-public abstract class AbstractLuceneSailSpinTest {
+public class LuceneSailSpinTest {
 
 	private static final String DATA = "org/eclipse/rdf4j/sail/220-example.ttl";
 
-	private static Logger log = LoggerFactory.getLogger(AbstractLuceneSailSpinTest.class);
+	private static Logger log = LoggerFactory.getLogger(LuceneSailSpinTest.class);
+
+	private File tempDir;
```
ansell: Use JUnit TemporaryFolder instead of recoding a tempDir pattern yourself.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -291,7 +300,39 @@ public void testDistanceFunction()
...
+	 * <code>
+	* <urn:test.org/data/rec2> a t:Data ;
+	*  rdfs:label "Sed ornare, leo" ;
+	*  t:number 2 .
+	* </code>
+	 * 
+	 * @throws Exception
+	 */
+	@Test
+	public void doubleLoading()
```
ansell: There are no assertions in this test to ensure that it has the expected side effects.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailTupleFunctionTest.java
```diff
@@ -51,16 +49,19 @@
...
-public abstract class AbstractLuceneSailTupleFunctionTest {
+public class LuceneSailTupleFunctionTest {
 
 	private Repository repository;
 
 	private static final String DATA = "org/eclipse/rdf4j/sail/220-example.ttl";
 
 	private RepositoryConnection connection;
+        
+        private File tempDir;
```
ansell: Ditto for using JUnit TemporaryFolder instead of recoding it here.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+	private void evaluate(String[] queries, ArrayList<List<Map<String, String>>> expectedResults)
+		throws MalformedQueryException, RepositoryException, QueryEvaluationException
+	{
+		for (int queryID = 0; queryID < queries.length; queryID++) {
+			String serql = queries[queryID];
+			List<Map<String, String>> expectedResultSet = expectedResults.get(queryID);
+
+			// fire the query
+			TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, serql);
+			TupleQueryResult tqr = query.evaluate();
```
ansell: Use try-with-resources on this TupleQueryResult object.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		String[] queries = new String[] {
+				"SELECT \n" + "  Resource, Score, Snippet \n" + "FROM \n" + "  {Resource} <" + MATCHES
+						+ "> {} \n" + "    <" + QUERY + "> {\"one\"}; \n" + "    <" + SCORE + "> {Score}; \n"
+						+ "    <" + SNIPPET + "> {Snippet}",
+				"SELECT \n" + "  Resource, Score, Snippet \n" + "FROM \n" + "  {Resource} <" + MATCHES
+						+ "> {} \n" + "    <" + QUERY + "> {\"five\"}; \n" + "    <" + SCORE + "> {Score}; \n"
+						+ "    <" + SNIPPET + "> {Snippet}" };
+
+		ArrayList<List<Map<String, String>>> results = new ArrayList<List<Map<String, String>>>();
+		ArrayList<Map<String, String>> resultSet = null;
```
ansell: Do not reuse the same variable names for different assertions, it makes it difficult to verify whether there are side effects between assertions.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		buffer.append("FROM \n");
+		buffer.append("  {Resource} <" + MATCHES + "> {} ");
+		buffer.append("    <" + QUERY + "> {\"one\"}; ");
+		buffer.append("    <" + PROPERTY + "> {<" + PREDICATE_1 + ">}; ");
+		buffer.append("    <" + SCORE + "> {Score} ");
+		String q = buffer.toString();
+
+		// fire the query
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, q);
+		TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources on this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+	/**
+	 * Test if the snippets do not accidentially come from the "text" field while we actually expect them to
+	 * come from the predicate field.
+	 */
+	@Test
+	public void testSnippetLimitedToPredicate()
+		throws MalformedQueryException, RepositoryException, QueryEvaluationException
+	{
+		// more test-data
+		RepositoryConnection myconnection = repository.getConnection();
```
ansell: Use try-with-resources and rename this variable to indicate its purpose as distinct from "connection"

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		// "come" and "unicorn"
+		// and 'poor' should not be returned if we limit on predicate1
+		// and watch http://www.youtube.com/watch?v=Q5im0Ssyyus like 25mio others
+		myconnection.add(SUBJECT_1, PREDICATE_1, vf.createLiteral("come charly lets go to candy mountain"));
+		myconnection.add(SUBJECT_1, PREDICATE_1, vf.createLiteral("but the unicorn charly said to goaway"));
+		myconnection.add(SUBJECT_1, PREDICATE_2, vf.createLiteral("there was poor charly without a kidney"));
+		myconnection.commit();
+		myconnection.close();
+
+		{
```
ansell: RDF4J does not use brace blocks like this. If the code is difficult to understand otherwise you need to split it up into smaller sub-methods

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+			buffer.append("  {Resource} <" + MATCHES + "> {} ");
+			buffer.append("    <" + QUERY + "> {\"charly\"}; ");
+			buffer.append("    <" + PROPERTY + "> {<" + PREDICATE_1 + ">}; ");
+			buffer.append("    <" + SNIPPET + "> {Snippet}; ");
+			buffer.append("    <" + SCORE + "> {Score} ");
+			String q = buffer.toString();
+
+			// fire the query
+			TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, q);
+			TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources on this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+			buffer.append("FROM \n");
+			buffer.append("  {Resource} <" + MATCHES + "> {} ");
+			buffer.append("    <" + QUERY + "> {\"charly\"}; ");
+			buffer.append("    <" + SNIPPET + "> {Snippet}; ");
+			buffer.append("    <" + SCORE + "> {Score} ");
+			String q = buffer.toString();
+
+			// fire the query
+			TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, q);
+			TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources on this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+				String snippet = ((Literal)bindings.getValue("Snippet")).stringValue();
+				boolean foundexpected = false;
+				for (Iterator<String> i = expectedSnippetPart.iterator(); i.hasNext();) {
+					String expected = i.next();
+					if (snippet.contains(expected)) {
+						foundexpected = true;
+						i.remove();
+					}
+				}
+				if (!foundexpected)
```
ansell: RDF4J always puts braces after "if", even for single line bodies

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		query.append("    {r} lucenesail:matches {match} lucenesail:query {\"four\"}; \n");
+		query.append("                                   lucenesail:score {s}, \n");
+		query.append("    {r} <" + PREDICATE_3.toString() + "> {r2} \n");
+		query.append("USING NAMESPACE\n");
+		query.append("    lucenesail = <" + LuceneSailSchema.NAMESPACE + "> \n");
+
+		int r = 0;
+		int n = 0;
+		GraphQuery gq = connection.prepareGraphQuery(QueryLanguage.SERQL, query.toString());
+		GraphQueryResult result = gq.evaluate();
```
ansell: Use try-with-resources for this GraphQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+
+	@Test
+	public void testQueryWithSpecifiedSubject()
+		throws RepositoryException, MalformedQueryException, QueryEvaluationException
+	{
+		// fire a query with the subject pre-specified
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, QUERY_STRING);
+		query.setBinding("Subject", SUBJECT_1);
+		query.setBinding("Query", vf.createLiteral("one"));
+		TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources for this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		queryStr += "          search:property <urn:predicate1> . }";
+		queryStr += " UNION ";
+		queryStr += "{ ?result search:matches ?match2 . ";
+		queryStr += "  ?match2 search:query 'one' ; ";
+		queryStr += "          search:property <urn:predicate2> . } ";
+		queryStr += "} ";
+
+		// fire a query with the subject pre-specified
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SPARQL, queryStr);
+		TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources for this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+	@Test
+	public void testContextHandling()
+		throws Exception
+	{
+		connection.add(SUBJECT_4, PREDICATE_1, vf.createLiteral("sfourponecone"), CONTEXT_1);
+		connection.add(SUBJECT_4, PREDICATE_2, vf.createLiteral("sfourptwocone"), CONTEXT_1);
+		connection.add(SUBJECT_5, PREDICATE_1, vf.createLiteral("sfiveponecone"), CONTEXT_1);
+		connection.add(SUBJECT_5, PREDICATE_1, vf.createLiteral("sfiveponectwo"), CONTEXT_2);
+		connection.add(SUBJECT_5, PREDICATE_2, vf.createLiteral("sfiveptwoctwo"), CONTEXT_2);
+		connection.commit();
```
ansell: Need documentation here to describe why you are committing

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		assertQueryResult("sfourponecone", PREDICATE_1, SUBJECT_4);
+		assertQueryResult("sfourptwocone", PREDICATE_2, SUBJECT_4);
+		assertQueryResult("sfiveponecone", PREDICATE_1, SUBJECT_5);
+		assertQueryResult("sfiveponectwo", PREDICATE_1, SUBJECT_5);
+		assertQueryResult("sfiveptwoctwo", PREDICATE_2, SUBJECT_5);
+		// blind test to see if this method works:
+		assertNoQueryResult("johannesgrenzfurthner");
+		// remove a context
+		connection.clear(CONTEXT_1);
+		connection.commit();
```
ansell: There was no call to connection.begin between this call and the commit above.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		assertNoQueryResult("sfourponecone");
+		assertNoQueryResult("sfourptwocone");
+		assertNoQueryResult("sfiveponecone");
+		assertQueryResult("sfiveponectwo", PREDICATE_1, SUBJECT_5);
+		assertQueryResult("sfiveptwoctwo", PREDICATE_2, SUBJECT_5);
+	}
+
+	@Test
+	public void testConcurrentReadingAndWriting()
```
ansell: It isn't clear how this method aims to test concurrent reading and writing.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		throws Exception
+	{
+
+		connection.add(SUBJECT_1, PREDICATE_1, vf.createLiteral("sfourponecone"), CONTEXT_1);
+		connection.add(SUBJECT_2, PREDICATE_1, vf.createLiteral("sfourponecone"), CONTEXT_1);
+
+		connection.commit();
+		// prepare the query
+
+		{
```
ansell: Remove these brace blocks

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+			String queryString = "SELECT Resource " + "FROM {Resource} <" + MATCHES + "> {} " + " <" + QUERY
+					+ "> {\"sfourponecone\"} ";
+			TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, queryString);
+			TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources for this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+			assertTrue(result.hasNext());
+			bindings = result.next();
+			result.close();
+			connection.commit();
+		}
+		{
+			String queryString = "SELECT Resource " + "FROM {Resource} <" + MATCHES + "> {} " + " <" + QUERY
+					+ "> {\"sfourponecone\"} ";
+			TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, queryString);
+			TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources for this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		assertQueryResult("sfourponecone", PREDICATE_1, SUBJECT_4);
+		assertQueryResult("sfourptwocone", PREDICATE_2, SUBJECT_4);
+		assertQueryResult("sfiveponecone", PREDICATE_1, SUBJECT_5);
+		assertQueryResult("sfiveponectwo", PREDICATE_1, SUBJECT_5);
+		assertQueryResult("sfiveptwoctwo", PREDICATE_2, SUBJECT_5);
+		// blind test to see if this method works:
+		assertNoQueryResult("johannesgrenzfurthner");
+		// remove a context
+		connection.clear((Resource)null);
+		connection.commit();
```
ansell: No call to connection.begin between the commit above and this call.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		buffer.append("  Resource, Score \n");
+		buffer.append("FROM \n");
+		buffer.append("  {Resource} <" + MATCHES + "> {} ");
+		buffer.append("    <" + QUERY + "> {\"one~0.8\"}; ");
+		buffer.append("    <" + SCORE + "> {Score} ");
+		String q = buffer.toString();
+
+		// fire the query
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, q);
+		TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources for this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		assertEquals(3, results);
+
+		result.close();
+	}
+
+	@Test
+	public void testReindexing()
+		throws Exception
+	{
+		sail.reindex();
```
ansell: Not clear what the expected side effects of this are in relation to "testComplexQueryTwo"

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+		buffer.append("  Resource, Property \n");
+		buffer.append("FROM \n");
+		buffer.append("  {Resource} <" + MATCHES + "> {} ");
+		buffer.append("    <" + QUERY + "> {\"one\"}; ");
+		buffer.append("    <" + PROPERTY + "> {Property} ");
+		String q = buffer.toString();
+
+		// fire the query
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, q);
+		TupleQueryResult result = query.evaluate();
```
ansell: Use try-with-resources for this TupleQueryResult

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+	}
+
+	protected void assertQueryResult(String literal, IRI predicate, Resource resultUri)
+		throws Exception
+	{
+		// fire a query for all subjects with a given term
+		String queryString = "SELECT Resource " + "FROM {Resource} <" + MATCHES + "> {} " + " <" + QUERY
+				+ "> {\"" + literal + "\"} ";
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, queryString);
+		TupleQueryResult result = query.evaluate();
```
ansell: Attach this to the try block below using try-with-resources

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
```diff
@@ -0,0 +1,914 @@
...
+	}
+
+	protected void assertNoQueryResult(String literal)
+		throws Exception
+	{
+		// fire a query for all subjects with a given term
+		String queryString = "SELECT Resource " + "FROM {Resource} <" + MATCHES + "> {} " + " <" + QUERY
+				+ "> {\"" + literal + "\"} ";
+		TupleQuery query = connection.prepareTupleQuery(QueryLanguage.SERQL, queryString);
+		TupleQueryResult result = query.evaluate();
```
ansell: Attach this to the try block below using try-with-resources

core/sail/fts/pom.xml
```diff
@@ -18,5 +19,6 @@
  		<module>lucene</module> 
 		<module>elasticsearch</module>
 		<module>solr</module>
-	</modules>
-</project>
+    <module>lucene-spin</module>
```
ansell: The indentation for these three lines is broken.

jgrzebyta:
@ansell @hmottestad The connection class content comes from `LuceneSailConnection` so I will change both classes parallely. Is that ok?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSail.java
jgrzebyta: @ansell  super.SetDataDir calls baseSail.setDataDir. I will add super.SetDataDir(path) in the beginning of this method but still need to create the path for index location. By default it is not passed using setter but only as parameter.

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSail.java
jgrzebyta: @ansell  I have separated the check from the indexes manipulation: first check and after (if successful) than getting connection from the baseSail and creating new one.

throw new SailExceptio() breakes code execution so I dont need to wrap the code after the -if- check within -else-, does it?

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
jgrzebyta: I have left previous version but because searchIndex was removed into LuceneSailSpinConnection I simplified the block.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
jgrzebyta: It is not necessary. Commit doesnt close the transaction. It just flushes buffer content into repository and clean it afterwards.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -37,100 +40,114 @@
...
-public abstract class AbstractLuceneSailSpinTest {
+public class LuceneSailSpinTest {
 
 	private static final String DATA = "org/eclipse/rdf4j/sail/220-example.ttl";
 
-	private static Logger log = LoggerFactory.getLogger(AbstractLuceneSailSpinTest.class);
+	private static Logger log = LoggerFactory.getLogger(LuceneSailSpinTest.class);
+
+	@ClassRule
+	public static TemporaryFolder tempDir = new TemporaryFolder();
```
ansell: This is only referenced in a non-static context. If this were non-static, using @Rule instead of @ClassRule, JUnit could clean it up between tests instead of just at the end of the test(s).

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -37,100 +40,114 @@
...
+		}
 
-		// validate population
-		int count = countStatements(connection);
-		log.info("storage contains {} triples", count);
-		assert count > 0;
+		// testing connection
+		connection = repository.getConnection();
+		connection.begin();
+		assert connection.isActive() : "connection is not active";
```
ansell: The issue with using assert rather than the JUnit assertTrue, is that assert is only tested if the JVM is running in server mode. This potentially creates the possibility of this test passing on one JVM but failing in another. Best just to use the JUnit assertTrue/etc. to avoid those issues. Also need to fix other uses of assert in this test.

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
jgrzebyta: For both `buffer` and `SailIindex` operations are thread safe. I assume all that things are solved there. That method also was copied from another class and I believe it was tested. If not maybe it is worth to create unit test for that.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
jgrzebyta: I guess the intention is to run complex query after the reindexing. Added documentation.

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
jgrzebyta: As above...

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -37,100 +40,114 @@
...
+		}
 
-		// validate population
-		int count = countStatements(connection);
-		log.info("storage contains {} triples", count);
-		assert count > 0;
+		// testing connection
+		connection = repository.getConnection();
+		connection.begin();
+		Assert.assertTrue("connection is not active", connection.isActive());
```
hmottestad: is this simply testing the connection? Or can you call connection.commit() so that the transaction isn't left open? I know the .close() should clean up the transaction in the @After.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
```diff
@@ -37,100 +40,114 @@
...
 		// load resources
-		URL resourceURL = AbstractLuceneSailSpinTest.class.getClassLoader().getResource(DATA);
-		log.info("Resource URL: {}", resourceURL.toString());
+		assert repoConn.isActive();
+		URL resourceURL = LuceneSailSpinTest.class.getClassLoader().getResource(DATA);
+		log.debug("Resource URL: {}", resourceURL.toString());
 		Model model = Rio.parse(resourceURL.openStream(), resourceURL.toString(), RDFFormat.TURTLE);
-		searchIndex.begin();
 		for (Statement stmt : model) {
 			repoConn.add(stmt);
```
hmottestad: Why was the searchIndex.addStatement removed here?

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,255 @@
...
+		}
+		catch (IOException e) {
+			logger.error("Rolling back", e);
+			luceneIndex.rollback();
+			throw e;
+		}
+	}
+
+	@Override
+	public synchronized void removeStatements(Resource arg0, IRI arg1, Value arg2, Resource... arg3)
```
hmottestad: Rename variable names.

core/sail/fts/lucene-spin/src/test/java/org/eclipse/rdf4j/lucene/spin/LuceneSailSpinTest.java
jgrzebyta: Yes. That is simply testing if the connection and transaction was created properly after data loading to the repository. Nothing more. Just double check of that. The transaction should be left open (active) after that.

jgrzebyta:
Because lucene index (search index) is from now managed within the
LuceneSpinSail and relevant SpinSailConnection.

hmottestad:
I'm starting to get a good feeling about this pr now. Still a bit unsure of how well it handles failures. 

Next step for me is to run it up locally and try it out.

@jgrzebyta could you tell me which of the tests would give me a minimal code example to get me started with local testing?

jgrzebyta:
I'll add unit test for local native storage. If you do not mind memory
storage based examples are in the unit tests already.

jgrzebyta:
@hmottestad  Examples of use are added for both `MemoryStore` and `NativeStore` in commit 3485bcc4c3705a78f80594e9a6cc460b6a581a4f .

Examples are located in method `setUp` of `LuceneMemoryStoreTest` and `LuceneNativeStore` classes respectively.

core/sail/fts/lucene-spin/src/main/java/org/eclipse/rdf4j/lucene/spin/LuceneSpinSailConnection.java
```diff
@@ -0,0 +1,261 @@
...
+		if (closed.compareAndSet(false, true)) {
+			try {
+				super.close();
+			}
+			finally {
+				try {
+					luceneIndex.endReading();
+				}
+				catch (IOException e) {
+					logger.warn("could not close IndexReader or IndexSearcher " + e, e);
```
jgrzebyta: done

core/sail/fts/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/AbstractGenericLuceneTest.java
jgrzebyta: done

jamesrdf:
Why doesn't LuceneSpinSailConnection do anything with SearchQueryInterpreters when evaluating queries (LuceneSailConnection does)?

jgrzebyta:
My intention was to build a wrapper for SpinSail. It binds a Lucene index
with the context of SpinSailConnection. All querise are processed by
SpinSail.

The index is used in 2 moments. First time when the commit is done. And
second time when the query with the magic property is processed by
SpinSailConnection. Needed index is restored from the context.

I hope my explanation is clear.

PR #776, opened 2017-03-02
Fix #535: Made timeout configurable

### LINKED ISSUE #535: SailBase default connection timeout cannot be changed.
(Migrated from https://openrdf.atlassian.net/browse/SES-673)

SailBase default connection timeout cannot be changed. The constant DEFAULT_CONNECTION_TIMEOUT is copied into a member variable (connectionTimeOut) but this variable is not used and there is no setter method. the shutDown() method should use the connectionTimeOut variable and there should be a setter method for it.

barthanssens:

This PR addresses GitHub issue: #535  .

* added setter, so one can change the timeout instead of always using the default
* used the variable in wait() instead of default


jamesrdf:
In this going to be in the next patch release 2.2.1 or the next minor release 2.3?

hmottestad:
@jamesrdf technically this is feature and 2.2.1 is a bug fix release. What do you think @jeenbroekstra?

Issue #775, opened 2017-03-02
All SPARQL hash functions like MD5 ignores the trailing zeros


vassilmomtchev:
The SPARQL query below skips the trailing zeros for all hash functions:

`select *
where {
    BIND(md5("363") as ?md5)
}`

Here is what I get:

00411460f7c92d2124a67ea0f4cb5f85 (with 'md5 -s 363')
___411460f7c92d2124a67ea0f4cb5f85 (with the SPARQL query)

Check the suspected code:

https://github.com/eclipse/rdf4j/blob/2.2/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java#L30-L36

barthanssens:
Message digest length should be checked... I'm looking into it

barthanssens:
pull request added, might be a bit too low-level / bit-banging (can also use String.format, or Arrays.fill... many ways to do it)

PR #777, opened 2017-03-03
Convert bytes to hex string with leading zeros

barthanssens:

This PR addresses GitHub issue: #775  .

* fixed bug (missing leading zeros)
* convert bytes to hexstring without BigInteger
* added javadoc and override

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java
@@ -7,7 +7,6 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.function.hash;
 
-import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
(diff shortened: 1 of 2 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java
```diff
@@ -23,20 +22,31 @@
...
-			return "0" + result;
+
+		// convert to hexadecimal representation, with leading zeros
+		char[] hex = new char[hash.length * 2];
+		for (int i = 0, j = 0; i < hex.length; ) {
+			hex[i++] = HEX.charAt((hash[j] & 0xF0) >>> 4);
+			hex[i++] = HEX.charAt(hash[j++] & 0x0F);
 		}
-		return result;
+		return new String(hex);
```
jeenbroekstra: As you said, there's many possible impls. Slightly worried that this approach is tricky to maintain. What do you think of this as an alternative fix:

     return String.format("%0" + (hash.length << 1) + "x", bi); 

(where `bi` is the `BigInteger` object from before the change)

jeenbroekstra: Also: can we add a unit test for this bug?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java
```diff
@@ -23,20 +22,31 @@
  * @author jeen
  */
 public abstract class HashFunction implements Function {
-
+private final String HEX = "0123456789abcdef";
```
ansell: Indentation

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java
```diff
@@ -23,20 +22,31 @@
...
-		BigInteger bi = new BigInteger(1, hash);
-		String result = bi.toString(16);
-		if (result.length() % 2 != 0) {
-			return "0" + result;
+
+		// convert to hexadecimal representation, with leading zeros
+		char[] hex = new char[hash.length * 2];
+		for (int i = 0, j = 0; i < hex.length; ) {
+			hex[i++] = HEX.charAt((hash[j] & 0xF0) >>> 4);
+			hex[i++] = HEX.charAt(hash[j++] & 0x0F);
```
ansell: This seems hard to maintain compared to reusing one of the many existing implementations of hex encoding that exist.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/hash/HashFunction.java
barthanssens: Ok, make senses to use the String.format and to add a unit test

barthanssens:
Ok, fixed

Issue #773, opened 2017-02-24
RDFJson parser checks for null instead of not null in the parse method for reader


nikolavp:
Look at line https://github.com/eclipse/rdf4j/blob/master/core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONParser.java#L157

jeenbroekstra:
It looks as if three different people have now independently implemented a fix for this. Only barthanssens has actually put up a Pull Request though.

PR #783, opened 2017-03-19
#773 fixed by checking not null insted of null

heshanjse:
Signed-off-by: Heshan Jayasinghe <shanujse@gmail.com>

This PR addresses GitHub issue: # 773 .

* fixed 773 by checking not null instead of null in RDFJSONParser class void parse(Reader reader,
String baseURI) method
* added two test-cases to AbstractParserHandlingTest class to test that issue

### DIFF
--- a/core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONParser.java
+++ b/core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONParser.java
@@ -154,7 +154,7 @@ public class RDFJSONParser extends AbstractRDFParser implements RDFParser {
 	public void parse(final Reader reader, final String baseUri)
 		throws IOException, RDFParseException, RDFHandlerException
 	{
-		if (rdfHandler == null) {
+		if (rdfHandler != null) {
 			rdfHandler.startRDF();
 		}
 		JsonParser jp = null;
--- a/testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
+++ b/testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
@@ -7,18 +7,6 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.rio;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Locale;
-
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Model;
 import org.eclipse.rdf4j.model.ValueFactory;
@@ -28,11 +16,6 @@ import org.eclipse.rdf4j.model.util.Models;
 import org.eclipse.rdf4j.model.vocabulary.DC;
 import org.eclipse.rdf4j.model.vocabulary.RDFS;
 import org.eclipse.rdf4j.model.vocabulary.XMLSchema;
-import org.eclipse.rdf4j.rio.DatatypeHandler;
-import org.eclipse.rdf4j.rio.ParserConfig;
-import org.eclipse.rdf4j.rio.RDFParseException;
-import org.eclipse.rdf4j.rio.RDFParser;
-import org.eclipse.rdf4j.rio.RioSetting;
 import org.eclipse.rdf4j.rio.helpers.BasicParserSettings;
 import org.eclipse.rdf4j.rio.helpers.ParseErrorCollector;
 import org.eclipse.rdf4j.rio.helpers.StatementCollector;
@@ -41,6 +24,15 @@ import org.junit.Before;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Locale;
+
+import static org.junit.Assert.*;
 
 /**
  * Abstract tests to confirm consistent behaviour for the datatype and language handling settings.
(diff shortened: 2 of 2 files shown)


hmottestad:
@heshanjse, It seems like the test is failing for some of the other parsers. I think the first step is to change the input to try to force the other parsers to error out with RDFParseException. Then we can have a look at the left over ones to see if they should be failing or if their behaviour is correct. If the behaviour is correct, then we need to move the test so that it doesn't run for those parsers.

testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
```diff
@@ -948,4 +951,15 @@ private final Model getTestModel(String languageValue, String languageTag) {
 				vf.createLiteral(languageValue, languageTag)));
 		return result;
 	}
+
+	@Test(expected = RDFParseException.class)
```
ansell: There need to be other tests where a valid document stream is sent in and parsing succeeds even with the rdfHandler being set to a non-null value.

ansell:
Travis is saying that the tests that are in this PR are not passing.

testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
hmottestad: Purpose of the test is to make parsing in presence of a null rdfHandler consistent. So I propose to give it input other than "" to force it to never be valid RDF across all parsers. Maybe something like "{(.;".

jeenbroekstra:
This is the output from Travis:

```
Failed tests: 
  NQuadsHandlingTest.testNullRdfHandlerInputStream Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  NQuadsHandlingTest.testNullRdfHandlerReader Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  NTriplesHandlingTest.testNullRdfHandlerInputStream Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  NTriplesHandlingTest.testNullRdfHandlerReader Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  TriGHandlingTest.testNullRdfHandlerInputStream Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  TriGHandlingTest.testNullRdfHandlerReader Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  TurtleHandlingTest.testNullRdfHandlerInputStream Expected exception: org.eclipse.rdf4j.rio.RDFParseException
  TurtleHandlingTest.testNullRdfHandlerReader Expected exception: org.eclipse.rdf4j.rio.RDFParseException
Tests in error: 
  BinaryHandlingTest.testNullRdfHandlerReader Â»  Unexpected exception, expected<...
```

testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
```diff
@@ -948,4 +951,30 @@ private final Model getTestModel(String languageValue, String languageTag) {
 				vf.createLiteral(languageValue, languageTag)));
 		return result;
 	}
-}
+//
+//	@Test(expected = RDFParseException.class)
```
hmottestad: clean up

testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
```diff
@@ -948,4 +951,30 @@ private final Model getTestModel(String languageValue, String languageTag) {
...
+//		getParser().parse(new InputStreamReader(new ByteArrayInputStream("".getBytes())) , "");
+//
+//	}
+
+	@Test
+	public void testNullRdfHandlerReader(){
+		try {
+			 getParser().parse(new InputStreamReader(new ByteArrayInputStream("".getBytes())), "");
+		}catch (Exception e){
+			assertTrue("Did not find expected rdfHandler", (!(e instanceof NullPointerException)));
```
hmottestad: Change your message to be more useful.

testsuites/rio/src/main/java/org/eclipse/rdf4j/rio/AbstractParserHandlingTest.java
```diff
@@ -948,4 +951,30 @@ private final Model getTestModel(String languageValue, String languageTag) {
...
+//	public void testNullRdfHandlerReader() throws IOException {
+//
+//		getParser().parse(new InputStreamReader(new ByteArrayInputStream("".getBytes())) , "");
+//
+//	}
+
+	@Test
+	public void testNullRdfHandlerReader(){
+		try {
+			 getParser().parse(new InputStreamReader(new ByteArrayInputStream("".getBytes())), "");
```
hmottestad: force the rdfhandler to be null. Maybe you can store getParser() in a variable and call something like .setHandler(null).

core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONParser.java
```diff
@@ -154,7 +154,7 @@ protected void reportFatalError(String msg, JsonLocation location)
 	public void parse(final Reader reader, final String baseUri)
 		throws IOException, RDFParseException, RDFHandlerException
 	{
-		if (rdfHandler == null) {
+		if (rdfHandler != null) {
```
hmottestad: pull and merge master into this branch.

jamesrdf:
I'd like to get this into 2.2.2. @heshanjse will you be able resolve the conflicts this week by pulling down the latest from the master branch?

ansell:
The issue isn't just conflicting files, as #773 has already been fixed:

https://github.com/eclipse/rdf4j/blob/master/core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONParser.java#L77

The core of this pull request would be implementing some tests that directly test whether all parsers can cope with handlers being null, as permitted by the API contract. The current tests do not do that directly.

In a future major version I would like the API to change to guarantee that getRDFHandler never returns null, and instead returns a stub RDFHandler if the user has not set one, which would avoid these issues in general and cleanup the parser code a little.

heshanjse:
@ansell I created these test for current situation. with your future implmentation i think this test can be not correct. 

@jamesrdf I can resolve this with in this week but with the future implmentation i think this test can be useless.I haven't more experience in rdf4j so please tell me most suitable way.

ansell:
Hi @heshanjse, nothing will be changing until the next major version (3.x) at least, so there needs to be a test that verifies null handlers are supported in the meantime.

Issue #637, opened 2016-11-01
Can we make repository id to be only alphanumerics plus underscore


nikolavp:
We are a hitting a problem where on some operating systems(Windows) some characters are not allowed in the filesystem. Sadly there is one to one correlation to repository ids and a storage folder in the data directory. Also the error messages are not descriptive enough to be useful, files in that directories cannot be created.

@damyan-ognyanov @avataar

jeenbroekstra:
Actually, `RepositoryManager` has a public method `getNewRepositoryID` which is exactly intended for this kind of sanitizing of a repository id string. But the strange thing is that this method is not actually used in either the Console or the Workbench app.

PR #791, opened 2017-03-27
issue #637: Replace special characters in new repository ID with dashes

jamesrdf:

This PR addresses GitHub issue: #637 .


nikolavp:
I would rather not change the repository id behind user's back. Imagine that someone does automation on top of repository creation and then inserts some data in the repositoryID that he thought was just created. Instead I think it's better to signal with an error that he is using invalid characters and he should probably try the sanitized version instead. Something like

```java
String newRepositoryID = getNewRepositoryID(repositoryID);
if (!repositoryID.equals(newRepositoryID) {
    throw new IllegalArgumentException("You are using invalid characters in your repository id, try using " + newRepository + " instead!");
}
```

PR #796, opened 2017-03-31
Fix #637: Prompt user if using incompatible chars in repository ID

### LINKED ISSUE #637: Can we make repository id to be only alphanumerics plus underscore

jamesrdf:

This PR addresses GitHub issue: #637 .


core/http/workbench/src/main/webapp/scripts/create.js
```diff
@@ -19,8 +19,9 @@ var workbench;
  */
 function checkOverwrite() {
     var submit = false;
+    var id = workbench.create.id.val();
```
jeenbroekstra: Please note that our javascript files are compiled from the corresponding typescript (in webapp/scripts/ts). You should not edit the javascript directly, but instead make the change in the typescript file and then recompile the ts to javascript (using the compileTypescript.sh found in the workbench root dir)>

jeenbroekstra:
Travis build failed but it looks like it was a problem with the cloud instance, not an actual test failure. If you push a change it should restart the build automatically, and (hopefully) succeed.

PR #797, opened 2017-03-31
Fix #695: Don't treat path modifiers as subqueries

### LINKED ISSUE #695: optional property paths (*, ?) are very slow
(Created from [OWLIM-1104](https://jira.ontotext.com/browse/OWLIM-1104)).

Prop paths using optional constructs (`p?` and `p*`) are extremely slow.
These paths **alone** are necessarily slow because they must return all nodes.
But nobody uses them alone, people always use them in combination with a non-optional path.
So one has to use workarounds, eg rewrite `q/p?` to `q|q/p`, to get good perfomance.

Use cases:
1. In [Getty queries](http://vocab.getty.edu/doc/queries/) Vladimir Alexiev uses many rewritten patterns like this:
```sparql
gvp:placeType/gvp:broaderGenericExtended? ->
gvp:placeType|(gvp:placeType/gvp:broaderGenericExtended)
```
2. In [OWLIM-2627](https://jira.ontotext.com/browse/OWLIM-2627) Barry Norton suggests a similar but wordier workaround for BM data:
```sparql
?x q/p? ?y ->
{?x q ?y} union {?x q ?z . ?z p ?y}
```
3. SHACL uses rdf:List for many of its constructs (sometimes unnecessarily) and uses `rdf:rest*/rdf:first` to unroll the list, eg see [ClosedConstraintComponent](http://w3c.github.io/data-shapes/shacl/#ClosedConstraintComponent).
   Unless fixed, one can't use the SHACL SPARQL Definitions to implement SHACL in rdf4j

4. Another typical example is poor man's RDFS inference: `rdf:type/rdfs:subClassOf*`

5. [American Art Collaborative](http://review.americanartcollaborative.org/cookbook/How-do-i-record-ordered-lists-of-things) discusses using `rdf:List` for "people depicted, with order". In this case a natural query would be
`crm:P62_depicts/(rdf:rest*/rdf:first)?`
but should be rewritten to
`crm:P62_depicts | crm:P62_depicts/rdf:first | crm:P62_depicts/rdf:rest+/rdf:first`

It seems to me that this can be fixed by internally rewriting to an alternative (union) path: one without the optional property and another with it.

    p?/q -> q|p/q
    p*/q -> q|p+/q
    q/p? -> q|q/p
    q/p* -> q|q/p+

Implementaton suggestion:
- add a path characteristic "positive" (cannot be empty) vs "non-negative" (can be empty); the names are by analogy with integers.
- design rewrite rules to produce alternatives where the positive path is always present, but the non-negative path alternates

More examples:

    p?/q? -> ()|p|q|p/q

This alone will be slow since it's non-negative (`()` is the bad part), but when combined with a positive path, it can be made fast again, eg:

    p?/q?/r ->
      r | (()|p|q|p/q)/r ->
      r | r|p/r|q/r|p/q/r ->
      r|p/r|q/r|p/q/r

jamesrdf:
This PR uses LH join bindings to eval zero or one path modifiers and greatly improves the performance. Previously zero or one path modifiers were treated like a subqueries and used a hash join operation. Note that zero or more path modifiers were already improved with #689

This PR addresses GitHub issue: #695 .

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
@@ -202,7 +202,7 @@ public class QueryJoinOptimizer implements QueryOptimizer {
 			List<TupleExpr> subselects = new ArrayList<TupleExpr>();
 
 			for (TupleExpr expr : expressions) {
-				if (TupleExprs.containsProjection(expr)) {
+				if (TupleExprs.containsSubquery(expr)) {
 					subselects.add(expr);
 				}
 			}
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
@@ -911,7 +911,7 @@ public class StrictEvaluationStrategy
 			return new ServiceJoinIterator(leftIter, (Service)join.getRightArg(), bindings, this);
 		}
 
-		if (TupleExprs.containsProjection(join.getRightArg())) {
+		if (TupleExprs.containsSubquery(join.getRightArg())) {
 			return new HashJoinIteration(this, join, bindings);
 		}
 		else {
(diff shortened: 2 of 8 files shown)


jamesrdf:
```
With http://vocab.getty.edu/dataset/aat/full.zip
Using release 2.2
PREFIX gvp:<http://vocab.getty.edu/ontology#>
ASK { ?x gvp:broader|(gvp:broader/gvp:broaderGenericExtended) ?y } # ~25ms
ASK{ ?x gvp:broader/gvp:broaderGenericExtended* ?y } # ~25ms
ASK{ ?x gvp:broader/gvp:broaderGenericExtended? ?y } # ~2500ms

Using jamesrdf:issues/#695-join-projection 2.3-SNAPSHOT
PREFIX gvp:<http://vocab.getty.edu/ontology#>
ASK { ?x gvp:broader|(gvp:broader/gvp:broaderGenericExtended) ?y } # ~25ms
ASK{ ?x gvp:broader/gvp:broaderGenericExtended* ?y } # ~25ms
ASK{ ?x gvp:broader/gvp:broaderGenericExtended? ?y } # ~25ms
```

Issue #800, opened 2017-04-03
Support for SHACL vocabulary


barthanssens:
A vocabulary class for W3C Shapes Constraint Language (SHACL, http://www.w3.org/ns/shacl# )

Related to (or part of) #743 , but this ticket only covers generating the vocabulary class, not implementing a SHACL engine

PR #802, opened 2017-04-04
Initial commit SHACL vocab

barthanssens:

This PR addresses GitHub issue: #800  .

* Vocabulary class with constants for W3C Shapes Constraint Language (SHACL)

SHACL should be fairly stable now, the EC has just started a work package on using OWL and SHACL for DCAT-AP validation.

Class was generated using [VocGen](https://github.com/Fedict/lod-vocgen)

This commit does not provide any code for running / checking SHACL rules (i.e. does not contain a SHACL engine)

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SHACL.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SHACL.java
@@ -1,0 +1,822 @@
+/**
+ * Copyright (c) 2017 Eclipse RDF4J contributors, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.eclipse.rdf4j.model.vocabulary;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Namespace;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleNamespace;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+
+
+/**
+ * Constants for the Shapes Constraint Language.
+ *
+ * @see <a href="https://www.w3.org/TR/2017/WD-shacl-20170303/">Shapes Constraint Language</a>
+ *
+ * @author Bart Hanssens
+ */
+public class SHACL {
+	/**
+	 * The SHACL namespace: http://www.w3.org/ns/shacl#
+	 */
+	public static final String NAMESPACE = "http://www.w3.org/ns/shacl#";
+
+	/**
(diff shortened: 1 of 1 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SHACL.java
```diff
@@ -0,0 +1,822 @@
...
+	public static final IRI QUALIFIED_MIN_COUNT_CONSTRAINT_COMPONENT_QUALIFIED_VALUE_SHAPES_DISJOINT;
+
+	/** sh:ResultAnnotation */
+	public static final IRI RESULT_ANNOTATION;
+
+	/** sh:SPARQLAskExecutable */
+	public static final IRI SPARQLASK_EXECUTABLE;
+
+	/** sh:SPARQLAskValidator */
+	public static final IRI SPARQLASK_VALIDATOR;
```
hmottestad: Should this be SPARQL_ASK_VALIDATOR? Same for the other ones with SPARQL in them.

core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SHACL.java
```diff
@@ -0,0 +1,822 @@
...
+	public static final IRI HAS_VALUE_CONSTRAINT_COMPONENT;
+
+	/** sh:HasValueConstraintComponent-hasValue */
+	public static final IRI HAS_VALUE_CONSTRAINT_COMPONENT_HAS_VALUE;
+
+	/** sh:IRI */
+	public static final IRI IRI;
+
+	/** sh:IRIOrLiteral */
+	public static final IRI IRIOR_LITERAL;
```
hmottestad: IRI_OR_LITERAL instead? Same as with SPARQL. I guess you have used an algorithm that looks for lowercase followed by uppercase to add underscore.

hmottestad:
@barthanssens I've added two comments. Other than this things look good, and similar to how RDF4J has solved it for other vocabs. Personally I think it's bad practice to use all caps to represent IRIs, it's only by java convention because they are final. Being able to type rdf.type, or RDF.type instead of RDF.TYPE just looks much closer to what is actually being represented, rdf:type. Although this is more a rant than anything, since I doubt anyone would be willing to change this.

barthanssens:
Thanks for the review. 

Indeed I've used all caps (also in previous contributions) because it was already the case for vocabularies like RDFS, FOAF... in RDF4J. The downside is that one has to be careful with vocabularies containing lowercase properties and uppercase classes (dcat:distribution and dcat:Distribution)

For the record, and as you probably know, Apache Jena does not follow the uppercase Java convention..

core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/SHACL.java
barthanssens: Indeed, the quick&dirty tool needs some finetuning. I'll make the suggested changes

hmottestad:
@ansell @jeenbroekstra This looks good to me. Can I merge?

Issue #778, opened 2017-03-09
Blank node collision when loading JSON-LD


skodapetr:
When loading JSON-LD, the blank node's ID is taken from the jsonld-java library. The library uses a counter to generate blank node ID's. So the blank nodes returned by jsonld-java are same for different files - it. they are "unique" just in the scope of a file . The rdf4j however uses these names to identify the blank nodes even between files.

The problem with this situation can be illustrated on a simple example:
File ```in1.jsonld``` : ``` { "http://localhost/ontology/filePredicate": "in1.jsonld" } ```
File ```in2.jsonld``` : ``` { "http://localhost/ontology/filePredicate": "in2.jsonld" } ```
The Java code: 
```    
public static void main(String[] args) throws Exception {
        Model model = new LinkedHashModel();
        try (InputStream stream = new FileInputStream("in1.jsonld")) {
            model.addAll(Rio.parse(stream, "http://localhost/base", RDFFormat.JSONLD));
        }
        try (InputStream stream = new FileInputStream("in2.jsonld")) {
            model.addAll(Rio.parse(stream, "http://localhost/base", RDFFormat.JSONLD));
        }
        try (OutputStream stream = new FileOutputStream("rdf4j.ttl")) {
            Rio.write(model, stream, RDFFormat.TURTLE);
        }
    }
```
Content of the output file (```rdf4j.ttl```): 
``` 
_:b0 <http://localhost/ontology/filePredicate> "in1.jsonld" , "in2.jsonld" . 
```
while it should be 
```
_:b0 <http://localhost/ontology/filePredicate> "in1.jsonld" .
_:b1 <http://localhost/ontology/filePredicate>  "in2.jsonld" .
```
We would get the same result with Repository, but the code for Model is shorter. 

I also tested similar code with Jena, where I got the expected result.

This behaviour is also not compatible with other loaders. For example, if I load two Turtle files, with two blank nodes of the same identifiers, I should got different blank nodes.  This works as Turtle loader give unique IDs to blank nodes even in different files. However as the jsonld-java "reuse" the identifiers, this does not hold for the JSON-LD.

One solution is to use a wrap for the RdfHandler and modify the blank node's ID (for example add a file based prefix).

The used version of rdf4j is 2.2

jeenbroekstra:
[...]

Issue #745, opened 2017-01-31
Turtle parser returns equivalent bnodes for successive parse operations


ansell:
The Turtle parser does not appear to be updating the bnode assignment state between calls to parse to make sure that successive parse operations do not return the same bnodes. Do we formally support multiple calls to RDFParser.parse btw? I thought that parsers in general were one-use but now I am not sure.

jeenbroekstra:
I don't think we document one way or the other, really. How difficult would it be to support multiple uses? Apart from this particular issue there's little I can think of in any of the parsers that is stateful and/or needs to be reinitialized. 

Advantage of reusing a parser is that you get to retain any custom parser configuration.

ansell:
There is a protected method call on the abstract parser class that should be done at the start of every parse method if we want to support it, IIRC.

ansell:
https://github.com/eclipse/rdf4j/blob/master/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/AbstractRDFParser.java#L340

ansell:
it is actually documented that it needs to be called at the end of parse, but I would prefer if it were (also?) called at the start, given we aren't storing the bnodes in a map anymore and the rest of the storage doesn't explode in terms of memory usage.

ansell:
The PR for this is ready for review/merging, and would fix both this issue and #778 if merged, so could be useful to do before the next patch release

PR #817, opened 2017-04-24
issue #745 : Clear parser state before/after parsing

ansell:
This PR addresses GitHub issue: #745 .

* Consistently call AbstractRDFParser.clear before and after parsing to ensure state is consistent with the API goals of safely (not concurrently, just in serial) reusing RDFParser instances within the RDF-1.1 Abstract Model expectations
* Adds a test to ensure that the same blank node is parsed as two distinct objects by the same RDFParser across successive calls to RDFParser.parse


ansell:
This also fixes #778

PR #830, opened 2017-05-23
Fix #93 : Add non-transactional versions of Repositories methods

### LINKED ISSUE #93: Repositories.tupleQuery requires transactions to be created for SPARQLRepository
(Migrated from https://openrdf.atlassian.net/browse/SES-2344 )

The Repositories.tupleQuery function is used to execute tuple query - this type of query is read only and thus should not need write permission on the target server. 

However the Repositories.tupleQuery function use [Repository.get|https://bitbucket.org/skodape/sesame/src/0f17ed322456ba4ea095606cc5db4fab39ad78a3/core/repository/api/src/main/java/org/openrdf/repository/util/Repositories.java?at=master&fileviewer=file-view-default#Repositories.java-139] function. This function call begin and commit on used connection. 

In combination with implementation of [SPARQLConnection|https://bitbucket.org/skodape/sesame/src/0f17ed322456ba4ea095606cc5db4fab39ad78a3/core/repository/sparql/src/main/java/org/openrdf/repository/sparql/SPARQLConnection.java?at=master&fileviewer=file-view-default#SPARQLConnection.java-327] this cause execution of update query, causing exception:
{{Exception in thread "main" org.openrdf.repository.RepositoryException: error executing transaction}}
as the repository refuse to perform the update.

May a solution be introduction of new "read-only" method into the Repositories utility class and usage of this method from the SPARQL select and construct helper methods?

ansell:
This PR addresses GitHub issue: #93 .

* Adds non-transactional versions of methods in Repositories helper utility class

Some Repository implementations do not silently succeed if transactions are attempted to be committed on read-only endpoints when there were no changes queued up for the transaction.

Hence, it is necessary to add versions of all of the Repositories methods that do not attempt to begin, commit, or rollback transactions.

### DIFF
--- a/core/repository/api/src/main/java/org/eclipse/rdf4j/repository/util/Repositories.java
+++ b/core/repository/api/src/main/java/org/eclipse/rdf4j/repository/util/Repositories.java
@@ -34,9 +34,9 @@ import org.eclipse.rdf4j.rio.RDFHandler;
 public final class Repositories {
 
 	/**
-	 * Opens a {@link RepositoryConnection} to the given Repository, sends the connection to the given
-	 * {@link Consumer}, before either rolling back the transaction if it failed, or committing the
-	 * transaction if it was successful.
+	 * Opens a {@link RepositoryConnection} to the given Repository within a transaction, sends the connection
+	 * to the given {@link Consumer}, before either rolling back the transaction if it failed, or committing
+	 * the transaction if it was successful.
 	 * 
 	 * @param repository
 	 *        The {@link Repository} to open a connection to.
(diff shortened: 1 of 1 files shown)


core/repository/api/src/main/java/org/eclipse/rdf4j/repository/util/Repositories.java
```diff
@@ -142,9 +208,43 @@ public static void consumeSilent(Repository repository, Consumer<RepositoryConne
...
+	 *        The {@link Repository} to open a connection to.
+	 * @param processFunction
+	 *        A {@link Function} that performs an action on the connection and returns a result.
+	 * @return The result of applying the function.
+	 * @throws RepositoryException
+	 *         If there was an exception dealing with the Repository.
+	 * @throws UnknownTransactionStateException
+	 *         If the transaction state was not properly recognised. (Optional specific exception)
+	 */
+	public static <T> T getNoTransaction(Repository repository,
```
ansell: This is the core new method added, all of the other NoTransaction methods link back to here.

PR #837, opened 2017-05-31
Fix #69: IRI Parser for validation and resolution

### LINKED ISSUE #69: Provide utility methods for IRI validation
(Migrated from https://openrdf.atlassian.net/browse/SES-2191 )

We should provide utility methods to allow checking that a given string is a valid IRI according to RFC3987.

jamesrdf:
This PR addresses GitHub issue: #69 .

* Creats ParsedIRI for RFC 3987 IRI syntax validation
* Deprecates ParsedURI in favor of ParsedIRI

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/BaseDeclProcessor.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/BaseDeclProcessor.java
@@ -7,6 +7,9 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.parser.sparql;
 
+import java.net.URISyntaxException;
+
+import org.eclipse.rdf4j.common.net.ParsedIRI;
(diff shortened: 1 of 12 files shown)


jamesrdf:
I have since run some performance tests against ParsedIRI and then in lined some common character checks to speed it up a bit. ParsedURI is still faster though, since ParsedIRI also validates characters when parsing.

jeenbroekstra:
@grexe you currently do, yes. Part of the reason is simply that `ParsedIRI` is defined in the `rdf4j-util` library, while the `IRI` class is defined in the `rdf4j-model` library, which is dependent on `rdf4j-util`. We can't use the `IRI` class in `rdf4j-util` without introducing a cyclic dependency.

grexe:
Ok I see, thanks for clarification @jeenbroekstra, but wouldn't it make more sense to move `ParsedIri` next to `IRI` then?

jeenbroekstra:
Possibly. I don't think use outside of direct internal usage by the parsers was really envisioned, which is why it is where it is. But feel free to log a feature request for this, and we can take a look at it.

PR #845, opened 2017-06-07
Fix #712: Add ValidatingValueFactory wrapper

### LINKED ISSUE #712: Provide optional syntax validation for IRIs when using ValueFactory.createIRI
I had a bug where my problem was, that I created an IRI with a leading whitespace so:

ValueFactory.createIri(" http://something.com") 

instead of 

ValueFactory.createIri("http://someting.com")

In my opinion it would be a good idea if the createIri method would trim whitespaces before creating the IRI or maybe throw an error if an IRI is created with whitespaces.

jamesrdf:

This PR addresses GitHub issue: #712 .

* Validates literals with known datatypes, literal languages, blank node IDs, and IRIs before passing the values to the underlying ValueFactory

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
@@ -10,6 +10,8 @@ package org.eclipse.rdf4j.model.datatypes;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.URISyntaxException;
+import java.util.IllformedLocaleException;
+import java.util.Locale;
 import java.util.StringTokenizer;
 
 import javax.xml.datatype.DatatypeConfigurationException;
@@ -273,6 +275,9 @@ public class XMLDatatypeUtil {
 		}
 		else if (datatype.equals(XMLSchema.ANYURI)) {
 			result = isValidAnyURI(value);
+		}
+		else if (datatype.equals(XMLSchema.LANGUAGE)) {
+			result = isValidLanguage(value);
 		}
 
 		return result;
@@ -672,6 +677,15 @@ public class XMLDatatypeUtil {
 			return true;
 		}
 		catch (URISyntaxException e) {
+			return false;
+		}
+	}
+
+	public static boolean isValidLanguage(String value) {
+		try {
+			new Locale.Builder().setLanguageTag(value);
+			return true;
+		} catch (IllformedLocaleException e) {
 			return false;
 		}
 	}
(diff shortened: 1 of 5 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
```diff
@@ -676,6 +681,15 @@ public static boolean isValidAnyURI(String value) {
 		}
 	}
 
+	public static boolean isValidLanguage(String value) {
```
jeenbroekstra: I seem to remember we had a different util method for this somewhere else. Can't remember the details though @ansell do you know what I'm thinking of?

ansell: Currently in core/rio/languages:

https://github.com/eclipse/rdf4j/blob/master/core/rio/languages/src/main/java/org/eclipse/rdf4j/rio/languages/BCP47LanguageHandler.java#L38

Advantage of the existing version was that it distinguished BCP47 from the previous RFC3066, but everything should have switched over to RDF-1.1 by now so the abstraction there may be a bit redundant now.

XMLDatatypeUtil isn't the ideal location for long term reuse of this particular function, but would be good to have it in core/model somewhere.

ansell: I am bit off today, the main part of this function is already in core/model in Literals, but currently it throws an exception rather than returning false:

https://github.com/eclipse/rdf4j/blob/master/core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java#L587

Good location for this function could be beside that in Literals

core/model/src/main/java/org/eclipse/rdf4j/model/impl/ValidatingValueFactory.java
```diff
@@ -138,8 +139,8 @@ public Literal createLiteral(String label, IRI datatype) {
 	}
 
 	public Literal createLiteral(String label, String language) {
-		if (!XMLDatatypeUtil.isValidLanguage(language)) {
-			throw new IllegalArgumentException("Not a validate language tag");
+		if (!Literals.isValidLanguageTag(language)) {
+			throw new IllegalArgumentException("Not a validate language tag: " + language);
```
ansell: Spelling: validate -> valid

jamesrdf:
We could either create ValidatingRepository(Manager/Connection) wrappers that override getValueFactory() and have them ultimately returned by RepositoryProvider, or bake it right into SailRepository and SPARQL/HTTPRepository, which could cause the parsers to use it as well.

PR #856, opened 2017-06-22
Fix #25: Execute each sequence of update operations in a single transaction

### LINKED ISSUE #25: SPARQL endpoint implementation should treat update sequences as atomic
The current SPARQL endpoint implementation handles update sequences by sending them down to the underlying Repository. Since at the level of SPARQL protocol no transactions are supported, this effectively means that transaction handling is left to the Repository API. 

The Repository API handles SPARQL update sequences by treating each operation in the sequence as a separate update, which is conform the [SPARQL 1.1 Update specification (section 3)](https://www.w3.org/TR/sparql11-update/#updateLanguage):

> Implementations MUST ensure that the operations of a single request are
> executed in a fashion that guarantees the same effects as executing them
> sequentially in the order they appear in the request.

In effect the SPARQL endpoint implementation handles update sequence requests as several transactions. The SPARQL spec, however, also has the following soft requirement (see [section 2.2](https://www.w3.org/TR/sparql11-update/#updateServices)):

> SPARQL 1.1 Update requests are sequences of operations. Each request SHOULD
> be treated atomically by a SPARQL 1.1 Update service. The term 'atomically'
> means that a single request will result in either no effect or a complete
> effect, regardless of the number of operations that may be present in the
> request.

While the current implementation does not break the spec, it does deviate from this recommended pattern. To change this, we should add a flag to the RDF4J REST protocol that allows our service implementation to distinguish between requests coming from a SPARQL endpoint client, and requests coming from an RDF4J client. In the former case, the service can choose to explicitly start a transaction before executing the sequence, so that the sequence is treated as an atomic update.

jamesrdf:
Fix #25: Execute each sequence of update operations in a single transaction

This PR addresses GitHub issue: #25 .

* If an error is detected it ignores subsequent operations
* If an error is detected all operations are rolledback

### DIFF
--- a/core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailUpdate.java
+++ b/core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailUpdate.java
@@ -54,7 +54,7 @@ public class SailUpdate extends AbstractParserUpdate {
 
 		boolean localTransaction = false;
 		try {
-			if (isLocalTransaction()) {
+			if (!getConnection().isActive()) {
 				localTransaction = true;
 				beginLocalTransaction();
 			}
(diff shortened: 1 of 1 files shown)


hmottestad:
It’s easier to understand when you inline it.

Issue #440, opened 2016-09-07
Support queries on vocabularies in model module


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-1850)

The vocabularies in the model module contain hardcoded static URIs, which are not able to be automatically queryable.

In the simplest case it would be useful to be able to query for all of the URIs in the vocabulary. Possibly: 

```
Set<URI> allURIs = OWL.getAllURIs();
```

This should all still be possible while maintaining backwards compatibility with the current fields, as the changes would be done in new super-classes/interfaces for the most part, and singletons would be used to create the public static final URI constants.

barthanssens:
Could perhaps be implemented using reflection, assuming that the vocabulary's classes and properties are the only public IRIs in the java class, e.g.

```
Class vocab = OWL.class;
List<IRI> lst = new ArrayList<>();
for(Field f: vocab.getFields()) {
    if (f.getType().equals(IRI.class) && Modifier.isPublic(f.getModifiers())) {
        lst.add((IRI) f.get(f));
    }
}
```

Would it be useful to turn this into a org.eclipse.rdf4j.model.util.VocabUtil or is it a nice-to-have/wont-fix ?

jeenbroekstra:
I think it's certainly useful as a utility class. If you're thinking of picking this up, please name it `Vocabularies` - we try to stick to the 'XXXs' convention (as opposed to 'XXXUtil') for naming utility classes.

barthanssens:
OK, that should be rather trivial to add 

It might be overkill, but for a future RDF4J release (3.0 ?) would it be worthwhile to rework  the vocabularies to make a distinction between properties and classes IRI (either add them to a list / array in the vocab class, or subclass IRI into e.g. PropertyIRI and ClassIRI) ?

PR #868, opened 2017-07-21
Issues/#440 vocabutil

barthanssens:
This PR addresses GitHub issue: #440 .

* Added a VocabularyUtil class with a function to get the values of all IRI fields from a java class, e.g. to get the IRIs of all the RDF classes and properties from a o.e.r.m.vocabulary java class.
* Currently there is no way to make a distinction between RDF classes and properties: both are instances of o.e.r.m.IRI

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/Vocabularies.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/Vocabularies.java
@@ -25,7 +25,7 @@ public class Vocabularies {
 	 * @param vocabulary RDF vocabulary
 	 * @return set of IRIs
 	 */
-	public static Set<IRI> getAllIRIs(Class vocabulary) {
+	public static Set<IRI> getIRIs(Class vocabulary) {
 		Set<IRI> iris = new LinkedHashSet<>();
 
 		for (Field f : vocabulary.getFields()) {
@@ -34,6 +34,7 @@ public class Vocabularies {
 					iris.add((IRI) f.get(vocabulary));
 				} catch (IllegalAccessException ex) {
 					// should not happen
+					throw new RuntimeException("Cannot access vocabulary field", ex);
 				}
 			}
 		}
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/Vocabularies.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/Vocabularies.java
@@ -8,7 +8,7 @@
 package org.eclipse.rdf4j.model.util;
 
 import java.lang.reflect.Field;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/util/VocabularyUtil.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class VocabularyUtil {
+	/**
+	 * Get all the {@link IRI IRIs} of the classes and properties of a vocabulary.
+	 *
+	 * @param vocabulary RDF vocabulary
+	 * @return set of IRIs
+	 */
+	public static Set<IRI> getAllIRIs(Class vocabulary) {
+		Set<IRI> iris = new HashSet<>();
```
jamesrdf: How about using a LinkedHashSet to preserve the (usually alpha) order that most vocabs use? This could help make the result easier to visually parse.

barthanssens:
Ok, done

core/model/src/main/java/org/eclipse/rdf4j/model/util/Vocabularies.java
```diff
@@ -0,0 +1,42 @@
...
+	 */
+	public static Set<IRI> getAllIRIs(Class vocabulary) {
+		Set<IRI> iris = new LinkedHashSet<>();
+
+		for (Field f : vocabulary.getFields()) {
+			if (f.getType().equals(IRI.class)) {
+				try {
+					iris.add((IRI) f.get(vocabulary));
+				} catch (IllegalAccessException ex) {
+					// should not happen
```
jeenbroekstra: If it should not happen, perhaps wrap in RuntimeException rather than swallowing?

core/model/src/main/java/org/eclipse/rdf4j/model/util/Vocabularies.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author Bart Hanssens
+ */
+public class Vocabularies {
+	/**
+	 * Get all the {@link IRI IRIs} of the classes and properties of a vocabulary.
+	 *
+	 * @param vocabulary RDF vocabulary
+	 * @return set of IRIs
+	 */
+	public static Set<IRI> getAllIRIs(Class vocabulary) {
```
jeenbroekstra: This is really minor, but perhaps just call it `getIRIs`? A matter of taste, so no biggie if you prefer to stick with this.

jamesrdf:
@jeenbroekstra should this be scheduled for a patch release or a minor release?

jeenbroekstra:
If you go [by the book](https://www.eclipse.org/projects/handbook/#release) it should be part of a minor. Patch releases (or 'service releases' as Eclipse calls them) are for bug fixes only, and should include no significant new functionality - though you can quibble over whether this counts as significant I guess.

PR #875, opened 2017-07-26
Fix #874: Create composite rdf4j-client/store.jar for use in distribution

### LINKED ISSUE #874: Classpath has Too Many JARs
RDF4J should package a jar for client side usage and a jar for server side and leave the 50~ module jars on maven central.

Users periodically complain about classpath errors cause from missing rdf4j jars. There are so many jars that are needed it is difficult for users to know if they have them all. Artifacts like rdf4j-onejar, rdf4j-bom, rdf4j-runtime, and rdf4j-runtime-osgi all help, but cast too wide of a net.

jamesrdf:

This PR addresses GitHub issue: #874.

* rdf4j-client.jar has all parsers/writers, repository-api, and http clients
* rdf4j-storage.jar has all sail dependent modules, excluding console/server/workbench/config/runtime
* changed runtime(-osgi), console, server, workbench, sdk, and onejar to use rdf4j-client/storage.jar


kenwenzel:
Would it be possible to make rdf4j-client.jar and rdf4j-store.jar compatible OSGi bundles? 

Instead of using the shade plugin you could also use the maven-bundle-plugin to create the jar files as runtime-osgi does.

kenwenzel:
This looks good - thank you.

Would it also be possible to add something like the following instructions to the bundle plugin:
`<Import-Package>
                                                org.slf4j.*,
                                                javax.*;resolution:=optional,
                                                org.apache.*;resolution:=optional,
                                                org.springframework.*;resolution:=optional,
                                                com.spatial4j.*;resolution:=optional,
                                                *
                                            </Import-Package>`

Currently, the runtime-osgi bundle does not resolve if org.springframework or com.spatial4j are not available.

jamesrdf:
springframework is not part of rdf4j-store, but I change lucene and spatial4j to be optional. Let me know if there is any other improvements I can make!

Issue #881, opened 2017-08-12
missing jdk.tool dependency for rdf4j-sail-solr


seralf:
Hi 
after a new clone/import of the latest source in eclipse, I've found this error:

![missing_jdk_rdf4j_solr](https://user-images.githubusercontent.com/88188/29243459-c1ab4d6a-7f9f-11e7-9110-97f8f1a1b09a.png)

then, searching on the issues/pr I suppose it could be a sort of side-effect of issue #137.

jamesrdf:
What version are you using? Can you provide steps to reproduce this? What version of jdk are you using? The jdk7 backport seems to be the only place this artifact is mentioned in rdf4j. Are you using the jdk7 backport branch?

seralf:
I'm using the main branch from last version on github.
The Screenshot is on Eclipse after a maven import, using latest JDK 8 build.

The same issue is cited in other places for other libraries, I suppose is related to changes from jdk 1.7 and later.
For example: http://hadoopinrealworld.com/missing-artifact-jdk-tools-jar/

jamesrdf:
Hadoop uses the path ${java.home}/../lib/tools.jar in their pom file, which works on my machine. What OS are you using?

I used the following command to find the value of ${java.home}:
$ jrunscript -e 'println(java.lang.System.getProperty("java.home"))'

Does ${java.home}/../lib/tools.jar point to tools.jar on your system?

seralf:
The problem was using windows, while on linux it seems to work ok, very
strange.
The path you proposed seems to work well on my ubuntu

2017-08-18 17:38 GMT+02:00 James Leigh <notifications@github.com>:

jamesrdf:
If the path Hadoop provides is not working then this sounds like an upstream issue and you should speak with Hadoop community directly to sort it out. I don't want to introduce OS specific workarounds here for potential upstream bugs.

PR #882, opened 2017-08-12
added jdk.tools for resolving issue #881

seralf:
Signed-off-by: Alfredo Serafini <seralf@gmail.com>

This PR addresses GitHub issue: #881  .

I've re-added the dependency only for `rdf4j-sail-solr` only, in order to avoid breaking other component.

The dependency is added as follows:

```
<dependency>
	<groupId>jdk.tools</groupId>
	<artifactId>jdk.tools</artifactId>
	<version>1.8</version>
	<scope>system</scope>
	<systemPath>${JAVA_HOME}/lib/tools.jar</systemPath>
</dependency>
```
to be honest I'n not a big fan of those systemPath dependencies, but in this case it seems something very conventional, so I think it could be a good compromise to re-add it here only.

### DIFF
--- a/core/sail/fts/solr/pom.xml
+++ b/core/sail/fts/solr/pom.xml
@@ -1,4 +1,5 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
 	<modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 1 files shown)


core/sail/fts/solr/pom.xml
```diff
@@ -42,6 +43,15 @@
 			</exclusions>
 		</dependency>
 
+		<!-- jdk.tools -->
+		<dependency>
+			<groupId>jdk.tools</groupId>
+			<artifactId>jdk.tools</artifactId>
+			<version>1.8</version>
+			<scope>system</scope>
+			<systemPath>${JAVA_HOME}/lib/tools.jar</systemPath>
```
jamesrdf: Will this path work on Mac computers?

seralf:
I've tested it on ubuntu.
We could avoid introducing this by manually setting things on the user
machine, but it could be a bit annoying

2017-08-18 17:35 GMT+02:00 James Leigh <notifications@github.com>:

core/sail/fts/solr/pom.xml
jeenbroekstra: It will work if `$JAVA_HOME` is correctly configured (pointing to something like `/Library/Java/JavaVirtualMachines/jdk<version>.jdk/Contents/Home`). But a standard install of the JDK doesn't set that var, I don't think.

jamesrdf:
This PR should be applied to the Hadoop codebase first and if needed we can backport something once they have a tested solution.

Issue #883, opened 2017-08-13
RepositorySail: wrapping third party in a Sail object?


seralf:
Hi
While experimenting with `SailRepository` and `LuceneSail` for a project I'm working on, I've notice that while a `SailRepository` instance expose a `getSail()` method for obtain the underlying implementation, there seems to be no way to wrap a `Repository` back into a working `Sail` instance.

I know it could probably introduce a lot of issues in some cases, it could be anyway very useful on some specific scenario: for example I'm trying to adopt virtuoso into a `LuceneSail`, and then it seems that if I could write something like:

```java
Sail sailVirtuoso = new RepositorySail(new VirtuosoRepository(...))
```

I could then try to log all the activities on the new `sailVirtuoso` object (a sort of naive transaction log) and also try to use it as a base sail for `LuceneSail`.

If I'm not wrong something similar was implemented in Ripple, but since as far as i know it was not updated to RDF4J, I'm trying to re-start a partial implementation for my tests.

What do you think about it?

jamesrdf:
Post questions to the mailing list. This is for requesting changes to the RDF4J project.

seralf:
It is a request indeed, with attached a basic idea for an implementation.
My apologies if it looks not in the right form

jamesrdf:
Thanks for the code. However, I think this needs more discussion on the mailing list. You can post to the rdf4j-dev mailing list to get more detailed feedback.

PR #884, opened 2017-08-13
RepositorySail basic (implemented) idea

seralf:
This PR addresses GitHub issue: #883 .

This class should offer a way for handling external third-party Repository implementation, wrapping them  for convenience into a working Sail instance.

Possible usage scenarios:
* using third-party repository implementations inside a sail stack, permitting to use them with:
  + inferencing sail
  + validators sail
  + lucene/fts sail
* exposing events at the NotifyingSail level: this way it should be possible to write custom transaction logs handlers.

```java
Repository repository = new SailRepository(new new MemoryStore());
repository.initialize();
...
Sail sail = new RepositorySail(repository);
sail.initialize();
...
```
Note this is only a first (very WIP) attempo for an implementation: it seems to work on basic tests, but any idea for improving it and introuduce better testing is very welcome.


jamesrdf:
Repositories should not wrap a Sail. The query performance would be horrible for many scenarios. A better solution would be to create a LuceneRepository that wraps a Repository and does query manipulation to be more efficient.

PR #892, opened 2017-08-27
Feature/#743 add support for shacl

heshanjse:
This PR addresses GitHub issue: # 743-add-support-for-SHACL

* AST for  SHACL min count validation
* SHACL Min Count validation


hmottestad:
This is the kind of plan I hope to support in the future. "New" is an internal MemoryStore that stores all changes made in a transaction while "exisiting" is the wrapped Sail with _all_ the data.

img_1960

core/shacl/src/main/java/org/eclipse/rdf4j/Main.java
```diff
@@ -0,0 +1,92 @@
...
+import org.eclipse.rdf4j.sail.memory.MemoryStore;
+import org.eclipse.rdf4j.validation.ShaclSail;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class Main {
+
+	public static void main(String[] args)
```
jeenbroekstra: This class should probably be in src/test (or in one of the test modules).

core/shacl/src/main/java/org/eclipse/rdf4j/plan/GroupBy.java
```diff
@@ -0,0 +1,98 @@
...
+			{
+				return hashmapiterator.hasNext();
+			}
+
+			@Override
+			public List<Tuple> next()
+					throws SailException
+			{
+				return hashmapiterator.next().getValue().stream().map(Tuple::new).collect(
+						Collectors.toList());
```
jeenbroekstra: This materializes the entire grouping in main memory, which will likely not scale well (though I'm not claiming the existing SPARQL impl is any better in that respect). I'm also slightly at a loss as to why you need a separate implementation of this (and several other operators): why doesn't any of this reuse the existing algebra classes?

hmottestad: There is no reuse at the moment no. Biggest factor is that it's still unclear what is needed by the SHACL engine. The future of the SHACL engine will require a query engine that can work against three datasources: added triples, removed triples, all triples. I have used Stardog's validation engine quite a lot recently, and it converts OWL to SPARQL and validates the entire database. The number one thing that is impossible to do efficiently is maxCount and minCount with large numbers (>10). These types of validations need to take into account the changed data in the transaction so that a minimal amount of data needs to be fetched from the underlying store for validation.

As for the materialization, this isn't the only place that this is being done. Heshan didn't have time to implement the volcano model properly, so I let him cut some corners for both group by and join so that he would be able to complete support for minCount within the 3-4 months that has been Google Summer of Code.

hmottestad:
@jeenbroekstra Thank you for taking time to look over the code. @heshanjse Can you take a look at Jeen's comments?

hmottestad:
@jeenbroekstra I would like to continue working with Heshan on the SHACL Sail. Do you agree with the current direction? Which is to follow the guidelines of a query engine, treating SHACL as a query language that returns invalid triples, with an abstract syntax tree, query plan generator, plan optimiser and executor. @jamesrdf @ansell What do you guys think?

heshanjse:
@jeenbroekstra Thank you so much for your code review. i fixed some issues in the code and i will send those changes with my next commit. can you please tell us are you guys ok with our direction?

jamesrdf:
Any tests that have similar dependencies as the main code (looks like these do), please put the tests in the same module as shacl main (provided they don't introduce a circular dependency in eclipse). Any tests that do introduce a circular dependency (in eclipse) should be move to their own compliance module, such would be the case if it has a repository or server dependency. A shacl testsuites module should be its own module and should be based on https://github.com/shexSpec/shexTest/ that can be used by shacl main or shacl compiance modules.

The-Alchemist:
Out of curiosity, is SHACL support now in RDF4J (and this PR can be abandoned), or is it worth fixing up this PR?

jamesrdf:
This is the only SHACL implementation right now for RDF4J and still needs to be cleaned up.

hmottestad:
Closing this PR. Code was moved into https://github.com/eclipse/rdf4j-storage/pull/34 which has been merged.

Issue #915, opened 2017-09-12
Split RDF4J Project


jamesrdf:
Split RDF4J Project into the following
 * [/eclipse/rdf4j](/eclipse/rdf4j)           rio, repository-api, and http client
 * [/eclipse/rdf4j-storage](/eclipse/rdf4j-storage)   repository-sail, and all sail impls
 * [/eclipse/rdf4j-tools](/eclipse/rdf4j-tools)     server, workbench, console, runtime, bom, assembly
 * [/eclipse/rdf4j-testsuite](/eclipse/rdf4j-testsuite) benchmark and testsuites
 * [/eclipse/rdf4j-doc](/eclipse/rdf4j-doc)       remains the home of documentation guides

ansell:
Will concurrent releases of all of the RDF4J artifacts still be made?

jgrzebyta:
@jamesrdf From the practical point of view today all parts are just forks of the same repository. For example code for `server` exists in all repositories but only one is still actively developed. Is it planned to clean the code - remove all undeveloped code from the parts? If so do you have any "scenario"/recommendations/plan for the cleaning?

jamesrdf:
Great question!

The develop branch of [/eclipse/rdf4j](/eclipse/rdf4j), [/eclipse/rdf4j-storage](/eclipse/rdf4j-storage), [/eclipse/rdf4j-tools](/eclipse/rdf4j-tools), and [/eclipse/rdf4j-testsuite](/eclipse/rdf4j-testsuite) don't overlap with one another. For any 2.3 code changes these develop branches for the corresponding github project is to be used.

We will use [/eclipse/rdf4j/tree/master](/eclipse/rdf4j/tree/master) for any 2.2.x code changes until 2.3 is released in the new year. The master branch of storage/tools/testsuites should not be modified (or used) in any way until after 2.3 is released.

Once 2.3 is released the develop branch of all these github projects will be copied to their master branches and used for any 2.3.x patch releases.

New issues and new pull-requests for 2.3 should be created in a corresponding github project. No need to close/reopen issues, instead reference existing issues/PRs from [/eclipse/rdf4j](/eclipse/rdf4j) using the cross referencing syntax: https://help.github.com/articles/autolinked-references-and-urls/

2.2.x shouldn't have much changes before 2.3 is released, so I don't expect porting 2.2.x PRs to 2.3 to be much of an issue. However, contributors will have to be careful to target their PRs to the develop branch of the correct github project!

[...]

PR #916, opened 2017-09-12
Fix eclipse/rdf4j#915: split project

jamesrdf:
This PR addresses GitHub issue: eclipse/rdf4j#915


jamesrdf:
The develop branches of the three other github repositories are ready to go. I had some unexpected trouble with network connectivity to download the snapshots (from the other builds), but I think it is sorted out now by attempting to pre-downloading dependencies as an earlier step (that won't fail the build). Take a look at these this week and we can merge this PR.
* [rdf4j-storage](/eclipse/rdf4j-storage/tree/develop)
* [rdf4j-tools](/eclipse/rdf4j-tools/tree/develop)
* [rdf4j-testsuite](/eclipse/rdf4j-testsuite/tree/develop)

hmottestad:
@jamesrdf I haven't followed the whole split project discussion. I agree that it's a great idea, since building the code today takes so long. 

For running the tests in the test suite, would i have to first `mvn install` the code in rdf4j-storage and then do `mvn test" in the test suite project?

jamesrdf:
Unless you plan on modifying the tests, you wouldn't even need to touch the rdf4j-testsuite project. The tests to be run are in their corresponding project, just run mvn test in either rdf4j, rdf4j-storage, and/or rdf4j-tools. rdf4j-testsuite contains a test library used by the compliance modules.

Issue #184, opened 2016-06-06
Adds xsd:duration arithmetics to SPARQL


Tpt:
It would be nice to add support of "+", "-", "<", ">", ">=" and "<=" operators on xsd:duration literal to SPARQL.

Jena already support such operators. see https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/NodeValueOps.java

Probably blocked by #164

jamesrdf:
Looks like a good idea, would need to target 2.3.0 though as it is a new feature. A Pull Request should be created in [/eclipse/rdf4j-storage/pulls](/eclipse/rdf4j-storage/pulls), as we are currently migrating storage/eval code to a separate project.

PR #926, opened 2017-10-13
issue #184: Adds some duration operators to SPARQL extended mode

Tpt:
I have only implemented operations that are provided by javax.xml.datatype and that should be conform to the XPath and XQuery operators specification. 

This PR addresses GitHub issue: #184 .


Tpt:
Moved to https://github.com/eclipse/rdf4j-storage/pull/10

PR #931, opened 2017-10-30
Suggest a few more valid xsd:gYear values

jimkont:
A library that was using a previous version of this library (openRDF)  was failing to parse xsd:gYear values greater than 9999. Even though I did not test the current version behavior, I thought it was a good idea to add these unit tests in the code base.

This PR adds the following tests in the rdf4j unit tests.

* checks that `20000` is a valid `xsd:gYear` value
* checks that `-20000` is a valid `xsd:gYear` value
* checks that `0000` is a valid `xsd:gYear` value


jamesrdf:
Thanks for sharing this. However, this would be better described in an issue. If you do want to provide complete code changes for this please, a PR is best, but please ensure that you follow the [CONTRIBUTING](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md) guide. I am going to close this PR as the code changes do not pass the unit tests, please create an issue to discuss possible solutions or discuss it on the mailing list.

Thanks!

PR #979, opened 2018-01-26
adding spanqit module

anqit:
This PR addresses GitHub issue: no issue created. This is to include a project described in the google grou[ (https://groups.google.com/forum/?utm_source=digest&utm_medium=email#!topic/rdf4j-users/pd2lR54CgzY) and discussed in email with @jamesrdf

Inclusion of https://github.com/anqit/spanqit and https://github.com/anqit/spanqit-examples
into rdf4j project


jeenbroekstra:
Thanks for getting started on this @anqit ! I am not fully across what you and James discussed but I definitely think having the spanqit functionality available as a module for RDF4J is a great idea.

I reckon this will need some further cleanup and discussion though, on how to best fit this in with the rest of RDF4J (without losing the ability to use it as an independent module of course). 

I am not sure how to deal with a code contribution under a separate license and with its own name (it may not be allowed under Eclipse Foundation rules). Perhaps, since this is such a large contribution, it might be best to first submit it for legal review by the Eclipse IP gurus.

anqit:
Hi Jeen,

Thanks for the reply. I didn't mean to leave the license in the PR, but now
that you bring up the point, I guess there are some protections I'd like to
ensure for the project as an rdf4j module. Do you know how I can get the
discussion started with the Eclipse IP people? Or is that something you
will have to do?

Thanks!
Ankit Prakash

jeenbroekstra:
Well let's discuss it here first, and see where that takes us. What kind of protections are you thinking of? 

A code contribution to RDF4J (or any Eclipse project) is governed under the ECA (https://www.eclipse.org/legal/ECA.php), which (given that there are no error markers on this PR) you presumably signed. IANAL so please read the actual ECA, but the gist of it is that you assert that you have the right to submit this code (e.g. you own it), and you grant the Eclipse Foundation a (non-exclusive) license to use and redistribute it to its downstream consumers. You, as the original owner, retain the copyright on your own work, and are free to distribute your own work through any other channel you see fit. 

In addition, RDF4J distributes code submissions that are part of its code base under the terms of the Eclipse Distribution License (which is a BSD-style license). See https://github.com/eclipse/rdf4j/blob/master/edl-v1.0.txt . Code submissions are reformatted to fit in with the RDF4J style guide and are required to include this standard copyright header:

`/*******************************************************************************
 * Copyright (c) 2018 Eclipse RDF4J contributors.
 *******************************************************************************/`

Do you think this is workable for you?

jeenbroekstra:
As an aside, the reason I said that we should let the Eclipse legal team review things is that for contributions of a certain size, simply signing the ECA is no longer enough - they want to do their own IP due diligence on the submission to make it doesn't inadvertently contain code that is owned by third parties.

anqit:
I guess the main thing I wanted to maintain was my own copyright on the
work, which appears protected by the ECA, so I can remove the Apache
license from the PR. Is there anything else I can do for now to continue to
move forward?

jeenbroekstra:
Cool, thanks. I think the first step is to get this code "into shape", which mainly just means removing the non-project license, and instead making it explicit that you are submitting this under the terms of the project license, which is the Eclipse Distribution License (EDL), v1.0. 

Simplest way to do this is to remove the Apache license file from your PR, and stick the source code header I posted above on top of every class.

Even after that we won't be able to immediately accept the PR. I'll need to check with someone in the know in Eclipse to figure out the exact next steps, but if you fix the licensing thing in this PR that's a good first step. 

PS sorry if this seems overly complicated - it's the first time we are accepting a third-party contribution of this size though. The rules are a bit different from a "normal" contribution like a few hundred lines of code :)

jeenbroekstra:
I've just verified what needs to happen, and it's simply this: once this PR is in good shape wrt to the license and headers, we (either myself or James) create a zip of your contribution, and submit it to the legal team as part of a Contribution Questionnaire. They have a good look at it, and once they approve it, we can merge in your contribution. We make a note in our project's IP log, and that's all - you're onboard :)

anqit:
Hey Jeen, finally figured out a script to add the headers to all the files
:) Let me know if this looks ok.

jamesrdf:
IP discussion CQ#16000 at https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16000

anqit:
I don't seem to be able to login to IPZilla, do I need some special
permission on my eclipse foundation account to be set?

jamesrdf:
It might just be for committers. I'll let you know when it is approved.

anqit:
Awesome, looking forward to the release!

Issue #1002, opened 2018-03-25
jjMoveNfa_0 in SyntaxTreeBuilderTokenManager is unusually slow


hmottestad:
The jjMoveNfa_0 in SyntaxTreeBuilderTokenManager is taking up 40% of the execution time, it seems, for executing SHACL rules that generate a lot of SPARQL queries.

We need to:
 - create SPARQL parsing benchmark
 - see how we can optimize jjMoveNfa_0

hmottestad:
Benchmark results before changes:

## Java 8
```
Benchmark                  Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.temp  avgt   40  74.169 ± 1.959  ms/op
```

## Java 9
```
Benchmark                  Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.temp  avgt   40  74.248 ± 1.105  ms/op
```

hmottestad:
Benchmark results *after* changes:

## Java 8
```
Benchmark                  Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.temp  avgt   40  32.390 ± 0.519  ms/op
```

## Java 9
```
Benchmark                  Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.temp  avgt   40  36.420 ± 0.727  ms/op
```

PR #1003, opened 2018-03-25
Issues/#1002  jj move nfa 0 performance

hmottestad:
This PR addresses GitHub issue: #1002 .

* Split jjMoveNfa_0 into multiple smaller methods, this makes the benchmark run in half the time. Presumably splitting the code allows the Java JIT to compile the code better, though unsure why. 
* 
*


hmottestad:
@jamesrdf found out that parsing SPARQL queries was making the SHACL engine slow. After a bit of digging I discovered that jjMoveNfa_0 was the culprite. So I've made a small benchmark and split the method up into smaller methods, this seemed sufficient in making it fast. Probably something to do with how Java optimized the method.

hmottestad:
@jamesrdf I have the benchmarks in SPARQLParseBenchmark. I installed the JMH plugin in Intellij which allows me to run the benchmarks right from there. Have added a main method for you now, and reordered the commits so you can `git checkout 84cafc7915213dd46d6f6f74f84b6716c1d693c9` to get just the benchmark, and then `git checkout "issues/#1002_-jjMoveNfa_0_performance"` to see the performance differences.

If I add @Test to the methods, then Intellij tries to run them as tests instead of as benchmarks with JMH. Also bumped the JMH version, for the JMH that I pull in with `<scope>test</scope>`, reckon that we don't need to involve eclipse compliance when we use test scope, do you know for sure?

Btw. warmup of 2 and 4 iterations is a bit too little to allow for JIT. I usually see JIT on the third warmup iteration.

jamesrdf:
What are the specs of the computer you tested this on? What is the version of the jdk? I have not been able to reproduce any significant improvement using 4x Intel(R) Core(TM) i5-2410M CPU @ 2.30GHz with openjdk version "1.8.0_151"
OpenJDK Runtime Environment (build 1.8.0_151-8u151-b12-0ubuntu0.16.04.2-b12)
OpenJDK 64-Bit Server VM (build 25.151-b12, mixed mode)

hmottestad:
Could have sworn you had a comment about not seeing any difference before-after the changes, but can't find it now. Did you manage to get it working @jamesrdf ?

jamesrdf:
Yes, I managed to reproduce the issue (not sure why I had trouble at first). I since removed that comment. Please take a look at the patch file I included. I reduced the changes down to the essentials. The PR looks good to me.

hmottestad:
@jamesrdf I did a pull and performance still looks good.

hmottestad:
@jeenbroekstra any comments?

jeenbroekstra:
Checking on my phone (traveling) but looks ok. I gather the changes have been applied in such a way that a new javacc run won't regress, so all good. If sparql testsuies all succeed I'm happy with it. Good work!

PR #1021, opened 2018-04-17
Use AssertJ

josephw:
Java Hamcrest development has [largely stalled](https://github.com/hamcrest/JavaHamcrest/issues/131). Although it's only lightly used in rdf4j's tests, I believe it's worth tracking a currently maintained framework.

This PR switches all use of `org.hamcrest` over to [AssertJ](http://joel-costigliola.github.io/assertj/). As a side benefit, some assertions are substantially more concise.

(Partly addresses #1030.)

### DIFF
--- a/compliance/manager/pom.xml
+++ b/compliance/manager/pom.xml
@@ -43,5 +43,10 @@
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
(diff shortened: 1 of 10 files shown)


jeenbroekstra:
Looks good, I'm fine this. However, there is no issue logged for this improvement, could you create one please? 

Also, since this introduces a new 3rd party dependency, we can't merge this until we've gotten a CQ for that dependency approved. I will log a CQ immediately.

jeenbroekstra:
CQ logged: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16249

Issue #1017, opened 2018-04-15
ParsedIRI.create() throws errors for perfectly valid urls


HansBrende:
ParsedIRI.create("http://2017familyandconsumerscienceeduc.sched.com/")

throws:

    Exception in thread "main" java.net.URISyntaxException: Invalid IPv4 address U+66 at index 11: 
    http://2017familyandconsumerscienceeduc.sched.com/
	    at org.eclipse.rdf4j.common.net.ParsedIRI.error(ParsedIRI.java:1098)
	    at org.eclipse.rdf4j.common.net.ParsedIRI.parseHost(ParsedIRI.java:979)
	    at org.eclipse.rdf4j.common.net.ParsedIRI.parse(ParsedIRI.java:858)
	    at org.eclipse.rdf4j.common.net.ParsedIRI.<init>(ParsedIRI.java:314)

Additionally: 

    ParsedIRI.create("http://69.agendaculturel.fr/concert/")

throws: 

    Exception in thread "main" java.lang.NumberFormatException: For input string: ""
	    at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	    at java.lang.Integer.parseInt(Integer.java:592)
	    at java.lang.Integer.parseInt(Integer.java:615)
	    at org.eclipse.rdf4j.common.net.ParsedIRI.parseHost(ParsedIRI.java:977)
	    at org.eclipse.rdf4j.common.net.ParsedIRI.parse(ParsedIRI.java:858)
	    at org.eclipse.rdf4j.common.net.ParsedIRI.<init>(ParsedIRI.java:314)

barthanssens:
ParsedIRI in 2.3.x assumes  that, when an IRI starts with a number, it is an IP-address
This might also breaks (sub)domains like http://2017.rdf.org,  http://256.be, http://127.0.0.1.me :-/

manuelfiorelli:
I encountered a seemingly related problem using the RDF4J OSGi runtime.

In this scenario, a URL to a resource inside a bundle (a JAR) may look like this: `bundle://21.0:0/META-INF/test/test.ttl`. When I use that URL as base URI for `RDFLoader`, I get the following stacktrace.
```java
java.lang.IllegalArgumentException: Invalid IPv4 address U+3A at index 13: bundle://21.0:0/META-INF/test/test.ttl
        at org.eclipse.rdf4j.common.net.ParsedIRI.create(ParsedIRI.java:266)
[...]

PR #1024, opened 2018-04-19
Issues/#1017 parsediri numdomain

barthanssens:
This PR addresses GitHub issue: #1017  .

* Commented out the IPv4 specific test
* Added a test for a numeric subdomain

Not a real fix, rather a quick and dirty one: NOT testing if a host is a valid IPv4, because the current test breaks on e.g. 2017.rdf4j.org... 

Drawback: ParsedIRI won't catch invalid IPv4 adresses anymore (like 257.0.0.1)

To really fix it, more logic is needed (e.g. 127.0.0.256 would be invalid, but 257.1.2.org would be valid)


jeenbroekstra:
@barthanssens PR #1027 addresses this same issue, but at first glance seems a more comprehensive solution. Can you review that one and if it's ok, perhaps we can close this PR and merge that solution instead?

barthanssens:
The other solution looks better indeed (mine is just quick & dirty).

PR #1027, opened 2018-05-04
#1017 Fix an issue with ParsedIRI and IRIs with host starting with digit

yasengmarinov:
Issues/#1017 Fix an issue with ParsedIRI and IRIs with host starting with a digit
   
Currently in the parseHost() method if the host starts with a digit then we consider that the host
is IPv4 address. However, there are no restrictions on using digits in host names so the parseHost() method
throws exceptions on perfetcly fine host names which start with a digit.
The suggested approach is: deltect a leading digit in the host name -> try to parse as IPv4
If the parsing fails -> parse the host as a regular host name (ireg-name according RFC-3987).
Once the host is parsed check if it is valid (it is considered invalid if its TLD is not made of letters only).
If the host is invalid we consider that indeed IPv4 address was intended and we throw.

### DIFF
--- a/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
+++ b/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
@@ -1006,8 +1006,8 @@ public class ParsedIRI implements Cloneable, Serializable {
 				// Reset position and parse host
 				pos = startPos;
 				String host = parsePctEncoded(hchar, ':', '/');
-
-				if (isTLDValid(start)) {
+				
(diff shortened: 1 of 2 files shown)


barthanssens:
Would it perhaps be possible to add a test for the issue mentioned by @manuelfiorelli  ?  https://github.com/eclipse/rdf4j/issues/1017#issuecomment-383099330 (i.e. test if "bundle://21.0:0/META-INF/test/test.ttl" gets parsed)

jeenbroekstra:
I've pushed an exception for the bundle: URI case to this branch, will merge as soon as build completes.

Issue #1035, opened 2018-06-07
Update jackson and http client dependencies


aschwarte10:
Currently RDF4J ships an outdated version of Jackson (2.8.8). For security reasons this should be updated to the latest release. See e.g. https://www.cvedetails.com/cve/CVE-2018-7489/ for a security issue in the jackson databind dependency.

The latest version currently is 2.9.5

Also http client and http client core should be updated to the latest stable patch releases

aschwarte10:
@jeenbroekstra are also requests for the updated dependencies in the Eclipse bug tracker required? If I followed the previous conversion correctly this is only required for the minor version update of jackson and not for the patch level update of http components, right?

jeenbroekstra:
@aschwarte10 that is correct, only needed for minor/major version bumps, not for patch updates. So I'll need this for the jackson dependencies, but not the others.

By the way, for what it's worth: I don't think it's really necessary to open separate tickets for this kind of upgrade for each repo. We'll leave it as-is for now, but in future feel free to just use one ticket for an upgrade across the subprojects.

barthanssens:
Probably less urgent for RDF4j, but there is a Jackson 2.9.6 release which (amongst others) fixes 2 security issues (remote execution possible oracle jdbc or jodd-db driver is in the classpath) when https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.6

aschwarte10:
@jeenbroekstra did you already file the CQ issues in the eclipse bug tracker? 
Otherwise (or if you are ok with this) I would suggest to switch to 2.9.6

@barthanssens thanks for pointing this out.

jeenbroekstra:
@aschwarte10 if you want to jump to 2.9.6 that's fine, don't need new CQs for patch-level updates. I'll go through the list of dependencies and log the necessary CQs ASAP.

jeenbroekstra:
Jackson core: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16713
Jackson databind: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16793
Jackson annotations: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16794

jeenbroekstra:
CQs approved, so feel free to merge version bumps

PR #1036, opened 2018-06-07
#1035: Update jackson to 2.9.5, http client and http core

aschwarte10:
This PR addresses GitHub issue: #1035  .

Update jackson to 2.9.5, http client and http core

jackson: 2.8.8 -> 2.9.5
http client: 4.5.2 -> 4.5.5
http core: 4.4.4 -> 4.4.9

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -211,9 +211,9 @@
 		<slf4j.version>1.7.10</slf4j.version>
 		<logback.version>1.1.2</logback.version>
 		<spring.version>4.2.1.RELEASE</spring.version>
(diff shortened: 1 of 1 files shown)


aschwarte10:
@jeenbroekstra are also requests for the updated dependencies in the Eclipse bug tracker required? If I followed the previous conversion correctly this is only required for the minor version update of jackson and not for the patch level update of http components, right?

jeenbroekstra:
@aschwarte10  that's correct. And I suspect that the jackon libraries have already been reviewed for other projects, so we can piggyback off those.

Issue #1041, opened 2018-06-13
Parsing NTriple with bnode fails if bnode lable contains "-"


RavinderSinghMaan:
As per N triple spec https://www.w3.org/TR/n-triples/#BNodes below is allowed

> 2.4 RDF Blank Nodes
> RDF blank nodes in N-Triples are expressed as _: followed by a blank node label which is a series of name characters. The characters in the label are built upon PN_CHARS_BASE, liberalized as follows:
> 
> The characters _ and [0-9] may appear anywhere in a blank node label.
> The character . may appear anywhere except the first or last character.
> The characters -, U+00B7, U+0300 to U+036F and U+203F to U+2040 are permitted anywhere except the first character. 
> 

But if I try to parse below 
```
Rio.parse(new StringReader("<http://example.org/people/alice> <http://www.perceive.net/schemas/relationship/enemyOf> _:genid-7047e9358ae44bd190d9f22a25762c2b-b1 ."), "", RDFFormat.NTRIPLES)
```
it fails with exception
```
org.eclipse.rdf4j.rio.RDFParseException: Expected '.', found: - [line 1]

	at org.eclipse.rdf4j.rio.helpers.RDFParserHelper.reportError(RDFParserHelper.java:322)
	at org.eclipse.rdf4j.rio.helpers.AbstractRDFParser.reportError(AbstractRDFParser.java:684)
	at org.eclipse.rdf4j.rio.ntriples.NTriplesParser.reportError(NTriplesParser.java:636)
	at org.eclipse.rdf4j.rio.ntriples.NTriplesParser.parseTriple(NTriplesParser.java:298)
	at org.eclipse.rdf4j.rio.ntriples.NTriplesParser.parse(NTriplesParser.java:177)
	at org.eclipse.rdf4j.rio.Rio.parse(Rio.java:343)
	at org.eclipse.rdf4j.rio.Rio.parse(Rio.java:254)
```

RavinderSinghMaan:
@jeenbroekstra any thoughts on this? If you guys agree that parser should allow extra characters mentioned in spec then I can try to create a patch for it.

jeenbroekstra:
Yeah that seems like a bug. A patch would be most welcome! Can you add a test case as well so we don't regress? Thanks!

RavinderSinghMaan:
@jeenbroekstra below are the test cases and code change required to fix it. If it looks ok then I can signup and do proper PR etc.
```
   @Test
[...]

PR #1045, opened 2018-06-24
issues/#1041 Changes in NTriplesParser to allow liberlaized character…

RavinderSinghMaan:
…s in blank node label. As NQuadsParser extends from NTriplesParser and spec for bnode label is same for both this change also applies to NQuadsParser.

Signed-off-by: RavinderSinghMaan <ravinder.singh@pearson.com>

This PR addresses GitHub issue: #1041  .

* As per the spec for NTriples blank node label can contain characters other than letters or numbers.
* This change contains fix for same.
* This change is also applicable to NQuadParser as blank node label spec is same for both NQuads and Ntriples.

### DIFF
--- a/rio/ntriples/src/main/java/org/eclipse/rdf4j/rio/ntriples/NTriplesParser.java
+++ b/rio/ntriples/src/main/java/org/eclipse/rdf4j/rio/ntriples/NTriplesParser.java
@@ -10,6 +10,7 @@ package org.eclipse.rdf4j.rio.ntriples;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.PushbackReader;
 import java.io.Reader;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
@@ -45,7 +46,7 @@ public class NTriplesParser extends AbstractRDFParser {
 	 * Variables *
 	 *-----------*/
 
-	protected Reader reader;
+	protected PushbackReader reader;
 
 	protected long lineNo;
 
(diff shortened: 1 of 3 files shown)


rio/ntriples/src/main/java/org/eclipse/rdf4j/rio/ntriples/NTriplesParser.java
```diff
@@ -470,22 +472,59 @@ else if (c != ':') {
...
+			reportError("Expected a letter or number or underscore, found: " + new String(Character.toChars(c)),
 					NTriplesParserSettings.FAIL_ON_NTRIPLES_INVALID_LINES);
 		}
 		name.append(Character.toChars(c));
 
 		// Read all following letter and numbers, they are part of the name
 		c = readCodePoint();
-		while (c != -1 && NTriplesUtil.isLetterOrNumber(c)) {
+		while (c != -1 && NTriplesUtil.isValidCharacterForBNodeLabel(c)) {
+			if(NTriplesUtil.isDot(c) && !NTriplesUtil.isValidCharacterForBNodeLabel(peekCodePoint())) {
```
jeenbroekstra: Shouldn't you also do an `unread` here on the dot char? Unless I misunderstand, in this situation, we should not consider the dot part of the node id, so it should be put back on the reader to be processed outside the context of the nodeId, I think.

jeenbroekstra: Ah never mind, of course not, the last-read char is returned by the method.

RavinderSinghMaan:
@jeenbroekstra is the PR for tests ok or you want me to do something else?

jeenbroekstra:
@RavinderSinghMaan all good, thanks again for the patch! There's a 2.4.0 release coming up soon, this fix will be included in that release.

PR #1046, opened 2018-06-27
Fix #1051: non-standard keys with arrays in JSON SPARQL results

### LINKED ISSUE #1051: (Follow-up of #994) JSON SPARQL parser fails if non-standard keys with arrays are present
After the fix for #994 the org.eclipse.rdf4j.query.resultio.sparqljson.AbstractSPARQLJSONParser can process non-standard keys with string values (e.g., "other": "Other keys, with different names, may be present in the JSON Results Object"), but it fails if the value is an array: (e.g., "otherArray": [ "Other keys, with different names, ", "may be present in the JSON Results Object" ]).

andriy-nikolov:
Signed-off-by: Andriy Nikolov <an@metaphacts.com>

This PR addresses GitHub issue: #1051 .

* Objects with non-standard keys in SPARQL JSON results get consumed.
* Extended the test file `other-keys.srj`

### DIFF
--- a/compliance/queryresultio/src/test/resources/sparqljson/other-keys.srj
+++ b/compliance/queryresultio/src/test/resources/sparqljson/other-keys.srj
@@ -2,6 +2,10 @@
     "head": { "vars": [ "book" , "title" ]
     } ,
     "other": "Other keys, with different names, may be present in the JSON Results Object",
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
Thanks @andriy-nikolov, looks good!

Issue #1065, opened 2018-08-08
Literal sets wrong datatype when language is empty (not null)


grexe:
When wrongly constructing a Literal (e.g. `SimpleLiteral`) with an *empty* (not-null) language, the wrong dataType is used and no Exception is thrown, because `setLanguage` only checks with `Objects.requireNonNull(language);` when it should also check for an empty String.

This contract is enforced in the constructor with IRI and datatype:
```
protected SimpleLiteral(String label, IRI datatype) {
        this.setLabel(label);
        if (RDF.LANGSTRING.equals(datatype)) {
            throw new IllegalArgumentException("datatype rdf:langString requires a language tag");
        } else {
            if (datatype == null) {
                datatype = XMLSchema.STRING;
            }
            this.setDatatype(datatype);
        }
    }
```
But not in:
```
    protected SimpleLiteral(String label, String language) {
        this.setLabel(label);
        this.setLanguage(language);
    }
```
in the latter case, `setLanguage` does not check for empty language:
```
    protected void setLanguage(String language) {
        Objects.requireNonNull(language);  // FIXME: should also check for empty language!
        this.language = language;
        this.setDatatype(RDF.LANGSTRING);
    }
```

So in our case, due to a UI bug, we ended up with some invalid labels that caused a bigger followup-issue. RDF4J correctly checking this would have helped us *fail fast* and detect that UI bug earlier, before resulting in data errors.

PR #1073, opened 2018-08-23
Issues/#1065 literal emptylang

barthanssens:

This PR addresses GitHub issue: #1065 .

* Throw an InvalidArgumentException when SimpleLiteral is created with an empty language tag 
* Added unit test

Another option would be to change the type to xsd:string when an empty language tag is provided, but that would probably be confusing

### DIFF
--- a/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
+++ b/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
@@ -113,12 +113,16 @@ public class SimpleLiteral implements Literal {
 		this.label = label;
 	}
 
+	@Override
 	public String getLabel() {
 		return label;
 	}
 
 	protected void setLanguage(String language) {
 		Objects.requireNonNull(language);
+		if (language.isEmpty()) {
+			throw new IllegalArgumentException("Language tag cannot be empty");
+		}
 		this.language = language;
 		setDatatype(RDF.LANGSTRING);
 	}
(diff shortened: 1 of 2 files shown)


grexe:
sorry for being so late with my feedback, but I could only now see the effect in an issue that popped up here recently. I am afraid this is not fully spec compliant, see:
https://www.w3.org/TR/swbp-xsch-datatypes/#sec-values-easy-examples
> In addition, [RDF Semantics] explicitly sanctions identification of RDF plain literals without language tags with corresponding typed literals with datatype xsd:string.

So, I'd really prefer a transparent type fix than an Exception, which forces the caller to handle simple literal assignments in a try-catch block.

jeenbroekstra:
I don't think this has anything to do with compliance to the spec really: RDF4J allows you to create plain literals without language tags, and it identifies those with corresponding xsd:string-typed literals.

If you are calling `ValueFactory.createLiteral(text, language)` it's your responsibility to check that that language tag is legal (for performance reasons the ValueFactory doesn't do full language tag validation itself). The empty string is a bit of a corner case, admittedly, but at the end of the day it's the caller's responsibility.

jeenbroekstra:
Fwiw there is a feature request for convience "static factory methods" (see #1325), I'm thinking that more flexible support for this kind of case could be part of that.

grexe:
Yes I agree @jeenbroekstra, just in our use case it was a bit awkward, since we wanted to call ValueFactory.createLiteral() always with a datatype, and that is actually forbidden and results in an Exception, because (this is true to spec) when using a language tag, the datatype is anyway fixed to rdfs:langstring. So yes, it was the callers mistake after all.

PR #1082, opened 2018-09-13
minor editorial fix in pom

jeenbroekstra:
Don't mind this, it's not a very interesting fix, but I am looking into configuring Jenkins to auto-verify our PRs, and needed a new PR to play with.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -772,7 +772,7 @@
 				</configuration>
 			</plugin>
 			<plugin>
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
Looks good. It works fine, is more configurable, and it looks like it's a lot faster too.  I'm enabling this for all our repos, and have asked the Eclipse webmaster to remove the Travis integration.

Issue #107, opened 2016-05-11
SparqlQueryRenderer does not support SPARQL 1.1


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-2206 )

SPARQL 1.1 constructs cannot be rendered via the SparqlQueryRenderer class.  

For example If I parse this query with QueryParserUtil.parseQuery:

{code}
SELECT * WHERE {
  ?s ?p ?o .
  BIND(uri("http://test-graph.com/") AS ?foo) .
}
{code}

And then render the ParsedQuery back out again with SPARQLQueryRenderer, it appears to lose the binding clause, returning the following string:

{code}
select ?g ?s ?p ?o ?g2
where {
  GRAPH ?g {
    ?s ?p ?o.
}}
{code}

Looking at renderTupleExpr in SparqlTupleExprRenderer I can see that the following lines are commented out:

{code:java}
        // aRenderer.mProjection = new ArrayList<ProjectionElemList>(mProjection);
        // aRenderer.mDistinct = mDistinct;
        // aRenderer.mReduced = mReduced;
        // aRenderer.mExtensions = new HashMap<String, ValueExpr>(mExtensions);
        // aRenderer.mOrdering = new ArrayList<OrderElem>(mOrdering);
        // aRenderer.mLimit = mLimit;
        // aRenderer.mOffset = mOffset;
{code}

With the following commented out in SPARQLQueryRenderer:

{code:java}
                        // SPARQL does not support this, its an artifact of copy and
                        // paste from the serql stuff
                        // aQuery.append(mRenderer.getExtensions().containsKey(aElem.getSourceName())
                        // ?
                        // mRenderer.renderValueExpr(mRenderer.getExtensions().get(aElem.getSourceName()))
                        // : "?"+aElem.getSourceName());
                        //
                        // if (!aElem.getSourceName().equals(aElem.getTargetName()) ||
                        // (mRenderer.getExtensions().containsKey(aElem.getTargetName())
                        // &&
                        // !mRenderer.getExtensions().containsKey(aElem.getSourceName())))
                        // {
                        // aQuery.append(" as ").append(mRenderer.getExtensions().containsKey(aElem.getTargetName())
                        // ?
                        // mRenderer.renderValueExpr(mRenderer.getExtensions().get(aElem.getTargetName()))
                        // : aElem.getTargetName());
                        // }
{code}

[...]

PR #1088, opened 2018-09-21
GH-107 reuse Files.createTempDirectory, deprecate FilUtil method

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j-storage#107 .

* reuse core Java `Files.createTempDirectory` method
* deprecate home-grown `FileUtil.createTempDir` as it's just redundant.

### DIFF
--- a/util/src/main/java/org/eclipse/rdf4j/common/io/FileUtil.java
+++ b/util/src/main/java/org/eclipse/rdf4j/common/io/FileUtil.java
@@ -14,7 +14,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.util.Collection;
 import java.util.StringTokenizer;
-import java.util.UUID;
 
 /**
  * Utility methods for operations on Files.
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
@hmottestad that's part of the plan yes. Unfortunately it's hard to reproduce locally or in a test setup, as it's likely the actual cause is multiple JVMs creating tempdirs at more or less the same time - so even locking doesn't help and the only fix is to use a larger random number for the dir suffix. All in all I'd rather let the core java library handle it.

hmottestad:
If it fails on Travis I don’t think it’s multiple jvms. On Travis your build runs on a vm or in a container, neither shared. So unless the tests are set up to run in parallel in maven, I don’t see how there would be a race condition.

jeenbroekstra:
I don't think I've seen this in Travis to be honest, it seems to be Jenkins issue. Anyway let's give this a try and see if it helps 😁

Issue #1087, opened 2018-09-21
Cleanup: annotations and javadoc


barthanssens:
"meta"-ticket for some non-code changes like adding Override annotations, removing unused imports and adding / correcting javadoc

barthanssens:
See also #790

barthanssens:
Closing this one, creating new ticket(s) when needed.

PR #1125, opened 2018-10-09
Issues/#1087 javadoc annotations

barthanssens:
This PR addresses GitHub issue: #1087  .

* I've used  Netbeans 8 Inspect and Transform to find and add missing Override annotations (over a 1000 of them in total...)
* Code still builds, so nothing seems to be broken

Arguably this does not add much when using a modern IDE (which should detect overrides anyway), but personally I prefer to add these annotations to the codebase


barthanssens:
Thanks, I'll run the same check on the other subprojects as well

PR #1146, opened 2018-10-25
Fix #1056 xxe vulnerability

### LINKED ISSUE #1056: XXE in XML Parser
# The Issue
An XML External Entity attack is a type of attack against an application that parses XML input. This attack occurs when XML input containing a reference to an external entity is processed by a weakly configured XML parser. This attack may lead to the disclosure of confidential data, denial of service, server side request forgery, port scanning from the perspective of the machine where the parser is located, and other system impacts.

# Where the Issue Occurred
The following code snippets display the usage of `XMLReader` without disabling entities:
https://github.com/eclipse/rdf4j/blob/e0718b2f53aea1385848c36bd8819b775ec5edb0/rio/trix/src/main/java/org/eclipse/rdf4j/rio/trix/TriXParser.java#L188

https://github.com/eclipse/rdf4j/blob/e0718b2f53aea1385848c36bd8819b775ec5edb0/rio/rdfxml/src/main/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLParser.java#L328

jeenbroekstra:
This PR addresses GitHub issue: #1056  .

* RdfXmlParser and TriXParser have a shared superclass that takes care of XML reader config
* doctype declaration disallowed by default, external text/param entities disabled
* SPARQLResultsXMLParser enforces the same settings
* added testcases

PS I consistently used the wrong issue number in my commit message. Some day soon I may get this right, but today's not that day...

### DIFF
--- a/queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/AbstractSPARQLXMLParser.java
+++ b/queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/AbstractSPARQLXMLParser.java
@@ -278,6 +278,9 @@ public abstract class AbstractSPARQLXMLParser extends AbstractQueryResultParser 
 	public Collection<RioSetting<Boolean>> getCompulsoryXmlFeatureSettings() {
 		Set<RioSetting<Boolean>> results = new HashSet<RioSetting<Boolean>>();
 		results.add(XMLParserSettings.SECURE_PROCESSING);
+		results.add(XMLParserSettings.DISALLOW_DOCTYPE_DECL);
+		results.add(XMLParserSettings.EXTERNAL_GENERAL_ENTITIES);
+		results.add(XMLParserSettings.EXTERNAL_PARAMETER_ENTITIES);
 		return results;
 	}
 
--- a/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLReaderBasedParser.java
+++ b/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLReaderBasedParser.java
@@ -53,7 +53,6 @@ public abstract class XMLReaderBasedParser extends AbstractRDFParser {
 		results.add(XMLParserSettings.DISALLOW_DOCTYPE_DECL);
 		results.add(XMLParserSettings.EXTERNAL_GENERAL_ENTITIES);
 		results.add(XMLParserSettings.EXTERNAL_PARAMETER_ENTITIES);
-		results.add(XMLParserSettings.LOAD_EXTERNAL_DTD);
 		return results;
 	}
 
@@ -80,7 +79,9 @@ public abstract class XMLReaderBasedParser extends AbstractRDFParser {
 	 *         {@link XMLReader#setFeature(String, boolean)}.
 	 */
 	public Collection<RioSetting<Boolean>> getOptionalXmlFeatureSettings() {
-		return Collections.<RioSetting<Boolean>> emptyList();
+		Set<RioSetting<Boolean>> results = new HashSet<RioSetting<Boolean>>();
+		results.add(XMLParserSettings.LOAD_EXTERNAL_DTD);
+		return results;
 	}
 
 	/**
--- a/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLReaderBasedParser.java
+++ b/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLReaderBasedParser.java
@@ -1,0 +1,168 @@
+package org.eclipse.rdf4j.rio.helpers;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.eclipse.rdf4j.common.xml.XMLReaderFactory;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.rio.RioSetting;
+import org.eclipse.rdf4j.rio.helpers.XMLParserSettings;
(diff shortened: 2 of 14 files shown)


queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/AbstractSPARQLXMLParser.java
```diff
@@ -278,6 +278,9 @@ protected void reportWarning(String msg) {
 	public Collection<RioSetting<Boolean>> getCompulsoryXmlFeatureSettings() {
 		Set<RioSetting<Boolean>> results = new HashSet<RioSetting<Boolean>>();
 		results.add(XMLParserSettings.SECURE_PROCESSING);
+		results.add(XMLParserSettings.DISALLOW_DOCTYPE_DECL);
+		results.add(XMLParserSettings.EXTERNAL_GENERAL_ENTITIES);
+		results.add(XMLParserSettings.EXTERNAL_PARAMETER_ENTITIES);
```
hmottestad: I see this is in line with OWASP: https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet#XMLReader

Sp that's good.

hmottestad: Btw. Would you like to add LOAD_EXTERNAL_DTD too, it's noted in OWASP that it shouldn't be needed but is still used.

A question about naming. You have DISALLOW_DOCTYPE_DECL but not DISALLOW_EXTERNAL_GENERAL_ENTITIES. This feels a bit inconsistent to me.

rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLReaderBasedParser.java
```diff
@@ -0,0 +1,169 @@
...
+	 * 
+	 * @return A collection of {@link RioSetting}s that indicate which boolean settings will always be setup
+	 *         using {@link XMLReader#setFeature(String, boolean)}.
+	 */
+	public Collection<RioSetting<Boolean>> getCompulsoryXmlFeatureSettings() {
+		Set<RioSetting<Boolean>> results = new HashSet<RioSetting<Boolean>>();
+		results.add(XMLParserSettings.SECURE_PROCESSING);
+		results.add(XMLParserSettings.DISALLOW_DOCTYPE_DECL);
+		results.add(XMLParserSettings.EXTERNAL_GENERAL_ENTITIES);
+		results.add(XMLParserSettings.EXTERNAL_PARAMETER_ENTITIES);
```
hmottestad: same comment here as for above

rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLReaderBasedParser.java
```diff
@@ -0,0 +1,169 @@
...
+	 * Returns a collection of settings that will be used, if set in {@link #getParserConfig()}, as XML parser
+	 * features using {@link XMLReader#setFeature(String, boolean)}.
+	 * <p>
+	 * Subclasses can override this to specify more supported settings.
+	 * 
+	 * @return A collection of {@link RioSetting}s that indicate which boolean settings can be setup using
+	 *         {@link XMLReader#setFeature(String, boolean)}.
+	 */
+	public Collection<RioSetting<Boolean>> getOptionalXmlFeatureSettings() {
+		Set<RioSetting<Boolean>> results = new HashSet<RioSetting<Boolean>>();
```
hmottestad: FYI these sets should probably be unmodifiable and maybe declared outside of the method so that we don't instantiate a new set every time. Also, since java 7, you can declare these without the redundat type spec: Set<RioSetting<Boolean>> results = new HashSet<>();

Here is another way of declaring it which I prefer:

```
Set<RioSetting<Boolean>> results = new HashSet<>(
    Arrays.asList(
        XMLParserSettings.LOAD_EXTERNAL_DTD
    )
);
```

rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLParserTest.java
```diff
@@ -72,8 +79,7 @@ public void tearDown()
 	public void rdfXmlLoadedFromInsideAJarResolvesRelativeUris()
 		throws Exception
 	{
-		URL zipfileUrl = this.getClass().getResource(
-				"/org/eclipse/rdf4j/rio/rdfxml/sample-with-rdfxml-data.zip");
+		URL zipfileUrl = this.getClass().getResource("/org/eclipse/rdf4j/rio/rdfxml/sample-with-rdfxml-data.zip");
```
hmottestad: I've experienced some oddities when mixing windows/linux/mac when calling getResource() on getClass(), so I now always call getClassloader() between those calls.

hmottestad:
Good fix @jeenbroekstra and great you've added some tests.

queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/AbstractSPARQLXMLParser.java
jeenbroekstra: @hmottestad I agree that the naming is inconsisent, but this is how they are named in the Xerces parser features documentation as well (see http://xerces.apache.org/xerces2-j/features.html). I figured it was better to stick to their naming conventions.

As for LOAD_EXTERNAL_DTD, that was already in there, but is implemented as an optional feature, not a compulsory one. I decided not to touch this and only changed its default to `false`.

PR #1152, opened 2018-10-29
GH-1056 minor code cleanup

jeenbroekstra:
This PR addresses GitHub issue: #1056  .

* minor code cleanup based off previous PR review


hmottestad:
Looks good. 

You could make the initial sets unmodifiable. Unless you were thinking of making an add/remove feature. 

Should we have a setter? For those users who rely on loading external resources. It’s only a security hole if you allow that for user submitted data. 

Could make the sets unmodifiable but non-final. And have a setter. If it’s likely someone will need that.

barthanssens:
Looks good, I assume this is also scheduled for 2.4.1 ?

Issue #1185, opened 2018-11-24
Update jsonld-java to 0.12.3


ansell:
There is a bug in the new code for #1162 (PR #1186 ) that requires an update of the jsonld-java dependency to 0.12.3 which was released today. 

This is required to pick up a bug fix for non-standard JSON parsing that is optionally allowed by the new code.

I think there is some sort of internal Eclipse legal review required for all upgrades, so separating this into a new issue so it can be tracked separately, but the version bump is in #1186 as it is required for code review.

barthanssens:
Thanks. It's a patch-level update (0.12.1 -> 0.12.3), so a legal review is not required.

Issue #1162, opened 2018-11-08
Use JsonUtils.fromJsonParser to allow configurability inside of RDF4J


ansell:
It would be ideal to be calling into JSONLD-Java using ``JsonUtils.fromJsonParser`` rather than ``JsonUtils.fromInputStream``, to allow the standard RDF4J configuration options to be used to setup the JsonParser, rather than using the default options provided by JSONLD-Java.

Related to this issue: https://github.com/jsonld-java/jsonld-java/issues/237

ansell:
I extended the scope of this issue to also include the other two Jackson based RIO parsers, RDF/JSON and SPARQL/JSON, based in the same way on the new `JsonParserSettings`

PR #1186, opened 2018-11-25
issue #1162 : Expose Jackson settings to allow configurability for JSON parsers

ansell:
This PR addresses GitHub issue: #1162 and #1185 .

* Exposes the various Jackson JSON parser features as RDF4J settings so they can be set by RDF4J users for JSONLD, RDF/JSON, and SPARQL/JSON. All of the new RDF4J settings are all set to the default Jackson values, so there is no change in behaviour for existing users
* Bumps jsonld-java version to pickup a fix that is exposed by the new, non-standard, off-by-default, ability to parse `NaN` as a native JSON double value

I closed the request that was targeted at master, and this PR is now rebased and targeted at develop


ansell:
@jeenbroekstra I don't have commit permissions so someone else will need to merge this in.

Issue #1205, opened 2018-12-16
rdf4j Turtle validation too loose for quoted strings?


rickhg12hs:
I have a Turtle file that `console.sh` reports as fine. (`console.sh` seems to use `org.eclipse.rdf4j.rio...`) However, `Jena`'s `riot` reports problems.

```bash
$ /path/to/rdf4j/eclipse-rdf4j-2.4.2/bin/console.sh 
20:18:35.779 [main] DEBUG o.e.r.c.platform.PlatformFactory - os.name = linux
20:18:35.791 [main] DEBUG o.e.r.c.platform.PlatformFactory - Detected Gnome window manager on Posix platform
Connected to default data directory
RDF4J Console 2.4.2+285c2b9

2.4.2+285c2b9
Type 'help' for help.
> verify data.ttl
RDF Format is Turtle
Verifying data...
Data verified, no errors were found
File contains 85499 statements
```

```bash
$ riot --validate ./data.ttl 
20:52:59 ERROR riot                 :: [line: 87123, col: 1 ] Broken token (newline): The reported vulnerability is fixed in version 4.0.0.39 of nPrtoect AVS.<br />

```
There is a newline (ASCII `x0a`) on line 87123 of `data.ttl` in the middle of a triple's string object.

In addition, I'm wondering why `rdf4j/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/Turtle grammar.txt` isn't identical to `https://www.w3.org/TR/turtle/#sec-grammar-grammar`.

barthanssens:
I assume the grammar.txt is still based on a very early version of Turtle... (the file itself dates back to at least OpenRDF Sesame 2.0, which is 11 years old... see also https://github.com/ansell/openrdf-sesame/blob/2.0/core/openrdf-rio/openrdf-rio-turtle/src/main/java/org/openrdf/rio/turtle/Turtle%20grammar.txt)

IIRC a newline is allowed if the object is surrounded by three quotes (""") , but I guess that's not the case in your data.ttl file ?

rickhg12hs:
> IIRC a newline is allowed if the object is surrounded by three quotes (""") , but I guess that's not the case in your data.ttl file ?

Correct.

Is the fix to just simply update the grammar to the current standard?

barthanssens:
I don't think so, looks like it's just included for reference. 

Method to be changed is probablye the `org.eclipse.rdf4j.rio.turtle.TurtleParser#parseString`

[...]

PR #1207, opened 2018-12-17
Check for CR or NL in single quoted object + test case

barthanssens:
This PR addresses GitHub issue: #1205  .

* Check for new line inside object literal
* Added test for new line inside quoted (not allowed) and triple quoted (allowed) literal
* Minor code formatting in tests

(FYI, the `writeLiteral` method in the TurtleWriter checks for CR / NL and uses triple quotes when the literal contains a CR / NL, so writing should already be covered)


jeenbroekstra:
Never mind, this PR is off the develop branch so a rebase sucks way too much other stuff with it. Given that 2.5 release date is relatively close we can just merge this in develop. If we decide we want a fix for 2.4.x later on we can always cherrypick.

Issue #1210, opened 2018-12-18
zip_slip


360CodeSafe:
Hi all,
There is a path traversal vulnerability found by Qihoo360 CodeSafe Team.
Details as bellow:
![default](https://user-images.githubusercontent.com/39950310/50139352-690c8680-02dc-11e9-9612-56394c620268.png)

When decompressing zip files, entries are not checked, resulting in overwriting arbitrary files by traversing directories using “.. /”

barthanssens:
Well, it sure is a small effort to backport it.
I'll take care of it this evening or tomorrow morning.

PR #1211, opened 2018-12-18
Verify that zip file entries don't try to escape the parent dir + test

barthanssens:
This PR addresses GitHub issue: #1210 .

* Check if zip file entries try to 'escape' / 'break out' of the directory  (i.e. tricks like `../../../etc/`)
* Added test to check if a file 
* Minor code formatting

### DIFF
--- a/util/src/main/java/org/eclipse/rdf4j/common/io/ZipUtil.java
+++ b/util/src/main/java/org/eclipse/rdf4j/common/io/ZipUtil.java
@@ -26,9 +26,14 @@ public class ZipUtil {
 	 */
 	private final static byte MAGIC_NUMBER[] = { (byte)0x50, (byte)0x4B, (byte)0x03, (byte)0x04 };
 
-	public static boolean isZipStream(InputStream in)
-		throws IOException
-	{
+	/**
+	 * Test if an input stream is a zip input stream by checking the "magic number"
+	 * 
+	 * @param in input stream
+	 * @return true if start of input stream matches magic number
+	 * @throws IOException 
+	 */
+	public static boolean isZipStream(InputStream in) throws IOException {
 		in.mark(MAGIC_NUMBER.length);
 		byte[] fileHeader = IOUtil.readBytes(in, MAGIC_NUMBER.length);
 		in.reset();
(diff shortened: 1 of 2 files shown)


util/src/main/java/org/eclipse/rdf4j/common/io/ZipUtil.java
```diff
@@ -45,9 +50,7 @@ public static boolean isZipStream(InputStream in)
 	 * @throws IOException
 	 *         when something untoward happens during the extraction process
 	 */
-	public static void extract(File zipFile, File destDir)
-		throws IOException
-	{
+	public static void extract(File zipFile, File destDir) throws IOException {
```
hmottestad: Unrelated to the security fix. It seems you’re not following the style guide. Mind you, I think the style guide is old and terrible, designed for the screens we had in the 1980s.

util/src/main/java/org/eclipse/rdf4j/common/io/ZipUtil.java
```diff
@@ -87,15 +88,16 @@ public static void extract(ZipFile zipFile, File destDir)
 	 * @throws IOException
 	 *         if the entry could not be processed
 	 */
-	public static void writeEntry(ZipFile zipFile, ZipEntry entry, File destDir)
-		throws IOException
-	{
+	public static void writeEntry(ZipFile zipFile, ZipEntry entry, File destDir) throws IOException {
 		File outFile = new File(destDir, entry.getName());
 
+		if (! outFile.getCanonicalFile().toPath().startsWith(destDir.toPath())) {
```
hmottestad: Looks like a decent way of checking for escape attempts. Btw. shouldn’t both paths be compared as canonical?

hmottestad:
Great that you made tests!

util/src/main/java/org/eclipse/rdf4j/common/io/ZipUtil.java
barthanssens: Thanks, I'll fix that. Probably best to switch to Eclipse IDE after all, as I'm heavily influenced by the more compact Sun Java code conventions (and Netbeans + the Eclipse-formatter don't seem to work as expected)

util/src/main/java/org/eclipse/rdf4j/common/io/ZipUtil.java
barthanssens: good catch, added

Issue #1237, opened 2019-01-08
TurtleParser throws fatal error on invalid datatype URI even when configured not to


jeenbroekstra:
```
		
		String ttl = "\n" + 
				"<http://example.org/foo> <http://example.org/prop> \"30.0\"^^<http://example.org/weird datatype> .\n" +
				"<http://example.org/bar> a <http://example.org/Foo> .";
		
		Model model = new TreeModel();
		RDFParser parser = Rio.createParser(RDFFormat.TURTLE);
		parser.getParserConfig().addNonFatalError(BasicParserSettings.VERIFY_URI_SYNTAX);
		parser.setRDFHandler(new StatementCollector(model));
		parser.setParseErrorListener(new ParseErrorLogger());
		parser.parse(new StringReader(ttl), "");
		
		for (Statement st: model) {
			System.out.println(st);
		}
```

The above code results in a fatal error, where it should just report a warning and skip the first line, correctly processing the second statement. 

The cause is in TurtleParser line 684 and further, where just immediately a fatal error is thrown.

PR #1238, opened 2019-01-08
GH-1237 invalid datatype IRI only fatal is VERIFY_URI_SYNTAX is enabled and fatal

jeenbroekstra:
This PR addresses GitHub issue: #1237 .

* if VERIFY_URI_SYNTAX is set but non-fatal, an illegal datatype URI should result in a skipped line, not a fatal error
* distinguish between unparsable datatype IRIs and datatypes that are something else (e.g. literals)
* regression tests added to cover cases (renamed test class for consistency)

### DIFF
--- a/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
+++ b/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
@@ -39,17 +39,15 @@ import org.eclipse.rdf4j.rio.helpers.BasicParserSettings;
 import org.eclipse.rdf4j.rio.helpers.TurtleParserSettings;
 
 /**
- * RDF parser for <a href="https://www.w3.org/TR/turtle/">RDF-1.1 Turtle</a>
- * files. This parser is not thread-safe, therefore its public methods are
- * synchronized.
+ * RDF parser for <a href="https://www.w3.org/TR/turtle/">RDF-1.1 Turtle</a> files. This parser is not
+ * thread-safe, therefore its public methods are synchronized.
  * <p>
- * <li>Normalization of integer, floating point and boolean values is dependent
- * on the specified datatype handling. According to the specification, integers
- * and booleans should be normalized, but floats don't.</li>
- * <li>Comments can be used anywhere in the document, and extend to the end of
- * the line. The Turtle grammar doesn't allow comments to be used inside triple
- * constructs that extend over multiple lines, but the author's own parser
- * deviates from this too.</li>
+ * <li>Normalization of integer, floating point and boolean values is dependent on the specified datatype
+ * handling. According to the specification, integers and booleans should be normalized, but floats
+ * don't.</li>
+ * <li>Comments can be used anywhere in the document, and extend to the end of the line. The Turtle grammar
+ * doesn't allow comments to be used inside triple constructs that extend over multiple lines, but the
+ * author's own parser deviates from this too.</li>
  * </ul>
  * 
  * @author Arjohn Kampman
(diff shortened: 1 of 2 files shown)


rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -680,19 +698,31 @@ protected Literal parseQuotedLiteral() throws IOException, RDFParseException, RD
 			// Read datatype
 			Value datatype = parseValue();
 			if (datatype instanceof IRI) {
-				return createLiteral(label, null, (IRI) datatype, getLineNumber(), -1);
-			} else {
+				return createLiteral(label, null, (IRI)datatype, getLineNumber(), -1);
+			}
+			else if (datatype != null) {
```
jeenbroekstra: I don't know why the Eclipse code formatter keeps throwing these curve balls, but the important bit of the change is here and in lines 707-719.

jeenbroekstra:
@ansell still can't assign you as a reviewer but would like your feedback on this if you have time :)

rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -680,19 +698,31 @@ protected Literal parseQuotedLiteral() throws IOException, RDFParseException, RD
...
 				reportFatalError("Illegal datatype value: " + datatype);
 				return null;
 			}
-		} else {
+			else {
+				if (getParserConfig().get(BasicParserSettings.VERIFY_URI_SYNTAX)
```
ansell: This logic should be available using AbstractRDFParser.reportError(String, RioSetting), which then checks whether it is a fatal error to throw an exception or return silently/log the error. This enables the ParseErrorListener to see the non-fatal error, which isn't the case in this code.

https://github.com/eclipse/rdf4j/blob/master/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/AbstractRDFParser.java#L635

https://github.com/eclipse/rdf4j/blob/master/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFParserHelper.java#L359

Similar to this code:

https://github.com/eclipse/rdf4j/blob/master/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/AbstractRDFParser.java#L387

rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -679,20 +702,23 @@ protected Literal parseQuotedLiteral() throws IOException, RDFParseException, RD
 
 			// Read datatype
 			Value datatype = parseValue();
-			if (datatype instanceof IRI) {
-				return createLiteral(label, null, (IRI) datatype, getLineNumber(), -1);
-			} else {
+			if (datatype instanceof Literal || datatype instanceof BNode) {
```
jeenbroekstra: @ansell new simplified logic. I distinguish between the datatype being an invalid IRI vs not being an IRI at all (e.g. a literal or bnode value instead). The latter should always result in a fatal error. The former only based on whether VERIFY_URI_SYNTAX is set. See the unit tests to understand how this behaves in practice.

ansell: That logic looks okay to me and more maintainable. The fixed set of negative `instanceof` checks aren't an issue because the RDF model isn't going to be changing anytime soon (and `instanceof` checks are relied on throughout the whole codebase anyway)

Issue #1263, opened 2019-01-26
RepositoryManager should use factory for model creation


jeenbroekstra:
In order to allow greater control over how RepositoryManagers read and manipulate config data, `RepositoryManager` should allow injection of a `ModelFactory` instead of just directly calling a specific `Model` impl constructor. 

This will also allow us to write more robust testing as we will be able to stub/mock the handling of config data without relying on a specific Sail implementation.

PR #1264, opened 2019-01-26
GH-1263 temporary workaround for test dependency issue

jeenbroekstra:
This PR addresses GitHub issue: #1302.

* temporary workaround for testing circular dependency: test dependencies use a fixed older version of rdf4j. 
* documented necessity to stub out these dependencies in unit tests as per #1263
* documented need to allow injection of a `ModelFactory` for test stubbing/mocking purposes as per #1263.

### DIFF
--- a/repository/manager/pom.xml
+++ b/repository/manager/pom.xml
@@ -67,16 +67,22 @@
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
 
+		<!-- testing dependencies -->
+
 		<dependency>
-			<groupId>${project.groupId}</groupId>
-			<artifactId>rdf4j-sail-memory</artifactId>
-			<version>${project.version}</version>
+			<groupId>org.eclipse.rdf4j</groupId>
+			<artifactId>rdf4j-repository-sail</artifactId>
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
Note that the PR verify failed on the compliance module, which is expected (as there is no 3.0-SNAPSHOT testsuite yet), but this won't affect the build/deploy as those modules are completely skipped in that phase. 

It's all a bit of a song and dance to be honest. For now we can deal with the compliance tests failing, but I think we should consider if _all_ compliance tests should live in the testsuite repo, instead of in their respective impl's repos.

jeenbroekstra:
Turns out this same issue also affects release deployment on the master branch :( I may need to cherrypick this across so we can get the 2.5 milestone out the door.

hmottestad:
I miss the single repo

jeenbroekstra:
> I miss the single repo

I do too, sometimes, but arguably in this case it's a good thing things are separated, it keeps us on our toes. This kind of circular dependency is bad for long-term maintenance, even if everything is in a single repo. The solution is better test isolation, which is Good Coding Practice(tm) :)

Issue #1038, opened 2018-06-09
SparqlBuilder to work with rdf4j model objects


jeenbroekstra:
Currently, the SparqlBuilder module has its own set of interfaces for RDF value objects: `RdfResource`, `Iri`, `Prefix`, etc. To facilitate better integration with the rest of the RDF4J framework we should make sure that the builder API (also) accepts RDF4J model objects, like `org.eclipse.rdf4j.model.IRI` etc.

anqit:
Can this be done while maintaining the module's ability to be used independently?

jeenbroekstra:
Well it will need a maven dependency on rdf4j-model, but we can extend the API in such a way that you don't _have_ to use rdf4j model objects if you don't want to, I think.

reckart:
> Well it will need a maven dependency on rdf4j-model, but we can extend the API in such a way that you don't have to use rdf4j model objects if you don't want to, I think.

Having to carry around a `ValueFactory` in RDF4J is quite a pain. The way the query builder code does it using static factory methods is much more convenient.

jeenbroekstra:
> Having to carry around a `ValueFactory` in RDF4J is quite a pain. The way the query builder code does it using static factory methods is much more convenient.

This (a closed issue) isn't really the best place to be having a conversation about this, but if you have a specific use case in mind where some other pattern than a `ValueFactory` would be better, please open a feature request (or alternatively, send a message to the user forum). We are currently mapping out our plans for rdf4j 3.0, so this is an ideal time to come with proposals for improvements, even if they are not necessarily backward compatible.

reckart:
@jeenbroekstra see #1325

PR #1265, opened 2019-01-27
GH-1038 sparqlbuilder with RDF4J Model objects

jeenbroekstra:
This PR addresses GitHub issue: #1038 .

* various conversions allow use of RDF Model objects directly in the SparqlBuilder
* added example variants to demonstrate its use

A todo on all of this is that the tests for the sparqlbuilder are not actually tests - they just print out the result, there are no assertions.


jeenbroekstra:
@anqit I finally had time to revisit this, have a look. Feedback welcome.

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/graphpattern/GraphPatterns.java
```diff
@@ -43,6 +48,76 @@ public static TriplePattern tp(RdfSubject subject, RdfPredicate predicate, RdfOb
...
+	 * @param predicate the triple pattern predicate as a {@link IRI}
+	 * @param objects the triples pattern object(s)
+	 * 
+	 * @return a new {@link TriplePattern}
+	 * 
+	 * @see <a
+	 *      href="http://www.w3.org/TR/2013/REC-sparql11-query-20130321/#QSynTriples">
+	 *      Triple pattern syntax</a>
+	 */
+	public static TriplePattern tp(Resource subject, IRI predicate, RdfObject... objects) {
```
hmottestad: I would prefer more verbose method names. Unless this method existed already.

sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/examples/sparql11spec/Section2Test.java
```diff
@@ -88,6 +123,19 @@ public void example_2_3_3() {
 		p();
 	}
 	
+	@Test
+	public void example_2_3_3_model() {
+		String datatype = "specialDatatype";
+		Variable v = query.var(), p = query.var();
```
hmottestad: Is v==p or does query.var() return a new object each time?

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/RdfSubject.java
```diff
@@ -37,6 +39,24 @@ default TriplePattern has(RdfPredicate predicate, RdfObject... objects) {
...
+	 * @return a new {@link TriplePattern} with this subject, and the given predicate and object(s)
+	 * 
+	 * @see <a
+	 *      href="http://www.w3.org/TR/2013/REC-sparql11-query-20130321/#QSynTriples">
+	 *      Triple pattern syntax</a>
+	 */
+	default TriplePattern has(IRI predicate, RdfObject... objects) {
+		return has(Rdf.iri(predicate), objects);
+	}
+
```
anqit: I think this should include a couple methods that take `Value...` as well, no? One for `RdfPredicate` and `Value...` and one for `IRI` and `Value...`?

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/util/SparqlBuilderUtils.java
```diff
@@ -12,7 +12,9 @@
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
+import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.sparqlbuilder.core.QueryElement;
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Iri;
```
anqit: looks like unused imports

sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/examples/sparql11spec/Section2Test.java
anqit: it will return a new `Variable` that is unique to the query instance that it was called on

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/graphpattern/GraphPatterns.java
jeenbroekstra: Keep in mind that in a fluent API it kinda makes sense to keep method names short (though you're right that `tp` is _really_ short :)). And yes this was pre-existing. 

As for the style reformatting, sorry about that, not much I can do at this point though - it should have been reformatted when first added to the code base. Redefining our style guide to get our various tools in better sync is on the calendar, but until that has been done this will occasionally happen.

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/graphpattern/GraphPatterns.java
```diff
@@ -43,6 +48,76 @@ public static TriplePattern tp(RdfSubject subject, RdfPredicate predicate, RdfOb
...
+	 * @param objects the triples pattern object(s)
+	 * 
+	 * @return a new {@link TriplePattern}
+	 * 
+	 * @see <a
+	 *      href="http://www.w3.org/TR/2013/REC-sparql11-query-20130321/#QSynTriples">
+	 *      Triple pattern syntax</a>
+	 */
+	public static TriplePattern tp(RdfSubject subject, IRI predicate, RdfObject... objects) {
+		return new TriplesSameSubject(subject, Rdf.iri(predicate), objects);
```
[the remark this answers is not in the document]
jeenbroekstra: Good catch, thanks. Will fix.

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/RdfSubject.java
jeenbroekstra: Yeah, missed that one!

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/Rdf.java
```diff
@@ -21,42 +27,60 @@
...
 /**
- * A class with static methods to create basic RDF objects
+ * A class with static methods to create basic {@link SparqlBuilder} RDF objects from either string values or
+ * RDF4J Model objects.
  */
 public class Rdf {
+
 	// not sure if other protocols are generally used in RDF iri's?
-	private static final Set<String> IRI_PROTOCOLS = Stream.of("http://", "https://", "mailto:").collect(Collectors.toSet());
+	private static final Set<String> IRI_PROTOCOLS = Stream.of("http://", "https://", "mailto:").collect(
```
barthanssens: I've also used "tel:" actually (in vcard stuff, but admittedly very rare), but `urn:` is not uncommon (especially 
 `urn:uuid:`)

sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/examples/BaseExamples.java
```diff
@@ -39,6 +41,8 @@
 	protected static final String DC_NS = "http://purl.org/dc/elements/1.1/";
```
barthanssens: I'd rather use the DC.NAMESPACE and FOAF.NAMESPACE to avoid any mistakes, both are already defined in org/eclipse/rdf4j/model/vocabulary package

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/Rdf.java
jeenbroekstra: Yes, this is a problem in the current SparqlBuilder codebase. We should look into letting it use RDF4J's IRI validation codes instead of this ad-hoc string parsing.  

I'm currently in rather a hurry, sorry, if you have time could you log this as a new bug? It's not something I'll address in this PR but we don't want to forget about it.

barthanssens: Ok, I've created  #1270

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/Rdf.java
```diff
@@ -155,56 +199,96 @@ public static RdfPredicateObjectList predicateObjectList(RdfPredicate predicate,
...
-	 * Convert an array of {@link String}s to an array of {@link StringLiteral}s
+	 * Converts an array of object {@link Value}s to an array of {@link RdfObject}s.
 	 * 
-	 * @param literals the {@link String}s to convert
+	 * @param values
+	 *        an array of {@link Value}.
+	 * @return an array of {@link RdfObject}
+	 */
+	public static RdfObject[] objects(Value... values) {
+		final RdfObject[] result = new RdfObject[values.length];
```
anqit: I think it could be useful having a method that converts a single `Value` to a single `RdfObject` (so everything in the `for` loop) in case a user would like to do that and not get an array back. And then this could be one-lined using streams, something like `Arrays.stream(values).map(Rdf::convertValueToRdfObject).toArray(RdfObject[]::new)`

Issue #1271, opened 2019-01-31
Rdf4j-sesameclientimpl thread occupies 100% of the CPU (without any use)


tomas-knap:
When using rdf4j remote (so having a client connecting to RDF4j server), we experienced on couple of servers that rdf4j-sesameclientimpl-%d thread occupies 100% of the CPU (even if there is no load, no triples added etc.). 

I was investigating further and realized that the issue is because of the way how [thread pool for the client is created (see the link to see the particular line of code](https://github.com/eclipse/rdf4j/blob/master/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java#L59). 

The problem is that you defined the thread pool with 0 core threads, but there is a bug in Java 8 core library (the version of Java used by all our 150+ customers) - https://bugs.openjdk.java.net/browse/JDK-8129861

The workaround (suggested in that bug report) is to set number of core threads to 1. So I changed [that line of code](https://github.com/eclipse/rdf4j/blob/master/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java#L59) to:
```
Executors.newScheduledThreadPool(1, new ThreadFactory() ...
```
After that change, I saw that immediately the CPU usage went down from 100% to 1% or so. 

We are currently using 2.3.2. Can you please do that single line of code change for us and create a bugfix release 2.3.3?

Thank you! Tomas Knap, Semantic Web Company

barthanssens:
Also present in newer releases apparently

jeenbroekstra:
Apparently, Github does not allow more than one milestone on an issue :-1: 

@tomas-knap Thanks for reporting this. Let us know if the fix as implemented by @barthanssens is workable for you. We can accomodate doing a 2.3.3 hotfix release for this. This release will _only_ include a fix for this particular issue. We'll likely immediately follow that with a 2.4.4 patch release as well, so that this particular fix will also be available in the latest stable release.

tomas-knap:
@jeenbroekstra Thank you for the quick response! 

Yes, the fix suggested by @barthanssens is OK - it is even better than change it in the code to 1, as the workaround (1 core thread instead of 0) includes one more idle thread, which is probably waste of resources from jdk9+.  

So yes, please go ahead with the PR https://github.com/eclipse/rdf4j/pull/1272 and please let me know when 2.3.3 hotfix release is ready (please create also the one-jar version with all RDF4j modules).

Thank you!

[...]

PR #1272, opened 2019-01-31
Issues/#1271 client cpu

barthanssens:
This PR addresses GitHub issue: #1271 .

* Workaround for openjdk8 bug, changing executors cores to 1 when `org.rdf4j.client.executors.jdkbug `system property is set

### DIFF
--- a/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
+++ b/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
@@ -56,7 +56,8 @@ public class SharedHttpClientSessionManager implements HttpClientSessionManager,
 
 	public SharedHttpClientSessionManager() {
 		final ThreadFactory backingThreadFactory = Executors.defaultThreadFactory();
-		this.executor = Executors.newScheduledThreadPool(1, new ThreadFactory() {
+		int cores = (System.getProperty("org.rdf4j.client.executors.jdkbug") != null) ? 1 : 0;
+		this.executor = Executors.newScheduledThreadPool(cores, new ThreadFactory() {
 
 			@Override
 			public Thread newThread(Runnable runnable) {
--- a/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
+++ b/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
@@ -56,7 +56,7 @@ public class SharedHttpClientSessionManager implements HttpClientSessionManager,
 
 	public SharedHttpClientSessionManager() {
 		final ThreadFactory backingThreadFactory = Executors.defaultThreadFactory();
-		this.executor = Executors.newScheduledThreadPool(0, new ThreadFactory() {
+		this.executor = Executors.newScheduledThreadPool(1, new ThreadFactory() {
 
 			@Override
 			public Thread newThread(Runnable runnable) {


barthanssens:
For current code base, if this is actually something we want to have in the latest release.
Not backported (yet)

http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
```diff
@@ -56,7 +56,8 @@
 
 	public SharedHttpClientSessionManager() {
 		final ThreadFactory backingThreadFactory = Executors.defaultThreadFactory();
-		this.executor = Executors.newScheduledThreadPool(0, new ThreadFactory() {
+		int cores = (System.getProperty("org.rdf4j.client.executors.jdkbug") != null) ? 1 : 0;
```
hmottestad: I’m comfortable with this kind of change if you add a nice big juicy paragraph-long comment explaining what the value should be and which bug is causing the issue with a link to the jdk bug. 

Also. Can you pull “cores” out as a static final field? We don’t need to support this dynamically changing.

barthanssens: Ok, good point, I've moved core to static final and added a few lines with the link to the openjdk bug report.

http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
```diff
@@ -30,7 +30,15 @@
...
 public class SharedHttpClientSessionManager implements HttpClientSessionManager, HttpClientDependent {
-
+	/**
+	 * FIXME: issue #1271, workaround for OpenJDK 8 bug.
+	 * ScheduledThreadPoolExecutor with 0 core threads may cause 100% CPU usage.
+	 * Using 1 core thread instead of 0 (default) fixes the problem but wastes some resources.
+	 * 
+	 * @see <a href="https://bugs.openjdk.java.net/browse/JDK-8129861">JDK-8129861</a>
+	 */
+	private static final int cores = (System.getProperty("org.rdf4j.client.executors.jdkbug") != null) ? 1 : 0;
```
jeenbroekstra: I'd prefer org.eclipse.rdf4j.* , keeping in sync with the package naming.

jeenbroekstra:
Btw if we backport, we'll need to do that twice. Once for a 2.3 patch, once for a 2.4 one. We can release both simultaneously.

http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
barthanssens: ok, done

barthanssens:
ok, good point, just let me know if you want me to do the release(s) if the change is worthwhile, or if you prefer to do it yourself

jeenbroekstra:
Happy with this, I'll do a squash-and-merge this time around, makes it easier to cherrypick and backport.

jeenbroekstra:
@barthanssens if you're happy to do the 2.3.3 hotfix release, by all means go for it! Please take a look at http://docs.rdf4j.org/developer/#_hotfix_releases . It documents how to handle creating the release branch for this (which is different from a normal patch release), and how to handle backporting the fix code to that release branch without adding in every other change since 2.3 as well :) Let me know if you're ok with doing it. Release date is up to you, I'd just wait until the OP has confirmed that the fix as implemented is workable for them.  

I suggest you do this 2.3.3 hotfix release first (with _only_ this particular fix in it), and then either you or I can follow with a 2.4.4 release the day after - if we're going to do a patch version anyway, we might as well include a few other fixes in that one, like the seconds bug fix. But I'd prefer to keep the 2.3.3 release minimal.

barthanssens:
Ok, I'll wait for the OP, and I'll do the 2.3.3 release

tomas-knap:
As I wrote [here](https://github.com/eclipse/rdf4j/issues/1271#issuecomment-460191743), the changes proposes in this PR makes sense, so please go ahead with 2.3.3 hotfix release, which includes only this bugfix (please create also the one-jar version with all RDF4j modules).

Please let me know when ready, thanks!

Issue #1294, opened 2019-02-13
NativeStore: Txncache files do not get cleaned up on Windows (file system leak)


aschwarte10:
We encountered a leak in the file system causing _txncache.dat_ files to not be cleaned up. On Windows the deletion of these _temporary_ files fails apparently because the file is still locked. 

In our case we had systems with thousands of these files lying around during the runtime of the application. This caused performance issues in various scenarios (e.g. our backups, cleanup etc)

Note that on Linux / Unix systems the files are getting deleted due to a different file locking mechanism.

Note also that the DB state is consistent and correct from what we can see, i.e. there only these files left in the native store's data dir.

Example unit test to reproduce (only Windows):

`public class NativeStoreTest {

	@Rule
	public TemporaryFolder tempFolder = new TemporaryFolder();

	protected Repository repo;

	protected ValueFactory vf = SimpleValueFactory.getInstance();

	protected IRI ctx = vf.createIRI("http://ex.org/ctx");

	@Before
	public void before()
		throws Exception
	{

		File dataDir = tempFolder.newFolder("dbmodel");
		repo = new SailRepository(new NativeStore(dataDir, "spoc,posc"));
		repo.init();
	}

	@After
	public void after()
		throws Exception
	{
		repo.shutDown();
	}

	@Test
	public void testTxncacheCleanup()
		throws Exception
	{

		/*
		 * Test for issue # On windows the txncacheXXX.dat files did not get properly deleted, as the file is
		 * locked when the deletion is attempted.
		 */

		IRI res = vf.createIRI("http://ex.org/s");

		addStmt(vf.createStatement(res, RDF.TYPE, FOAF.PERSON));

		Statement st1 = vf.createStatement(res, RDFS.LABEL, vf.createLiteral("test"));
		addStmt(st1);

		Statement remStmt = vf.createStatement(res, RDFS.LABEL, vf.createLiteral("test"), ctx);
		removeStmt(remStmt);

		File repoDir = repo.getDataDir();
		System.out.println("Data dir: " + repoDir);

		for (File file : repoDir.listFiles()) {
			System.out.println("# " + file.getName());
		}
		Assert.assertEquals(14, repoDir.listFiles().length);

		// make sure there is no txncacheXXX.dat file
		Assert.assertFalse(Files.list(repoDir.getAbsoluteFile().toPath()).anyMatch(
				file -> file.toFile().getName().matches("txncache[0-9]+.*dat")));

[...]

PR #1295, opened 2019-02-13
#1294: Fix resource leak in txncache file deletion

aschwarte10:
This PR addresses GitHub issue: #1294

The temporary txncache.dat file cannot be deleted on Windows as the file
is still locked while attempting the deletion.

Closing the resource handle properly makes sure that deletion works.

### DIFF
--- a/util/src/main/java/org/eclipse/rdf4j/common/io/NioFile.java
+++ b/util/src/main/java/org/eclipse/rdf4j/common/io/NioFile.java
@@ -137,12 +137,9 @@ public final class NioFile implements Closeable {
 	public boolean delete()
 		throws IOException
 	{
-		try {
-			return file.delete();
-		}
-		finally {
-			close();
-		}
+		// make sure to close file handles prior to deletion
+		close();
+		return file.delete();
 	}
 
 	/**


barthanssens:
Since the PR was based upon develop, I guess it should be cherry-picked as well for master ?

jeenbroekstra:
Yes. You'll need to cherry-pick to master for inclusion in 2.5, and you'll need to cherry-pick _again_ to a new 2.4 release branch for inclusion in a 2.4 hotfix release. I'd advise very strongly against _also_ doing a 2.3 hotfix release again. 

We should make it more clear to contributors that 99 out of a 100 times, they should be basing their PRs off of the master branch, not develop.

barthanssens:
OK.
And indeed, a 2.3 hotfix would be a bit too much, with 2.5 (and then 3.0) coming up, ...

aschwarte10:
@jeenbroekstra  I will make sure to use the "master" branch as a basis for PRs next time. Sorry if this cause inconvenience

jeenbroekstra:
@aschwarte10 no worries, fwiw we changed our branching policy a while ago, and using develop _used_ to be the recommendation. But it's easier to work off master because if we want _we_ can retarget a PR from master to develop, while the other way around is hard because it will include things that should not yet be in the master branch.

Issue #1299, opened 2019-02-19
Random order of conditions in where clause prevents query optimization


reckart:
The where clause uses a QueryElementCollection which by default internally uses a HashSet to store its elements. This causes the order of patterns in the where clause to be random. However, practically it makes a great difference in which order the patterns are provided because not all triple stores are equally able to optimize the query. In order to provide client code with some control on the order of patterns (and also in general on the order of things in a QueryElementCollection), the default HashSet should be replaced with a LinkedHashSet.

PR #1303, opened 2019-02-20
#1299 - Random order of conditions in where clause prevents query optimization

reckart:
This PR addresses GitHub issue: #1299.

Replace default HashSet with a LinkedHashSet - same set behavior but elements remain in the general order in which the user code adds them. This permits the user code to optimize queries by placing fast patterns before slow patterns.


barthanssens:
Hi,

thank you for the patch.

However, the Eclipse Foundation is rather strict about the IPR of contributions, so would it be possible to fill out the Eclipse Contributor Agreement form (you only have to do this once, and you may have to create an Eclipse login), and to signoff the commit (git -s) ?

(see also https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md , step 1)

Thanks

reckart:
@jeenbroekstra Jumped through the IP hoops

barthanssens:
Excellent, thanks

Issue #1308, opened 2019-02-24
STR function missing from SparqlFunctions


reckart:
SparqlFunctions does not declare the STR function.

jeenbroekstra:
Thanks for the fix @reckart ! It will be included in release 2.5.0, scheduled for later this week.

PR #1309, opened 2019-02-24
#1308 - STR function missing from SparqlFunctions

reckart:
This PR addresses GitHub issue: #1308.

- Added STR function

### DIFF
--- a/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/SparqlFunction.java
+++ b/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/SparqlFunction.java
@@ -54,7 +54,7 @@ public enum SparqlFunction implements SparqlOperator {
 	SHA256("SHA256"),
 	SHA384("SHA384"),
 	SHA512("SHA512"),
-    STR("STR"),
+	STR("STR"),
 	STRAFTER("STRAFTER"),
 	STRBEFORE("STRBEFORE"),
 	STRDT("STRDT"),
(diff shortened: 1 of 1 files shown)


sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/SparqlFunction.java
```diff
@@ -54,6 +54,7 @@
 	SHA256("SHA256"),
 	SHA384("SHA384"),
 	SHA512("SHA512"),
+    STR("STR"),
```
jeenbroekstra: Minor thing but can you please fix the indentation? We standardize on tabs-only.

reckart:
Done

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/SparqlFunction.java
reckart: Done

reckart:
Would be good if RDF4J came with Checkstyle such that the IDE can show such problems immediately. 

Given that tabs resolve to different sizes depending on where one views the file, I'd personally also vote for a spaces policy instead.

jeenbroekstra:
@reckart we've got code formatting settings for Eclipse (see the contributor guidelines), but having something independent from the IDE, like checkstyle, is under discussion.

And I'm not getting involved in a tabs vs spaces war except to point out that the official Eclipse code guidelines standardize on tabs 🙂

Issue #1326, opened 2019-03-07
Support for property paths in the SparqlBuilder


reckart:
It would be nice if the SparqlBuilder had at least some basic support for property paths.

PR #1328, opened 2019-03-07
#1326 - Support for property paths in the SparqlBuilder

reckart:
This PR addresses GitHub issue: #1326.

- Very basic property path support "proposal" which already works for my use cases

This should get some refinement and testing before it gets merged. Consider it a proposal to potentially work on together.

This is a PR is a "fork" of the `Path` class I'm developing for https://github.com/inception-project/inception because we need property path support there now.


reckart:
I was thinking if it might be a sensible idea to have a `SparqlBuilder.path(...)` method instead of `Path.of(...)` - in particular given that `Path` is also the name of a Java standard class.

jeenbroekstra:
> I was thinking if it might be a sensible idea to have a `SparqlBuilder.path(...)` method instead of `Path.of(...)` - in particular given that `Path` is also the name of a Java standard class.

Good point, hadn't thought of that. A suggestion: rename the `Path` class you introduced here to `PropertyPath`, and in addition introduce `SparqlBuilder.path` as a convenience shortform.

jeenbroekstra:
add to whitelist

jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
@reckart I have left this open because you mentioned you wanted to do further improvements to it. I'm happy to accept what you have you as-is, but if you want to improve it further soon I'll leave this open.  Keep in mind that we are planning a 2.5.1 patch release for this friday or early next week.

reckart:
@jeenbroekstra I have applied the suggestions you made. Additionally, I have changed the signatures to accept only `RdfPredicate` instead of `QueryElement`. At least for the cases I am using this, it seems to be sufficient. Some thoughts:

* there are no unit tests
* the `PropertyPath` class contains only static methods. Maybe `PropertyPaths` would be a name for the class more in-line with the general naming conventions, e.g. `GraphPatterns`?
* I have added the `SparqlBuilder.path` method as you suggested, but I am not sure it makes too much sense since the user in many cases will find other methods like `zeroOrMore` in `PropertyPath`. So I think it might be more sensible to rename `PropertyPath.of` to `PropertyPaths.path(...)` method and leave this out of `SparqlBuilder`.

reckart:
Actually... I have applied the thoughts from the last two bullet points now.

reckart:
So if you are fine with untested code, I think this looks ok now.

jeenbroekstra:
Yeah, the unit tests for sparqlbuilder in general are missing - this is something we missed when we accepted the original contribution. Improving test coverage for it is on our radar. See #1385

sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/core/PropertyPaths.java
```diff
@@ -0,0 +1,43 @@
+/*******************************************************************************
+Copyright (c) 2019 Eclipse RDF4J contributors.
```
jeenbroekstra: This looks malformed, it's missing the leading asterisk.

jeenbroekstra: Huh. I just noticed that this consistently done wrong in the sparqlbuilder classes. No biggie. But I'll fix it.

jeenbroekstra:
@rdf4j-bot run tests

Issue #1240, opened 2019-01-09
Update code formatting guidelines


jeenbroekstra:
The code formatting guidelines for RDF4J are "enshrined" in an [eclipse formatting config file](https://github.com/eclipse/rdf4j/blob/master/eclipse-settings/rdf4j-formatting.xml). This file is somewhat out of date with newer versions of Eclipse it seems, and occasionally causes weird reformatting issues.

Moreover, not everybody uses Eclipse IDE - some heretics (j/k) prefer IntelliJ or other editors. We should codify the formatting conventions in a way that is easier to transport across multiple tools.

One way to do that is to adopt standard code conventions (like the Oracle java conventions, or the Google conventions) - these are pretty universally supported. If we really feel we need to, we can document minor departures (line length, tabs vs. spaces, etc), and still provides a custom eclipse formatting file for RDF4J - just with fewer departures from common conventions. 

It's more important to have a commonly shared standard than to have something that I (or anyone else) personally find the most pleasing layout. Git diffs would become a lot less messy this way, for one.

barthanssens:
Related to #822

barthanssens:
Admittedly I'm probably one of the worst offenders, being used to the Oracle Java guidelines (way less newlines e.g. no newline before else and the { on the same line as method name... and thus more lines of codes per screen)  _and_ having joined the dark side a.k.a. Netbeans ;-)

I don't mind using the RDF4J coding rules, but it would definitely help if there would be a few good, short examples on how to format the code, without Eclipse. 

(side note: I know there is an Eclipse formating plugin  for Netbeans, but the results where mixed - literally and figuratively)

jeenbroekstra:
Well I'm actually with you on the newlines thing. A lot of these guidelines were chosen when it was just me and Arjohn working on this code base, both in the same office, using the same tools, and still using 4:3 CRT monitors :)

So I'm happy to revisit and adapt. I don't want to get into a whole nit-picky discussion about individual formatting tweaks though, which is why I'd advocate something along the lines of just adopting [Eclipse Coding Conventions](https://wiki.eclipse.org/Coding_Conventions). 

These are basically just Sun/Oracle Java conventions with a few modifications, notably:

- standardize on 4-space tab-only indentation (Sun/Oracle allows mixed indents, that's a mess)
- Eclipse license header guidelines for beginning comment
- some more explicit package/class [naming conventions](https://wiki.eclipse.org/Naming_Conventions)

[...]

Issue #822, opened 2017-05-04
Investigate formatter-maven-plugin for CI builds


ansell:
Investigate adding formatter-maven-plugin to CI builds to validate formatting and fail CI builds (after compile/tests complete) if it has not been applied:

http://code.revelc.net/formatter-maven-plugin/formatter-maven-plugin/plugin-info.html

This will help with issues where the user has accidentally forgot to apply the formatting before submitting their pull request.

ansell:
In particular, what are peoples thoughts on exporting the formatter configuration as an artifact to Maven Central and putting it on a different build number (similar to some parent pom systems where they only use 1/2/3/4/etc.) so it can be consistently used and tracked separate to the rest of RDF4J?

That is the recommended method for using the formatter-maven-plugin with multi module projects:s

http://code.revelc.net/formatter-maven-plugin/formatter-maven-plugin/examples.html#Multimodule_Configuration

barthanssens:
Related to #790

jeenbroekstra:
Fwiw I have previoulsy briefly looked at using the maven checkstyle plugin for this purpose, but got mixed results. It's easy enough to set up but I had a hard time getting it to play nice with the Eclipse built-in code formatter. This plugin looks interesting though, and more immediately suited to our purpose. Link to github repo: https://github.com/revelc/formatter-maven-plugin

jeenbroekstra:
I had a bit of a play with this formatter and I really like it. Easy to integrate in our build process, and also plays nice with Eclipse's internal formatter.

hmottestad:
@jeenbroekstra I tested the formatter on fluent APIs like .stream() and it's a bit aggressive:

```
return m
    .stream()
    .map(st -> st.getObject())
    .filter(o -> o instanceof Literal)
    .map(l -> (Literal) l)
    .findAny();
```

Gets converted to a single line, because it's shorter than 120 characters!

I strongly prefer the multiline version of fluent apis, unless there is a particular need for using a one-liner (eg. in an IF condition).

jeenbroekstra:
The problem with that is that that's not a distinction the formatter can make. I can configure it to forcible _always_ turn qualified invocations like this into multiline statements (even if < 120 chars) but I can't make an exception for "when it is part of an IF-condition". The upshot is that if I change this, we also get this:

	if (book.stringValue()
					.equals("http://example.org/book/book6")) {
				assertEquals("Harry Potter and the Half-Blood Prince", b.getValue("title")
[...]

PR #1340, opened 2019-03-12
Updated code conventions and formatter plugin

jeenbroekstra:
This PR addresses GitHub issue: #822 and #1240 .

* introduced new code conventions based on Eclipse code formatting style (as per #1240), with 120-char linewrapping, unix line endings, and a slight tweak to format qualified invocations (e.g. fluent api calls) in a more readable fashion
* reformatted code base with new conventions.
* Maven formatter plugin added to automatically validate code against the code conventions (#822). it will fail the build if formatting is not compliant. `mvn formatter:format` will reformat existing code to the configured conventions if necessary.

Note that the supplied xml file is an eclipse formatting file, and can therefore be imported in Eclipse to use the IDE formatter directly. 

I'd like to get this out of the way as soon as possible - the master and develop branch are currently not much different, which means merging this change over both branches will be less difficult (but still a bit painful). The longer we wait, the more awkward this becomes.

Have a look at the new code formatting. I'm reluctant to diverge from the Eclipse conventions too much, but if there's anything standing out that definitely needs fixing, we'd better fix it now.


jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
Hm, it appears my clever trick to have a relative reference to the xml file is giving jenkins trouble. Odd as it works perfectly on my machine. Running a test to see if it has to do with the maven version.

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
```diff
@@ -82,48 +78,38 @@ public void testBindings1()
...
+		assertEquals("http://example.org/book/book6", handler.getBindingSets().get(0).getValue("book").stringValue());
 
 		for (BindingSet b : handler.getBindingSets()) {
 			assertNotNull(b.getValue("book"));
 			assertNotNull(b.getValue("title"));
 			assertTrue(b.getValue("book") instanceof IRI);
 			assertTrue(b.getValue("title") instanceof Literal);
 
-			IRI book = (IRI)b.getValue("book");
+			IRI book = (IRI) b.getValue("book");
```
hmottestad: This looks better

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
```diff
@@ -82,48 +78,38 @@ public void testBindings1()
 		assertEquals(7, handler.getBindingSets().size());
 
 		// Results are ordered, so first should be book6
-		assertEquals("http://example.org/book/book6",
-				handler.getBindingSets().get(0).getValue("book").stringValue());
```
hmottestad: I would prefer if it breaks on all the arguments to the method, rather than arbitrarily on one.

Eg. like this:

```
assertEquals(
    "http://example.org/book/book6",
    handler.getBindingSets().get(0).getValue("book").stringValue()
);
```

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
```diff
@@ -82,48 +78,38 @@ public void testBindings1()
...
 			assertTrue(b.getValue("book") instanceof IRI);
 			assertTrue(b.getValue("title") instanceof Literal);
 
-			IRI book = (IRI)b.getValue("book");
+			IRI book = (IRI) b.getValue("book");
 			if (book.stringValue().equals("http://example.org/book/book6")) {
 				assertEquals("Harry Potter and the Half-Blood Prince", b.getValue("title").stringValue());
-			}
-			else if (book.stringValue().equals("http://example.org/book/book7")) {
+			} else if (book.stringValue().equals("http://example.org/book/book7")) {
```
hmottestad: I like this.

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
```diff
@@ -82,48 +78,38 @@ public void testBindings1()
...
 			}
 		}
 
 	}
 
 	@Test
-	public void testBindings2()
-		throws Exception
-	{
+	public void testBindings2() throws Exception {
```
hmottestad: This is a lot cleaner too. The above is actually how the old eclipse one used to be.

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLServiceEvaluationTest.java
```diff
@@ -50,8 +50,8 @@
 import org.slf4j.LoggerFactory;
```
hmottestad: Are there any `import` rules in the style guide?

jeenbroekstra:
> Can we change the wrap width to 180? I have around a 220 character wide windows when I program. If I use fullscreen I have around 310.

I think we need to cater to a lowest common denominator/accepted conventions here. We're actually already going out on a limb a bit with 120 since most coding guidelines still (!) prescribe 80 columns. I also don't think that going wider will necessarily improve readability: it's one thing to have the screen real-estate, but it also needs to be code that is easy to read and understand. 

Moreover if we changed it to 180 or more, I would personally have to scroll horizontally all the time : my laptop monitor is not that hi-res...

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
barthanssens: indeed, one line of scarce screen real estate saved :-)

barthanssens:
> Can we change the wrap width to 180? I have around a 220 character wide windows when I program.
> If I use fullscreen I have around 310.

You have very good eyesight and/or a nice monitor over there, I barely get 150 on my machine (15" 
 laptop)...
So if you don't mind I'd also go for 120...

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
jeenbroekstra: It's likely that our previous code conventions inherited it from that. I personally kinda prefer having the throws clause on a new line, but I'm oldfashioned that way.

barthanssens: No strong feelings about the Exception.

I'm happy as long as the '{' isn't sitting lonely on a new line, that is really a waste of space IMHO

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLServiceEvaluationTest.java
jeenbroekstra: I believe it has settings to add empty lines between groups. But other things are not set in the current xml file. For example, we have a guideline to not use wildcard imports but this is not enforced by this plugin.

hmottestad:
I think we will need to be able to mark a section as "don't reformat", like the javadocs I have in the ShaclSail.

jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
> I think we will need to be able to mark a section as "don't reformat", like the javadocs I have in the ShaclSail.

That's actually catered for:


//@formatter:off
custom-formatted stuff
//@formatter:on

This should be used only in special circumstances though.

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
jeenbroekstra: Fwiw in this particular case I kind of agree with you, but the way it currently wraps method arguments is "where necessary", which is how the Eclipse conventions define it. I had a quick look to change this but modifying this would require tweaking several wrapping settings ( = deviating further from the sensible defaults) and even then I can't quite get it how you prefer.

So I think I'll leave this as-is and live with the compromise.

jeenbroekstra:
Alright if there are no further issues I'll merge this in (Jenkins is now happy as well) and I will also immediately try and sync the develop branch. I'll try and get to the other repos later in the week.

compliance/queryresultio/src/test/java/org/eclipse/rdf4j/query/resultio/sparqljson/SPARQLJSONTupleBackgroundTest.java
hmottestad: ok

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -65,31 +64,30 @@ public static Value anyObject(Model m) {
...
-	 * 		the model from which to retrieve an object Literal value.
-	 * @return an object Literal value from the given model, or {@link Optional#empty()} if no such value
-	 * exists.
+	 * @param m the model from which to retrieve an object Literal value.
+	 * @return an object Literal value from the given model, or {@link Optional#empty()} if no such value exists.
 	 */
 	public static Optional<Literal> objectLiteral(Model m) {
-		return m.stream().map(st -> st.getObject()).filter(o -> o instanceof Literal).map(
-				l -> (Literal)l).findAny();
+		return m.stream().map(st -> st.getObject()).filter(o -> o instanceof Literal).map(l -> (Literal) l).findAny();
```
hmottestad: Here is an example of the formatter fixing where the old format did a line break in the wrong place.

Question. What happens if I format it myself as: 

```
return m
    .stream()
    .map(st -> st.getObject())
    .filter(o -> o instanceof Literal)
    .map(l -> (Literal) l)
    .findAny();

```

Will the formatter turn that back into a single line?

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -65,31 +64,30 @@ public static Value anyObject(Model m) {
...
 	 * @see Model#objects()
 	 */
 	public static Set<Literal> objectLiterals(Model m) {
-		return m.stream().map(st -> st.getObject()).filter(o -> o instanceof Literal).map(
-				l -> (Literal)l).collect(Collectors.toSet());
+		return m.stream()
+				.map(st -> st.getObject())
+				.filter(o -> o instanceof Literal)
+				.map(l -> (Literal) l)
+				.collect(Collectors.toSet());
```
hmottestad: this looks so much better :)

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
hmottestad: @jeenbroekstra I was just a couple of seconds late. Do you know how the above would work?

jeenbroekstra: It will _only_ turn it back into a single line if it would actually fit in 120 chars. Otherwise it would reformat to something _almost_ identical to what you have here. This is actually the case I was talking about when I mentioned "qualified invocations": it's been modified to put each successive call on a new line (at least if it needs linebreaks), instead of just chaining the whole thing.

jeenbroekstra: I just gave it a quick go and how it formats is as follows. Your original line is < 120 chars so it would just fit it on one line:

    return m.stream().map(st -> st.getObject()).filter(o -> o instanceof Literal).map(l -> (Literal) l).findAny();


But if we tweak it slightly to make it too long, it would reformat like this:

		return modelWithAVeryLongName.stream()
				.map(st -> st.getObject())
				.filter(o -> o instanceof Literal)
				.map(l -> (Literal) l)
				.findAny();

hmottestad: I would like it to be a one way thing. Long one-liners get split, but already multiline formatted fluent api stays multiline.

Issue #389, opened 2016-09-07
Remove deprecated Graph


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-2077)

The Graph API was deprecated in Sesame 2.7.0 and will be removed in RDF4J 3.0, being then fully replaced by the Model API.

classes and interfaces removed:

- `org.eclipse.rdf4j.model.Graph` 
- `org.eclipse.rdf4j.model.impl.GraphImpl`
- `org.eclipse.rdf4j.model.util.GraphUtil`
- `org.eclipse.rdf4j.model.util.GraphUtilException`

PR #1406, opened 2019-05-09
GH-389 removed Graph and related classes/interfaces

jeenbroekstra:
This PR addresses GitHub issue: #389 .

* removed `Graph`, `GraphImpl`, `GraphUtil` and `GraphUtilException`.
* modified several places where this stuff was still used


rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFHandlerWrapper.java
```diff
@@ -42,8 +47,22 @@
 	 *                    <tt>null</tt>.
 	 */
 	public RDFHandlerWrapper(RDFHandler... rdfHandlers) {
+		this(new TreeModelFactory(), rdfHandlers);
```
hmottestad: Does the RDFHandlerWrapper need a sorted Model? If not it should be swapped for the hash model.

model/src/main/java/org/eclipse/rdf4j/model/impl/GraphImpl.java
```diff
@@ -1,173 +0,0 @@
...
- * @deprecated since release 2.7.0. Use a {@link org.eclipse.rdf4j.model.Model} implementation (e.g. {@link TreeModel}
- *             or {@link LinkedHashModel} instead.
- * @author Arjohn Kampman
- */
-@Deprecated
-public class GraphImpl extends AbstractCollection<Statement> implements Graph {
-
-	private static final long serialVersionUID = -5307095904382050478L;
-
-	protected LinkedList<Statement> statements;
```
hmottestad: GraphImpl was using a LinkedList, which is usually slow compared to an ArrayList for inserts until the List grows very large (gigabytes). 

The Model implementations use either a TreeSet or a HashSet approach. I think both of these are slower than both LinkedList and ArrayList for workloads with mostly inserts. 

How does the workload compare for the places where you are now switching out GraphImpl for one of the Model implementations?

rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFHandlerWrapper.java
jeenbroekstra: That would depend on the type of wrapper I think. This is just a default, specific implementations can override (ideally, it should offer a way to inject as well, but that goes a bit beyond the scope of this ticket).

model/src/main/java/org/eclipse/rdf4j/model/impl/GraphImpl.java
jeenbroekstra: No idea, I haven't done any performance tests on this. 

It wouldn't be hard to introduce a new list-based Model implementation to facilitate quicker inserts, if there's a need for that.

rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFHandlerWrapper.java
hmottestad: Should go with the LinkedHashModelFactory instead then. Since it's usually faster.

Issue #1409, opened 2019-05-10
Unnecessary parsing of IRIs every time an RDF parser is used


avataar:
Every time a parser is used AbstractRDFParser.resolveURI() uses new ParsedIRI() to determine if the IRI is relative. This is an expensive operation and has a noticeable overhead when many IRIs need to be created. A simpler check such as indexOf(':') will suffice to determine if something is a relative IRI.

avataar:
Sorry for that. It seems I only ran "test" instead of "verify". I've modified the test to match the new behaviour, which I believe is very much expected since the IRI instance is created directly via SimpleValueFactory. The previous behaviour was a side effect of passing the string through new ParsedIRI(string).toString() before calling the value factory.

PS I can't seem to run the full integration suite. Tests that rely on HTTP fail with java.lang.NoSuchMethodError: org.springframework.beans.factory.config.ConfigurableListableBeanFactory.hasEmbeddedValueResolver()Z. I'm not sure why that is.

Issue #780, opened 2017-03-13
IRI with space


skodapetr:
File ```input.ttl``` contains IRI with encoded space:
```
_:b25978837	a <http://purl.bioontology.org/ontology/UATC/\u0020SERINE\u0020\u0020> .
```
The Turtle parser is able to read this IRI (with encoded spaces) as a valid IRI.  
However Turtle writer does not encode the space on  output, resulting in:
```
_:genid-ca45343d5be841429acf1ca16a6fe928-b25978837 a <http://purl.bioontology.org/ontology/UATC/ SERINE  > .
```
When we try to load this statement the parser fail with:
```
Exception in thread "main" org.eclipse.rdf4j.rio.RDFParseException: IRI included an unencoded space: '32' [line 2]
	at org.eclipse.rdf4j.rio.helpers.RDFParserHelper.reportError(RDFParserHelper.java:322)
	at org.eclipse.rdf4j.rio.helpers.AbstractRDFParser.reportError(AbstractRDFParser.java:601)
...
```
A simple program can be utilized to reproduce this error:
```
    public static void main(String[] args) throws Exception {
        try (InputStream input = new FileInputStream(new File("input.ttl"));
                OutputStream output = new FileOutputStream(new File("output.ttl"))) {
            RDFParser parser = Rio.createParser(RDFFormat.TURTLE);
            parser.setRDFHandler(Rio.createWriter(RDFFormat.TURTLE, output));
            parser.parse(input, "http://localhost/base/");
        }
        try (InputStream input = new FileInputStream(new File("output.ttl"))) {
            RDFParser parser = Rio.createParser(RDFFormat.TURTLE);
            parser.parse(input, "http://localhost/base/");
        }
    }
```
This problem also propagates to the querying remote SPARQL endpoints.

The same issue can be experienced when using N-Triples, however for JSON-LD and RDF/XML no error is given. 
The comment  in [TurtleParser](https://github.com/eclipse/rdf4j/blob/master/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java#L921) suggest that the \n and similar (spaces?) should be handled as errors.

It's not clear what is an intended behaviour.

Tested with rdf4j 2.2 .

barthanssens:
I assume this is related to #69 (and that there is no strict checking on generating/writing IRIs because it impacts performance)

jeenbroekstra:
Also related to #712

[...]

PR #1410, opened 2019-05-10
GH-1409: Unnecessary parsing of IRIs every time an RDF parser is used

avataar:

This PR addresses GitHub issue: #1409

Also fixes an issue in AbstractRDFParser.createURI() where non-fatal validation of IRIs didn't return null (this was masked by old code in resolveURI()).

### DIFF
--- a/compliance/rio/src/test/java/org/eclipse/rdf4j/rio/turtle/CustomTurtleParserTest.java
+++ b/compliance/rio/src/test/java/org/eclipse/rdf4j/rio/turtle/CustomTurtleParserTest.java
@@ -440,7 +440,7 @@ public class CustomTurtleParserTest {
 		assertEquals(1, model.size());
 		model.filter(null, RDF.TYPE, null)
 				.objects()
-				.forEach(obj -> assertEquals("http://purl.bioontology.org/ontology/UATC/%20SERINE%20%20",
+				.forEach(obj -> assertEquals("http://purl.bioontology.org/ontology/UATC/ SERINE  ",
 						obj.stringValue()));
 	}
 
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
@rdf4j-bot whitelist

jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
Thanks for the fix @avataar. It appears a regression test is failing. There may be a corner case overlooked here.

compliance/rio/src/test/java/org/eclipse/rdf4j/rio/turtle/CustomTurtleParserTest.java
```diff
@@ -440,8 +440,7 @@ public void test780IRISpace() throws Exception {
 		assertEquals(1, model.size());
 		model.filter(null, RDF.TYPE, null)
 				.objects()
-				.forEach(obj -> assertEquals("http://purl.bioontology.org/ontology/UATC/%20SERINE%20%20",
-						obj.stringValue()));
+				.forEach(obj -> assertEquals("http://purl.bioontology.org/ontology/UATC/ SERINE  ", obj.stringValue()));
```
jeenbroekstra: This changes the test case to conform to the changed behaviour, and will result in syntactically illegal IRIs being produced. Looking at the original fix that introduced this test case (#780, see PR #839 ) the %-encoding of encoded spaces when syntax validation is disabled was a deliberate feature.  Are you quite sure you're correct to modify this test?

jeenbroekstra: Looking a bit further, the original issue complained about the round-trip: the writer serializing such IRIs with the spaces un-encoded. I've just verified that the TurtleWriter has been fixed to deal with that. So I guess the motivation for this test case (and the behavior that it tested) has changed, and you may be right that this is the correct behavior.

Issue #1357, opened 2019-03-23
SHACL in workbench?


robertbbb:
Hi, Very happy to see SHACL coming into RDF4J. However, the documentation does not make it clear how SHACL support must be activated in the server and workbench. Is it part of every type of repository? Will it replace the SPIN support?

hmottestad:
@jeenbroekstra will know what needs to be added to make it work in the server and workbench.

SHACL does not replace SPIN, at least not all of it. SHACL supports restrictions on your data, but not arbitrary restrictions like SPIN does with SPARQL. This makes SHACL a lot faster, but also more restrictive. The SHACL engine is incremental, so it works best if you have a large database and make small changes to it. If your database is 100 triples or 100 million triples doesn't matter much for performance.

In the next release of RDF4J (3.0), SHACL will get even faster and support more of the SHACL standard (currently passing 46/98 compliance tests). In the works is also support for modifying the loaded SHACL shapes. 

SPIN is also a lot faster in 3.0, you should try out the 3.0-SNAPSHOT build if you use SPIN a lot!

hmottestad:
https://github.com/eclipse/rdf4j/issues/1262

hmottestad:
Previous discussions on SPIN vs SHACL: https://github.com/eclipse/rdf4j/issues/1262

hmottestad:
@jeenbroekstra

jeenbroekstra:
There's several possible ways we could enable SHACL support in the workbench. The simplest way is just to offer it as one possible sail configuration for a new repository. We'd need variants on top of a memory store, native store, and possibly combined with an RDFS inferencer as well?   

Another option is as a standalone tool inside the workbench - basically an option to quickly load a file, check it against a schema to see if it's consistent. This will be a lot more work to build though.

hmottestad:
Two new options in the workbench:
 - memory-shacl
 - native-shacl

Shapes need to be "added" to the SHACL context. You have to use the "add" feature and not do it with SPARQL. 

This will all be available in the 3.0 release.

PR #1415, opened 2019-05-17
Issues/1357 shacl workbench

hmottestad:
This PR addresses GitHub issue: #1357 .

* turtle config files for native-shacl and memory-shacl
* 
* 

Related:
 * https://github.com/eclipse/rdf4j-tools/pull/132
 * https://github.com/eclipse/rdf4j-storage/pull/240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ dependency-reduced-pom.xml
 **/nb-configuration.xml
 **/nbactions*.xml
 *.jfr
+compliance/repository/overlays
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
Hm. I just had a look at the build failures and I believe Jenkins has issues when simultaneously running two jobs that both spin up a test server on port 18080. Looks like jobs are not fully containerised as I thought they were. We'll need to pick different port numbers, or look into configuring Jenkins to run fully independent job agents.

hmottestad:
I’ve run into that before. Jenkins is not very well isolated. Travis is much better in that way. If we used runtime random ports it shouldn’t be an issue though.

Issue #1421, opened 2019-05-22
Upgrade japicmp-maven-plugin


hmottestad:
Upgrade japicmp-maven-plugin to 0.14.0 because this one is compatible with Java 11.

hmottestad:
relates to https://github.com/eclipse/rdf4j/issues/1148

barthanssens:
To be checked if there is a need to be a CQ for the japicmp or not (IIRC there was a mail about maybe putting all build-tools in one request)

hmottestad:
I merged the one on RDF4J storage already.

PR #1422, opened 2019-05-22
Issues/1421 upgrade japicmp

hmottestad:
This PR addresses GitHub issue: #1421 .

* upgrade japicmp
* include jaxb dependencies
*


hmottestad:
I reckon this will require CQ.

barthanssens:
Not sure if a CQ is actually required (only used in building), and I didn't find a CQ for the current 0.11 version in IPZilla (@jeenbroekstra  ?)

hmottestad:
@barthanssens "include jaxb dependencies" those might need a CQ.

barthanssens:
Good point, I'll create a request for jaxb in IPzilla

barthanssens:
CQ20163 created (PB) for JAXB 2.3.x

hmottestad:
So we are good to go then?

barthanssens:
Not sure about the japicmp though, I _think_ it does not require a CQ, because it's not distributed with RDF4j, just used as a dependency

barthanssens:
So far I've only created a request for javax.xml.bind/jaxb-api 2.3, not for jaxb-core/jaxb-impl

On my Debian ARM system, running `mvn install` of the rdf4j subproject actually works without jaxb-core and jaxb-impl.

But it does choke on the argLine `--add-modules java.xml.bind`, so limiting that argument to JDK 9 and 10 gets rid of the `Error occurred during initialization of boot layer` fork error

Have to double check (e.g. on Intel, and to verify if there are no traces of older JDKs on my system), I assume you are using OpenJDK 11 on Intel ?

hmottestad:
Yes. 11 on intel.

barthanssens:
FWIW, there is an Eclipse licensed JAXB: jakarta.xml.bind-api

jeenbroekstra:
> Not sure about the japicmp though, I _think_ it does not require a CQ, because it's not distributed with RDF4j, just used as a dependency

For maven plugins, we don't usually log CQs. I am seeking clarification on the incubation ML whether this is necessary (if it turns out it is, that's on me - but I kind of expect that it isn't).

barthanssens:
May I suggest to use v2.3.2 ?
See  https://github.com/eclipse-ee4j/jaxb-ri#maven-artifacts

Eclipse is now maintaining the JAXB, so that might make it easier in the future

```
<dependency>
    <groupId>jakarta.xml.bind</groupId>
    <artifactId>jakarta.xml.bind-api</artifactId>
</dependency>

<!-- Runtime -->
<dependency>
    <groupId>org.glassfish.jaxb</groupId>
    <artifactId>jaxb-runtime</artifactId>
</dependency>
```

hmottestad:
@barthanssens go for it! Make any changes you want to this branch ;)

hmottestad:
I've switched this to the older jaxb imoplementation:

```
		<dependency>
 			<groupId>javax.xml.bind</groupId>
 			<artifactId>jaxb-api</artifactId>
 			<version>${jaxb.version}</version>
 		</dependency>
```

Since this is the one already approved for the CQ, and the one used in rdf4j-storage.

It looks like everything builds and all tests run on jdk 11 on my laptop. So I will merge this now.

Issue #439, opened 2016-09-07
toString method in DatasetImpl still with "USING"


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-1852)

I noticed that the {{toString()}} method in {{org.openrdf.query.impl.DatasetImpl}} still returns a value starting with "USING"

However the parser no longer accepts "USING" as an alternative to "FROM".
Version 2.6.3 did.

PR #1436, opened 2019-05-29
Changed USING to FROM

barthanssens:
This PR addresses GitHub issue: #439  .

* Use "FROM" and "FROM NAMED" instead of "USING"

I assume this is the only changed that is needed, since the (old) ticket mentions it is a "good first issue", so I guess this is the only trivial change that was required ?


jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
@barthanssens there are a number of failing tests in the JsonLdParser - I don't see how that could be caused by this PR though.

jeenbroekstra:
Ah, in fact the master branch is currently unstable. Looks like this was caused by the jsonld version bump. See https://ci.eclipse.org/rdf4j/job/rdf4j-client-master-verify/86/

barthanssens:
My bad, I’ll take a look at it…

barthanssens:
@rdf4j-bot run tests

Issue #1444, opened 2019-06-08
Abstract classes with public constructors


woutermont:
Not really an issue, but seems bad practice to declare public constructors for abstract classes. They behave like protected constructors anyway, so they might as well be named so.

Following classes contain one (or so a quick search would have me believe):
```
org.eclipse.rdf4j.OpenRDFException
org.eclipse.rdf4j.common.iteration.ConvertingIteration
org.eclipse.rdf4j.common.iteration.DelayedIteration
org.eclipse.rdf4j.common.iteration.ExceptionConvertingIteration
org.eclipse.rdf4j.common.iteration.FilterIteration
org.eclipse.rdf4j.common.iteration.IterationWrapper
org.eclipse.rdf4j.common.iteration.LookAheadIteration
org.eclipse.rdf4j.common.iteration.QueueIteration
org.eclipse.rdf4j.common.iteration.QueueIteration
org.eclipse.rdf4j.common.iteration.TimeLimitIteration
org.eclipse.rdf4j.common.iterator.DelayedIterator
org.eclipse.rdf4j.common.iterator.LookAheadIterator
org.eclipse.rdf4j.http.client.query.AbstractHTTPQuery
org.eclipse.rdf4j.http.client.query.AbstractHTTPUpdate
org.eclipse.rdf4j.model.impl.AbstractValueFactory
org.eclipse.rdf4j.model.impl.FilteredModel
org.eclipse.rdf4j.query.algebra.AbstractAggregateOperator
org.eclipse.rdf4j.query.algebra.AggregateOperatorBase
org.eclipse.rdf4j.query.algebra.BinaryTupleOperator
org.eclipse.rdf4j.query.algebra.BinaryValueOperator
org.eclipse.rdf4j.query.algebra.CompareSubQueryValueOperator
org.eclipse.rdf4j.query.algebra.NAryValueOperator
org.eclipse.rdf4j.query.algebra.SubQueryValueOperator
org.eclipse.rdf4j.query.algebra.UnaryTupleOperator
org.eclipse.rdf4j.query.algebra.UnaryValueOperator
org.eclipse.rdf4j.query.parser.ParsedOperation
org.eclipse.rdf4j.query.parser.ParsedQuery
org.eclipse.rdf4j.query.parser.serql.ast.ASTBooleanExpr
org.eclipse.rdf4j.query.parser.serql.ast.ASTGraphQuery
org.eclipse.rdf4j.query.parser.serql.ast.ASTGraphQuerySet
[...]

PR #1445, opened 2019-06-08
Issues/#1444 abstract classes protected constructors

woutermont:
This PR addresses GitHub issue: #1444 .

*  Changes public constructors to protected constructors in all abstract classes
   * Commit `7cc9870` contains all changes to non-deprecated classes
   * Commit `9e1dc43` contains all changes to deprecated classes, should these changes not be needed/wanted.

### DIFF
--- a/http/client/src/main/java/org/eclipse/rdf4j/http/client/query/AbstractHTTPQuery.java
+++ b/http/client/src/main/java/org/eclipse/rdf4j/http/client/query/AbstractHTTPQuery.java
@@ -31,7 +31,7 @@ public abstract class AbstractHTTPQuery extends AbstractQuery {
 
 	protected final String baseURI;
 
-	public AbstractHTTPQuery(SPARQLProtocolSession httpClient, QueryLanguage queryLanguage, String queryString,
+	protected AbstractHTTPQuery(SPARQLProtocolSession httpClient, QueryLanguage queryLanguage, String queryString,
 			String baseURI) {
 		super();
 		this.httpClient = httpClient;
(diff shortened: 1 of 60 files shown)


woutermont:
Sorry for the mess! Was working on another system, where I sign off with my institutional mail instead of my personal one.

http/client/src/main/java/org/eclipse/rdf4j/http/client/query/AbstractHTTPQuery.java
```diff
@@ -31,7 +31,7 @@
 
 	protected final String baseURI;
 
-	public AbstractHTTPQuery(SPARQLProtocolSession httpClient, QueryLanguage queryLanguage, String queryString,
+	protected AbstractHTTPQuery(SPARQLProtocolSession httpClient, QueryLanguage queryLanguage, String queryString,
```
jeenbroekstra: I haven't checked - does this influence how this is shown in the Javadoc?

woutermont:
> This seems more a matter of taste than really bad practice to me - YMMV. I'm happy to change it though, since as you say the actual behavior doesn't change.

It's up to you.

http/client/src/main/java/org/eclipse/rdf4j/http/client/query/AbstractHTTPQuery.java
woutermont: Indeed, it would have been unhandy if it stopped showing up in the javadoc, but it still does.

hmottestad:
The only issue I see down the line with this is the slight confusion when a new abstract class gets a public constructor and we all start wondering why the others are protected.

woutermont:
Well, we should take care not to (i.e. put it in guidelines). If you think it's really just a matter of taste, maybe it's better to just leave them public. I.m.o. it's still a (minor) bad practice, though. After all, if it acts like a duck, better call it a duck.

hmottestad:
Can our code formatter force this @jeenbroekstra ?

jeenbroekstra:
I doubt it. We'll have to keep ourselves honest.

jeenbroekstra:
I'm fine with merging this in and just keeping an eye on us being consistent in the future. Fwiw we _already_ have several abstract classes with protected constructors in the code base, so it's not as if we are being consistent right now...

Issue #1473, opened 2019-07-01
.remove() method optimisation on Model


hmottestad:
.remove() method is optimized in LinkedHashModel to no-op if the model is empty. This should be the case for all Models, including the TreeModel and the one that supports overflow to disk.

hmottestad:
The overflow to disk model uses the TreeModel as a base, so in the case that there is no overflow the remove() method will use the one from TreeModel. If it overflows to disk the remove() method does not need to no-op since it is obviously not empty.

PR #1474, opened 2019-07-01
issues/1473 check if model is empty before continuing with remove()

hmottestad:
This PR addresses GitHub issue: #1473 .

* check if model is empty before continuing with remove() method, for performance
* 
*

### DIFF
--- a/model/src/main/java/org/eclipse/rdf4j/model/impl/LinkedHashModel.java
+++ b/model/src/main/java/org/eclipse/rdf4j/model/impl/LinkedHashModel.java
@@ -168,23 +168,6 @@ public class LinkedHashModel extends AbstractModel {
 	}
 
 	@Override
-	public boolean remove(Object o) {
-		if (o instanceof Statement) {
-			if (statements.isEmpty()) {
-				return false;
-			}
-
-			Iterator iter = find((Statement) o);
-			if (iter.hasNext()) {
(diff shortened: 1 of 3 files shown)


model/src/main/java/org/eclipse/rdf4j/model/impl/LinkedHashModel.java
```diff
@@ -167,23 +167,6 @@ public void clear() {
 		statements.clear();
 	}
 
-	@Override
-	public boolean remove(Object o) {
```
hmottestad: removed this because it's implemented in AbstractModel.java.

jeenbroekstra: I notice the implementation is different though. I haven't looked in detail but is the way AbstractModel performs a remove as efficient as the more specific LinkedHashModel implementation?

hmottestad: My understanding is that is just converts the Statements into its parts (subject, predicate...) and calls the other remove(...) method with those parts. 

I will double check though.

hmottestad: Double checked. Essentially identical.

Issue #1441, opened 2019-06-05
Stack overflow in Models.matchModels


ArkanEmre:
Hi everyone, 

I wanted to compare two RDF Datasets with `Models.isomorphic` function. They both have around 50000 Statements, unfortunately this triggers a stack overflow due to a recursive call in the Models class. [Here](https://github.com/eclipse/rdf4j/blob/82d52a34d60e58ff9b3cda2691e2664682f912be/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java#L621)  is the link to the line where the recursive call happens.

It is also commented by the developer that it should be fixed. Do you know of any solutions?

Best regards,
Emre

jeenbroekstra:
I haven't looked in detail but generally speaking the solution would be to rewrite as an iterative algorithm. A common way to do that is to introduce an explicit stack (instead of relying on the programming stack).

hmottestad:
I've had huge issues with `Models.isomorphic` in RDF4J. In my previous project we used RDF4J everywhere except for isomorphic. Here we converted our model to a JSON-LD string and parsed it into Jena and ran isomorphic there. This was considerably faster.

jeenbroekstra:
I don't think a performance refactor would be particularly difficult. I'll try and take a stab at it over the weekend.

jeenbroekstra:
>  I'll try and take a stab at it over the weekend.

Ok so that didn't happen. If anyone else wants to have a go be my guest :)

hmottestad:
I've created some benchmarks for us for reference: https://github.com/eclipse/rdf4j/pull/1476

All based on files I made or found in other places. The `spin-full-forwardchained.ttl` file actually triggers the stack overflow, so that's good.

hmottestad:
I found another case of stack overflow. A lot of properties seems to be the trigger. So now we have two stack overflow cases and a bunch of other benchmarks. Here are the results on my ageing macbook pro.

```
# Run complete. Total time: 00:05:02

Benchmark                                 Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes            avgt   10    0.614 ±  0.007  ms/op
IsomorphicBenchmark.bsbm                  avgt   10  810.974 ± 60.225  ms/op
IsomorphicBenchmark.empty                 avgt   10    0.132 ±  0.002  ms/op
IsomorphicBenchmark.internallyIsomorphic  avgt   10   43.071 ±  0.833  ms/op
[...]

PR #1476, opened 2019-07-05
Issues/1441 models isomorphic benchmarks

hmottestad:
This PR addresses GitHub issue: #1441 .

* benchmarks for model isomorphic
* 
*


hmottestad:
@jeenbroekstra could you take a quick look so I can merge this?

jeenbroekstra:
Looks good in principle, though I still have a hard time following how to use these benchmark tests in anger. You occasionally post numbers but I must admit I don't really understand what those numbers tell me other than "on my machine, running test X goes at speed Y". Unless I have a trend or a comparison to a previous run that's not very helpful. 

Remind me again: if we run a normal verification build, these benchmark tests will not be automatically executed, correct? 

Sorry if I'm being a bit hazy, I've been swamped at work and my head is all over the place ATM.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
```diff
@@ -0,0 +1,123 @@
...
+
+	@Benchmark
+	public void blankNodes() throws IOException {
+
+		isomorphic("blankNodes.ttl");
+
+	}
+
+	@Benchmark
+	public void shacl() throws IOException {
```
jeenbroekstra: Could you add a short javadoc comment that explains what each test actually does? E.g. "check performance on a typical SHACL shape with N blank nodes in a (chain/spread/tree/whatever) pattern". For some it's obvious but for others (e.g. the sparql test case) I'm not sure what you're actually testing.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	private void isomorphic(String name) throws IOException {
+		Model m1 = getModel(name);
+		Model m2 = getModel(name);
+
+		boolean isomorphic = Models.isomorphic(m1, m2);
+		if (!isomorphic) {
+			throw new IllegalStateException("Not isomorphic");
```
jeenbroekstra: Would it be helpful to also have a few test cases for two models that are _not_ isomorphic? 

We can add those later and then adapt this I guess.

hmottestad:
I use this plugin for IntelliJ: https://plugins.jetbrains.com/plugin/7529-jmh-plugin

Then you can right click on a benchmark (or the whole file) and run it. 

My typical approach is to make some benchmarks to identify things I think are slow. Then run those benchmarks (remembering to turn the fans up, stop time machine from backing up and stop any other cpu or io intensive tasks). I’ll store the numbers for later, so I can see if my future code changes actually make any difference. 

If I wonder what’s going on, then I’ve set up the second @Fork line to run java flight recorder. If you want to use this it’s important to only run a single benchmark (and not the entire file). It will generate a file that you can open in Java Mission Control to see things like hot methods. 

My laptop is fairly old (6 years) and I’ve noticed that a work laptop I used to have which was newer would sometimes show that certain changes made more of less difference. This was due to things like differences in cpu caching and support for vector operations (SSE and AVX). I would love to have a set of machines with new and old consumer processors and some server processors like Xeon to do a broader benchmark. Would also be amazing if someone made a product like Jenkins or Travis but for benchmarks instead of tests.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
hmottestad: Would be helpful.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
hmottestad: I can do that.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
hmottestad: Would also help to benchmark early termination. Eg. when two models are obviously different. I wonder if we can do something simple like m1.size() != m2.size()?

hmottestad:
After I made the new benchmarks to test for cases were two models were not isomorphic I found out that some of my benchmarks were bound by IO and how fast it could parse RDF and create a model. So I extracted that into fields, so now we only benchmark isomorphic(...).

```
# Run complete. Total time: 00:07:23

Benchmark                                 Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes            avgt   10    0.201 ±  0.088  ms/op
IsomorphicBenchmark.bsbm                  avgt   10   23.595 ±  0.400  ms/op
IsomorphicBenchmark.bsbmArrayList         avgt   10  310.037 ±  8.122  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic     avgt   10   25.226 ±  0.796  ms/op
IsomorphicBenchmark.bsbmTree              avgt   10  135.948 ±  3.899  ms/op
IsomorphicBenchmark.empty                 avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.emptyNotIsomorphic    avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.internallyIsomorphic  avgt   10   30.016 ±  0.545  ms/op
IsomorphicBenchmark.list                  avgt   10    5.872 ±  0.453  ms/op
IsomorphicBenchmark.longChain             avgt   10    0.016 ±  0.001  ms/op
IsomorphicBenchmark.manyProperties2       avgt   10  180.771 ± 48.912  ms/op
IsomorphicBenchmark.shacl                 avgt   10    2.543 ±  0.052  ms/op
IsomorphicBenchmark.sparqlTestCase        avgt   10    0.041 ±  0.001  ms/op
```

Here you can also see that `bsbm` (which uses a LinkedHashModel) is considerably faster than `bsbmTree` which uses a TreeModel. I have yet to see a case were TreeModel is faster than LinkedHashModel. Btw. `bsbmArrayList` uses an arraylist to call isomorphic(...) which gets converted to a TreeModel by that method.

I need to make some benchmarks for loading data into LinkedHashModel vs. TreeModel to see which is faster. Maybe LinkedHashModel will get slower once it passes 1-2 GB, since it will need to rehash to grow.

hmottestad:
I've now implemented an ArrayDeque based approach without recursion. Seems to be working fine, but we will see what jenkins says.

hmottestad:
Ran a final benchmark and also compared it to Jena. Jena is sometimes faster and sometimes slower.

```
Benchmark                                       Mode  Cnt    RDF4J      Jena    Units
IsomorphicBenchmark.blankNodes                  avgt   10    0.165     0.250    ms/op
IsomorphicBenchmark.bsbm                        avgt   10   24.869    65.654    ms/op
IsomorphicBenchmark.bsbmNotIsomorphic           avgt   10   25.697    23.458    ms/op
IsomorphicBenchmark.bsbmArrayList               avgt   10  322.951    ------    ms/op
IsomorphicBenchmark.bsbmTree                    avgt   10  138.642    ------    ms/op
IsomorphicBenchmark.empty                       avgt   10   ≈ 10⁻⁴    ≈ 10⁻³    ms/op
IsomorphicBenchmark.emptyNotIsomorphic          avgt   10   ≈ 10⁻⁵    ≈ 10⁻⁴    ms/op
IsomorphicBenchmark.internallyIsomorphic        avgt   10   30.933     9.786    ms/op
IsomorphicBenchmark.list                        avgt   10    6.121     1.580    ms/op
IsomorphicBenchmark.longChain                   avgt   10    0.012     0.037    ms/op
IsomorphicBenchmark.manyProperties              avgt   10  127.931   319.048    ms/op
IsomorphicBenchmark.manyProperties2             avgt   10  173.765     8.245    ms/op
IsomorphicBenchmark.shacl                       avgt   10    2.508     0.637    ms/op
IsomorphicBenchmark.sparqlTestCase              avgt   10    0.045     0.070    ms/op
IsomorphicBenchmark.spinFullForwardchained      avgt   10  764.446    15.503    ms/op
```

In some cases Jena is 2-3x slower than RDF4J. In other cases RDF4J is slower, up to 50x slower.

I think the biggest issue for RDF4J is the choice of TreeModel, so I reckon it'll be useful to swap that to LinkedHashModel.

hmottestad:
Results after optimising LinkedHashModel.contains(Object o):

```
Benchmark                                   Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes              avgt   10    0.188 ±  0.004  ms/op
IsomorphicBenchmark.bsbm                    avgt   10   34.417 ±  0.919  ms/op
IsomorphicBenchmark.bsbmArrayList           avgt   10  114.878 ±  2.485  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic       avgt   10   17.743 ±  0.532  ms/op
IsomorphicBenchmark.bsbmTree                avgt   10  141.476 ±  2.152  ms/op
IsomorphicBenchmark.empty                   avgt   10   ≈ 10⁻⁴           ms/op
IsomorphicBenchmark.emptyNotIsomorphic      avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.internallyIsomorphic    avgt   10   35.329 ±  0.968  ms/op
IsomorphicBenchmark.list                    avgt   10    6.852 ±  0.120  ms/op
IsomorphicBenchmark.longChain               avgt   10    0.009 ±  0.001  ms/op
IsomorphicBenchmark.manyProperties          avgt   10  138.511 ±  1.791  ms/op
IsomorphicBenchmark.manyProperties2         avgt   10  169.942 ±  6.942  ms/op
IsomorphicBenchmark.sameModel               avgt   10   ≈ 10⁻⁶           ms/op
IsomorphicBenchmark.shacl                   avgt   10    3.236 ±  0.129  ms/op
IsomorphicBenchmark.sparqlTestCase          avgt   10    0.050 ±  0.001  ms/op
IsomorphicBenchmark.spinFullForwardchained  avgt   10  811.222 ± 19.786  ms/op
```

hmottestad:
After optimising to use HashSet where available:

```
Benchmark                          Mode  Cnt   Score   Error  Units
IsomorphicBenchmark.bsbmArrayList  avgt   10  71.239 ± 1.532  ms/op
```

Down from 114.878 ms/op.

model/src/main/java/org/eclipse/rdf4j/model/impl/LinkedHashModel.java
```diff
@@ -183,6 +188,10 @@ public boolean remove(Object o) {
 	@Override
 	public boolean contains(Object o) {
 		if (o instanceof Statement) {
+			if (o instanceof ModelStatement) {
+				return statements.contains(o);
```
hmottestad: this helps performance

model/src/main/java/org/eclipse/rdf4j/model/impl/LinkedHashModel.java
```diff
@@ -79,8 +79,12 @@ public LinkedHashModel(Collection<? extends Statement> c) {
...
 	public LinkedHashModel(int size) {
 		super();
-		values = new HashMap<>(size * 2);
-		statements = new LinkedHashSet<>(size);
+
+		// assume fewer unique values than statements (eg. some reuse of nodes instatements)
+		values = new HashMap<>(size, 0.75f);
+
+		// compensate for the loadfactor by multiplying by 2
+		statements = new LinkedHashSet<>(size * 2, 0.75f);
```
hmottestad: this helps performance, though I don't have any explicit benchmarks just for this

model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleBNode.java
```diff
@@ -78,8 +78,19 @@ public boolean equals(Object o) {
...
-			return this.getID().equals(otherNode.getID());
+			String a = getID();
+			String b = ((BNode) o).getID();
+
+			if (a.length() != b.length()) {
+				return false;
+			}
+
+			// faster when bnode IDs are incremental
+			if (a.length() > 0 && a.charAt(a.length() - 1) != b.charAt(b.length() - 1)) {
```
hmottestad: this also makes a small improvement

model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleIRI.java
```diff
@@ -115,7 +115,20 @@ public boolean equals(Object o) {
...
+
+			String a = toString();
+			String b = o.toString();
+
+			if (a.length() != b.length()) {
+				return false;
+			}
+
+			// faster if using same namespace and UUID
+			if (a.length() > 0 && a.charAt(a.length() - 1) != b.charAt(b.length() - 1)) {
```
hmottestad: I don't have any benchmarks for this, so it's more an assumption that anything else.

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -510,7 +514,11 @@ public static Model setProperty(Model m, Resource subject, IRI property, Value v
 	 *      3.6 (Graph Comparison)</a>
 	 */
 	public static boolean isomorphic(Iterable<? extends Statement> model1, Iterable<? extends Statement> model2) {
-		Model set1 = toModel(model1);
+		if (model1 == model2) {
+			return true;
```
hmottestad: check if two models actually are the same model

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -538,7 +546,7 @@ public static boolean equals(Iterable<? extends Statement> model1, Iterable<? ex
 	 */
 	public static boolean isSubset(Iterable<? extends Statement> model1, Iterable<? extends Statement> model2) {
 		// Filter duplicates
-		Model set1 = toModel(model1);
+		Set<Statement> set1 = toSet(model1);
```
hmottestad: We don't actually need a full-blown model here, using a HashSet is much faster.

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -663,55 +696,15 @@ private static boolean statementsMatch(Statement st1, Statement st2, Map<Resourc
...
-					// 'subj2' is already mapped to some other value.
-					return false;
-				}
-			}
-		} else {
-			// subjects are not (both) bNodes
-			if (!subj1.equals(subj2)) {
-				return false;
-			}
+		if (bnodeValueMatching(bNodeMapping, subj1, subj2)) {
```
hmottestad: Extracted out this method, mostly because intellij told me that there were 3 duplicates. Also gave a small performance improvement, since the smaller method is easier to optimise for the jvm.

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -724,51 +717,77 @@ private static boolean statementsMatch(Statement st1, Statement st2, Map<Resourc
...
-			}
+			// objects are not (both) bNodes
+			return !obj1.equals(obj2);
 		}
-
-		return true;
 	}
 
 	private static boolean isBlank(Value value) {
+		if (value instanceof BNode) {
```
hmottestad: Checking for instanceOfBnode first also gave a nice improvement. I think it might be mostly due to the jvm being able to do better escape analysis, but that's just a hunch from looking at the flight recordings.

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -724,51 +717,77 @@ private static boolean statementsMatch(Statement st1, Statement st2, Map<Resourc
...
 	private static Model toModel(Iterable<? extends Statement> iterable) {
 		if (iterable instanceof Model) {
 			return (Model) iterable;
 		}
-		final Model set = new TreeModel();
-		StreamSupport.stream(iterable.spliterator(), false).filter(Objects::nonNull).forEach(st -> set.add(st));
+
+		final Model set;
+		if (iterable instanceof Collection) {
+			int size = ((Collection<? extends Statement>) iterable).size();
```
hmottestad: Getting the size if possible is great for the LinkedHashModel.

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -724,51 +717,77 @@ private static boolean statementsMatch(Statement st1, Statement st2, Map<Resourc
...
+			set = new LinkedHashModel(size);
```
hmottestad: Also I swapped from the TreeModel, which is very slow.

model/src/main/java/org/eclipse/rdf4j/model/util/PatternIterator.java
```diff
@@ -86,7 +86,6 @@ public void remove() {
 	 * 
 	 * @param st The statement to be tested.
 	 * @return <tt>true</tt> if the object should be returned, <tt>false</tt> otherwise.
-	 * @throws X
```
hmottestad: not sure why this was here, intellij said it was wrong so I removed it :)

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
hmottestad: this is done

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
hmottestad: this is still missing

hmottestad:
Lastest numbers:

```
Benchmark                                   Mode  Cnt    Score     Error  Units
IsomorphicBenchmark.blankNodes              avgt   10    0.134 ±   0.005  ms/op
IsomorphicBenchmark.bsbm                    avgt   10   39.182 ±   0.845  ms/op
IsomorphicBenchmark.bsbmArrayList           avgt   10   88.719 ±   8.680  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic       avgt   10   21.488 ±   2.624  ms/op
IsomorphicBenchmark.bsbmTree                avgt   10  144.218 ±   2.983  ms/op
IsomorphicBenchmark.empty                   avgt   10   ≈ 10⁻⁴            ms/op
IsomorphicBenchmark.emptyNotIsomorphic      avgt   10   ≈ 10⁻⁵            ms/op
IsomorphicBenchmark.internallyIsomorphic    avgt   10   30.692 ±   0.455  ms/op
IsomorphicBenchmark.list                    avgt   10    4.565 ±   0.126  ms/op
IsomorphicBenchmark.longChain               avgt   10    0.016 ±   0.001  ms/op
IsomorphicBenchmark.manyProperties          avgt   10  147.418 ±   2.799  ms/op
IsomorphicBenchmark.manyProperties2         avgt   10  177.336 ±   2.343  ms/op
IsomorphicBenchmark.sameModel               avgt   10   ≈ 10⁻⁶            ms/op
IsomorphicBenchmark.shacl                   avgt   10    3.069 ±   0.045  ms/op
IsomorphicBenchmark.sparqlTestCase          avgt   10    0.046 ±   0.001  ms/op
IsomorphicBenchmark.spinFullForwardchained  avgt   10  706.181 ± 168.265  ms/op
```

model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
jeenbroekstra: We should really be using an injectable ModelFactory for this kind of thing. But good find.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
```diff
@@ -0,0 +1,217 @@
+package org.eclipse.rdf4j.model.util;
```
jeenbroekstra: Copyright header is missing here, and also on the new IsomorphicTest class.

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
```diff
@@ -0,0 +1,217 @@
...
+	private Model list_2 = getModel("list.ttl");
+	private Model internallyIsomorphic_2 = getModel("internallyIsomorphic.ttl");
+	private Model manyProperties_2 = getModel("manyProperties.ttl");
+	private Model manyProperties2_2 = getModel("manyProperties2.ttl");
+
+	@Setup(Level.Iteration)
+	public void after() {
+		System.gc();
+	}
+
```
jeenbroekstra: What would generally make benchmarks like these easier to use for those of us who don't use IntelliJ, is if they had a main method, like so:


```
    public static void main(String[] args) throws RunnerException {

        Options opt = new OptionsBuilder()
                .include(IsomorphicBenchmark.class.getSimpleName())
                .forks(1)
                .build();

        new Runner(opt).run();
    }
```

If this is present, running these benchmarks directly from Eclipse becomes a lot easier.

PR #1492, opened 2019-08-05
Public API clarifications

jeenbroekstra:
This PR addresses GitHub issue: #619 .

* marked SPARQL AST and its internal classes as deprecated
* added two custom annotations (may remove those again if we decide using deprecated is enough) 

The idea for a lot of the classes marked deprecated here is that in future we can make them package-internal. But I want to go slow and give our users time to react to this change - there may be legit use cases for direct use of the AST out there that we may still need to accomodate somehow.  I'm also still on the fence over whether it's even a good idea to make this package internal completely. 

If we agree that this is a good general way to handle the API clarification issue, we can go ahead and do this for as many relevant classes and packages as possible before the 3.0 release.


jeenbroekstra:
@ansell would also like your views on this approach if you have time.

jeenbroekstra:
> My usual practice is to use both the `@deprecated` javadoc tag and the `@Deprecated` annotation based on this guidance from Java-5 https://docs.oracle.com/javase/1.5.0/docs/guide/javadoc/deprecation/deprecation.html Not familiar with the implications of using one or both, however.

Good point.

> In addition to deprecations, it could be nice to experiment with `moditect` to emulate `module-info` while staying on Java-8 bytecode/APIs. https://github.com/moditect/moditect
> 
> Even if `module-info` isn't implemented for 3.0, it would be great to ensure at the 3.0 mark that the main design prerequisites for using it in the near future (JDK-8 will maintained until 2022/2023, so it has a fair lifetime left) are put in place. By that I mean,
> 
>     * Ensure that packages are completely contained in a single JAR file; and,
> 
>     * `META-INF/services` only referring to classes that are in the same JAR file to avoid moving things around when implementing `module-info`
>

It sounds like a good approach, but 3.0 is scheduled for August 15. I think it's unlikely we can do this before that time (unless someone with time and experience can step up to the plate). At this point, I'm just looking for quick wins/low hanging fruit to help set us up at the documentation / developer expectations level.

avataar:
There are legitimate uses, for example when you work with queries supplied by the end user and you want to do some preprocessing on them before evaluation or you want to generate a different query that reuses some of the supplied query (e.g. the prefixes). I'd rather have a proper parser than resort to crude regular expression matching in such cases.

How about splitting the RDF4J code into a public, documented, guaranteed to stay compatible part and a non-public (in the sense of guarantees for compatibility) internal part but leave the internal part public in the Java sense. This way it will stay accessible to those who need it for corner cases and yet make it easily maintainable. Public vs internal is just a naming convention. It could also be stable vs experimental.

Marking something as deprecated is certainly a way to discourage people from using something but if you provide no alternative to the deprecated bit they'll still use it if they need it.

jeenbroekstra:
> There are legitimate uses, for example when you work with queries supplied by the end user and you want to do some preprocessing on them before evaluation or you want to generate a different query that reuses some of the supplied query (e.g. the prefixes). I'd rather have a proper parser than resort to crude regular expression matching in such cases.

Excellent point, and the more I think about it, the more I agree that we should _not_ make AST and its processing classes "off-limits".

> How about splitting the RDF4J code into a public, documented, guaranteed to stay compatible part and a non-public (in the sense of guarantees for compatibility) internal part but leave the internal part public in the Java sense. This way it will stay accessible to those who need it for corner cases and yet make it easily maintainable. Public vs internal is just a naming convention. It could also be stable vs experimental.

Yes, I think that is actually more or less what we're trying to do here, broadly speaking. But I think I'll roll it back for the SPARQL AST at least. 

> Marking something as deprecated is certainly a way to discourage people from using something but if you provide no alternative to the deprecated bit they'll still use it if they need it.

Well this is generally true of course - if we make it _technically_ accessible, people will use it in a pinch, even if we put red flashing lights on it :) To me, the main point of the exercise is to make it clear what parts of the code people can rely on remaining stable between versions, and what parts we may decide to change if we need to (without immediately having to go for big honking major new release and massive upgrade notes).

hmottestad:
Just a little hacky tip. If you make helper classes in the same java package you get access to everything that isn’t private. I’ve used this for other libraries when I needed to get hold of package-local or protected methods/variables.

ansell:
> Just a little hacky tip. If you make helper classes in the same java package you get access to everything that isn’t private. I’ve used this for other libraries when I needed to get hold of package-local or protected methods/variables.

That is the reason behind closing down that method with Java-9+ modules, it prevents people from using your code in ways that you didn't want, who then complain when it breaks.

jeenbroekstra:
> That is the reason behind closing down that method with Java-9+ modules, it prevents people from using your code in ways that you didn't want, who then complain when it breaks.

For what it's worth, at this point I subscribe to the Python mantra: "we're all consenting adults". It's one thing to use a public class that was "intended" to be internal-only, but if someone is deliberately creating helpers to get access to package-protected code, it's no longer our responsibility as maintainers. 

At the end of the day there's no technical way to _force_ people not to use a particular piece of open-source code. Even with Java 9 modules, what's to stop someone from, say, writing a script that automatically forks and opens up the code base, or use some weird reflection loophole, and then complain when we make a change that makes their approach stop working? We can only signal intent.

jeenbroekstra:
@ansell I have raised your point about ensuring each package is in a single jar as a new issue, see https://github.com/eclipse/rdf4j/issues/1494. I am not sure we can still do this before the 3.0 release, but I'll have a look and see what I can do. 

Unless I hear anything to the contrary, I will merge this PR in by the end of the day.

Issue #1148, opened 2018-10-25
Build on JDK 11


barthanssens:
Oracle JDK 11 is an LTS version, so OpenJDK 11 might also be an interesting version to support

Some EE classes like JAXB are removed in JDK 11, e.g. maven complains about missing javax.xml.bind.annotation.adapters.HexBinaryAdapter when building rdf4j-tools etc

More info can be found here: https://blog.codefx.org/java/java-11-migration-guide/

And here (modular development) https://www.youtube.com/watch?v=gtcTftvj0d0&feature=youtu.be&t=32m19s

Todo:
- [x] use Maven 3.6.1
- [x] workaround for javadoc errors / NPE
- [x] upgrade maven plugins
  - [x] Upgrade japicmp-maven-plugin (e.g. 0.14.0)
- [x] upgrade Apache HttpCore to 4.4.11
- [x] upgrade AssertJ to 3.11
- [x] upgrade Mockito to 2.23.4
  - [x] ignore module.info when using 'maven enforce bytecode JDK 1.8' (modules in byte-buddy dependency)
- [x] upgrade server/workbench to Spring 5.1
- [x] add JAXB-api
- [x] add other EE modules removed in recent JDK
  - [x] javax.annotation (jakarta)

Status:
- [x] `mvn clean install -DskipTests`
- [x] `mvn clean install -Passembly -DskipTests`
- [x] `mvn clean install`
- [x] `mvn clean install -Passembly`
- [x] `mvn clean install -Pcompliance`

hmottestad:
- [ ] Upgrade japicmp-maven-plugin

hmottestad:
bumping the maven-javadoc-plugin version causes this: 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.0:aggregate (create-aggregate-javadoc) on project rdf4j-storage: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /Users/havardottestad/Documents/Programming/RDF4J/rdf4j-storage/elasticsearch/src/main/java/org/eclipse/rdf4j/sail/elasticsearch/ElasticsearchIndex.java:36: error: package org.elasticsearch.action does not exist
[ERROR] import org.elasticsearch.action.ActionRequestBuilder;
[ERROR]                                ^
[...]

PR #1501, opened 2019-08-13
Issues/#1148 jdk11 javadoc

barthanssens:
This PR addresses GitHub issue: #1148 .

* Seems like the -Xdoclint:none is not needed (or not anymore, at least not on JDK11), though there are some warnings

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,15 +192,6 @@
 				<module>compliance</module>
 			</modules>
 		</profile>
-		<profile>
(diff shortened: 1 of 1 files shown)


barthanssens:
No real need to remove it I guess, if Jenkins can't build without disabling doclint.
So maybe it's better to leave it as it was, and just create another profile for checking the javadocs with doclint enabled.

Issue #1467, opened 2019-06-26
Use a single repository for the entire rdf4j code base (merging storage, tools, and testsuite back into rdf4j)


jeenbroekstra:
## Motivation

Back in 2017 (see https://www.eclipse.org/lists/rdf4j-dev/msg00410.html ) we made a decision to split the rdf4j project over multiple repositories. The main motivation for this was that a full build + verification of the project was taking too long, and this encouraged contributors to take shortcuts. The theory was that by splitting the project, we could get verification time down.
 
However, that expected speed gain has not really materialized. It's true that individual repo builds are quicker, but when we make a change in, for example, the rdf4j repo, we still need to run verification in rdf4j-storage and -tools - and can turn out to still break things.
 
A further downside is that compliance tests in the rdf4j repo often use code from rdf4j-storage (e.g. a sail impl) or from rdf4j-tools (e.g. to spin up an rdf4j server) - however due to the order of dependencies, those modules are not built yet when rdf4j repo does its verification. While that's no big deal as long as we're on a develop branch, and everything just uses "the latest SNAPSHOT", it seriously messes up things when release time rolls around and we have to set fixed versions: suddenly when rdf4j repo tries to build it fails because its tests can't find rdf4j-sail-memory 3.0.0. I have been attempting to mitigate this by moving tests around in the project, as well as setting fixed (older) versions for these kinds of test dependencies - but it's not ideal.
 
Finally: the Jenkins pipeline we currently have to build, verify and deploy all of this is just incredibly convoluted. We now have 24(!) separate Jenkins jobs to coordinate all of this, and it's painful to maintain tbh.
 
## Proposed change 

We move back to a mono-repo for the entire codebase, which will live in the rdf4j github repository. rdf4j-doc will remain a separate repository, where the project website and documentation are maintained.

To make sure we get decent build and verification times, we will make the following improvements:
 
- culling and cleaning in our compliance and integration tests (there are quite a few tests in there that are either very slow, or redundant, or both).
- better unit testing with mocking and stubbing instead of cramming all our verification into massive compliance/integration test suites that spin up full servers every time.

## Tasks

- [x] merge rdf4j-storage (with tags and history) into rdf4j
- [x] merge rdf4j-tools (with tags and history) into rdf4j
- [x] merge rdf4j-testsuites (with tags and history) into rdf4j
- [x] reconfigure maven to handle the single-repo build
- [x] reconfigure Jenkins to handle the simplified build pipeline
[...]

PR #1512, opened 2019-08-25
monorepo integration

jeenbroekstra:
This PR addresses GitHub issue: #1467 .

* modules from rdf4j-storage, rdf4j-tools, and rdf4j-testsuite re-integrated in main rdf4j repo, preserving history where possible
* all core library modules are in the `core` directory, all tools/applications are in `tools`. Testsuite and compliance tests are top-level and are compiled and run _after_ all libraries and tools are compiled (so they are becoming more system/end-to-end tests). 
* maven version in pom uses patch snapshot version again (3.0.1-SNAPSHOT). 
* Verified that tests run and succeed, and that SDK and onejar can be built (with javadoc included)

One thing to note:

Compliance tests _do not run_ when verifying a PR. It does run unit tests, but running the full compliance suite takes close to an hour, so instead that gets run _after_ something gets merged into master.
 
There are likely still minor glitches with this setup, and it will need some further tinkering, but this should be good enough to get us going with development on the master branch again.


jeenbroekstra:
This is a huge PR of course, so I'm not expecting a full review, but I'd like it if you could give it a glance, maybe try it out locally, and give me a green tick if you're happy for this to be merged. 

note that Jenkins has not yet been reconfigured to deal with this.

barthanssens:
mvn clean install -DskipTests worked just fine (with JDK8), so that's already a good sign, nice work
So I'll give it a shot with tests enabled...

jeenbroekstra:
> "mvn clean install" worked perfectly on Debian Linux with OpenJDK8 (8u222)

Minor point: you (currently) have to include `-Pcompliance` for it to execute the compliance tests. We can discuss whether this is a good approach or not but I thought I'd give a heads up at least.

Issue #1511, opened 2019-08-22
Console: NPE when just typing "prefix" 


barthanssens:
NPE when just entering "prefix" at the command prompt (Console is expecting a full query)

See message on forum: https://groups.google.com/forum/#!topic/rdf4j-users/x-dPGe53Wk4

barthanssens:
```
Exception in thread "main" java.lang.NullPointerException
        at org.eclipse.rdf4j.query.parser.QueryPrologLexer.getRestOfQueryToken(QueryPrologLexer.java:161)
        at org.eclipse.rdf4j.query.parser.QueryParserUtil.removeSPARQLQueryProlog(QueryParserUtil.java:167)
        at org.eclipse.rdf4j.query.parser.QueryParserUtil.parseOperation(QueryParserUtil.java:45)
        at org.eclipse.rdf4j.console.command.QueryEvaluator.parseAndEvaluateQuery(QueryEvaluator.java:323)
        at org.eclipse.rdf4j.console.command.QueryEvaluator.executeQuery(QueryEvaluator.java:194)
        at org.eclipse.rdf4j.console.Console.executeCommand(Console.java:344)
        at org.eclipse.rdf4j.console.Console.start(Console.java:304)
        at org.eclipse.rdf4j.console.Console.main(Console.java:143)
```

PR #1516, opened 2019-08-27
Fix for NPE with malformed PREFIX query

barthanssens:
This PR addresses GitHub issue: #1511 .

* Make sure the QueryPrologLexer.getRestOfQueryToken does not throw NPE (e.g. when just entering "PREFIX" in the console, without anything else).

Workaround is to just use an empty string (which should be OK because the method  - with any input - does not guarantee that the returned string is syntactically correct).


barthanssens:
It's not just the Console that uses this QueryPrologLexer, but should be save to use in other parts of RDF4J

jeenbroekstra:
I'm merging this, because I want to use it to check something in the Jenkins setup.

Issue #1529, opened 2019-08-31
SHACL benchmark execution as part of unit tests is flaky


jeenbroekstra:
> Error Message
> 
> ERROR: Unable to acquire the JMH lock (/tmp/genie.rdf4j/jmh.lock): already taken by another JMH instance, exiting. Use -Djmh.ignoreLock=true to forcefully continue.
> 
> Stacktrace
> 
> org.openjdk.jmh.runner.RunnerException: ERROR: Unable to acquire the JMH lock (/tmp/genie.rdf4j/jmh.lock): already taken by another JMH instance, exiting. Use -Djmh.ignoreLock=true to forcefully continue.
> 	at org.eclipse.rdf4j.sail.shacl.benchmark.BenchmarkJunitTests.test(BenchmarkJunitTests.java:36)

This needs to be fixed. Possibilities to eplore:

- set `jmh.ignoreLock=true`
- disable running of these benchmarks as part of unit tests

A longer-term solution to the issue of benchmarks being regularly run would be a completely separate Jenkins job that executes these benchmarks as a nightly job. But for the immediate future we need a solution that stops this from making the main build unstable.

jeenbroekstra:
See also #1515

jeenbroekstra:
I'm having a look at configuring the surefire plugin to use that jhm.ignoreLock setting.

jeenbroekstra:
It _looks_ as if that fixed the problem.

PR #1530, opened 2019-08-31
JMH runner ignores lock to avoid parallel builds failing

jeenbroekstra:
This PR addresses GitHub issue: #1529  .

* configured surefire plugin set `jmh.ignoreLock` to `true` to avoid benchmark execution failing
* ran local parallel test, seems to solve the issue (but proof is in Jenkins obviously)

### DIFF
--- a/core/sail/shacl/pom.xml
+++ b/core/sail/shacl/pom.xml
@@ -107,4 +107,22 @@
 
 	</dependencies>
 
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
Merging this in immediately in a "let's see if it actually works" kinda way.

Issue #1541, opened 2019-09-07
Support retrieval of repository configuration in REST API


jeenbroekstra:
Followup of #1389 

Although the REST API now officially supports (re)configuration of a repository by means of upload of a configuration RDF model, it has no official way to retrieve the existing configuration of a repository. 

The (obsolete) way in which the client manager currently handles this is to use a GET request against `/rdf4j-server/repositories/SYSTEM/statements` to retrieve config data for all repos and then use `RepositoryConfigUtil.getRepositoryConfig` to filter out the specific repository we're interested in.

We should add an official per-repository endpoint to retrieve its current configuration. 

Possible design alternatives for the REST resource:

### Alternative 1 
    GET /repositories/<repId>

Pro: 

- consistent with endpoint for configuration change (see #1389)

Con: 

- overloads the endpoint since this is the same resource for (SPARQL) querying. 
    
### Alternative 2

- `GET /repositories/<repId>/config` 
  Retrieval of configuration of existing repository
- `POST /repositories/<repId>/config`
  Modify configuration of existing repository 

In this alternative, we'd adapt the definition of the PUT on `/repositories/<repId>` as introduced in #1389 slightly: it would _only_ be used to create a new repository. For updating the config of an existing repo, we would require the use of this new POST method.
  
Pro:

-  conceptually clean separation between repository and _config of_ the repository.

Con:

- not symmetric with PUT operation for repository config change (though we can of course align this).

jeenbroekstra:
I'm inclined to go with the first alternative. Despite the downside, it seems to fit with rest of the existing API best (not just the PUT to create a repository or change its config, but also the DELETE to remove a repository).

PR #1545, opened 2019-09-15
new endpoint for repo config retrieval

jeenbroekstra:
This PR addresses GitHub issue: #1541 .

* new endpoint for retrieval of repository config
* new endpoint for changing existing repo config
* modified `PUT /repo/<ID>` endpoint to only accept if repository does not yet exist.


hmottestad:
I don’t really have any comments except that this needs documentation :)

jeenbroekstra:
> I don’t really have any comments except that this needs documentation :) User documentation.

I've added that now, it's in rdf4j-doc on the develop branch. See https://github.com/eclipse/rdf4j-doc/pull/106
 
> Would be great if we could integrate this into the docker solution so you could easily spin up a server with a NativeStore.

Not a bad plan!

Issue #1538, opened 2019-09-05
Integrate FedX into rdf4j


jeenbroekstra:
Move fedx codebase into rdf4j as a module `org.eclipse.rdf4j.federated`, under `tools/`.

CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20667

aschwarte10:
@jeenbroekstra in PR https://github.com/eclipse/rdf4j/pull/1622 I have succesfully tested to create a FedX federation in the RDF4J workbench. Currently however this is a bit cumbersome as the integration is not very tight (i.e. the federation binary is not bundled with the rdf4j-server war and moreover there does not exist a template).

In my manual test I basically added a binary jar to the server webapp, and created a federation repository in the file system. This example test was successful.

For the release and for our users I would however like to pursue a more tight integration, i.e. make the federation feature more out-of-box usable. What would be the steps from your perspective to make the federation available also in the rdf4j-server webapp? Do you think we can aim for this in the release?

jeenbroekstra:
Off the cuff I'd say that it's simply a matter of adding the dependency to the rdf4j-server's pom. Now that I think about it, we'll probably also need to add it to the bill of materials and to the rdf4j-runtime pom. I think we should deifnitely try and include that in the next release (we have some time anyway, at least another 4-6weeks before we can do a release as we'll need to finalize legal review for FedX as well as go through our own release review).

jeenbroekstra:
CQ has been approved! :tada:

PR #1547, opened 2019-09-16
Integrate FedX into RDF4j

aschwarte10:
This PR addresses GitHub issue: #1538 .

In this PR we integrate the source of the FedX project into RDF4J.

FedX is a practical framework for transparent access to Linked Data sources through a federation. It incorporates various sophisticated optimization techniques combined with effective variants of existing techniques and is thus a highly scalable solution for practical federated query processing. The implementation is based on the RDF4J SAIL infrastructure.

The project was initially developed under the affiliation of fluid Operations as part of my Master Thesis in 2011. In 2018 the copyrights went over to Veritas (after the aquisition of fluid Operations by Veritas). As part of the merge we were permitted to change the license of FedX to Apache License and publish it on Github: https://github.com/VeritasOS/fedx Veritas agreed to contribute FedX to the Eclipse RDF4J project

In this PR the integration has happenend in the following steps
- migrate source code into the maven structure
- integrate the project as maven module into RDF4j
- add RDF4J copyright headers
- apply formatting
- minor refinements (logging verbosity in tests, resolving deprecated methods)


jeenbroekstra:
@rdf4j-bot whitelist

jeenbroekstra:
@rdf4j-bot run tests

tools/federation/build/README.txt
```diff
@@ -0,0 +1,15 @@
+Readme
+
+First steps and important information:
+
+ * Documentation can be found at https://github.com/VeritasOS/fedx/wiki
+ * Required libraries are in the lib folder
+ * Github Page: https://github.com/VeritasOS/fedx
+ * Contact: Andreas.Schwarte (at) veritas.com / aschwarte10 (at) gmail.com
+ * Project Website: https://github.com/VeritasOS/fedx
+ * FedX is licensed under Apache License 2.0 (see attached license).
```
jeenbroekstra: I'm guessing this file should be changed or removed?

tools/federation/build/test/rdf4j-server/WEB-INF/rdf4j-http-server-servlet.xml
```diff
@@ -0,0 +1,150 @@
...
+		<property name="basenames">
+			<list>
+				<value>org.eclipse.rdf4j.http.server.messages</value>
+				<value>org.eclipse.rdf4j.common.webapp.system.messages</value>
+				<value>org.eclipse.rdf4j.common.webapp.messages</value>
+			</list>
+		</property>
+	</bean>
+
+    <!-- Customized FedX: use custom init to avoid initialization of logging framework -->
```
jeenbroekstra: What's this for? Is this to avoid a lot of STDOUT output during test runs, or something else?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/exception/FedXException.java
```diff
@@ -0,0 +1,35 @@
...
+ *******************************************************************************/
+package org.eclipse.rdf4j.federated.exception;
+
+/**
+ * Base class for any FedX Exception.
+ * 
+ * @author Andreas Schwarte
+ *
+ */
+public class FedXException extends RuntimeException {
```
jeenbroekstra: Could we make this extend RDF4JException?

tools/federation/build/test/rdf4j-server/WEB-INF/rdf4j-http-server-servlet.xml
aschwarte10: From my memories this was required for my stripped-down rdf4j workbench to work properly w.r.t logging while executing the test. I think also it was to reduce dependencies (as I typically used log4j). Maybe this could now be change a bit again.

I also remember that there is an issue in the RDF4J tracker to make logging in the workbench more configurable, as this currently strictly requires logback.

For now I would suggest to keep this (as it is for testing only), if this is OK with you.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/exception/FedXException.java
aschwarte10: sure, that makes perfectly sense. will be done in the next commit as part of the PR.

tools/federation/build/README.txt
aschwarte10: Let's remove it, and add a section on the history of FedX in the documentation later.

tools/federation/pom.xml
```diff
@@ -0,0 +1,173 @@
...
+               <goal>test</goal>
+             </goals>
+             <configuration>
+               <systemPropertyVariables>
+                 <repositoryType>NATIVE</repositoryType>
+               </systemPropertyVariables>
+             </configuration>
+           </execution>
+           
+           <execution>
```
aschwarte10: @jeenbroekstra to save time during the build (or during quickcheck) we can also skip the tests against a HTTP remote repository.
Basically we run the same test set (with a few exceptions) in three environments: SPARQL repository, Native Store, and HTTP repository.
SPARQL and HTTP repository are pretty much similar. In the native store different optimization strategies are applied.
What's your opinion?

aschwarte10:
@jeenbroekstra I cannot really nail down what's going on with the failing test. It seems to be an OS specific issue (on Mac it is working, on Unix apparently not). The same embeeded test server is also used for the other tests, so it cannot be a general problem. 

I do not want to try & error with commits, can you maybe do a short debugging locally (maybe with more logging enabled in the logback configuration file, see test resources). Hopefully this gives some more hints. Otherwise I need to run the tests in a docker container or a VM.

tools/federation/pom.xml
jeenbroekstra: > @jeenbroekstra I cannot really nail down what's going on with the failing test. It seems to be an OS specific issue (on Mac it is working, on Unix apparently not). The same embeeded test server is also used for the other tests, so it cannot be a general problem.
> 
> I do not want to try & error with commits, can you maybe do a short debugging locally (maybe with more logging enabled in the logback configuration file, see test resources). Hopefully this gives some more hints. Otherwise I need to run the tests in a docker container or a VM.

From what I can tell, for some reason the test expects a remote repository with id `myFederation` to be present on the server, but when I check the test server config, it only contains repositories `endpoint1`, `endpoint2`, `endpoint3`, and `endpoint4`. Does that help at all?

jeenbroekstra:
> @jeenbroekstra I cannot really nail down what's going on with the failing test. It seems to be an OS specific issue (on Mac it is working, on Unix apparently not). The same embeeded test server is also used for the other tests, so it cannot be a general problem.
> 
> I do not want to try & error with commits, can you maybe do a short debugging locally (maybe with more logging enabled in the logback configuration file, see test resources). Hopefully this gives some more hints. Otherwise I need to run the tests in a docker container or a VM.

From what I can tell, for some reason the test expects a remote repository with id myFederation to be present on the server, but when I check the test server config, it only contains repositories endpoint1, endpoint2, endpoint3, and endpoint4. Does that help at all?

tools/federation/pom.xml
jeenbroekstra: I think we can probably be a bit more efficient yeah, though currently (on my machine) the entire fedx test suite runs in about 24 seconds. So I'm not _too_ worried about this :) 

We're more generally, for the entire RDF4J framework, looking into streamlining the tests a bit more: doing more unit tests, and using more mocking and stubbing where possible to avoid spinning up servers all the time.

jeenbroekstra:
As an aside: don't ask me _why_ we use lowercase dir names on Unix/Posix please, that decision has been lost in the mists of time, and we're just stuck with it now...

aschwarte10:
@jeenbroekstra thanks for identifying and resolving the issue with the test. 

Anything else to be done on this PR? Any cleanup? Or should we merge and discuss next steps on the developers list.

What I see:
- check if we can get rid of the NativeStoreExt code (i.e. the optimization)
- check if we can deprecate the current federation sail and use this one as a drop-in replacement
- provide documentation
- refinements in the initialization / configuration (I have a few things in my mind, will create issues for these)
- do we want to stick to the current location in tools, or should we rather integrate into the main framework (not sure if this is require such that it is automatically available in the workbench)

jeenbroekstra:
> @jeenbroekstra thanks for identifying and resolving the issue with the test.
> 
> Anything else to be done on this PR? Any cleanup? Or should we merge and discuss next steps on the developers list.

Let's merge and continue on from there. Most of the followup improvements I'd like to be able to track (and implement) as separate issues. 

> What I see:
> 
> * check if we can get rid of the NativeStoreExt code (i.e. the optimization)

https://github.com/eclipse/rdf4j/issues/1553

> * check if we can deprecate the current federation sail and use this one as a drop-in replacement

https://github.com/eclipse/rdf4j/issues/1554

> * provide documentation

https://github.com/eclipse/rdf4j/issues/1555

> * refinements in the initialization / configuration (I have a few things in my mind, will create issues for these)

Great! I'll leave those to you.

> * do we want to stick to the current location in tools, or should we rather integrate into the main framework (not sure if this is require such that it is automatically available in the workbench)

I think we want to integrate further, and split things out into several modules that sit alongside existing rdf4j modules. At the very least I'd like the Repository implementation and the Sail implementation each in their own modules: rdf4j-repository-federated (see https://github.com/eclipse/rdf4j/issues/1556), and rdf4j-sail-federated  (we could deal with this as part of 
https://github.com/eclipse/rdf4j/issues/1554 )

Issue #1548, opened 2019-09-19
Major bug with DROP GRAPH <non-existing-graph-in-repo>executed over /statements Sparql Update protocol


vujasm:
Hello,

I discovered a major bug with RDF4J server, version 3.0.0 and version 
2.5.3+0adb5d9, that causes ALL named graphs to be permanently removed from a repository. 

On version 2.5.3 it is caused with these steps:

1. I have created few named graphs using INSERT DATA command via Sparql Update protocol

```
INSERT DATA
  { 
    GRAPH <urn:idea:1> 
      { 
triples
      } 
  }


INSERT DATA
  { 
    GRAPH <urn:idea:2> 
      { 
triples
      } 
  }

INSERT DATA
  { 
    GRAPH <urn:idea:3> 
      { 
triples
      } 
  }
```
And few more  with uri `<urn:idea:4> , <urn:idea:5> , <urn:idea:6>` , etc.

2. Then,  I wanted to drop/remove one of graphs , e.g. `<urn:idea:5>` , and I executed `DROP GRAPH <urn:idea:5>`
It works fine --` <urn:idea:5>` has been removed from the repository.

3. But now, if I use DROP GRAPH for a named  graph that is not in a repository, e.g.` DROP GRAPH <urn:idea:1000>`, the result is that all named graphs are removed and context set gets totally empty.

With version 3.0.0. happens something similar. If  I use DROP GRAPH for a  named  graph that is not in my repository, e.g.` DROP GRAPH <urn:idea:1000>`, then the context-set remains as it was before the drop command, as expected, BUT any next execution of DROP GRAPH command for named graphs that are in the repository, is not working at all (does not remove the named graph). Further, if I add a new named graph e.g.  `<urn:idea:25>`, and immediately after adding it I execute `DROP GRAPH <urn:idea:25>`, again ALL the named graphs are removed and context set gets totally empty.

Everything was executed over RD4J REST /statements endpoint for SPARQL UPDATE protocol. Repository is Native Store + RDFS and SPIN.

This is a serious and critical bug, in v2.5.3. and v3 Please try to analyse it and to provide a bugfix or explanation/workaround for it.

Please respond to this issue ASAP. Thanks in advance.

barthanssens:
Thanks for the detailed bug report, we'll look into it

barthanssens:
Seems to be an issue deeper in the library (using 3.0, ode listed below also removes all data)

```
public class Main {
	private final static String INS_1 = 
[...]

PR #1551, opened 2019-09-23
make invalid context different from unknown/all context + added test

barthanssens:
This PR addresses GitHub issue: #1548 .

Isn't a problem with invalid/not found, S/P/O, because removeStatements immediately returns when
a search for an S/P/O yields UNKNOWN_ID
But for contexts, zero or more contexts can be given + the UNKNOW_ID (-1) is used when zero contexts are selected, effectively removing statements from all contexts + zero is used for empty context.

* use MIN-INT number for context not in the store
* added a test

Another way to solve this would be using a List  + converting it back to int array, or  use System.arraycopy to return only the array of found contexts, but that seems less clear IMHO.

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeSailStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeSailStore.java
@@ -520,12 +520,18 @@ class NativeSailStore implements SailStore {
 				}
 
 				final int[] contextIds = new int[contexts.length == 0 ? 1 : contexts.length];
-				if (contexts.length == 0) {
+				if (contexts.length == 0) { // remove from all contexts
 					contextIds[0] = NativeValue.UNKNOWN_ID;
 				} else {
 					for (int i = 0; i < contexts.length; i++) {
 						Resource context = contexts[i];
-						contextIds[i] = context == null ? 0 : valueStore.getID(context);
+						if (context == null) {
+							contextIds[i] = 0;
+						} else {
+							int id = valueStore.getID(context);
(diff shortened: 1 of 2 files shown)


core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeSailStore.java
```diff
@@ -520,12 +520,18 @@ private long removeStatements(Resource subj, IRI pred, Value obj, boolean explic
...
 				} else {
 					for (int i = 0; i < contexts.length; i++) {
 						Resource context = contexts[i];
-						contextIds[i] = context == null ? 0 : valueStore.getID(context);
+						if (context == null) {
+							contextIds[i] = 0;
+						} else {
+							int id = valueStore.getID(context);
+							// unknown_id cannot be used (would result in removal from all contexts)
+							contextIds[i] = (id != NativeValue.UNKNOWN_ID) ? id : Integer.MIN_VALUE;
```
hmottestad: can `id` be negative, if not then I propse -1 instead of MIN_VALUE

barthanssens: id can indeed be negative, NativeStore.UNKNOWN_ID is -1

Which actually triggers the bug, becayse for contexts there is not distinction between not specifying the context at all (yields -1, so remove from all contexts) and a non-existing context....

hmottestad: thanks for explaining :)

hmottestad:
lgtm

PR #1552, opened 2019-09-24
Fix to allow auto-closing of linked issue

jeenbroekstra:
See https://help.github.com/en/articles/closing-issues-using-keywords. This phrasing will ensure the corresponding issue is automatically closed when the PR is merged.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,6 @@
-GitHub issue resolved: #<number>. 
+
+GitHub issue resolved: # <!-- add a Github issue number here, e.g #123. This line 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,4 @@
-
-This PR addresses GitHub issue: # .
+GitHub issue resolved: #<number>.
```
hmottestad: I think it should be possible to add comment html style to these templates where we can explain that this line will automatically close the issue when merged. Those comments will only show up in edit mode and not in viewing mode.

jeenbroekstra: Great idea, done!

Issue #1557, opened 2019-09-25
SHACL - race condition with jdk11 when using multiple threads to access a shacl sail


hmottestad:
With reference to the email from Bart there is a failing test that tests for race conditions.

org.eclipse.rdf4j.sail.shacl.MultithreadedTest#testDataAndShapes

We should try to figure out 
 - if this is only an issue when using parallel shacl validation?
 - if this is an issue with the shacl sail, with one of the added/removed repos or with the base sail

And also try to make it more reproducable!

hmottestad:
This seems to be happening against the base sail. So it makes sense that it only happens when using parallel validation (since this shares a single connection from the base sail between multiple threads).

hmottestad:
It doesn't seem to happen with `IsolationLevels.NONE`.

hmottestad:
Next step is to try to figure out where the read locks are being taken.

hmottestad:
Also, even after 8 hours the test still hangs!

barthanssens:
Well, I guess it will hang forever: a back-off mechanism is being used, each time doubling the interval between  attempts ...

hmottestad:
this seems like it might be caused by my read/write lock stuff directly in the shacl sail

hmottestad:
It's definitively in the shacl code. Looks like there is some code that isn't releasing a read lock.....but I can't figure out where. I might instead end up switching to the lock manger instead.

hmottestad:
<img width="1517" alt="Screenshot 2019-09-26 at 18 22 41" src="https://user-images.githubusercontent.com/797185/65706376-a4698c00-e08a-11e9-9daf-0234c976f9ca.png">

hmottestad:
This seems very odd. All the work threads from the common pool have single stacks except for this thread here that has two stacks.

hmottestad:
The top stack [2] is waiting for a lock that the bottom stack has acquired. I have never seen anything like this before!

barthanssens:
Related to #1148

hmottestad:
Maybe we should go ahead and deactivate parallel validation while I get this sorted. 

I have tried switching from stamped locks to the RDF4J common locks. It still hangs. 

Next is to use a custom thread pool. Maybe for the test rather than for the actuall validation.

PR #1558, opened 2019-09-25
#1557 use explicit executor service instead of .parallel

hmottestad:
This PR addresses GitHub issue: #1557 .

* change from .parallel to an executor service for parallel validation so that other uses of .parallel don't block because of validation
* 
*


hmottestad:
@barthanssens this is the PR I have for diagnosing the failing/stuck test due to race conditions.

hmottestad:
Performance is not impacted. However the previous solution used a stamped lock which could downgrade to a read-lock. However there were no real cases where that would be beneficial.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -307,11 +322,22 @@ public void shutDown() throws SailException {
 			shapesRepo = null;
 		}
 
+		if (executorService != null) {
+			executorService.shutdown();
```
hmottestad: We shutdown the executorService in shutDown(). This makes it very important to remember to shut down the sail.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -307,11 +322,22 @@ public void shutDown() throws SailException {
...
+		}
+
 		initialized.set(false);
 		nodeShapes = Collections.emptyList();
 		super.shutDown();
 	}
 
+	public synchronized <T> Future<T> submitRunnableToExecutorService(Callable<T> runnable) {
+		if (executorService == null) {
+			executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 2);
```
hmottestad: Lazy init of the executor service, so it's only created when we first need parallel validation.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -352,64 +378,85 @@ private void runInferencingSparqlQueries(SailRepositoryConnection shaclSailConne
...
 
-		long newStamp = lock.writeLock();
-		threadHoldingWriteLock = Thread.currentThread();
-		return newStamp;
-	}
-
-	boolean holdsWriteLock(long stamp) {
-		return lock.validate(stamp);
+		try {
+			Lock writeLock = lockManager.getWriteLock();
```
hmottestad: locks now use the rdf4j lock manager

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -352,64 +378,85 @@ private void runInferencingSparqlQueries(SailRepositoryConnection shaclSailConne
...
+			threadHoldingWriteLock = Thread.currentThread();
+			return writeLock;
+		} catch (InterruptedException e) {
+			throw new RuntimeException(e);
```
hmottestad: I don't really know what I should do with the interrupt exception. This should only happen if the users code calls interrupt, so I reckon the safest thing is just to re throw it. @jeenbroekstra do you have any recommendations?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -352,64 +378,85 @@ private void runInferencingSparqlQueries(SailRepositoryConnection shaclSailConne
...
-	synchronized long convertToReadLock(long writeLockStamp) {
-		assert writeLockStamp != 0;
-		long l = lock.tryConvertToReadLock(writeLockStamp);
-		assert l != 0;
-		threadHoldingWriteLock = null;
-		return l;
+		return null;
 	}
 
+//	Lock convertToReadLock(Lock writeLockStamp) {
```
hmottestad: The rdf4j lock manager doesn't support downgrading of locks, I might consider adding this in the future.....so I thought I would keep this code around for a bit longer.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -166,10 +169,12 @@ public void commit() throws SailException {
 			sail.setNodeShapes(nodeShapes);
 		}
 
-		if (sail.holdsWriteLock(writeLockStamp)) {
```
hmottestad: writeLockStamp should never both be "not null" and "not active". However we still double check, since locking is something that is easy to get wrong in subtle and unpredictable ways.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -342,63 +348,87 @@ ConnectionsGroup getConnectionsGroup() {
...
+						}
+					});
 
-						logger.info(
-								"SHACL not valid. The following experimental debug results were produced: \n\tNodeShape: {}\n\tPropertyShape: {} \n\t\t{}",
-								propertyShape.getNodeShape().getId(), propertyShape.getId(),
-								collect.stream()
-										.map(a -> a.toString() + " -cause-> " + a.getCause())
-										.collect(Collectors.joining("\n\t\t")));
+			if (sail.isParallelValidation()) {
```
hmottestad: here is the new code that handles parallel validation

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
hmottestad: Actually wonder if we shouldn't make the sails closable to indicate that they need to be closed/shutdown?

hmottestad:
Propose to merge this today.

Issue #1553, opened 2019-09-25
Merge/remove fedx nativestore extensions


jeenbroekstra:
Investigate if the fedx nativestore extensions are generally relevant and applicable, and if so, merge with the core native store implementation.

aschwarte10:
I had a look at the code and the extensions / overrides and it looks that they are no longer required. In 3.0 the optimizer configuration has been moved to the evaluation strategy and I think with this API enhancement I no longer require the extensions.

Will do a final evaluation and possibly the implementation for this.

aschwarte10:
PR with the relevant changes have been merged

PR #1577, opened 2019-10-01
#1553: remove NativeStore extensions and replace with strategy

aschwarte10:
GitHub issue resolved: #1553

The NativeStoreExt was used to evaluate pre-compiled TupleExpr without
applying any optimizer.

With the latest API changes to EvaluationStrategy we can solve the same
in the strategy.


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/SailTripleSource.java
```diff
@@ -181,27 +180,29 @@ public boolean usePreparedQuery() {
...
+
+				// optimization attempt: use precompiled query
+				PrecompiledQueryNode precompiledQueryNode = new PrecompiledQueryNode(preparedQuery);
+				res = (CloseableIteration<BindingSet, QueryEvaluationException>) sailConn.evaluate(precompiledQueryNode,
+						null, EmptyBindingSet.getInstance(), true);
+
+			} catch (Exception e) {
+				log.warn(
+						"Precompiled query optimization for native store could not be applied: " + e.getMessage());
+				log.debug("Details:", e);
```
jeenbroekstra: Good pattern, this (warn + debug with details). We should use that more often.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/algebra/PrecompiledQueryNode.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.rdf4j.query.algebra.TupleExpr;
+
+/**
+ * Node representing a precompiled query.
+ * 
+ * @author Andreas Schwarte
+ * @see SailSourceEvaluationStrategyFactory
+ * @see SailTripleSource
+ */
+public class PrecompiledQueryNode extends QueryRoot {
```
jeenbroekstra: Is it necessary for this class to be public?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/endpoint/provider/SailSourceEvaluationStrategyFactory.java
```diff
@@ -0,0 +1,150 @@
...
+ * optimization.
+ * <p>
+ * All other types of {@link TupleExpr} are optimized and evaluated through the configured delegate strategy, i.e.
+ * typically the one provided by the sail itself.
+ * </p>
+ * 
+ * @author Andreas Schwarte
+ * @see NativeStoreProvider
+ */
+public class SailSourceEvaluationStrategyFactory implements EvaluationStrategyFactory {
```
jeenbroekstra: Same here. If we can keep these package-private it saves backward compatibility headaches later on.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/algebra/PrecompiledQueryNode.java
aschwarte10: This class is used in two different sub packages (evaluation and endpoint.provider). I currently do not see a good way of restricting visibility. Since this is a simple agebra node I however do not have that much of a problem having this public

tools/federation/src/main/java/org/eclipse/rdf4j/federated/endpoint/provider/SailSourceEvaluationStrategyFactory.java
aschwarte10: Makes sense, will amend the PR

Issue #1578, opened 2019-10-01
Nativestore, new context cache without Exception


barthanssens:
See also https://www.eclipse.org/lists/rdf4j-dev/msg01242.html

When a new native store is created, the context cache file does not exist, so this files gets created. 
Unfortunately there is also an Exception being thrown, because there is currently no distinction between creating a cache for a new store, and recreating an invalid/corrupt cache.

It would be nicer to just log this, instead of throwing / propagating an Exception.

jeenbroekstra:
This thing is driving me nuts in running tests on master. I'm gonna have a quick look at backporting your fix to the master branch.

PR #1579, opened 2019-10-02
Produce less alarming log entry when context file does not yet exist

barthanssens:
GitHub issue resolved: #1578  

* Made a distinction between a "context file not found" and another IO error
* Less alarming log entry, at the expense of some minor code duplication in catch block

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
@@ -102,10 +102,10 @@ class ContextStore implements Iterable<Resource> {
 		try {
 			readContextsFromFile();
 		} catch (FileNotFoundException fe) {
-			logger.info("context index has not been created yet: " + fe.getMessage());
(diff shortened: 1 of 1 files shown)


core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
```diff
@@ -100,8 +101,13 @@
 
 		try {
 			readContextsFromFile();
-		} catch (IOException e) {
-			logger.info("could not read context index: " + e.getMessage(), e);
+		} catch (FileNotFoundException fe) {
+			logger.info("context index has not been created yet: " + fe.getMessage());
+			initializeContextCache();
+			writeContextsToFile();
+			logger.info("context index construction complete");
```
hmottestad: Propose .debug(...) since the user usually doesn't need to know about this.

barthanssens: good idea
FWIW, this PR also reduces the noise in the output of the unit tests, since quite a few NativeStores are created in temp directories

Issue #1585, opened 2019-10-06
Improve federation repository config to support inline definition of members


aschwarte10:
The FedX based federation module already provides a repository factory such that it can be used for instance in the RDF4J workbench using a ttl configuration file.

Currently however the federation members are configured in an external dataConfig.ttl file in the data directory. This approach requires file system access to the RDF4J web-servers working directory.

The goal of this task is to allow defining federation members as part of the configuration turtle file inline, as this can also be done from the RDF4J workbench UI.

This change includes revisiting the namespace prefixes used within the factories, as they still point to http://www.fluidops.com/config/fedx#

PR #1586, opened 2019-10-06
#1585: support inline definition of federation members

aschwarte10:
GitHub issue resolved: #1585 

As of this change tbe federation members can be specified inline in the
repository config turtle file, i.e. no separate dataConfig.ttl file is
required.

### DIFF
--- a/tools/federation/src/test/resources/tests/rdf4jserver/config.ttl
+++ b/tools/federation/src/test/resources/tests/rdf4jserver/config.ttl
@@ -3,7 +3,7 @@
 #
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
 @prefix rep: <http://www.openrdf.org/config/repository#>.
-@prefix fedx: <http://www.fluidops.com/config/fedx#>.
+@prefix fedx: <http://rdf4j.org/config/federation#>.
 
 [] a rep:Repository ;
    rep:repositoryImpl [
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXFactory.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXFactory.java
@@ -16,11 +16,15 @@ import org.eclipse.rdf4j.federated.cache.MemoryCache;
 import org.eclipse.rdf4j.federated.endpoint.Endpoint;
 import org.eclipse.rdf4j.federated.endpoint.EndpointFactory;
 import org.eclipse.rdf4j.federated.endpoint.ResolvableEndpoint;
+import org.eclipse.rdf4j.federated.endpoint.provider.NativeRepositoryInformation;
+import org.eclipse.rdf4j.federated.endpoint.provider.ResolvableRepositoryInformation;
+import org.eclipse.rdf4j.federated.endpoint.provider.SPARQLRepositoryInformation;
 import org.eclipse.rdf4j.federated.exception.FedXException;
 import org.eclipse.rdf4j.federated.repository.FedXRepository;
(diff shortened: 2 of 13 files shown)


aschwarte10:
Note: this PR is still WIP and requires one more iteration. This PR serves as a basis for discussion for namespace changes (i.e. prefixes for the configuration parameters)

tools/federation/src/test/resources/tests/rdf4jserver/config.ttl
```diff
@@ -9,7 +9,14 @@
...
       fedx:fedxConfig "fedxConfig.prop" ;
-      fedx:dataConfig "dataConfig.ttl" ;
+      fedx:member [
+         fedx:store "ResolvableRepository" ;
+         fedx:repositoryName "endpoint1" 
+      ] ,
+      [
+         fedx:store "ResolvableRepository" ;
+         fedx:repositoryName "endpoint2" 
+      ]
```
aschwarte10: @jeenbroekstra can you please provide your feedback on this configuration file, and particularly define which namespace prefix we want to use in the future for the configuration settings?

This PR  currently contains the functional part and I would like to use the opportunity to finalize namespace prefixes here.

Current prefix:

fedx: <http://www.fluidops.com/config/fedx#>

Proposals:

fedx: <http://www.openrdf.org/config/federation#>.

Note: FedX is implemented as a Repository, basically extending the SailRepository (because it requires some parts of the API that are not available to the SAIL). The main functional part is still a SAIL.

The existing federation sail uses properties like: http://www.openrdf.org/config/sail/federation#member

Once we have this final, I will also start with the write-up of the documentation.

jeenbroekstra: For new vocabulary, I'd prefer using rdf4j.org rather than openrdf.org (which is a domain we no longer own). How about http://rdf4j.org/config/federation# ?

jeenbroekstra:
Looks like there's a formatting problem with your code. Could you run `mvn formatter:format` and commit the changed files? 

I'll take a look at why the builds needs 20 minutes to determine this - it shouldn't run the full compile + test suite until _after_ it's check the syntax/formatting everywhere.

aschwarte10:
> Looks like there's a formatting problem with your code. Could you run `mvn formatter:format` and commit the changed files?
> 
> I'll take a look at why the builds needs 20 minutes to determine this - it shouldn't run the full compile + test suite until _after_ it's checked the syntax/formatting everywhere.

Yes, I noticed this. I don't know how this slipped through, as I have the Eclipse formatter active (particularly as an automatic save action). Seems to generally work as I didn't have other formatting problems. Will run the check in the future locally to see if there are any issues.

tools/federation/src/test/resources/tests/rdf4jserver/config.ttl
aschwarte10: Sounds good, will adjust the examples and provide a new commit

jeenbroekstra:
> Yes, I noticed this. I don't know how this slipped through, as I have the Eclipse formatter active (particularly as an automatic save action). Seems to generally work as I didn't have other formatting problems. Will run the check in the future locally to see if there are any issues.

Odd - looks like it was complaining about trailing whitespaces (or tabs) on empty lines. I'll take a quick look at the formatter settings. Out of curiosity: which version of Eclipse are you working with?

aschwarte10:
I recently switched to the latest version of Eclipse: 2019-09. 
After checking again and applying the formatter again on the file, the formatting was applied correctly.
In any case, it's working now, and I will have an eye on this in the future

Issue #1596, opened 2019-10-10
SPIN - Configurable feature set


hmottestad:
Make the feature set in SPIN configurable so the user can enable or disable:
 - validation
 - construct
 - ????

hmottestad:
Limiting thing to just configurable validation.

PR #1597, opened 2019-10-10
Issues/1596 configurable spin

hmottestad:
GitHub issue: #1596   

* added setting to enabled/disable SPIN constraint validation which is slow (half the time is used for this during a delete operation)
* added benchmark to compare with/without constraint validation
*


jeenbroekstra:
Looks good. But this should have into develop rather than master: it's new feature, not a bugfix.

hmottestad:
That is very true @jeenbroekstra . I can patch over this and merge it into develop over the weekend.

hmottestad:
I'm planning on doing the following.

 - Syncing develop up with master. 
 - Take a file copy of the "sail" directory in develop
 - Revert this branch on master
 - Merge master into develop
 - This feature will then be gone from both develop and master
 - Then I can paste inn my files from before  and create a new PR

Issue #1598, opened 2019-10-10
Automate release


hmottestad:
Automate the release process.

hmottestad:
Looks like I missed somethings.
 - tell user to run job in jenkins (since this isn't automatic 😞)
 - push tag, no need to ask the user if it's ok since it doesn't actually do anything.

jeenbroekstra:
I'll try and update the developer documentation to make this script the "go to" approach for release handling.

PR #1599, opened 2019-10-10
Issus/1598 automate release

hmottestad:
GitHub issue resolved: #1598   

* 
* 
*

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 echo "Enabling fail on error";
 set -e -o pipefail
@@ -10,6 +10,15 @@ case "$choice" in
   n|N ) exit;;
   * ) echo "unknown response, exitting"; exit;;
 esac
+
+
+
+ if  !  mvn -v | grep -q "Java version: 1.8."; then
+         echo "You need to use Java 8!";
+         echo "mvn -v";
+         echo "";
+        exit 1;
+ fi
 
 
  if  ! git status --porcelain --branch | grep -q "## master...origin/master"; then
@@ -40,9 +49,12 @@ MVN_VERSION=$(xmllint --xpath "//*[local-name()='project']/*[local-name()='versi
 
 echo "Your maven version is: ${MVN_VERSION}"
 
+read -n 1 -s -r -p "Press any key to continue (ctrl+c to cancel)"; printf "\n\n";
+
 #Remove backup files. Finally, commit the version number changes:
 mvn versions:commit
 
+git branch --delete --force releases/${MVN_VERSION}
 git checkout -b releases/${MVN_VERSION}
 git commit -s -a -m "release ${MVN_VERSION}"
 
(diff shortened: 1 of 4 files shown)


hmottestad:
what do you think of this kind of script @jeenbroekstra ?

release.sh
```diff
@@ -0,0 +1,96 @@
...
+read -p "Start the release process (y/n)?" choice
+case "${choice}" in
+  y|Y ) echo "";;
+  n|N ) exit;;
+  * ) echo "unknown response, exiting"; exit;;
+esac
+
+
+if  !  mvn -v | grep -q "Java version: 1.8."; then
+         echo "You need to use Java 8!";
```
jeenbroekstra: I don't think we need to be this strict, exactly. The maven compiler is set to target 1.8, so even if you use a newer version of Java, it should still produce Java 8-compatible bytecode.

hmottestad:
> Note, by the way, that our developer workflow documentation doesn't fully capture this. This is because previously, we got around the issue by just not recording the patch version in the snapshot - so we didn't have 3.0.1-SNAPSHOT and 3.0.2-SNAPSHOT, we just had 3.0-SNAPSHOT, so it doesn't need an update after release. I wanted to use more accurate snapshot versioning when we merged into a single repo, but perhaps we need to move back to that approach.

I think maybe we can handle this  creating a merge branch. After creating a release in master and incrementing the snapshot, we can create  a new branch where we set the version to the same as in develop and merge that branch into develop. Then merge master into develop. 

Need to test it, but in my head it sounds like it should work.

jeenbroekstra:
> > Note, by the way, that our developer workflow documentation doesn't fully capture this. This is because previously, we got around the issue by just not recording the patch version in the snapshot - so we didn't have 3.0.1-SNAPSHOT and 3.0.2-SNAPSHOT, we just had 3.0-SNAPSHOT, so it doesn't need an update after release. I wanted to use more accurate snapshot versioning when we merged into a single repo, but perhaps we need to move back to that approach.
> 
> I think maybe we can handle this by creating a merge branch. After creating a release in master and incrementing the snapshot, we can create a new branch where we set the version to the same as in develop and merge that branch into develop. Then merge master directly into develop just to make git look pretty.
> 
> Need to test it, but in my head it sounds like it should work.

Another alternative approach is this: _before_ updating the pom version in master, merge master into develop (so that all meaningful changes are there). Then, update the version on master, and then merge with develop again using `git merge -s ours`.

release.sh
jeenbroekstra: As always this is not completely true as the build currently fails when using Java 11, of course. Keep it in, we can change it once we have build running on java 11 :)

jeenbroekstra:
> Another alternative approach is this: _before_ updating the pom version in master, merge master into develop (so that all meaningful changes are there). Then, update the version on master, and then merge with develop again using `git merge -s ours`.

FWIW I just did this manually for the 3.0.1 release and it works perfectly.

hmottestad:
One question about develop. I noticed that the <parent> tag points to 3.0.1-SNAPSHOT in develop even though the rest of the Pom references 3.1.0-SNAPSHOT. Is this the way it’s supposed to be?

jeenbroekstra:
> One question about develop. I noticed that the tag points to 3.0.1-SNAPSHOT in develop even though the rest of the Pom references 3.1.0-SNAPSHOT. Is this the way it’s supposed to be?

I'm not sure what you mean. What tag are you referring to?

hmottestad:
Huh. I was seeing it when I was creating the PR from the script. But now I can't reproduce it.

PR #1603, opened 2019-10-12
Sync develop with master

hmottestad:
Sync develop with master


hmottestad:
I guess the build failure here is related to https://github.com/eclipse/rdf4j/pull/1601 @jeenbroekstra

Issue #1604, opened 2019-10-12
Wiremock random port


hmottestad:
Wiremock port is hardcoded. Use random port provided by wiremock instead.

PR #1605, opened 2019-10-12
#1604 wiremock random port

hmottestad:
GitHub issue resolved: #1604  

* random port in wiremock
* 
*


hmottestad:
@jeenbroekstra I was thinking we should use random ports for wiremock. What do you think?

Issue #1600, opened 2019-10-10
Maven warning/error


hmottestad:
Getting an error about a dependency using a beta version of javax.el. Seems strange:

```
[INFO] ----------------< org.eclipse.rdf4j:rdf4j-http-server >-----------------
[INFO] Building RDF4J: HTTP server 3.0.1-SNAPSHOT                       [70/83]
[INFO] --------------------------------[ war ]---------------------------------
[WARNING] The POM for org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 is missing, no dependency information available
[WARNING] Failed to retrieve plugin descriptor for org.eclipse.m2e:lifecycle-mapping:1.0.0: Plugin org.eclipse.m2e:lifecycle-mapping:1.0.0 or one of its dependencies could not be resolved: Failure to find org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced
[WARNING] The POM for org.glassfish:javax.el:jar:3.0.1-b06-SNAPSHOT is missing, no dependency information available
[WARNING] The POM for org.glassfish:javax.el:jar:3.0.1-b07-SNAPSHOT is missing, no dependency information available
[WARNING] The POM for org.glassfish:javax.el:jar:3.0.1-b08-SNAPSHOT is missing, no dependency information available
[WARNING] The POM for org.glassfish:javax.el:jar:3.0.1-b11-SNAPSHOT is missing, no dependency information available
[INFO] 
```

I also don't know what org.eclipse.m2e:lifecycle-mapping is or if it is related to the javax.el issue.

barthanssens:
Which maven version (and Java VM) are you using ?

hmottestad:
```
$ mvn -v
Apache Maven 3.6.1 (d66c9c0b3152b2e69ee9bac180bb8fcc8e6af555; 2019-04-04T21:00:29+02:00)
Maven home: /usr/local/Cellar/maven/3.6.1/libexec
Java version: 1.8.0_144, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk1.8.0_144.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.14.6", arch: "x86_64", family: "mac"
```

hmottestad:
[...]

PR #1612, opened 2019-10-15
GH-1600 m2e lifecycle plugin hack no longer necessary for eclipse m2e

jeenbroekstra:
GitHub issue addressed: #1600  

* pom entry for m2e-lifecycle-plugin (which is not an actual plugin but just a dummy entry for Eclipse m2e) is no longer necessary in more recent Eclipse.
* tested with Eclipse 2019-09 
* removes annoying warning when using commandline maven.
* doesn't fully resolve the issue as that is also about warnings related to glassfish dependencies


jeenbroekstra:
Looks like I reformatted the pom.xml to use spaces instead of tabs. Not sure what happened there, will try to fix.

jeenbroekstra:
Experimentally using squash and merge. Not yet advising that everyone follows suit.

Issue #1611, opened 2019-10-14
rdf4j-runtime packaging is too low level in 3.0.1 compared to 3.0.0


hmottestad:
Look at the difference for the compile dependencies between 3.0.0 and 3.0.1.

## 3.0.1:
https://mvnrepository.com/artifact/org.eclipse.rdf4j/rdf4j-runtime/3.0.1
<img width="536" alt="Screenshot 2019-10-14 at 12 51 23" src="https://user-images.githubusercontent.com/797185/66746144-54a80480-ee81-11e9-9088-54a77702ab91.png">

## 3.0.0:
https://mvnrepository.com/artifact/org.eclipse.rdf4j/rdf4j-runtime/3.0.0
<img width="600" alt="Screenshot 2019-10-14 at 12 51 55" src="https://user-images.githubusercontent.com/797185/66746166-68536b00-ee81-11e9-9de5-977fa2be3ec0.png">

Looks like the way that 3.0.1 is creating a fat jar is very different from 3.0.0. 3.0.0 seems to just depend on a bunch of other modules, while 3.0.1 is just including all the code from those modules directly.

This makes a big difference for end users because the rdf4j-runtime could previously be used with a bunch of exclusions for things like spin and lucene modules if the user didn't want these (usually because they cause conflicts). Now the rdf4j-runtime can pretty much only be used if you are willing to include everything.

hmottestad:
@jeenbroekstra any chance you know how to convert back to how the runtime module was built for 3.0.0?

jeenbroekstra:
I'm having a look. It's surprising to me that rdf4j-runtime is producing a fat jar at all. I may goofed when merging the various pom configurations as part of the repo merge.

hmottestad:
https://mvnrepository.com/artifact/org.eclipse.rdf4j/rdf4j-runtime/3.0.2

Your fix worked @jeenbroekstra

PR #1615, opened 2019-10-17
GH-1611 remove unnecessary fat jar build from runtime pom

jeenbroekstra:
GitHub issue resolved: #1611   

* rdf4j-runtime is now a pom-packaged dependency
* removed shaded jar build process (I still think that got in there by mistake, it's never intentionally been a shaded jar before)
* tested that this behaves as expected by having a local scratchpad app that used rdf4j-runtime as its only dependency.


hmottestad:
I'm not sure how to test this myself other than what you did. I can see that there was no jar created for the runtime, which I think is correct. LGTM.

After you've merged this I can try to create a new release 3.0.2 with my script.

jeenbroekstra:
> I'm not sure how to test this myself other than what you did. I can see that there was no jar created for the runtime, which I think is correct. LGTM.
> 
> After you've merged this I can try to create a new release 3.0.2 with my script.

Thanks Havard. Happy for you to do a 3.0.2 release with the script, but, but there's one or two other fixes I'd like to get into this release. Shall we say Thursday? Let's continue this on the dev list.

Issue #1405, opened 2019-05-05
SPARQL: Variables from outside group are visible inside group


cygri:
The following query's query pattern has two group graph patterns:
```sparql
SELECT * {
  { BIND ("a" AS ?a) }
  { BIND (?a AS ?b) } 
}
```
According to the SPARQL spec, the result of the query is defined as the result of evaluating each group independently, and then joining the results:

- Evaluating the first group should yield a single solution with `?a` bound to `"a"`.
- Evaluating the second group should yield a single solution with no variables bound. Each group is evaluated independently, and `?a` is not bound within the group, so the `BIND` leaves `?b` unbound.
- Joining both results should yield a single solution with `?a` bound to `"a"`.

In RDF4J, however, the query returns a single solution with both `?a` and `?b` bound to `"a"`. This is incorrect.

Similar problems exists for other language constructs. For example, changing the second group to an `OPTIONAL` shows the same problem. But changing it to a subquery does not.

jeenbroekstra:
Can you give an example query with an optional pattern that shows this issue? Reason I ask is that first glance the issue appears to be specific to how BIND is processed, but I may be overlooking something.

cygri:
What I meant is, it's not just group graph patterns that are affected, but if the second group was replaced by, say, `OPTIONAL { BIND (?a AS ?b) }`, it shows the same problem (or so I seem to remember—haven't tried it again now).

It may well be specific to `BIND`.

jeenbroekstra:
I have just realized that this fix is incomplete: while all the mechanics are put in place, the query planner doesn't actually _use_ the new logic to determine what join strategy to use. :man_facepalming: 

That'll teach me for adding a fix without a regression test.

PR #1616, opened 2019-10-18
GH-1405 fixed scope handling in join iterator

jeenbroekstra:
GitHub issue resolved: #1405   

* use merge-join for group patterns to handle scope, with special case
exception for FILTER (NOT) EXISTS

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/JoinIterator.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/JoinIterator.java
@@ -17,6 +17,7 @@ import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.query.Binding;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.QueryEvaluationException;
+import org.eclipse.rdf4j.query.algebra.Filter;
 import org.eclipse.rdf4j.query.algebra.Join;
 import org.eclipse.rdf4j.query.algebra.TupleExpr;
 import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategy;
@@ -69,7 +70,7 @@ public class JoinIterator extends LookAheadIteration<BindingSet, QueryEvaluation
 
 				if (leftIter.hasNext()) {
 					TupleExpr rightArg = join.getRightArg();
-					if (TupleExprs.isGraphPatternGroup(rightArg) && !TupleExprs.isFilterExistsFunction(rightArg)) {
+					if (TupleExprs.isGraphPatternGroup(rightArg) && !(rightArg instanceof Filter)) {
 						// leftiter bindings are out of scope for the right arg, so we merge afterward.
 						BindingSet next = leftIter.next();
 						rightIter = new MergeIteration(next, new BindingSetFilterIteration(next,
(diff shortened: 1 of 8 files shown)


jeenbroekstra:
The fix seems to cause a regression in the federation sail, got two failing tests there.

jeenbroekstra:
@aschwarte10 is this something you might be able to shed some light on? I've got two failing tests in the (old) federation sail due a change I made in the JoinIterator algorithm, put I can't put my finger on why. Is there an implicit assumption somewhere in the federated evaluation strategy that left argument results automatically get injected into right argument evaluation?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/JoinIterator.java
```diff
@@ -61,7 +68,15 @@ protected BindingSet getNextElement() throws QueryEvaluationException {
 				rightIter.close();
 
 				if (leftIter.hasNext()) {
-					rightIter = strategy.evaluate(join.getRightArg(), leftIter.next());
+					TupleExpr rightArg = join.getRightArg();
+					if (TupleExprs.isGraphPatternGroup(rightArg) && !TupleExprs.isFilterExistsFunction(rightArg)) {
+						// leftiter bindings are out of scope for the right arg, so we merge afterward.
+						BindingSet next = leftIter.next();
+						rightIter = new MergeIteration(next, new BindingSetFilterIteration(next,
```
aschwarte10: I did a bit of debugging for the no longer working query pattern from org.eclipse.rdf4j.sail.federation.FederationQueryTest.test(), i.e.

```
{ ?person b:name ?name ; b:desc ?desc }
```

What I could observe by adding the following additional debugging code in line 73:

```
System.out.println(Iterations.asList(
	strategy.evaluate(rightArg, new EmptyBindingSet())));
```

and moreover the following debugging output between after line 74

```
System.out.println(Iterations.asList(
	new BindingSetFilterIteration(next,
	strategy.evaluate(rightArg, new EmptyBindingSet()))));
```

I see that we are loosing the second relevant bindingset from the inner join because of the BindingSetFilterIteration.

I am not sure if it is correct at all in the first place that this Join is interpreted as GraphPatternGroup, my feeling is that this is already the problem. However, I do not yet understand what GraphPatternGroups really represent.

Does this help to get a step further?

jeenbroekstra: It definitely does, thanks. And you're probably right about the cause as well. A graph pattern group is mostly a syntactical construct so it has no real place in the algebra. However, the SPARQL algebra is based on bottom-up evaluation while RDF4J's engine typically uses a more top-down approach. This is mostly not a problem but in some cases (such as this one) it messes with variable scoping, which is why we have some exception cases built-in. 

I'll take a closer look at why this is marked as a group pattern and see if we can somehow bypass that without the code becoming too obscure.

jeenbroekstra: I've discovered something relevant: in breaking down the query to send to various federation members, the Federation sail produces this:
```
SELECT ?name ?person ?desc
WHERE {
{?person <urn:schema:b:name> ?name  .}
{?person <urn:schema:b:desc> ?desc  .}
}
```

Notice the extra braces around each statement pattern. It's this that causes the parser to mark each pattern as a separate group graph pattern. I'll dig a bit deeper to find out why it's necessary to produce the query in this way. Perhaps these extra braces can be eliminated.

jeenbroekstra:
This still has a number of failing compliance tests and will need more time.

jeenbroekstra:
Verified locally that this now passes all compliance tests.

Issue #1618, opened 2019-10-18
AssertionError thrown from AbstractSailConnection when using Federation


ieurich:
When adding multiple repositories to a federation and evaluating a query which contains a NaryJoin the query is executed over multiple threads. This can result in AbstractSailConnection.startUpdate throwing an assertion error because multiple threads attempt to flush the same federation connection instance concurrently. This also results in the query failing to return the expected results.

Here is an example test which reproduces the problem. On my machine this test reproduces the issue maybe once in ten runs on average. 

```
public void testAssertionErrorReproduction() {
        
        final ValueFactory vf = SimpleValueFactory.getInstance();
        final BNode address = vf.createBNode();
        final BNode anotherAddress = vf.createBNode();
        
        final ModelBuilder builder = new ModelBuilder();
        builder
            .setNamespace("ex", "http://example.org/")
            .subject("ex:Picasso")
                .add(RDF.TYPE, "ex:Artist")
                .add(FOAF.FIRST_NAME, "Pablo")
                .add("ex:homeAddress", address)
            .subject("ex:AnotherArtist") 
                .add(RDF.TYPE, "ex:Artist")
                .add(FOAF.FIRST_NAME, "AnotherArtist")
                .add("ex:homeAddress", anotherAddress)
            .subject(address)           
                .add("ex:street", "31 Art Gallery")
                .add("ex:city", "Madrid")
                .add("ex:country", "Spain")
            .subject(anotherAddress)           
                .add("ex:street", "32 Art Gallery")
                .add("ex:city", "London")
                .add("ex:country", "UK");

        final Model model = builder.build();
        final Repository repo1 = new SailRepository(new MemoryStore());
        repo1.initialize();
        repo1.getConnection().add(model);
        
        final Repository repo2 = new SailRepository(new MemoryStore());
        repo2.initialize();
        
        final Federation fed = new Federation();
        fed.addMember(repo1);
        fed.addMember(repo2);

        final String ex = "http://example.org/";
        final String queryString =
                "PREFIX rdf: <" + RDF.NAMESPACE + ">\n" +
                "PREFIX foaf: <" + FOAF.NAMESPACE + ">\n" +
                "PREFIX ex: <" + ex + ">\n" +  
                "select (count(?persons) as ?count) {\n" +
                "   ?persons rdf:type ex:Artist ;\n"
                + "          ex:homeAddress ?country .\n"
                + " ?country ex:country \"Spain\" . }";
        
[...]

PR #1619, opened 2019-10-18
Issues/1618 assertion error in abstract sail connection

hmottestad:
GitHub issue resolved: #1618

 - Added synchronization to flushPendingUpdates() because this method is called from read operations and should be thread safe. The federation sail has a parallel join operation that assumes the all read operations against the base sails are thread safe.

### DIFF
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
@@ -343,7 +343,7 @@ public abstract class AbstractSailConnection implements SailConnection {
 	}
 
 	@Override
-	synchronized public void flush() throws SailException {
+	public void flush() throws SailException {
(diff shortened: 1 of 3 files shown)


hmottestad:
Need to run compliance tests.

hmottestad:
@jeenbroekstra @ieurich this is my proposed fix. I have a good test based on code from @ieurich that reproduces the issue every time on my laptop. The issue is some multithreaded use of the same base connection by the FederationSail, this triggers a race condition because not all "read" methods are thread safe. There might still be more issues, as I've discovered during my ShaclSail development, but I think synchronizing flushPendingUpdates() should do the trick for this particular issue.

hmottestad:
@jeenbroekstra can i merge this?

jeenbroekstra:
> @jeenbroekstra can i merge this?

Yep, lgtm, sorry for the delay.

jeenbroekstra:
@hmottestad  Don't forget to squash!

Issue #1621, opened 2019-10-20
Improve FedX federation configuration to support injection of RepositoryResolver


aschwarte10:
With the current implementation it is quite cumbersome to use FedX within the RDF4J workbench, particularly if it is desired to use repositories managed by the repository manager as federation members. Particularly, the RepositoryResolver (i.e. the reference to the repository manager) needs to be injected manually (e.g. by modifying the RDF4J webapp).

Using the interface RepositoryResolverClient it is possible to mark Repositories to accept a RepositoryResolver. This interface makes the RepositoryManager aware to automatically inject the RepositoryResolver.

The goal of this task is to automatically inject the repository resolver if FedX is managed by a RepositoryManager, e.g. from the RDF4J workbenchj

aschwarte10:
Implementation has been merged as of above PRs

PR #1622, opened 2019-10-20
#1621: inject RepositoryResolver into FedX repository

aschwarte10:
GitHub issue resolved: #1621

As of this change the RepositoryResolver (e.g. the reference to the
repository manager) is automatically injected to the FedXRepository, if
FedX is managed from a RepositoryManager.

This is for instance the case when FedX is used in the RDF4J workbench.

### DIFF
--- a/tools/federation/build/test/rdf4j-server/WEB-INF/rdf4j-http-server-servlet.xml
+++ b/tools/federation/build/test/rdf4j-server/WEB-INF/rdf4j-http-server-servlet.xml
@@ -32,9 +32,9 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 		</constructor-arg>
 	</bean>
 	
-	<!-- Inject the local repository manager as RepositoryResolver into FedX -->
+	<!-- Inject the local repository manager as RepositoryResolver into FedX for tests -->
 	<bean id="fedxRepositoryResolver"
-		class="org.eclipse.rdf4j.federated.repository.FedXRepositoryResolverBean"
+		class="org.eclipse.rdf4j.federated.server.FedXRepositoryResolverBean"
 		scope="singleton" init-method="init" destroy-method="destroy">
 		<property name="repositoryResolver" ref="rdf4jRepositoryManager" />
 	</bean>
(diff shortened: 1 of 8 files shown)


aschwarte10:
Just a note for the review to make it simpler: the inner class "RepositoryWrapper" is now externalized into FedXRepositoryWrapper such that it can extend additional interfaces. 

Note that in a next step (separate change) I will try to avoid the FedXRepositoryWrapper class entirely and do the initialization properly in FedXRepository. This wrapper was helper structure.

As of this change it is basically possible to just put the rdf4j-tools-federation jar into the WEB-INF/lib of the workbench server webapp and configure a federation from the UI. The RepositoryManager (i.e. resolver is now automatically injected).

@jeenbroekstra as a remark for my plans: ideally for the release we can achieve to have the FedX binary code directly available in the workbench / server WAR package. Then user's can just apply a template and have the federation feature. I will track this as an issue and we can follow up on how to best achieve this in the issue.

aschwarte10:
Quick update: I have successfully tested the binary rdf4j-tools-federation-3.1.0-SNAPSHOT.jar within a rdf4j-server / rdf4j-workbench.

Note: for my test I have used a 3.0.1 deployment, and manually put the above jar into the WEB-INF/lib folder of the server webapp.

As currently there is no repository template defined (i.e the federation cannot be selected from the UI), I created it in the file system with the following configuration (my-federation/config.ttl)

```
#
# RDF4J configuration template for a FedX Repository
#
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix rep: <http://www.openrdf.org/config/repository#>.
@prefix fedx: <http://rdf4j.org/config/federation#>.

[] a rep:Repository ;
   rep:repositoryImpl [
      rep:repositoryType "fedx:FedXRepository" ;
      fedx:member [
         fedx:store "ResolvableRepository" ;
         fedx:repositoryName "repo1" 
      ] ,
      [
         fedx:store "ResolvableRepository" ;
         fedx:repositoryName "repo2" 
      ]
   ];
   rep:repositoryID "my-federation" ;
   rdfs:label "FedX Federation" .
```

Moreover, I created two NativeStore repositories within RDF4J workbench. The federation is working fine.

We should consider a closer integration of the FedX federation with the RDF4J server and workbench.

jeenbroekstra:
@aschwarte10 I pushed a minor fix in your pom, it didn't include a dependency for repository-manager.

PR #1624, opened 2019-10-22
Sync develop up with master

hmottestad:


hmottestad:
@jeenbroekstra could you take a look at the conflict?

jeenbroekstra:
@hmottestad fixed. Looks like it choked on a whitespace char somewhere. Odd.

Issue #1262, opened 2019-01-26
Deprecate SPIN support


jeenbroekstra:
Now that SHACL support is maturing I think we need to make a decision on whether we choose to continue support SPIN reasoning. From my perspective everything SPIN can do, SHACL can do as well (or better), there's a fairly straightforward migration from SPIN notation to SHACL, and there's no real need to continue to support both. 

We also need to focus our resources: there's only so many developers on this project with so many hours, and the lead developer of the original SPIN engine is no longer very actively involved. 

In my opinion: we should deprecate SPIN and focus any future development and maintenance on SHACL instead.

barthanssens:
To be clear, by deprecated, do you mean "bug-fixes only", or "to be removed in 3.0" ?

jeenbroekstra:
I mean bug fixes only, really. I wouldn't remove until the next major release, and not until we are sure that we can fully support everything SPIN offers in the new SHACL engine. 

@hmottestad pointed out somewhere else that there are use cases for SPIN that SHACL doesn't really cover. I don't quite understand why that is the case, as [this article](http://spinrdf.org/spin-shacl.html) seems to suggest that SHACL can do everything SPIN can. Havard, maybe I misunderstood and you only meant to suggest that our _current_ SHACL implementation doesn't fully cover SPIN use cases?

hmottestad:
The sparql based “inferencer” in SPIN was not standardized in SHACL as far as I can tell. The features your article mentions are not mentioned in the standard. I think TopQuadrant got a lot of pushback from the community and ended up having to focus SHACL on the usecase of validation. From what I can tell there is no rule/sparql based “inference” in SHACL.

hmottestad:
In general I’m positive to deprecating and removing SPIN. I would like to develop a lightweight alternative though, for those who still want to use SPARQL for “inference”. A simple SAIL that takes UPDATE queries and runs them until no more changes are produced. 

The test suite for SPIN is the ultimate proof of why SPIN should be removed. If I remember correctly part of it is still @ignored because it’s too unstable and slow.

hmottestad:
Btw. This is the document that is referenced wrt. inference: 
https://www.w3.org/TR/shacl-af

“Publication as a Working Group Note does not imply endorsement by the W3C Membership. This is a draft document and may be updated, replaced or obsoleted by other documents at any time. It is inappropriate to cite this document as other than work in progress.”

[...]

PR #1625, opened 2019-10-22
#1262 deprecate the SpinSail

hmottestad:
GitHub issue resolved: #1262  

* Deprecated the SpinSail and added comments that explain why and recommend alternative solutions
* 
*

### DIFF
--- a/core/sail/sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
+++ b/core/sail/sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
@@ -32,8 +32,8 @@ import java.util.List;
 /**
  * <p>
  * The SpinSail is currently deprecated. If you are using SPIN to validate your data then it is recommended to move to
- * SHACL with the ShaclSail. SHACL does not support any of the SPARQL inference features of SPIN, however some were
- * proposed in "SHACL Advanced Features" but were not made a part of the W3C Recommendation in 2017.
+ * SHACL with the ShaclSail. Currently, the SHACL W3C Recommendation only supports validation, and has no
+ * equivalent to SPIN's inference features.
  * </p>
  * <p>
  * Deprecating the SpinSail has been discussed at https://github.com/eclipse/rdf4j/issues/1262 and can be summarized
--- a/core/sail/sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
+++ b/core/sail/sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
@@ -32,8 +32,8 @@ import java.util.List;
 /**
  * <p>
  * The SpinSail is currently deprecated. If you are using SPIN to validate your data then it is recommended to move to
- * SHACL with the ShaclSail. Currently, the SHACL W3C Recommendation only supports validation, and has no
- * equivalent to SPIN's inference features.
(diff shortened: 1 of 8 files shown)


core/sail/sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
```diff
@@ -29,6 +29,23 @@
...
+ * Deprecating the SpinSail has been discussed at https://github.com/eclipse/rdf4j/issues/1262 and can be summarized
+ * with there being no developers actively supporting it and that SPIN in itself is no longer recommended by
+ * TopQuadrant. Do not expect the SpinSail to scale in any way and expect simple delete operations to take seconds to
+ * complete. There are still a number of open issues in GitHub connected to the SpinSail, none of these are likely to
+ * get fixed.
+ * </p>
+ *
+ * @since 3.1.0 2019
+ */
+@Deprecated
```
hmottestad: Please read the above copy and let me know what you think.

core/sail/sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
```diff
@@ -29,6 +29,23 @@
 import java.util.ArrayList;
 import java.util.List;
 
+/**
+ * <p>
+ * The SpinSail is currently deprecated. If you are using SPIN to validate your data then it is recommended to move to
+ * SHACL with the ShaclSail. SHACL does not support any of the SPARQL inference features of SPIN, however some were
+ * proposed in "SHACL Advanced Features" but were not made a part of the W3C Recommendation in 2017.
```
jeenbroekstra: I'd rephrase this:

 "SHACL does not support any of the SPARQL inference features of SPIN, however some were proposed in "SHACL Advanced Features" but were not made a part of the W3C Recommendation in 2017." 

to something shorter, like: "Currently, the <a href="https://www.w3.org/TR/shacl/">SHACL W3C Recommendation</a> only supports validation, and has no equivalent to SPIN's inference features."

hmottestad:
Also need to change this: https://rdf4j.org/documentation/programming/spin/ to reflect that SPIN is no longer recommended.

core/sail/sail-spin/src/test/java/org/eclipse/rdf4j/sail/spin/SpinSailTest.java
```diff
@@ -141,6 +147,39 @@ public void testSpinxRule() throws Exception {
...
+		conn.commit();
+		conn.begin();
+
+		conn.add(ex2, FOAF.KNOWS, ex3);
+		conn.add(ex3, FOAF.KNOWS, ex4);
+		conn.commit();
+
+		assertStatements("testDeepConstructRule-expected.ttl");
+
+	}
```
hmottestad: @jeenbroekstra 
I've added this test here to show the lack of transactional support in the SpinSail. Is there some way for us to make this kind of scenario function? Eg. by having the SpinSail always recompute every rule on every transaction?

hmottestad: Essentially, the current implementation only triggers a computation when a matching type statement is added. So if it is added in the first transaction, and later we add more data that should actually trigger the rule....then nothing is inferred.

hmottestad: It's actually a little bit better, it's triggered when a resource is modified if that resource has a matching type.

Issue #1623, opened 2019-10-22
Method equals() in BindingSetAssignment.java is implemented incorrectly.


uditsharma7:
According to the current implementation of [equals()](https://github.com/eclipse/rdf4j/blob/master/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java#L51), two BindingSetAssignment objects are always equal irrespective of bindingSets present in them. 

**Solution**
Modify the equals() and hashcode() method
```
@Override
public boolean equals(Object other) {
	if (this == other) {
		return true;
	}

	if (other == null || !(other instanceof BindingSetAssignment)) {
		return false;
	}
	BindingSetAssignment that = (BindingSetAssignment) other;

	if (!that.getBindingNames().equals(this.getBindingNames()))
		return false;

	Iterator<BindingSet> thisIt = this.getBindingSets().iterator();
	Iterator<BindingSet> thatIt = that.getBindingSets().iterator();

	while (thisIt.hasNext() && thatIt.hasNext()) {
		if (!thisIt.next().equals(thatIt.next())) {
			return false;
		}
	}
	if (thisIt.hasNext() || thatIt.hasNext())
		return false;

	return true;
}

@Override
public int hashCode() {
	int hashCode = "BindingSetAssignment".hashCode();
	Iterator<BindingSet> thisIt = this.getBindingSets().iterator();
	while (thisIt.hasNext()) {
		hashCode ^= thisIt.next().hashCode();
	}
	return hashCode;
}
```

PR #1629, opened 2019-10-23
#1623 Fixed equals() and hashcode()

uditsharma7:
Signed-off-by: Udit Sharma <usharma898@gmail.com>

GitHub issue resolved: #1623 

* Fixed equals and hashcode in BindingSetAssignment
* 
*

### DIFF
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
@@ -49,13 +49,34 @@ public class BindingSetAssignment extends AbstractQueryModelNode implements Tupl
 	}
 
 	@Override
-	public boolean equals(Object other) {
-		return other instanceof BindingSetAssignment;
+	public boolean equals(Object obj) {
+		if (this == obj)
+			return true;
+		if (obj == null)
+			return false;
+		if (getClass() != obj.getClass())
+			return false;
+		BindingSetAssignment other = (BindingSetAssignment) obj;
+		if (bindingNames == null) {
+			if (other.bindingNames != null)
+				return false;
+		} else if (!bindingNames.equals(other.bindingNames))
+			return false;
+		if (bindingSets == null) {
+			if (other.bindingSets != null)
+				return false;
+		} else if (!bindingSets.equals(other.bindingSets))
+			return false;
+		return true;
 	}
 
 	@Override
 	public int hashCode() {
-		return "BindingSetAssignment".hashCode();
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((bindingNames == null) ? 0 : bindingNames.hashCode());
+		result = prime * result + ((bindingSets == null) ? 0 : bindingSets.hashCode());
+		return result;
 	}
 
 	@Override
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
@@ -53,22 +53,11 @@ public class BindingSetAssignment extends AbstractQueryModelNode implements Tupl
 	public boolean equals(Object obj) {
 		if (this == obj)
 			return true;
-		if (obj == null)
-			return false;
-		if (getClass() != obj.getClass())
+		if (obj == null || getClass() != obj.getClass())
 			return false;
 		BindingSetAssignment other = (BindingSetAssignment) obj;
-		if (bindingNames == null) {
-			if (other.bindingNames != null)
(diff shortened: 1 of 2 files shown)


core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
```diff
@@ -49,13 +49,34 @@
...
 	}
 
 	@Override
 	public int hashCode() {
-		return "BindingSetAssignment".hashCode();
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((bindingNames == null) ? 0 : bindingNames.hashCode());
+		result = prime * result + ((bindingSets == null) ? 0 : bindingSets.hashCode());
+		return result;
```
hmottestad: This looks like a great catch! Any chance that Objects.hash(....) could be used? https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash(java.lang.Object...)

hmottestad:
@rdf4j-bot run tests

jeenbroekstra:
Thanks for this @uditsharma7! The build currently fails with a test failure in the SpinSail - from a glance it looks as if it is the test itself that is the issue here but it will need some eyeballs. I would also like to run this through a full compliance test run before merging as there may be side effects in the query planner that won't show up until we go through full end-to-end testing.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
uditsharma7: @hmottestad Updated the code. Also replaced Object.equals with Objects.equals in equals() implementation.

uditsharma7:
@jeenbroekstra I think [Test5](https://github.com/eclipse/rdf4j/blob/master/core/spin/src/test/resources/testcases/test5.ttl) is correct. It is also present [here](https://spinrdf.org/sp.html#sp-Values). 
In the [Unit Test](https://github.com/eclipse/rdf4j/blob/master/core/spin/src/test/java/org/eclipse/rdf4j/spin/SpinParserTest.java#L68) the text query is parsed as 
```
Projection
   ProjectionElemList
      ProjectionElem "country"
      ProjectionElem "code"
   Extension
      ExtensionElem (country)
         Var (name=country)
      ExtensionElem (code)
         Var (name=code)
      BindingSetAssignment ([[country=ex:Germany;code="de"], [country=ex:Norway;code="no"], [country=ex:Endor]])
```
and the spin query is parsed as 
```
Projection
   ProjectionElemList
      ProjectionElem "country"
      ProjectionElem "code"
   Extension
      ExtensionElem (country)
         Var (name=country)
      ExtensionElem (code)
         Var (name=code)
      BindingSetAssignment ([[country=ex:Germany;code="de"], [country=ex:Norway;code="no"], [country=ex:Endor;code=http://spinrdf.org/sp#undef]])
```
It seems there is another bug in [SpinParser.java](https://github.com/eclipse/rdf4j/blob/master/core/spin/src/main/java/org/eclipse/rdf4j/spin/SpinParser.java#L1383) due to which it is treating `sp:undef` as an IRI(http://spinrdf.org/sp#undef) while it should be treated as missing binding.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/BindingSetAssignment.java
```diff
@@ -49,13 +50,19 @@
 	}
 
 	@Override
-	public boolean equals(Object other) {
-		return other instanceof BindingSetAssignment;
+	public boolean equals(Object obj) {
+		if (this == obj)
+			return true;
+		if (obj == null || getClass() != obj.getClass())
```
hmottestad: Ref: https://stackoverflow.com/questions/1569688/an-efficient-equalsobject-o-implementation

Choice of .getClass() and instanceOf is relevant if we want to allow for subclassing or not. @jeenbroekstra any preference?

jeenbroekstra: I tend to prefer allowing subclasses, so I'd use `!(obj instanceof BindingSetAssignment)` instead.

jeenbroekstra:
> It seems there is another bug in [SpinParser.java](https://github.com/eclipse/rdf4j/blob/master/core/spin/src/main/java/org/eclipse/rdf4j/spin/SpinParser.java#L1383) due to which it is treating `sp:undef` as an IRI(http://spinrdf.org/sp#undef) while it should be treated as missing binding.

You're quite right @uditsharma7 . I am running some local tests on your issue branch and will push a fix for this directly to your branch (it's trivial enough that I don't want to bother with a separate bug report).

jeenbroekstra:
add to whitelist

PR #1633, opened 2019-10-25
#1598 adjustments for the script

hmottestad:
Signed-off-by: Håvard Ottestad <hmottestad@gmail.com>

GitHub issue resolved: #1598   

* 
* 
*

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -112,6 +112,7 @@ mvn clean;
 
 MVN_CURRENT_SNAPSHOT_VERSION=$(xmllint --xpath "//*[local-name()='project']/*[local-name()='version']/text()" pom.xml)
 
+# replace "SNAPSHOT" with ""
 MVN_VERSION_RELEASE="${MVN_CURRENT_SNAPSHOT_VERSION/-SNAPSHOT/}"
 
 MVN_NEXT_SNAPSHOT_VERSION="$(increment_version $MVN_VERSION_RELEASE 3)-SNAPSHOT"
@@ -238,5 +239,16 @@ mvn clean install -DskipTests
 
 echo "DONE!"
 
+# the news file on github should be 302 if the release is 3.0.2, so replace "." twice
+NEWS_FILE_NAME=$MVN_VERSION_RELEASE
+NEWS_FILE_NAME=${NEWS_FILE_NAME/./}
+NEWS_FILE_NAME=${NEWS_FILE_NAME/./}
 
+echo ""
+echo "You will now want to inform the community about the new release!"
+echo " - Go to https://github.com/eclipse/rdf4j-doc/tree/master/site/content/news and create rdf4j-${NEWS_FILE_NAME}.md"
+echo " - Edit the following file https://github.com/eclipse/rdf4j-doc/blob/master/site/content/release-notes/index.md"
+echo " - Edit the following file https://github.com/eclipse/rdf4j-doc/blob/master/site/content/download/_index.md"
+echo " - Post to Google Groups: https://groups.google.com/forum/#!forum/rdf4j-users"
+echo "     - Good example: https://groups.google.com/forum/#!topic/rdf4j-users/isrC7qdhplY"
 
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -159,19 +159,10 @@ esac
...
-
-echo "Starting utomated upload with sftp. Timeout is set to 1 hour!"
-
-./sftp-onejar-upload.expect $username $password $MVN_VERSION_RELEASE
-
-echo "";
-echo "Upload complete";
-echo "";
+echo "You need to tell jenkins to start the release process."
+echo "Go to: https://ci.eclipse.org/rdf4j/job/rdf4j-deploy-release-ossrh/ (if you are on linux or windows, remember to use CTRL+SHIFT+C)."
```
jeenbroekstra: "Remember to use Ctrl-Shift-C"? What's that for?

release.sh
```diff
@@ -223,6 +214,24 @@ echo "You want to merge 'merge_master_into_develop_after_release_${MVN_VERSION_R
...
 
+git checkout $MVN_VERSION_RELEASE
+
+echo "";
+echo "One-jar build takes several minutes, this is the last step and the script will complete by itself when you continue."
+read -n 1 -s -r -p "Press any key to continue to one-jar build (ctrl+c to cancel)"; printf "\n\n";
+
+# build one jar
+mvn -Passembly clean install -DskipTests
+
```
jeenbroekstra: This doesn't just build the onejar, it also builds the SDK zip. And the problem is that due to a glitch in our javadoc aggregate configuration, you need to run command twice: once in the project root, and then a second time _inside_ the `assembly` directory. If you skip that second step, the SDK zip still gets created, but it does not include the javadoc documentation.

release.sh
```diff
@@ -223,6 +214,24 @@ echo "You want to merge 'merge_master_into_develop_after_release_${MVN_VERSION_R
...
+echo "Starting utomated upload with sftp. Timeout is set to 1 hour!"
+
+./sftp-onejar-upload.expect $username $password $MVN_VERSION_RELEASE
```
jeenbroekstra: I also asked on the mailinglist, but just for completeness sake: what kind of tool is 'expect'?

release.sh
hmottestad: Copy to clipboard. Since Linux and Windows use CTRL-C to copy everywhere except when you are in a terminal.

release.sh
hmottestad: Ok. I’ll adjust that.

release.sh
hmottestad: The sftp tool doesn’t allow you to pass a username and password. So it’s hard to automate an sftp upload. There were a number of workarounds I found when googling it, using expect was one of those options. It seems to be a fairly well known tool (and was installed by default on my Mac). It lets you automate interaction through stdin and stdout. So for sftp it going “run command”, “wait for login prompt”, “type password” and so on.

release.sh
```diff
@@ -223,12 +211,50 @@ echo "You want to merge 'merge_master_into_develop_after_release_${MVN_VERSION_R
...
+echo " - Create a new milestone for ${MVN_NEXT_SNAPSHOT_VERSION/-SNAPSHOT/} : https://github.com/eclipse/rdf4j/milestones/new"
+echo " - Close the ${MVN_VERSION_RELEASE} milestone: https://github.com/eclipse/rdf4j/milestones"
+echo "     - Make sure that all issues in the milestone are closed, or move them to the next milestone"
+echo "     - Go to the milestone, click the 'closed' tab and copy the link for later"
+echo " - Edit the following file https://github.com/eclipse/rdf4j-doc/blob/master/site/content/release-notes/index.md"
+echo " - Edit the following file https://github.com/eclipse/rdf4j-doc/blob/master/site/content/download/_index.md"
+echo " - Go to https://github.com/eclipse/rdf4j-doc/tree/master/site/content/news and create rdf4j-${NEWS_FILE_NAME}.md"
+echo " - Post to Google Groups: https://groups.google.com/forum/#!forum/rdf4j-users"
+echo "     - Good example: https://groups.google.com/forum/#!topic/rdf4j-users/isrC7qdhplY"
+echo " - Upload the javadocs " #TODO Where should they be uploaded and how?
```
hmottestad: @jeenbroekstra can you take a look at this list and see if it is complete. Also, where do you upload the javadocs?

Issue #1638, opened 2019-10-28
SHACL - thread pool not closed correctly


hmottestad:
There are a number of log messages about threads not exiting when running the ComplexLargeBenchmark with all iterations set to 1.

```
<JMH had finished, but forked VM did not exit, are there stray running threads? Waiting 24 seconds more...>

Non-finished threads:

Thread[DestroyJavaVM,5,main]

Thread[pool-2-thread-3,5,main]
  at sun.misc.Unsafe.park(Native Method)
  at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
  at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
  at java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
  at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1074)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1134)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  at java.lang.Thread.run(Thread.java:748)
```

hmottestad:
This turned out to be a bug in the benchmark code, just forgot to shutdown all the sails.

Issue #1223, opened 2018-12-26
Remove the need to initialize a Sail / Repository


jeenbroekstra:
In the SAIL and Repository APIs, a separate call to `initialize()` is needed to get a usable Sail / Repository object. We should remove the need for this separate call, and just have object initialization happen as and when needed (for example, on the first call to `getConnection()`).

## Note for third party Sail / Repository implementors

This is marked as a breaking change in the API because the API no longer _requires_ a client to explicitly call `init()` (or `initialize()`) before using the Repository or Sail object. Those methods still exist of course, and they _can_ still continue to use them as before - so none of this will break existing client code. All rdf4j-provided `Repository` and `Sail` implementations have been adapted to automatically initialize when necessary. Third party implementations of `Repository` or `Sail` may need to update their code to provide the same functionality. 

Related issues: #552 and #1243

jeenbroekstra:
A possible approach: instead of forcing the user to call `initialize`, we initialize lazily on the first operation on the Repository that requires it (typically, that will be when the user calls `getConnection`).

ansell:
Performing initialisation of things that may generate an exception or that require IO is not encouraged inside of constructors due to the impossibility of cleanup by a user if the constructor doesn't complete successfully and doesn't cleanup its partially constructed resources itself completely. 

Lazily calling `initialize` automatically using a combination of `if(AtomicBoolean.compareAndSet(true, false))` to ensure it only happens once and a `CountDownLatch` to wait on it completing, is something I have done in the past.

jeenbroekstra:
I quite agree that doing this in the constructor is not a good idea. Apart from the reasons you mention it also makes it difficult to deal with any configuration that happens using setter methods. As you say, lazy auto-init on first need is the way forward I think.

PR #1639, opened 2019-10-29
#1638 fixed bechmark, added test, use shutdownNow

hmottestad:
GitHub issue resolved: #1638  

* fixes benchmark
* add test and shutdown thread pool correctly
* also changed a method from public to package private, it should not have been public to begin with!

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -189,7 +189,7 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	}
 
-	private volatile ExecutorService[] executorService = new ExecutorService[1];
+	private final ExecutorService[] executorService = new ExecutorService[1];
 
 	public ShaclSail(NotifyingSail baseSail) {
 		super(baseSail);
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -43,6 +43,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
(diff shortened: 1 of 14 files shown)


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -311,22 +312,37 @@ public void initialize() throws SailException {
...
 
 		initialized.set(false);
+		executorService = null;
 		nodeShapes = Collections.emptyList();
 		super.shutDown();
 	}
 
-	public synchronized <T> Future<T> submitRunnableToExecutorService(Callable<T> runnable) {
+	synchronized <T> Future<T> submitRunnableToExecutorService(Callable<T> runnable) {
+		if (!initialized.get()) {
```
hmottestad: this might be overkill

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/benchmark/ComplexLargeBenchmark.java
```diff
@@ -113,7 +113,7 @@ public void setUp() {
 		System.gc();
 	}
 
-	@TearDown
+	@TearDown(Level.Invocation)
```
hmottestad: this was what was causing the error messages in the log

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -675,4 +701,14 @@ private static String resourceAsString(String s) throws IOException {
...
+	@Override
+	protected void finalize() throws Throwable {
+		super.finalize();
+		if (initialized.get()) {
+			logger.error("ShaclSail was finalized (garbage collected) without shutdown() having been called first.");
+		}
+		if (executorService != null) {
+			executorService.shutdownNow();
+		}
+	}
```
hmottestad: Also considering finalize(). This helps clean up if the user has forgotten to shutdown the sail. Also logs an error. Maybe this should just be a warning?

hmottestad: finalize() is deprecated in java 9, and it's recommended anymore. Phantom references are the way to go now: https://docs.oracle.com/javase/8/docs/api/java/lang/ref/PhantomReference.html

hmottestad: Looks fairly complex compared to just using finalize()

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: Actually, we changed the contract on Sails in 3.0: it should auto-initialize when necessary, instead of throwing an exception. See #1223

hmottestad:
@rdf4j-bot run tests

Issue #1643, opened 2019-10-30
Automated code quality check


barthanssens:
Run Netbeans Inspect and Transform on the code base (unused imports, try-with-resources, possible issues ...)

Somewhat related to #790

- [x] unused imports
- [x] try-with-resources
- [x] diamond inference
- [x] @Override annotation
- [x] multi catch instead of repeating catch blocks
- [x] isEmpty() instead of size() == 0
- [x] isEmpty() instead of equals("")
- [x] lambda expression instead of anonymous class
- [x] auto (un)boxing

hmottestad:
For try-with-resource you should check if this case shows up:

<img width="1493" alt="Screenshot 2019-10-30 at 11 05 17" src="https://user-images.githubusercontent.com/797185/67848580-3d197e80-fb05-11e9-9611-ed27d7399ede.png">

This is a false positive and should not be refactored.

hmottestad:
@barthanssens  I hope you don't mind. After checking for try-with-resource with the intellij inspection I ended up fixing a couple of them.

hmottestad:
`isEmpty() instead of size() == 0` will probably lead to some performance improvements ;)

barthanssens:
Not at all :-) Most of the try-with-resources were fixed in a previous run
 
Most issues Netbeans I/T noticed are suggestions to use diamond inference, and using multi catch instead of multiple catch blocks repeating the same code

And a list of suggestions to make fields final

PR #1646, opened 2019-10-30
Issues/#1643 code quality

barthanssens:
GitHub issue resolved: #1643  

* Various code syntax changes (no change in functionality)


hmottestad:
lgtm

barthanssens:
It's a matter of taste I guess, but personally I prefer using lambdas instead of anonymous inner classes because it saves a few lines of boiler plate + often also a tab on vertical space, which makes things IMHO a bit more readable

(Same for multicatch instead of separate catches with same code)

hmottestad:
I'm all for lambdas. Also, the future is lambdas and not anonymous classes. Looking at Graal I also reckon that there will be a lot more optimisations heading for lambdas in the future.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
```diff
@@ -79,39 +79,36 @@ private ActiveTransactionRegistry() {
...
+						final Transaction entry = notification.getValue();
+						synchronized (primaryCache) {
+							if (!entry.hasActiveOperations()) {
+								// no operation active, we can decommission this entry
+								primaryCache.invalidate(transactionId);
+								logger.warn("deregistered expired transaction {}", transactionId);
+							} else {
+								// operation still active. Reinsert in secondary cache.
+								secondaryCache.put(transactionId, entry);
+							}
```
jeenbroekstra: Interesting: the Jenkins build fails on this, saying that (on line 105) `secondaryCache` may not have been initialized.

barthanssens: indeed, maybe a if not null may solve it, I'll look into it

hmottestad:
I think this should be targeted to master. But maybe a bit late now.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
hmottestad: It’s a scoping issue. Lambdas expect all variables in their scope to effectively be final. Because the secondaryCache is bound after the scope has been determined for the lambda, the lambda assumes that’s its uninitialised. 

Can you attach the removal listener after build() has bee called?

Also, I think whole approach is a bit funky. Is there no place to attach a willSoonBeRemovedListener instead?

barthanssens: There seems to be no method to attach the listener after building the cache... a workaround seems to be to create a getter, and use that inside the lambda instead of the variable.

Another dirty way would be to just surpress the warning, but IMHO that isn't a good practice...

hmottestad:
There is a risk we might get some conflicts when I merge master into develop. So I can hold off on that @barthanssens until you've merged this branch.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/ActiveTransactionRegistry.java
jeenbroekstra: I like the setter approach.

For the wider issue of the design of this cache, I'd happily revisit when we have time. I remember designing it this way because I was somewhat hampered by the limitations of the Guava cache.

Issue #1515, opened 2019-08-27
Testsuite slim down


jeenbroekstra:
The current tests run (on Jenkins) for close to an hour (54 minutes). Even when disabling compliance tests, building and executing all unit tests still takes over 20 minutes. 

We should look into "slimming down" the testuite: identify large/long-running tests, and determine if these tests are still necessary. If not, scrap them. If they are, see if they can be rewritten to run faster.

While we're at it we should also pay attention to the output from tests. Some tests spit out thousands of logger or even STDOUT lines, and it makes trawling through the build log near-impossible. 

- [x] look into updating/removing SailConcurrency test
- [ ] look into streamlining RepositoryConnection test suite
- [ ] look into not executing full RepositoryConnection test suite on ever possible combination of stacked + backend sail. 
- [ ] investigate if all maven build plugins currently used are still necessary
- [x] investigate use of Jenkins incremental or parallel builds
- [x] look into options for fully isolated Jenkins agents.

jeenbroekstra:
@hmottestad could you take a closer look at the SHACL tests? Executing the build+tests on the SHACL module takes almost 8 minutes (it's responsible for nearly half the build time, on its own).

I have a hunch this is related to the recent changes you made to include benchmark tests. 

I've also noticed that if Jenkins executes two builds in parallel, the JMH benchmarks become flaky: it complains about not being able to create a lock:

> Error Message
> 
> ERROR: Unable to acquire the JMH lock (/tmp/genie.rdf4j/jmh.lock): already taken by another JMH instance, exiting. Use -Djmh.ignoreLock=true to forcefully continue.
> 
> Stacktrace
> 
> org.openjdk.jmh.runner.RunnerException: ERROR: Unable to acquire the JMH lock (/tmp/genie.rdf4j/jmh.lock): already taken by another JMH instance, exiting. Use -Djmh.ignoreLock=true to forcefully continue.
> 	at org.eclipse.rdf4j.sail.shacl.benchmark.BenchmarkJunitTests.test(BenchmarkJunitTests.java:36)

I think we should make sure not to run JMH benchmarks as part of the Jenkins build, at least not as part of the regular build-verification jobs. There is an option available to set Jenkins up using fully independent agents, which we could potentially use to schedule regular (nightly?) runs of certain benchmark suites, instead.

hmottestad:
Why does each parallel build server in Jenkins share a temporary directory. Must be some way to enable some decent isolation in Jenkins.

[...]

PR #1648, opened 2019-10-31
GH-1515 moved store-related tests out of compliance

jeenbroekstra:
GitHub issue resolved: #1515  

* converting various "compliance" tests into unit tests
* removed rdf4j-store-testsuite and rdf4j-store-compliance

The idea is to make our test rig simpler and more easily ran incrementally.


core/sail/api/src/test/java/org/eclipse/rdf4j/sail/helpers/NotifyingSailConnectionWrapperTest.java
```diff
@@ -85,24 +82,8 @@ public void tearDown() throws SailException {
...
 		wrapper.addConnectionListener(listener);
-		addStatement("a");
-		assertThat(listener.getCount()).isEqualTo(1);
-		removeStatement("a");
-		assertThat(listener.getCount()).isEqualTo(0);
+		verify(delegate).addConnectionListener(listener);
 		wrapper.removeConnectionListener(listener);
-		addStatement("b");
-		assertThat(listener.getCount()).isEqualTo(0);
-	}
```
hmottestad: Do some of the other tests check that you can add statements after the listener has been removed?

core/sail/base/pom.xml
```diff
@@ -51,9 +51,49 @@
...
 	<build>
 		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<executions>
+					<execution>
+						<goals>
+							<goal>test-jar</goal>
+						</goals>
```
hmottestad: Do we need a test jar?

hmottestad: does this approach mean we need to build twice? Or that eclipse/IntelliJ won’t be able to build properly?

jeenbroekstra: We need a test jar because normally maven modules don't include the test code. There are tests which are _defined_ at a certain abstract level but which need to executed in concrete implemenations. To be able to inherit from a common set of  tests, we use these test jars. This basically replaces the testsuite module we had for this purpose, and make it easier to just execute these tests as part of the module itself, instead of in a separate compliance module.

And no, it doesn't mean we need to build twice, I've been careful (hopefully) not to introduce any cyclic dependencies. This is a pretty common maven approach, it builds fine in Eclipse and though I haven't tested in IntelliJ I'd be very surprised if it had trouble with it.

core/sail/api/src/test/java/org/eclipse/rdf4j/sail/helpers/NotifyingSailConnectionWrapperTest.java
jeenbroekstra: I honestly don't know. I thought the whole test was badly written and simplified it so that I could at least integrate it. The point of this particular test is to check that connection listeners are properly added/remved, not to check if  statements can be added. It might be good to add more tests to this class to check for those things - part of reason I'm restructuring all this is also just to make it easier to find the right place to add a test.

hmottestad: I have a bunch of tests for the notifying sail in the SHACL module. When I developed it I realized that there were a few edge cases that the notifying sail didn’t notify correctly. Things like adding and removing and re-adding the same statement in a single transaction.

jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
@rdf4j-bot run tests

PR #1652, opened 2019-11-01
Master to develop merge branch

hmottestad:


jeenbroekstra:
Why did you need a separate merge branch for this?

hmottestad:
Merge conflicts. This is the only way I know to make jenkins run the tests.

jeenbroekstra:
Let me have a quick look before you merge this. I mean it's fine to do it this way but I want to understand what the conflicts are and if there's perhaps an easier way to handle this in the future.

jeenbroekstra:
Ah I see the conflict - couple of variable renames in RemoteRepositoryManager - so nothing to do with the poms. I typically just do a local merge and conflict resolution, run compliance locally, and then push. But if you prefer handling via a merge branch that's fine. Just be aware that the Jenkins PR verify job does not run full compliance, it only runs the unit tests.

jeenbroekstra:
In retrospect I realize that these kinds of "parallel branch syncs" are probably still better done using merge commits rather than squash and merge....

hmottestad:
Yes. Never squash when merging master into develop.

Issue #1635, opened 2019-10-26
Refine repository initialization lifecycle of FedX federation


aschwarte10:
Currently the lifecycle of the FedXRepository federation does not clearly follow the RDF4J repository API, e.g. the FedXFactory returns an initialized repository (and initializing again does not work).

Because of this for use-cases such as RDF4J workbench we had to introduce workarounds like the FedXRepositoryWrapper.

In addition, we currently use static fields to manage the federation, having the consequence that there can only be a single configuration in one given JVM.

This issue has the goal overcome the current limitations by restructuring and refactoring the code appropriately. Note that this does not include any functional changes

aschwarte10:
With all above changes we are currently good. The lifecycle is now clean and this issue can be closed

Issue #552, opened 2016-09-07
initialize is a reserved method name in Ruby


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-461)

In the Ruby language 'initialize' is a reserved method name for constructors and methods with the name 'initialize' cannot be called from within Ruby. Ruby is the only language over the last couple of years that has shown significant increase in it user base. Sun has shown its commitment to running Ruby on the JVM (JRuby). JRuby allows Java classes/objects to act like Ruby classes/objects in Ruby. We may an increase in JRuby users wanting to use Sesame from within Ruby as it popularity increases. If the 'initialize' methods were renamed to 'init' this would not be a problem.

ansell:
Would JRuby baulk if we added init as a new default method and made it always call "initialize"? Ie, would JRuby refuse to acknowledge the call from init into initialize?

That wouldn't be something we could backport to the Java-7 code, but we would just say that it is only supported by RDF4J-2+.

jeenbroekstra:
We'll have to do this in phases. I suggest we introduce a new `init()` method and deprecate the old `initialize()`. Final removal of the offending method can then happen in a future major release (not 3.0, since that is too close to release, and we need to give devs time).

PR #1654, opened 2019-11-03
Issues/#1635 repository lifecycle

aschwarte10:
GitHub issue resolved: #1635 

This change cleans the initialization of the FedX Repository lifecycle.

Main features:
- follow semantics of RDF4J Repository API
-- particularly w.r.t to initialization and shutdown
- no static management of federation (=> multiple federations per JVM)
- properly support external RepositoryResolver and FederatedServiceResolver

Note: this change is a huge refactoring touching many classes. In most cases this is due to replacing the static access to all helper structures with instance data from the FederationContext.

I am proposing this change now, because this is basically the last chance for the required cleanup before the federation is released in RDF4J.

For the review it is recommended to go commit wise as there is a logical order. Commits by themselves should be ok to follow (with the explanations from the commit message)

Note: Either as of this PR or as a follow up I am also planning to remove the static Config class and make it a member of the federation context.


tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedX.java
```diff
@@ -113,16 +127,29 @@ public ValueFactory getValueFactory() {
 	public void initialize() throws SailException {
```
jeenbroekstra: `Sail#initialize` is deprecated, I think you should rename this to `Sail#init`.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FederationManager.java
```diff
@@ -85,124 +72,27 @@
...
-		this.cache = cache;
-		this.executor = executor;
-		QueryManager.instance = new QueryManager(this, repo); // initialize the singleton query manager
+	public FederationManager() {
+
 	}
 
-	public FedX getFederation() {
-		return federation;
+	public void initialize(FedX federation, FederationContext federationContext) {
```
jeenbroekstra: Same here - for the Sail, the Repository, and the RepositoryManager, we renamed `initialize` to `init`. See #552 for details.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/QueryManager.java
```diff
@@ -52,30 +57,43 @@
...
-	protected QueryManager(FederationManager federationManager, Repository repo) {
+	public QueryManager(FederationManager federationManager, Repository repo) {
 		this.federationManager = federationManager;
 		this.repo = repo;
+
+		BigInteger lastQueryId = new BigInteger("0");
+		this.nextQueryID = new AtomicBigInteger(lastQueryId);
+	}
+
+	public void initialize() {
```
jeenbroekstra: I realize this one is not overriding any existing methods, but given #552 perhaps we should rename this to `init` as well?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/endpoint/EndpointBase.java
```diff
@@ -145,11 +146,11 @@ public long size() throws RepositoryException {
 	}
 
 	@Override
-	public void initialize() throws RepositoryException {
+	public void initialize(FederationContext federationContext) throws RepositoryException {
```
jeenbroekstra: As above

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedX.java
aschwarte10: Good catch. I now make use of AbstractSail functionality and all the inherited base functionality. Required some more smaller adjustments, see last commit

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FederationManager.java
aschwarte10: Makes sense to align

tools/federation/src/main/java/org/eclipse/rdf4j/federated/QueryManager.java
aschwarte10: Done

tools/federation/src/main/java/org/eclipse/rdf4j/federated/endpoint/EndpointBase.java
aschwarte10: Done

aschwarte10:
All review comments addressed. @jeenbroekstra would you please have one more look at the changes?

Issue #1642, opened 2019-10-29
Join of BIND clauses with UNION incorrectly scoped


jeenbroekstra:
Followup of GH-1405

```
prefix wd: <http://www.wikidata.org/entity/> 

select * {
  bind(wd:Q156578 as ?co)
  bind(strafter(str(?co),str(wd:)) as ?WD)
  {
    bind(?WD as ?WD1)
  } union {
    bind(?WD as ?WD2)
  }
}
```

Expected:

```
+-------------------------------------+-------------------------------------+-------------------------------------+-------------------------------------+
| co                                  | WD                                  | WD1                                 | WD2                                 |
+-------------------------------------+-------------------------------------+-------------------------------------+-------------------------------------+
| <http://www.wikidata.org/entity/Q156578>| "Q156578"^^xsd:string               |           |                                     |
| <http://www.wikidata.org/entity/Q156578>| "Q156578"^^xsd:string               |                                     |      |
+-------------------------------------+-------------------------------------+-------------------------------------+-------------------------------------+

```
Actual:
```

+-------------------------------------+-------------------------------------+-------------------------------------+-------------------------------------+
| co                                  | WD                                  | WD1                                 | WD2                                 |
+-------------------------------------+-------------------------------------+-------------------------------------+-------------------------------------+
| <http://www.wikidata.org/entity/Q156578>| "Q156578"^^xsd:string               | "Q156578"^^xsd:string               |                                     |
| <http://www.wikidata.org/entity/Q156578>| "Q156578"^^xsd:string               |                                     | "Q156578"^^xsd:string               |
+-------------------------------------+-------------------------------------+-------------------------------------+-------------------------------------+
```

See https://github.com/w3c/sparql-12/issues/103 

Root cause appears to be that the JoinIterator doesn't recognize the union as a scoped argument (it is not parsed as a groupgraphpattern).

PR #1660, opened 2019-11-04
GH-1642 avoid injecting join leftarg bindings when rightarg is union

jeenbroekstra:
GitHub issue resolved: #1642   

* check if rightArg is a Union
* added regression test


jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
It appears this exposes a flaw in the code for the LuceneSpinSail - hence the failing test. I never really understood the use case for that module, and it's not documented either. Given our intention to sunset Spin support I'm included to just disable those tests. But I'll try and take a closer look later first, perhaps there's an easy fix.

Issue #790, opened 2017-03-26
Add code quality plugins to jenkins build


jeenbroekstra:
We should look into adding some code quality plugins (like FindBugs, PMD, and/or CheckStyle) to the Jenkins build. This will make it easier for contributors to understand where their contributions need to be updated and will help us produce overall better code. 

- [x] CQ22184: Spotbugs 4.0.3
- [x] CQ22185: Spotbugs maven plugin 4.0.0

barthanssens:
FWIW, I've started using Netbeans Inspect and Transform.
When enabling all the bells and whistles, there are a few thousands suggestions in the report, although most of them can be either ignored (e.g. using static imports, that's a matter of taste IMHO, I prefer not to use static imports) or are easy to "fix" (missing override annotation or javadoc param etc...)

barthanssens:
On a related note, there is also SpotBugs, available as Eclipse and maven plugin https://spotbugs.readthedocs.io/en/latest/maven.html

barthanssens:
FWIW, I'm currently looking into spotbugs, looks promising:
* there is a Jenkins plugin (https://wiki.jenkins.io/display/JENKINS/Warnings+Next+Generation+Plugin)
* supported by maven, Eclipse (and older Netbeans, unfortunately not yet for recent Netbeans)
* an additional plugin for extra security checks is also available (OWASP, CWE) https://find-sec-bugs.github.io/

barthanssens:
Created  CQ 21366 (PB) in IpZilla

barthanssens:
Creating a new one for more recent spotbugs version + plugins

jeenbroekstra:
I closed this because the associated PR was closed - I get the feeling there was little enthusiasm for the idea in the end. Feel free to correct me and reopen though.

PR #1666, opened 2019-11-06
GH-790 quality plugin

barthanssens:
GitHub issue resolved: #790  

This may partially solve the issue since I did not touch the Jenkins config, though there seems to be a plugin that can use the XML output from these plugins (https://wiki.jenkins.io/display/JENKINS/Warnings+Next+Generation+Plugin) to generate nicer looking reports.

See also https://spotbugs.readthedocs.io/en/latest/maven.html and https://spotbugs.github.io/

SpotBugs will generate reports in xml format, that can be loaded in some IDEs (apparently from within Eclipse/Intellij), or by invoking the standalone GUI via mvn spotbugs:gui and opening the files manually.

Note that SpotBugs is a _bytecode_ analyzer, which means that one has to compile the sources first, and only then the plugin can be used to perform an analysis (mvn spotbugs:spotbugs)

Apparently there might also be some false positives when "final" fields are involved, but it appears to provide some interesting feedback, though I haven't compared the results with e.g. PMD or other tools.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,6 +138,35 @@
 			<build>
 				<plugins>
 					<plugin>
+						<groupId>com.github.spotbugs</groupId>
+						<artifactId>spotbugs-maven-plugin</artifactId>
+						<version>3.1.12</version>
+						<configuration>
+							<effort>Max</effort>
+							<threshold>Low</threshold>
+							<failOnError>false</failOnError>
+							<plugins>
+								<plugin>
+									<groupId>com.mebigfatguy.fb-contrib</groupId>
+									<artifactId>fb-contrib</artifactId>
+									<version>7.4.7</version>
+								</plugin>
+								<plugin>
+									<groupId>com.h3xstream.findsecbugs</groupId>
+									<artifactId>findsecbugs-plugin</artifactId>
+									<version>1.9.0</version>
+								</plugin>
+							</plugins>
+						</configuration>
+						<executions>
+							<execution>
+								<goals>
+									<goal>check</goal>
+								</goals>
+							</execution>
+						</executions>
+					</plugin>
+					<plugin>
 						<groupId>org.apache.maven.plugins</groupId>
 						<artifactId>maven-source-plugin</artifactId>
 						<executions>
--- a/pom.xml
+++ b/pom.xml
@@ -138,35 +138,6 @@
 			<build>
 				<plugins>
 					<plugin>
-						<groupId>com.github.spotbugs</groupId>
-						<artifactId>spotbugs-maven-plugin</artifactId>
-						<version>3.1.12</version>
-						<configuration>
-							<effort>Max</effort>
-							<threshold>Low</threshold>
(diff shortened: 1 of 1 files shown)


hmottestad:
I'm fairly sceptical of static code analyzers because they tend to give a lot of false positives per true positive. How many false positives does this currently give? And how easy is it to mark something as a false positive?

barthanssens:
In attachment an example from one of the modules (as XML and as HTML).

I think it does a rather nice job for what it is designed for (it might have caught an issue #1666), though it will need manual analysis and/or tweaking, so I guess it's more a run-once-in-a-while tool (unlike the formatting plugin)

E.g using a "+" to concatenate a string instead of using StringBuilder is slower but might be more readable, do we want to use the Nullable annotation to document methods that may return null, do we really need to provide a toString for each and every class, and various other nitty-gritty "best" practices

And it obviously won't catch potential big architectural issues.

[spotbugs.zip](https://github.com/eclipse/rdf4j/files/3816176/spotbugs.zip)

hmottestad:
I had a look, I think I prefer using the static code analyzer in IntelliJ. It is a lot easier to understand that issue when you can click yourself through to the code. Does findbugs find more bugs because it uses bytecode analysis compared to Intellij?

Of those ~300 warnings. How many do you reckon are actually bugs that have gone undetected until now?

barthanssens:
Well, I haven't checked all the reports yet, but in terms of actual bugs I guess it will only be a few real ones. Most of the others will be suggestions for improvement, and occasionally (a small dozen ?) it seems worth to actually implement the suggestions or look into the details.

I assume the low-hanging fruit was already picked (since you are using IntelliJ and I've done 
some Netbeans Inspect and Transform, and maybe Jeen or someone else has a similar tool in Eclipse...)

So no strong feelings about whether or not using this (or another) tool in Jenkins :-)

jeenbroekstra:
@barthanssens thanks for looking into this, and apologies for not reviewing sooner - ok if I get back to this after we've done the patch release this weekend, and perhaps also have a clearer idea of the role of Jenkins vs the Github actions that we've been experimenting with?

barthanssens:
No rush :-)

jeenbroekstra:
FWIW I'd be happy to merge this in as it doesn't impact our default flow - the plugin is not tied to the build/test life cycle AFAICT, you have to manually invoke it. So if you don't want to use it, you don't have to.

barthanssens:
I probably have to file a CQ 'works-with' request. 

FWIW, I'm analyzing the reports on the dev branch, nothing spectacular (like discouraging string concatenation, some hints on possible NPEs and suggestion to add Nullable annotations and minor thingies), though there are a few good hints to improve some pieces of code

jeenbroekstra:
Let's not merge this before the 3.1 release then, so that the CQ doesn't block the release.

barthanssens:
Still have to file a few CQs for the additional rulesets, so not merging it yet

Issue #1662, opened 2019-11-04
Elasticsearch backed sail store


hmottestad:
Create a sail store that stores and queries triples in Elasticsearch 6.x.

Consider
 - Bulkloading
 - optimizing for "hasStatement"
 - consider using cursor
 - considering caching the transaction in a local memorystore
 - consider caching in general for read operations
 - target higher performance for read operations rather than writes
 - this is meant for storing small datasets where transaction support is not needed, like storing reference data used in consortion with your current use of elasticsearch

Use the following for testing:

```xml
        <dependency>
            <groupId>pl.allegro.tech</groupId>
            <artifactId>embedded-elasticsearch</artifactId>
            <version>2.10.0</version>
            <scope>test</scope>
        </dependency>
```

```java
        private static EmbeddedElastic embeddedElastic;

	private static File installLocation = Files.newTemporaryFolder();


	@BeforeClass
	public static void beforeClass() throws IOException, InterruptedException {


		String version = "6.8.3";

		embeddedElastic = EmbeddedElastic.builder()
			.withElasticVersion(version)
			.withSetting(PopularProperties.TRANSPORT_TCP_PORT, 9350)
			.withSetting(PopularProperties.CLUSTER_NAME, "cluster")
			.withInstallationDirectory(installLocation)
			.withDownloadDirectory(new File("tempDownloads"))
//			.withPlugin("analysis-stempel")
//			.withIndex("cars", IndexSettings.builder()
//				.withType("car", getSystemResourceAsStream("car-mapping.json"))
//				.build())
//			.withIndex("books", IndexSettings.builder()
//				.withType(PAPER_BOOK_INDEX_TYPE, getSystemResourceAsStream("paper-book-mapping.json"))
//				.withType("audio_book", getSystemResourceAsStream("audio-book-mapping.json"))
//				.withSettings(getSystemResourceAsStream("elastic-settings.json"))
//				.build())
			.withStartTimeout(5, TimeUnit.MINUTES)
			.build();


		embeddedElastic.start();
	}
```

hmottestad:
We can now add triples and query for them. Next will be support for blank nodes and for values. After than maybe bulk loading. Scroll for getting statements and removal of statements.

barthanssens:
So you're using ES 6.8.x ?
That's a good thing IMHO, though it will require another round of CQs in IPZilla.

hmottestad:
I believe I switched back to the one we already using. The code above was initially copied from another project I have.

PR #1670, opened 2019-11-10
Issues/1662 elasticsearch storage sail

hmottestad:
GitHub issue resolved: #1662  

* 
* 
*


hmottestad:
### What works at the moment
 - adding, getting and removing statements
 - sparql

### What doesn't work
 - deleting a lot of data is really really slow
 - stats probably don't work

hmottestad:
@barthanssens wanna take a look at my ElasticsearchStore implementation that uses Elasticsearch as a backend to store statements.

hmottestad:
@rdf4j-bot run tests

hmottestad:
@rdf4j-bot run tests

hmottestad:
@rdf4j-bot run tests

hmottestad:
@rdf4j-bot run tests

hmottestad:
@rdf4j-bot run test

hmottestad:
@rdf4j-bot run tests

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -741,7 +753,7 @@ protected void clearNamespacesInternal() throws SailException {
 
 	private IsolationLevel getIsolationLevel() throws UnknownSailTransactionStateException {
 		if (isActive()) {
-			return super.getTransactionIsolation();
```
hmottestad: the use of super here meant it couldn't be overridden

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -614,12 +614,24 @@ public boolean removeInferredStatement(Resource subj, IRI pred, Value obj, Resou
 
 	private boolean remove(Resource subj, IRI pred, Value obj, SailDataset dataset, SailSink sink, Resource... contexts)
 			throws SailException {
+
+//		(subj == null || pred == null || obj == null) &&
+		if (!hasConnectionListeners()
+				&& sink instanceof SailSinkVersion2 && getIsolationLevel().isCompatibleWith(IsolationLevels.NONE)) {
+			return ((SailSinkVersion2) sink).deprecateByQuery(subj, pred, obj, contexts);
+		}
```
hmottestad: Going to need more performance testing at some point to see if deprecateByQuery is faster.

hmottestad:
@jeenbroekstra @barthanssens I think this is starting to get close to merging to develop and maybe an alpha release. All the compliance tests are passing now.

I made a new kind of store that is used as the base for the elasticsearch one. It simplifies the implementation a lot and supports up to READ_COMMITTED. 

I've also made a new SailSink interface based on the old one with a couple new methods. I didn't want to change the original interface or add to it because it changes a lot of code and it could be considered a breaking change to modify such an essential interface.

One thing I haven't done yet is to implement the config code stuff.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSinkVersion2.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * An extension of the SailSink interface to support Statements directly for better performance.
+ *
+ * @author Håvard Ottestad
+ */
+public interface SailSinkVersion2 extends SailSink {
+
+	/**
+	 * Removes a statement with the specified subject, predicate, object, and context.
```
hmottestad: change javadoc

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSinkVersion2.java
```diff
@@ -0,0 +1,38 @@
...
+	 *
+	 * @param statement The statement that should be removed
+	 * @throws SailException If the statement could not be removed, for example because no transaction is active.
+	 */
+	void deprecate(Statement statement) throws SailException;
+
+	/**
+	 * Removes all statements with the specified subject, predicate, object, and context. All four parameters may be
```
hmottestad: change javadoc

core/sail/elasticsearch-store/src/main/java/org/eclipse/rdf4j/sail/elasticsearchstore/ClientPool.java
```diff
@@ -0,0 +1,20 @@
...
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.elasticsearchstore;
+
+import org.elasticsearch.client.Client;
+
+/**
+ * @author Håvard Mikkelsen Ottestad
+ */
+public interface ClientPool extends AutoCloseable {
```
hmottestad: does this need to be public?

core/sail/elasticsearch-store/src/main/java/org/eclipse/rdf4j/sail/elasticsearchstore/ClientPoolImpl.java
```diff
@@ -0,0 +1,72 @@
...
+		return closed;
+	}
+
+	@Override
+	synchronized public void close() throws Exception {
+		if (!closed) {
+			closed = true;
+			if (client != null) {
+				client.close();
+				client = null;
```
hmottestad: swap these statements

core/sail/elasticsearch-store/src/main/java/org/eclipse/rdf4j/sail/elasticsearchstore/ElasticsearchDataStructure.java
```diff
@@ -0,0 +1,738 @@
...
+ * @author Håvard Mikkelsen Ottestad
+ */
+class ElasticsearchDataStructure implements DataStructureInterface {
+
+	private static final String MAPPING;
+
+	static final int BUFFER_THRESHOLD = 1024 * 16;
+	private final ClientPool clientPool;
+	private Set<Statement> addStatementBuffer = Collections.synchronizedSet(new HashSet<>());
+	private Set<ElasticsearchId> deleteStatementBuffer = Collections.synchronizedSet(new HashSet<>());
```
hmottestad: The rest of the code is now synchronized, so this can be a normal set now.

core/sail/elasticsearch-store/src/main/java/org/eclipse/rdf4j/sail/elasticsearchstore/ElasticsearchDataStructure.java
```diff
@@ -0,0 +1,738 @@
...
+						statement.getPredicate());
+			} else {
+				elasticsearchIdStatement = vf.createStatement(id, statement.getSubject(), statement.getPredicate(),
+						statement.getPredicate(), statement.getContext());
+			}
+
+		}
+
+		if (deleteStatementBuffer.size() >= BUFFER_THRESHOLD) {
+			flushRemoveStatementBuffer(clientPool.getClient());
```
hmottestad: flushRemoveStatementBuffer shouldn't need a Client

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSinkVersion2.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.sail.SailException;
+
+/**
+ * An extension of the SailSink interface to support Statements directly for better performance.
+ *
+ * @author Håvard Ottestad
+ */
+public interface SailSinkVersion2 extends SailSink {
```
jeenbroekstra: I'd prefer adding the additional methods as default methods to the original interface instead.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSinkVersion2.java
```diff
@@ -0,0 +1,39 @@
...
+	 */
+	void deprecate(Statement statement) throws SailException;
+
+	/**
+	 * Removes all statements with the specified subject, predicate, object, and context. All four parameters may be
+	 * null.
+	 *
+	 * @throws SailException If statements could not be removed, for example because no transaction is active.
+	 */
+	boolean deprecateByQuery(Resource subj, IRI pred, Value obj, Resource[] contexts);
```
jeenbroekstra: I don't think you need this additional method. The standard `deprecate` already allows params to be null (though admittedly it's awkwardly formulated in the javadoc).

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSinkVersion2.java
jeenbroekstra: The default implementation would be to just redirect to the subject-predicate-object-context version of the method.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -614,12 +614,23 @@ public boolean removeInferredStatement(Resource subj, IRI pred, Value obj, Resou
 
 	private boolean remove(Resource subj, IRI pred, Value obj, SailDataset dataset, SailSink sink, Resource... contexts)
 			throws SailException {
+
+		if (!hasConnectionListeners() && sink instanceof SailSinkVersion2
+				&& getIsolationLevel().isCompatibleWith(IsolationLevels.NONE)) {
+			return ((SailSinkVersion2) sink).deprecateByQuery(subj, pred, obj, contexts);
+		}
```
jeenbroekstra: Once you fold SailSinkVersion2 into the "normal" SailSink, this code can just go. Same with the bit on lines 629 and further.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSinkVersion2.java
hmottestad: I took a look at this.

The standard deprecate is never used in the code with nulls. It also doesn't support context[] which is needed to support queries. After introducing the deprecate(Statement statement) method there are no longer any calls to deprecate(subj, pred, obj, context).

Should I introduce the new method in the SailSink and implement it everywhere? It can't be implemented as default in SailSink because it needs access to a getStatement method.

hmottestad:
I’ve integrated a lot of the new sink code into the rest of ref4j. I need to make some benchmarks for memory and native stores so that I can measure before and after performance.

hmottestad:
@rdf4j-bot run tests

hmottestad:
Made a few more performance improvements to the ElasticsearchStore. There is a big thing missing: WAL. Write Ahead Logging. Essentially, while committing a transaction it gets written to elasticsearch in bulk, if it fails....then you will have half the transaction visible when the failure is fixed. To handle this I would need to implement a transaction log in elasticsearch, and also handle duplicate statements across multiple transactions if one transaction were to fail.

hmottestad:
## After
```
Benchmark                                               Mode  Cnt    Score    Error  Units
QueryBenchmark.groupByQuery                             avgt   10   11.574 ±  6.295  ms/op
QueryBenchmark.removeByQuery                            avgt   10    4.679 ±  1.006  ms/op
QueryBenchmark.simpleUpdateQueryIsolationNone           avgt   10   89.914 ±  3.779  ms/op
QueryBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt   10  343.143 ± 15.074  ms/op
```

hmottestad:
@rdf4j-bot run tests

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
jeenbroekstra: I just realized that part of the problem is that this private method seems to assume deprecate only accepts a single triple at a time, and does a dataset retrieval over which it iterates to remove individual triples. I can see how that could conceivably be inefficient. From what I can tell though, the reason we do that is not that `deprecate` doesn't accept wildcards (it does), but that we need a list of the statements that are removed to send to the listener.

jeenbroekstra:
FWIW happy to (async) pair on this if necessary - I have some spare time this Sunday.

hmottestad:
I believe changeset assumes that the arguments to deprecate are non null: https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java#L249

It’s essentially adding a statement to the model on that line. 

For the rest of the code there was not a single place that assumed that deprecate supported nulls.

There also seems to be a need for a return value for if a statement was deleted. If not sure if this is actually needed, because it’s hard to see which paths actually utilize the returned value.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
hmottestad: Yes. It’s that we need a list of statements for the listener and that we need to return true if a statement was removed. 

The code:

```
if (!hasConnectionListeners() && getIsolationLevel().isCompatibleWith(IsolationLevels.NONE)
 				&& !(sink instanceof Changeset)) {
```

 checks if there is a listener, that we are in isolation level null (which might be a bit too restrictive) and that the sink is not a changeset, because  changeset doesn’t support deprecatebyquery.

hmottestad: Btw there is a bug in GitHub. It’s showing some old code in the code window at the start of this chain of comments.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
```diff
@@ -252,6 +270,26 @@ public synchronized void deprecate(Resource subj, IRI pred, Value obj, Resource
 		}
 	}
 
+	@Override
+	public boolean deprecateByQuery(Resource subj, IRI pred, Value obj, Resource[] contexts) {
+		throw new UnsupportedOperationException("Not implemented");
```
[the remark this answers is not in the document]
hmottestad: deprecate in changeset assumes all non nulls.

hmottestad:
@rdf4j-bot run tests

hmottestad:
@rdf4j-bot run tests

hmottestad:
Test time differences:

org.eclipse.rdf4j.sail.elasticsearchstore.compliance.ElasticsearchStoreConnectionTest

 - local: 45 seconds
 - github ci: 50 seconds
 - jenkins: 6 minutes

hmottestad:
I've now deprecated the old `deprecate(sub, pred, obj, ctx)` method, I also made it `default` and made the `deprecate(statement)` the main method. Removed all use of the old deprecate method. Also found a minor optimization for the memory store, it's a bit ugly but gives a 30% higher performance for the remove method while in  READ_COMMITTED or higher.

hmottestad:
@jeenbroekstra I’m starting to feel that this is getting good enough for a merge and a potential 3.1.0 release. The code works and is fairly performant. I would like to start using this on one of my own projects to test it in the wild. 

One thing I do have a question about is the config code. Can you take a look and see if it conforms you how the rest of RDF4J works? I’m also wondering about how to handle required config fields. I’ve given it a shot, but I’m not sure if it’s correct.

jeenbroekstra:
@hmottestad I'll try and look at this in detail later this week. Happy to go with early release - let's mark it as experimental though (if it isn't already).

core/sail/elasticsearch-store/src/main/resources/META-INF/services/org.eclipse.rdf4j.sail.config.SailFactory
```diff
@@ -0,0 +1 @@
+org.eclipse.rdf4j.sail.extensiblestore.config.ElasticsearchSailFactory
```
jeenbroekstra: I think this needs to be `org.eclipse.rdf4j.sail.elasticsearchstore.ElasticsearchSailFactory`.

core/sail/elasticsearch-store/src/main/java/org/eclipse/rdf4j/sail/elasticsearchstore/config/ElasticsearchStoreConfig.java
```diff
@@ -0,0 +1,168 @@
...
+
+	public ElasticsearchStoreConfig() {
+		super(ElasticsearchStoreFactory.SAIL_TYPE);
+	}
+
+	@Override
+	public Resource export(Model graph) {
+		Resource implNode = super.export(graph);
+
+		graph.setNamespace("ms", NAMESPACE);
```
jeenbroekstra: Why `ms`? Perhaps change to something like `ess` to make conflicts with other namespace prefixes less likely.

core/sail/elasticsearch-store/src/main/resources/META-INF/services/org.eclipse.rdf4j.sail.config.SailFactory
hmottestad: Good catch thanks :)

core/sail/elasticsearch-store/src/main/java/org/eclipse/rdf4j/sail/elasticsearchstore/config/ElasticsearchStoreConfig.java
hmottestad: I think this wast copy-paste from somewhere else. Will change :)

hmottestad: Copy paste from MemoryStore => "ms"

Issue #1679, opened 2019-11-17
Make rdf4j-runtime dependent on rdf4j-storage and rdf4j-client


jeenbroekstra:
To make the chance of discrepancies between the various umbrella dependencies smaller, we should make sure that rdf4j-runtime derives its dependencies from rdf4j-storage and rdf4j-client.

PR #1681, opened 2019-11-17
GH-1679 only include rdf4j-storage poms 

jeenbroekstra:
(and by extension, rdf4j-client as well)

This ensures rdf4j-runtime only includes the libraries necessary for
anyone building RDF4J-based applications, and _doesn't_ include
additional dependencies like the workbench or console - which are not
commonly necessary for app dev purposes. This is mostly to consolidate and remove discrepancies between the various umbrella poms we have in play. 

GitHub issue resolved: #1679  

* apps depending on rdf4j-runtime pom dependency now include everything in rdf4j-storage and rdf4j-client
* it does _not_ include rdf4j-console, workbench and server modules, and therefore also _not_ things like logback.


hmottestad:
Only seems to include rdf4j-storage, was that intentional?

jeenbroekstra:
@hmottestad that module includes a lot of other ones, transitively. Do you think there's something missing?

hmottestad:
I don’t know if anything else is needed, so I reckon this is fine :)

hmottestad:
You mentioned RDF4J-client in your PR, so I just wondered if it should also have been included or if it is included transitively. I don’t know if it is needed.

jeenbroekstra:
Ah fair point. It is included transitively, but it might be clearer to include it explicitly.

PR #1684, opened 2019-11-21
merge master into develop

hmottestad:


hmottestad:
@rdf4j-bot run tests

testsuites/benchmark/pom.xml
```diff
@@ -60,6 +60,21 @@
 			<scope>compile</scope>
 		</dependency>
     </dependencies>
+    <profiles>
```
jeenbroekstra: It looks as if the merge fails because this pom already contains a profiles section in the develop branch. We'll need to reconcile that (either by merging this PR and then fixing, or by using a separate merge branch - either way is fine with me).

Issue #1685, opened 2019-11-21
3.0.x subClassOf inferences disappear - major bug ?


vujasm:
Hi,
It seems there is a bug (v3.0.2, 3.0.1, 3.0.0) that somehow causes subClassOf inferences (inferred statements) to disappear . My config is Native + RDFS. 

Here are the steps that should help you to quickly replicate the issue and check it further.
 
1.
`INSERT data { graph <urn:test>   {<urn:A> <http://www.w3.org/2000/01/rdf-schema#subClassOf> <urn:B> .   <urn:B> <http://www.w3.org/2000/01/rdf-schema#subClassOf> <urn:C> .  <urn:C> <http://www.w3.org/2000/01/rdf-schema#subClassOf> <urn:D> .} }`

 2.   `SELECT * WHERE {?s <http://www.w3.org/2000/01/rdf-schema#subClassOf>  <urn:D>}`
Results with inferred statements:
<urn:A>
<urn:B>
<urn:D>
<urn:C>

3.  `DELETE data { graph <urn:test>  {<urn:A> <http://www.w3.org/2000/01/rdf-schema#subClassOf>  <urn:B>} }`

4. `SELECT * WHERE {?s <http://www.w3.org/2000/01/rdf-schema#subClassOf>  <urn:D>}`
Result (with  inferred statements true) is only <urn:C>. ??  It is expecting to get <urn:B>,  <urn:D>, <urn:C>. So there is some issue here.

Further,
5. I insert back what I just deleted (hoping inferred statements will get back)
 `INSERT data { graph <urn:test>  {<urn:A> <http://www.w3.org/2000/01/rdf-schema#subClassOf> <urn:B>} }`

6. `SELECT * WHERE {?s <http://www.w3.org/2000/01/rdf-schema#subClassOf>  <urn:D>}` , with  inferred statements true, again returns <urn:C> only, which is wrong.

7 I insert new triple with new resource urn:AA
`INSERT data { graph <urn:test>  {<urn:AA> <http://www.w3.org/2000/01/rdf-schema#subClassOf> <urn:B>} }`

and now `SELECT * WHERE {?s <http://www.w3.org/2000/01/rdf-schema#subClassOf>  <urn:D>}` , with  inferred statements true, gives back
<urn:A>
<urn:B>
<urn:D>
<urn:AA>
<urn:C>

Now run 
```
DELETE data { graph <urn:test> 
[...]

PR #1688, opened 2019-11-24
GH-1685 subclassof inferencer bug

jeenbroekstra:
GitHub issue resolved: #1685  

* added regression test that demonstrates the problem


hmottestad:
I think this solution makes the code easier to understand. There is however a rather large performance regression which needs to be looked into.

In `ReasoningBenchmark` method `forwardChainingSchemaCachingRDFSInferencerMultipleTransactions` with param `"moreRdfs::12180"` the master branch shows 85 ms per operation, while this branch shows 600 ms per opertion. So roughly 7x slower. 

I've added a class for you to run the benchmark yourself: https://github.com/eclipse/rdf4j/blob/1dc0252f299e5eae927ee12f65dc8ff4f66834c3/testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/Main.java

jeenbroekstra:
Performance regression is likely caused by the additional check in`statementAdded`. I think we can eliminate that and instead change the boolean flag as part of `processSchemaStatement`

hmottestad:
LGTM - performance is good now :)

hmottestad:
I've also downgraded the JMH version. I can make a new issue for upgrading it.

PR #1689, opened 2019-11-24
Issues/#1635 fedx config

aschwarte10:
GitHub issue resolved: #1635 

This PR re-designs and cleans the configuration of FedX.

The main goal of this change is to avoid a static singleton configuration class, and rather manage all settings in the context of a given federation instance. Thus after this change we allow to have multiple federations in the same JVM (e.g. in a RDF4J workbench).

Note that the Config class is entirely removed and the settings have been moved to the FedXConfig class (which is managed per federation instance).

Furthermore note that the base directory can no longer be configured via settings, instead we now rely fully on the Repository#getDataDir().


aschwarte10:
@jeenbroekstra I think this should be the last big required refactoring on FedX to get it fully in shape for RDF4J integration. After this the implementation should be on a good path for the future.
Would be good if you can scan through this change.  Might look a bit big, but it is mostly moving things around

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedX.java
```diff
@@ -52,6 +52,8 @@
 
 	protected RepositoryResolver repositoryResolver;
 
+	protected File dataDir;
```
jeenbroekstra: Can this be `private`?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/QueryManager.java
```diff
@@ -75,7 +75,7 @@ public QueryManager(FederationManager federationManager, Repository repo) {
 	public void init() {
 
 		// initialize prefix declarations, if any
-		String prefixFile = Config.getConfig().getPrefixDeclarations();
+		String prefixFile = federationManager.federationContext.getConfig().getPrefixDeclarations();
```
jeenbroekstra: For consistency's sake I'd prefer the use of a getter method for `federationContext`. Can we make the field private and provide an accessor method instead?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedX.java
aschwarte10: Sure, done

tools/federation/src/main/java/org/eclipse/rdf4j/federated/QueryManager.java
aschwarte10: Done. Also addressed this in a few other places. See last commit

Issue #1634, opened 2019-10-26
maven sdk creation javadoc aggregate issue


jeenbroekstra:
For some reason the way the javadoc is aggregated as part of the creation of the SDK zip (in the assembly module) fails when `mvn -Passembly clean install` is run from the project root: it needs to be run a second time from _inside_ the `assembly/` directory, otherwise the SDK zip file does not contain the javadoc.

This should be fixed so that the second pass is unnecessary.

jeenbroekstra:
Since there is a more serious problem with the javadoc aggregate creation in the latest patch release (it's missing most of the expected javadoc info) I'll use this issue to track changes made to project setup to get this to work again.

jeenbroekstra:
Investigating if the simplest fix is to move javadoc aggregate and SDK assembly to the project root, removing the need for a separate assembly module.

jeenbroekstra:
It looks as if the solution is to move the javadoc aggregate generation to the root pom, but to leave the assembly creation in its own module (cf. http://maven.apache.org/plugins/maven-assembly-plugin/examples/multimodule/module-binary-inclusion-simple.html# ).

PR #1708, opened 2019-11-30
GH-1634 javadoc aggregate

jeenbroekstra:
GitHub issue resolved: #1634   

* javadoc aggregate creation has been moved to root pom as per https://maven.apache.org/plugins/maven-javadoc-plugin/examples/aggregate.html
* removed obsolete assembly profile from core module pom
* release script modified to handle simpler way to generate javadoc and SDK zip.


release.sh
```diff
@@ -258,5 +255,6 @@ echo " - Edit the following file https://github.com/eclipse/rdf4j-doc/blob/maste
 echo " - Go to https://github.com/eclipse/rdf4j-doc/tree/master/site/content/news and create rdf4j-${NEWS_FILE_NAME}.md"
 echo " - Post to Google Groups: https://groups.google.com/forum/#!forum/rdf4j-users"
 echo "     - Good example: https://groups.google.com/forum/#!topic/rdf4j-users/isrC7qdhplY"
-echo " - Upload the javadocs " #TODO Where should they be uploaded and how?
+echo " - Upload the javadocs by adding them to rdf4j-doc project: site/static/javadoc/${MVN_VERSION_RELEASE}"
```
hmottestad: The location of the generated javadocs, is it: assembly/target/site/apidocs

jeenbroekstra: Actually it's target/site/apidocs (in the project root) now. Or, of course, you can get them by unzipping the SDK zip :)

hmottestad: Could you add that as an echo line to the script? In the future we could automate that too.

hmottestad:
Lgtm

Issue #1674, opened 2019-11-15
The artifact rdf4j-runtime-osgi.jar is almost empty


manuelfiorelli:
The artifact `rdf4j-runtime-osgi.jar` is almost empty: it doesn't contain most classes from RDF4J.

I noticed this issue in version 3.0.2, while version 3.0.0 seems fine.

hmottestad:
Did you try setting <type>pom</type>?

manuelfiorelli:
Unfortunately this issue is different from the one that we have already discussed on the ML:
https://groups.google.com/d/msg/rdf4j-users/eER3QQwoPOw/HpcYfadYAgAJ

Let me explain.

The issue on the ML concerns the project `rdf4j-runtime`, which is no longer available as a JAR artifact. This problem affects dependency management by Maven (which I use to build my projects).

This issue concerns the project `rdf4j-runtime-osgi` that should build an OSGi bundle containing all classes and resources corresponding to RDF4J's runtime: you can think of this bundle as a sort of fat JAR aggregating all RDF4J's jars (without the dependencies) plus some metadata (in `MANIFEST.MF`) of interest for an OSGi framework. I use this bundle when deploying my OSGi-based applications that use RDF4J.

I think that some change in the build process of  `rdf4j-runtime-osgi` has erroneously excluded most classes and resources from the aggregate.

jeenbroekstra:
Possibly related to #1611, though it's unclear to me why that should happen (there is no obvious direct dependency between rdf4j-runtime-osgi and rdf4j-runtime).

manuelfiorelli:
I was surprised to see that the pom of `rdf4-runtime-osgi` didn't change significantly between versions `3.0.0` and `3.0.1`. The latter seem almost empty, but at a closer inspection it seems to me that it just contains the classes of `rdf4j-http-server-spring`, which is the sole explicit dependency of the project.
Looking at the pom of `rdf4j-http-server-spring`, I noticed that a potentially relevant change.
In version 3.0.0
```
		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>rdf4j-client</artifactId>
			<version>${project.version}</version>
		</dependency>
		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>rdf4j-storage</artifactId>
			<version>${project.version}</version>
		</dependency>
		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>rdf4j-config</artifactId>
[...]

PR #1716, opened 2019-12-06
Updated maven-bundle-plugin to avoid the embedding of pom dependencie…

manuelfiorelli:
GitHub issue resolved: #1674

This PR updates `maven-bundle-plugin` from version to 3.3.0 to 4.2.1.
The rationale of the change is as follows:
* the runtime OSGi depends on the modules `rdf4j-client` and `rdf4j-storage`
* those dependencies were turned into pom-typed artifact (while they used to be simple `jar`s)
* the instructions guiding  `maven-bundle-plugin` excluded pom dependencies from being inlined (indeed, there is no jar to unpack)
* however, the version `3.3.0` of the plugin seems to ignore the transitive dependencies of the ignored artifact, thus producing an almost empty runtime-osgi
* conversely, the version `4.2.1` of the plugin (which is the latest one) seems to continue processing the transitive dependencies of the artifacts being excluded

 I've verified the produced  OSGi bundle, by using it as a temporary dependency of [VocBench 3](http://vocbench.uniroma2.it/).
Additionally, I compared the contents of the bundle with the contents of the version `3.0.0` (by using [Beyond Compare](https://www.scootersoftware.com/) to perform a folder comparison and only show orphan files).

Below you can find a manually edited report of the comparison.
immagine

Beyond some anonymous and inner classes, which might have changed between the two versions, I've noticed that the poms of `rdf4j-client` and `rdf4-storage` are no longer includes (because these dependencies have been excluded).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -643,7 +643,7 @@
           <plugin>
             <groupId>org.apache.felix</groupId>
             <artifactId>maven-bundle-plugin</artifactId>
-            <version>3.3.0</version>
+            <version>4.2.1</version>
             <extensions>true</extensions>
             <configuration>
               <instructions>
--- a/tools/runtime-osgi/pom.xml
+++ b/tools/runtime-osgi/pom.xml
@@ -60,13 +60,6 @@
 						<Embed-Transitive>true</Embed-Transitive>
 					</instructions>
 				</configuration>
-				<dependencies>
-					<dependency>
-						<groupId>biz.aQute.bnd</groupId>
-						<artifactId>bndlib</artifactId>
-						<version>2.4.0</version>
-					</dependency>
-				</dependencies>
 			</plugin>
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>


tools/runtime-osgi/pom.xml
```diff
@@ -60,13 +60,6 @@
 						<Embed-Transitive>true</Embed-Transitive>
 					</instructions>
 				</configuration>
-				<dependencies>
-					<dependency>
-						<groupId>biz.aQute.bnd</groupId>
-						<artifactId>bndlib</artifactId>
-						<version>2.4.0</version>
-					</dependency>
-				</dependencies>
```
jeenbroekstra: I'm not sure what this dependency was there for in the first place, do you know what it is and why it can be removed?

pom.xml
```diff
@@ -643,7 +643,7 @@
           <plugin>
             <groupId>org.apache.felix</groupId>
             <artifactId>maven-bundle-plugin</artifactId>
-            <version>3.3.0</version>
+            <version>4.2.1</version>
```
jeenbroekstra: This looks straightforward enough, thanks! We'll need to update our build-and-test dependency CQ since this version is not covered in the current one, I'll take care of that.

Also: the bundle plugin is used for more than just the osgi module. I will need to run a couple of local tests to make sure the rest of the build still works as expected.

tools/runtime-osgi/pom.xml
manuelfiorelli: Honestly, I removed that dependency without much thought and, in particular, I can't answer your questions. My idea was to get the updated `bndlib` used by the updated version of `maven-bundle-plugin`.

pom.xml
manuelfiorelli: I didn't noticed this fact. Thank you for the additional testing.

Issue #1718, opened 2019-12-08
Invalid cookie header warning in log when using SPARQLRepository against Wikidata query service


aschwarte10:
When using the SPARQLRepository to evaluate a query against the Wikidata query service, the log is filled with warnings of the following type:

```
[main] WARN org.apache.http.client.protocol.ResponseProcessCookies - Invalid cookie header: "Set-Cookie: WMF-Last-Access-Global=08-Dec-2019;Path=/;Domain=.wikidata.org;HttpOnly;secure;Expires=Thu, 09 Jan 2020 00:00:00 GMT". Invalid 'expires' attribute: Thu, 09 Jan 2020 00:00:00 GMT
```

An investigation has shown that this is due to the current (default) HTTP Client implementation which does not correctly parse the cookies. Further investigation has shown that users of the HTTP client are are advised to to explicitly use RFC 6265 compliant policies. 

This can be achieved using configuration.

See https://issues.apache.org/jira/browse/HTTPCLIENT-1763 and https://stackoverflow.com/questions/36473478/fixing-httpclient-warning-invalid-expires-attribute-using-fluent-api/40697322 for further details

The log can be seen by running a test like the following (with logs enabled):

```
@Test
	public void testWikidata() throws Exception {


		SPARQLRepository repo = new SPARQLRepository("https://query.wikidata.org/sparql");
		repo.init();

		try (RepositoryConnection conn = repo.getConnection()) {

			TupleQuery tq = conn.prepareTupleQuery("SELECT * WHERE { ?s ?p ?o } LIMIT 10");
			try (TupleQueryResult tqr = tq.evaluate()) {
				while (tqr.hasNext()) {
					System.out.println(tqr.next());
				}
			}
		}
		repo.shutDown();
	}
```

PR #1719, opened 2019-12-08
#1718: fix invalid cookie warnings in HttpClient from SPARQLRepository

aschwarte10:
GitHub issue resolved: #1718 

It turns out that the default configuration of the HttpClient has a
configuration for cookie parsing which does not follow todays standards,
thus yielding to warnings in the log for certain cookies.

According to the HttpClient documentation users are advised to
explicitly use RFC 6265 compliant policies.

This is done as of this change by adjusting the CookieSpec in the
default HttpClientBuilder used in the RDF4J SPARQLRepository.

Signed-off-by: Andreas Schwarte <aschwarte10@gmail.com>

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
@@ -17,6 +17,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.http.client.HttpClient;
+import org.apache.http.client.config.CookieSpecs;
+import org.apache.http.client.config.RequestConfig;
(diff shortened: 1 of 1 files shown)


aschwarte10:
We have run into this issue in our application. As the fix is actually quite simple, we may consider to cherry pick this also to the 3.0.4 release. What's your opinion?

hmottestad:
> We have run into this issue in our application. As the fix is actually quite simple, we may consider to cherry pick this also to the 3.0.4 release. What's your opinion?

If you rebase onto master and change the base in your pr we don’t need to cherry pick. Master is anyway merged into develop. 

https://stackoverflow.com/questions/10853935/change-branch-base

jeenbroekstra:
LGTM. Like @hmottestad said, best way to get this in the next patch release is to rebase on master and then change the PR to target master as well.

aschwarte10:
Done, the commit is now based on master and the PR is targeted to master. Thanks for the feedback

PR #1721, opened 2019-12-08
fix typo in source branch of github action

jeenbroekstra:
this job failed for me just now. See https://github.com/eclipse/rdf4j/commit/f8ead84c7f5c93c8518b0e7ff9f221ea77ea4e86/checks?check_suite_id=347622360 

Not sure if this typo was the cause but it looks incorrect to me.


hmottestad:
Good catch. Definitely a typo.

PR #1726, opened 2019-12-09
Streamline action workflow

jeenbroekstra:
- added badges on README for master and develop status (had to split master and develop compliance into separate jobs for this)
- modified pull request testing to do full compliance testing (safer, and only marginally slower)

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -1,4 +1,4 @@
-name: PR unit test
+name: pull request verify
 
 on: pull_request
 
@@ -21,4 +21,4 @@ jobs:
     - name: Build
       run: mvn -B clean install -DskipTests
     - name: Run tests
-      run: mvn -B verify --file pom.xml
+      run: mvn -B verify -Pcompliance --file pom.xml
--- a/README.md
+++ b/README.md
@@ -7,6 +7,12 @@ This is the main code repository for the Eclipse RDF4J project.
 
 Visit the [project website](https://rdf4j.org/) for news, documentation, and downloadable releases.
 
+## Installation and usage
+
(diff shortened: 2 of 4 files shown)


.github/workflows/pr-verify.yml
```diff
@@ -21,4 +21,4 @@ jobs:
     - name: Build
       run: mvn -B clean install -DskipTests
     - name: Run tests
-      run: mvn -B verify --file pom.xml
```
hmottestad: why do we need to specify "--file pom.xml" btw?

README.md
```diff
@@ -1,8 +1,11 @@
-# Welcome to the Eclipse rdf4j repository
+# Welcome to the Eclipse RDF4J repository
 
-This is the main code repository for the Eclipse rdf4j project. 
+This is the main code repository for the Eclipse RDF4J project. 
 
-user documentation and [downloads of the latest release](http://rdf4j.org/download).
+![](https://github.com/eclipse/rdf4j/workflows/master%20status/badge.svg)
+![](https://github.com/eclipse/rdf4j/workflows/develop%20status/badge.svg)
```
hmottestad: badge :D

README.md
```diff
@@ -23,7 +26,3 @@ The short version:
 
 These steps are explained in more detail in the [Contributor
 guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md).
-
-## Release
-
-Run `./release.sh` and follow the prompts to create a release.
```
hmottestad: We should have this info somewhere else? Maybe here: https://rdf4j.org/documentation/developer/#_developer_workflow_and_project_management ?

jeenbroekstra: Yeah good point. I took it out because it is only relevant to committers, not to contributors. But a link to our internal developer documents would be good to put in. I'll add something!

.github/workflows/pr-verify.yml
jeenbroekstra: I'm not sure it's necessary - I just took this from the default/example maven workflow.

Issue #1729, opened 2019-12-09
Could you add the xsd:dateTimeStamp datatype to the XMLSchema vocabulary?


tonicebrian:
We are using the OWL Time ontology for defining time objects and we are heavy users of [time:inXSDDateTimeStamp](https://www.w3.org/TR/owl-time/#time:inXSDDateTimeStamp) that explicitly relays on [xsd:dateTimeStamp](https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp) but this datatype doesn't appear in class ` org.eclipse.rdf4j.model.vocabulary.XMLSchema`. 

Is there something preventing to include that datatype in that file? I'm more than willing to provide a PR if some directions about where/how to place are provided.

barthanssens:
Ah, presumably just an oversight when the class was updated from RDF 1.0 to RDF 1.1...
(and the reference in the javadoc should point to https://www.w3.org/TR/rdf11-concepts/#xsd-datatypes or https://www.w3.org/TR/rdf11-concepts/#xsd-datatypes) 

By the way, 3.1.0 will also include o.e.r.m.vocabulary.TIME (as part of #1580)

barthanssens:
Might also have an impact on org.eclipse.rdf4j.model.datatypes.XMLDatatypeUtil

PR #1730, opened 2019-12-10
Fixes: #1729. Add dateTimeStamp to the XMLSchema vocabulary

tonicebrian:
GitHub issue resolved: #1729 

* Adds the definition for xsd:dateTimeStamp as defined in https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp 
* Provides some testing for the validation of such type.

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
@@ -84,7 +84,8 @@ public class XMLDatatypeUtil {
 				|| datatype.equals(XMLSchema.NON_NEGATIVE_INTEGER) || datatype.equals(XMLSchema.POSITIVE_INTEGER)
 				|| datatype.equals(XMLSchema.UNSIGNED_LONG) || datatype.equals(XMLSchema.UNSIGNED_INT)
 				|| datatype.equals(XMLSchema.UNSIGNED_SHORT) || datatype.equals(XMLSchema.UNSIGNED_BYTE)
-				|| datatype.equals(XMLSchema.DAYTIMEDURATION) || datatype.equals(XMLSchema.YEARMONTHDURATION);
+				|| datatype.equals(XMLSchema.DAYTIMEDURATION) || datatype.equals(XMLSchema.YEARMONTHDURATION)
+				|| datatype.equals(XMLSchema.DATETIMESTAMP);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


barthanssens:
Thanks for providing the patch !

Would it be possible to sign the Eclipse contributor agreement, so it can be integrated in RDF4J ? It's a small one-time effort, though it is required for Eclipse projects.

The steps are documented here: https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md
(basically it's creating an account on Eclipse and agreeing with a very reasonable policy)

tonicebrian:
I'm in an infinite loop trying to link my Github handle. Every time I try to input the handle the website requires to also put my password, and then when I submit it thinks that I'm trying to update my password and it is back to the sign in page with this message

image

barthanssens:
Autch, sorry for the hassle :/ 
Maybe you could try _not_ entering the github handle (hoping the system will accept/actually check email adres instead) ?
Also accepting all cookies (and pop-ups, at once occasion I had to solve a CAPTCHA) might perhaps help.

tonicebrian:
No luck. I already have the handle empty and it is not picking my email address. This is what I see in  my profile:

image

When I enter my Github handle WITHOUT providing the password in the blue box I get:

image

If I provide the password, I get a message (plus an email) saying that my password has been updated correctly and back to the sign in page. 

I think I cannot do anything about it.

barthanssens:
Hmz, you might need to contact emo(at)eclipse.org or webmaster(at)eclipse.org to get this solved :-/

hmottestad:
Sorry that the eclipse site sucks. It sucks for me too. But I have other issues, like the 5 minutes login delay and the fact that they didn't support utf-8 until earlier this year.

tonicebrian:
Ok, I give up. I'm spending more time with the auth than the time it took to create the PR. Please pick the patch and apply it yourselves so we can use it in future releases.

For the record, I was able to input my github handle by disabling the password manager since the web page was picking it from a different pane that hadn't the focus. But now I have no idea why it's not picking nor my email address or my github handle.

hmottestad:
You want to make a new PR @barthanssens ?

jeenbroekstra:
@tonicebrian sorry about the hassle. We are following up with the Eclipse webmaster to figure out the problem (especially since it might bit others too), and in the mean time we'll just apply the patch manually. Thanks again for providing it!

barthanssens:
OK, I'll make some minor changes and create a new PR

barthanssens:
See PR #1734

Issue #1731, opened 2019-12-10
Read cache for ElasticsearchStore


hmottestad:
Implement a configurable read cache for the ElasticsearchStore

PR #1732, opened 2019-12-10
ExtensibleStore read cache

hmottestad:
GitHub issue resolved: #1731  

* a read cache that wraps a DataStructureInterface and provides caching of all read operations and write through for write operations
* automatic cache clearing
* lazy caching (getStatement iterations are still lazy)
* memory sensitive cache that will be garbage collected if the JVM runs low on memory

### DIFF
--- a/core/sail/extensible-store/pom.xml
+++ b/core/sail/extensible-store/pom.xml
@@ -26,7 +26,6 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-collections4</artifactId>
-            <version>4.4</version>
         </dependency>
 
         <dependency>
--- a/core/sail/extensible-store/pom.xml
+++ b/core/sail/extensible-store/pom.xml
@@ -22,6 +22,13 @@
             <artifactId>rdf4j-sail-base</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>4.4</version>
+        </dependency>
+
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>rdf4j-sail-memory</artifactId>
@@ -29,9 +36,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.rdf4j</groupId>
+            <groupId>${project.groupId}</groupId>
             <artifactId>rdf4j-repository-sail</artifactId>
-            <version>3.1.0-SNAPSHOT</version>
+            <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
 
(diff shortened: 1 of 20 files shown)


hmottestad:
Look at https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/ReferenceMap.html

hmottestad:
CQ pending: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21262

hmottestad:
```
Benchmark                                        Mode  Cnt      Score   Error  Units
ReadCacheBenchmark.complexQueryWithCache         avgt        1292.071          ms/op
ReadCacheBenchmark.complexQueryWithCacheCleared  avgt       17811.627          ms/op
ReadCacheBenchmark.complexQueryWithoutCache      avgt       98220.901          ms/op
ReadCacheBenchmark.groupByQueryWithCache         avgt          11.648          ms/op
ReadCacheBenchmark.groupByQueryWithCacheCleared  avgt         255.904          ms/op
ReadCacheBenchmark.groupByQueryWithoutCache      avgt         233.095          ms/op
ReadCacheBenchmark.pathQueryWithCache            avgt          12.029          ms/op
ReadCacheBenchmark.pathQueryWithCacheCleared     avgt        3394.183          ms/op
ReadCacheBenchmark.pathQueryWithoutCache         avgt        3929.765          ms/op
```

*WithCache = use as much caching as possible
*WithoutCache = no cache at all
*WithCacheCleared = same as WithCache, except that it adds and removes a statement to begin with in order to clear the cache between runs making it act more like a single-shot benchmark

In general caching makes the Elasticsearch store a whole lot more useable. Sometimes caching even makes single runs of queries better. That is probably because the query plan contains a lot of duplicate evaluations.

hmottestad:
CQ approved

hmottestad:
@jeenbroekstra What do you think? This makes the ElasticsearchStore very useful as a read-heavy store. The cache is global for the ElasticsearchStore, however it wouldn't be too hard to implement a transaction level cache too.....but then we might end up caching things twice. So that might be a future feature.

hmottestad:
Would really like to merge this so I can get started on the WAL.

core/sail/extensible-store/pom.xml
```diff
@@ -22,16 +22,23 @@
             <artifactId>rdf4j-sail-base</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>4.4</version>
```
jeenbroekstra: We use older versions of commons-collections in several places - there's both 3.2.2 and 4.2 in use as (transitive) dependencies. Is it possible to reconcile this? Do you need 4.4 here specifically, or would 4.2 be good enough?

hmottestad: Thanks Jeen. I’ll take a look and see if I can’t upgrade everything. There wasn’t much of a release note. They just simple wrote something along the lines of “bug fixes”.

PR #1734, opened 2019-12-11
Added xsd:dateTimeStamp, initial patch provided by Toni Cebrián

barthanssens:
GitHub issue resolved: #1729  

* Added constant to XMLSchema vocabulary
* Added checks in XmlDateUtil
* Added tests

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
@@ -84,7 +84,8 @@ public class XMLDatatypeUtil {
 				|| datatype.equals(XMLSchema.NON_NEGATIVE_INTEGER) || datatype.equals(XMLSchema.POSITIVE_INTEGER)
 				|| datatype.equals(XMLSchema.UNSIGNED_LONG) || datatype.equals(XMLSchema.UNSIGNED_INT)
 				|| datatype.equals(XMLSchema.UNSIGNED_SHORT) || datatype.equals(XMLSchema.UNSIGNED_BYTE)
-				|| datatype.equals(XMLSchema.DAYTIMEDURATION) || datatype.equals(XMLSchema.YEARMONTHDURATION);
+				|| datatype.equals(XMLSchema.DAYTIMEDURATION) || datatype.equals(XMLSchema.YEARMONTHDURATION)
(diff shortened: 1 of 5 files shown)


hmottestad:
@tonicebrian

hmottestad:
One quick question. Does this also support sorting? ie. by using the default RDF4J comparator.

barthanssens:
Good catch, I'll write a compareDateTimeStamp method and add it to compare

barthanssens:
Hmz, JDK 8 (and 11) only supports XML Schema 1.0, not 1.1 (https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/javax/xml/datatype/DatatypeConstants.html)

Doesn't stop this to be useful, but there might be cases where functionality is limited.

barthanssens:
My main (minor) concern is that one has to use 
`createLiteral(String datetime, XMLSchema.DATETIMESTAMP)`

Because `javax.xml.datatype.XMLGregorianCalendar` and `javax.xml.datatype.DatatypeConstants` don't support XML Schema 1.1 datatypes

barthanssens:
@tonicebrian: I think it should be fairly complete now, so basically ready to merge unless I forgot something

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
```diff
@@ -1809,8 +1823,9 @@ private static String trimPlusSign(String s) {
 	}
 
 	/**
-	 * Maps a datatype QName from the javax.xml.namespace package to an XML Schema URI for the corresponding datatype.
-	 * This method recognizes the XML Schema qname mentioned in {@link DatatypeConstants}.
+	 * Maps a datatype QName from the javax.xml.namespace package to an XML Schema 1.0 URI for the corresponding
+	 * datatype. This method recognizes the XML Schema qname mentioned in {@link DatatypeConstants}. Note that Java 8 /
+	 * 11 do not have constants for XML Schema 1.0 datatypes like xsd:dateTimeStamp.
```
jeenbroekstra: Do you mean XML Schema 1.1 here?

barthanssens: ah indeed, thanks

tonicebrian:
@barthanssens LGTM, Thanks!!

Issue #1735, opened 2019-12-11
XMLDataTypeUtil: use precompiled regex


barthanssens:
A tiny performance enhancement in checking various datatypes could be to use Pattern.compile, instead of String.matches

PR #1743, opened 2019-12-12
XMLDateutil performance

barthanssens:
GitHub issue resolved: #1735   

* Added javadoc
* Use precompiled patterns


barthanssens:
Probably only useful in very specific cases, since the regex are only used for checking time/duration etc, so only users with a decent amount of those literals may benefit from it.

As a rough indication,  4800 K compares / second with precompiled Pattern instead of 700 K / sec with String on my i7 laptop, so it might not be worth the effort.

hmottestad:
Very much worth the effort. I've noticed before that sorting statements can be really really slow in RDF4J. So any effort made to make that faster is well worth it.

Issue #1747, opened 2019-12-14
Make RepositoryResult and QueryResult a java.lang.Iterable


jeenbroekstra:
Related to #96 .

As an RDF4J Repository API user, I want `RepositoryResult` and `QueryResult` to implement the `java.lang.Iterable` interface, so that I can directly use language constructions such as `forEach` or the enhanced for-loop, without having to first wrap the resulting using `QueryResults.asList`.

For example, the change will make it possible to simplify

	QueryResults.asList(conn.getNamespaces()).forEach(ns -> model.setNamespace(ns));

to 

         conn.getNamespaces().forEach(ns -> model.setNamespaces(ns)); 

or 

        List<BindingSet> result = QueryResults.asList(query.evaluate());
        for (BindingSet bs: result) {
               ...
        }

to 

        try(TupleQueryResult result = query.evaluate()) {
            for(BindingSet bs: result) {
                ....
             }
        }

Apart from improved readability of the code, an additional advantage is that the query result no longer needs to be fully materialized into a List object before we process it.

jeenbroekstra:
There is an unforeseen problem with this simple idea: it results in ambiguous method calls in existing code. For example:

 		try (RepositoryResult<Statement> stats = conn.getStatements(null, null, null)) {
			conn.remove(stats);
		}

The `remove` method is overloaded in `RepositoryConnection`: it accepts either an `Iterable` or an `Iteration`. However, because `RepositoryResult` implements _both_ `Iteration` and `Iterable`, the compiler complains that the method call is ambiguous.

jeenbroekstra:
A possible workaround for this is to add default methods to RepositoryConnection that explicitly take a RepositoryResult object as input.

jeenbroekstra:
We will need to care that we don't introduce code paths where iterations are not properly closed.

PR #1748, opened 2019-12-15
#1747 repositoryresult and queryresult iterable

jeenbroekstra:
GitHub issue resolved: #1747   

* RepositoryResult and QueryResult extend java.lang.Iterable
* added method overloads to avoid ambiguous calls in RepositoryConnection


hmottestad:
Looks awesome. Now I’m wondering if we could have a .stream() method too?

jeenbroekstra:
> Looks awesome. Now I’m wondering if we could have a .stream() method too?

You can already stream a result via `QueryResults.stream`. However I am starting to agree that for discoverability it is better to have these methods on the objects themselves. I'll create a followup issue.

Issue #1387, opened 2019-04-09
Pass JsonLdOptions through parser configs 


RavinderSinghMaan:
Hi all

We discussed [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/rdf4j-users/xrirAvWzcP8/2jXVDWsFBwAJ) if JsonLDOptions object should be allowed to pass through parser configs. We agree that it will help in some use cases to allow to do so hence raising this feature request.

Thanks
Ravinder

barthanssens:
Related to #395

PR #1749, opened 2019-12-16
Issues/#1387 jsonldoptions

barthanssens:
GitHub issue resolved: #1387  

* Removed duplicate code in JSONLDParser
* Added ClassRioSettings
* Added PRODUCE_GENERALIZED_RDF and DOCUMENT_LOADER options
* Pass DOCUMENT_LOADER to JSONLDParser
* Pass COMPACT_ARRAYS and PRODUCE_GENERALIZED_RDF to JSONLDWriter
* Added test

### DIFF
--- a/core/rio/api/pom.xml
+++ b/core/rio/api/pom.xml
@@ -34,6 +34,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>com.github.jsonld-java</groupId>
+			<artifactId>jsonld-java</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
 			<scope>test</scope>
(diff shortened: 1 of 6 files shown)


barthanssens:
@ansell , could you perhaps also take a look at this PR ? thanks

barthanssens:
I covered the original use case / request on https://groups.google.com/forum/#!msg/rdf4j-users/xrirAvWzcP8/_GT91RicAwAJ + some minor enhancements

Though not entirely happy with the new ClassRioSetting

core/rio/api/pom.xml
```diff
@@ -33,6 +33,11 @@
 			<version>${jaxb.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>com.github.jsonld-java</groupId>
+			<artifactId>jsonld-java</artifactId>
+		</dependency>
```
jeenbroekstra: I understand why you need this here and I'm ok with it, but it does pinpoint to me that in a future release we really should move the format-specific helpers out of the api module, and into their respective parser implementation modules.

barthanssens: Indeed, probably a todo for 4.0 ...

barthanssens: Created #1755

barthanssens:
The ClassRioSetting felt like stretching it a bit, but it does indeed do the job...

ansell:
I agree, having to hard-link a class (DocumentLoader) into the api module
is a sign that the format specific configurations are probably more suited
to being moved in the next major release.

The original rio config design had them all comfortably in the api module
because they were simple types that didn't hardlink into specific
implementations.

Issue #1740, opened 2019-12-11
CQ for embedded-elasticsearch


hmottestad:
A works-with CQ for using the embedded-elasticsearch for testing the new ElasticsearchStore. https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21260

jeenbroekstra:
CQ has been approved! @hmottestad I'll look at reverting the PR before the milestone build.

PR #1754, opened 2019-12-18
comment out embedded-elasticsearch dependency and remove all tests

hmottestad:
Related to #1740 

I was unable to get a CQ for the use of the embedded-elasticsearch in our tests. Commenting out the dependency and zipping the tests so we don't cause test failures.

This is a temporary fix.


hmottestad:
@jeenbroekstra I would like to merge this in before you make the milestone build!

jeenbroekstra:
Thanks for taking care of due diligence on this Havard! Fwiw I'm not actually 100% sure we need to do this before a milestone build (which doesn't count as an official release), but better safe than sorry.

hmottestad:
Messed up the CQ a bit too. I chose source, should have chosen binary. Was looking for works-with, but the CQ system doesn’t use the same words as the CQ docs.

jeenbroekstra:
> Messed up the CQ a bit too. I chose source, should have chosen binary. Was looking for works-with, but the CQ system doesn’t use the same words as the CQ docs.

Yeah, works-with is usually just indicated by adding a comment after you've created the CQ. It's a bit, erhm, manual.

Issue #1710, opened 2019-12-03
WAL for Elasticsearch Store


hmottestad:
Write Ahead Log for Elasticsearch Store to handle disconnects or crashes with Elasticsearch.

There are `@ignored` tests in the code already.

hmottestad:
Related to #1662

PR #1769, opened 2019-12-21
#1710 simple locking for increased isolation

hmottestad:
GitHub issue resolved: #  

* simple locking to increase transaction isolation
* 
*


core/sail/extensible-store/src/test/java/org/eclipse/rdf4j/sail/extensiblestoreimpl/TransactionIsolationAndWalTests.java
```diff
@@ -26,7 +26,6 @@
 	 * transaction should see either nothing added or everything added. Nothing in between.
 	 */
 	@Test
-	@Ignore
 	public void testReadCommittedLargeTransaction() throws InterruptedException {
```
jeenbroekstra: I don't understand why you need a custom test for this. We have a whole testsuite full of isolation compliance tests that you could reuse for this purpose. Why not simply extend  `SailIsolationLevelTest`?

core/sail/extensible-store/src/test/java/org/eclipse/rdf4j/sail/extensiblestoreimpl/TransactionIsolationAndWalTests.java
```diff
@@ -71,6 +70,8 @@ public void testReadCommittedLargeTransaction() throws InterruptedException {
 
 		assertFalse(failure.get());
 
+		repository.shutDown();
```
jeenbroekstra: On line 45 you are printing to stdout - could you replace that with using the logger, probably to output to logger.warn level?

core/sail/extensible-store/src/test/java/org/eclipse/rdf4j/sail/extensiblestoreimpl/TransactionIsolationAndWalTests.java
jeenbroekstra: (not saying you can't have more tests than what the testsuite already offers of course, but this particular thing you're testing here seems to me to be well-covered).

hmottestad: This test interacts with the underlying datastructure to consistently reproduce the isolation issue. The isolation tests we currently have do not seem to cover this issue here or any write-ahead-logging issues.

hmottestad: I will look into adding this to the suite instead.

hmottestad:
Screenshot 2019-12-27 at 14 54 07

@jeenbroekstra I moved the test but the SchemaCachingRDFS.... is failing. Seems that there is some sort of leak between transactions that I don't understand. Got any ideas?

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/UnionSailSource.java
```diff
@@ -64,8 +64,8 @@ public void prepare() throws SailException {
 
 	@Override
 	public void flush() throws SailException {
-		primary.flush();
 		additional.flush();
+		primary.flush();
```
hmottestad: @jeenbroekstra 

Here is the essence of the problem. Flushing either of `additional` or `primary` will flush the changes to the underlying memory store and also move to the next snapshot. The fact that it moves to the next snapshot will make the changes visible to other transactions, but this is problematic since `additional` and `primary` both operate on the same snapshot version.

The current algorithm can be flattened as something close to:

1. flush changes from `primary`
1. make current snapshot visible to other transactions
1. flush changes from `additional` to the visible snapshot
1. redundantly make the current snapshot visible again


It should be something along the lines of:

1. flush changes from `primary`
1. flush changes from `additional`
1. make current snapshot visible to other transactions

Swapping `primary` and `additional` like I did here doesn't fix the underlying issue, just masquerades it.

hmottestad:
Now it's failing in the `ShaclIsolationLevelTest`. So yes, this will likely need a proper fix.

hmottestad:
Screenshot 2019-12-27 at 17 03 39

I thought maybe the native store had the same issues as the memory store, so I made a nativestore based test for the SchemaCachingRDFS... and this reveals a new type of transaction isolation issue :(

jeenbroekstra:
Interesting find. Fwiw I had a quick look at the deprecated ForwardChainingRDFSInferencer, and it also fails these tests - just to confirm that this is clearly something of a problem in the UnionSailSource, and not anything specific to either ShaclSail or SchemaCaching inferencer.

I'm a bit confused by what is happening here - at first glance it _looks_ as if somehow the read transaction's snapshot number gets updated so it starts seeing uncommitted statements in the backing store. But I'm a bit lost on why UnionSailSource#flush plays a role in this. If I remember correctly, a flush on a `SailSource` is a no-op, only flush on a `SailSink` advances the snapshot....

jeenbroekstra:
Fwiw the problem appears not to be specific to large transactions - I can reproduce the same errors using a txn of just 10 statements.

hmottestad:
@jeenbroekstra i've reduced the tests to 1000 statements, which is small enough to be fast and still reproduces the issues. I've created two new issues for these test failures and commented out the tests so everything is currently passing.

Issue #1695, opened 2019-11-26
clean up maven output when running builds/tests


jeenbroekstra:
Running a build / executing tests puts a tremendous amount of text to STDOUT and / or the logger. We should look at cleaning that up where possible so that looking through the build log becomes easierf.

PR #1770, opened 2019-12-22
GH-1515 GH-1695 testsuite logging cleanup

jeenbroekstra:
GitHub issue resolved: #1515 #1695   

* make logback-classic logger of choice consistently for all tests
* have shared test-scoped dependencies
* tweak various test suites to output less stuff. These changes reduce the output from a full `mvn clean verify` run from ~38,000 lines to a "mere" ~9,000. 

Further improvements are possible for #1695, in particular in the sparqlbuilder examples which output to stdout, several tests in the ShaclSail and various Rio parsers/writers.


tools/console/src/test/java/org/eclipse/rdf4j/console/command/FederateTest.java
```diff
@@ -72,14 +72,14 @@
 	@InjectMocks
 	private Federate federate;
 
-	private Level originalLevel;
+//	private Level originalLevel;
 
 	@Before
 	public void setUp() throws Exception {
-		originalLevel = ((Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).getLevel();
+//		originalLevel = ((Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).getLevel();
```
jeenbroekstra: Note to self: remove instead of commenting out.

core/sail/shacl/src/test/resources/logback-test.xml
```diff
@@ -6,15 +6,12 @@
 			<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %msg%n</pattern>
 		</encoder>
 	</appender>
-	
+
+	<logger name="org.eclipse.rdf4j.sail.shacl" level="info"/>
+
 	<root>
-		<level value="info" />
+		<level value="warn" />
```
hmottestad: I prefer to keep the shacl logging fairly verbose. I've done a big cleanup for some of the logging now to reduce the output. I've also got a lot of tests set up to only log output if they fail, this greatly reduces the output.

hmottestad:
Hi Jeen. 

Good cleanup.

I am a fan of fairly verbose logging for tests. You almost never need it, but those times when a test fails it is really useful. Especially for the github CI, since there is no way to debug a failed run and intermittent failures are not necessarily easy to reproduce.

core/sail/shacl/src/test/resources/logback-test.xml
jeenbroekstra: Note that logging for anything in package `org.eclipse.rdf4j.sail.shacl` is set to INFO, so you should still get fairly detailed logging for anything shacl-specific.

jeenbroekstra: I'd also like to keep the default less verbose. Of course, when a test is giving us problems (intermittent or otherwise), we can always up the verbosity (either locally, or pushed to CI) temporarily to diagnose the issue.

hmottestad: I designed the main shacl tests to only output info if they fail. I want to check if your changes affect this or not before your merge the branch. Because of Christmas is might take a few days or a week before I have time to test it. Is that ok @jeenbroekstra ?

jeenbroekstra: Sure, no rush. I'm pretty certain it won't be affected as I didn't change anything in the report output writing - I just adapted the actual logging. But doesn't hurt to double-check.

Fwiw I would prefer to have the error reporting on failure also use the logger, instead of printing directly to stdout. But that's a separate issue, and one we can discuss later.

hmottestad: I had a look now and your changes look good.

hmottestad:
lgtm

Issue #1785, opened 2019-12-26
Remove IsolationLevel interface


hmottestad:
The use of a interface doesn't seem to be particularly useful when there is only a single class implementing the interface. It also hinders discoverability since it doesn't play well with code hinting.

jeenbroekstra:
The reason we have this as a separate interface is that third-party stores may choose to implement other, non-standard isolation levels.

jeenbroekstra:
As also suggested on the PR: instead of getting rid of the interface, we can look at removing the need for the enum, and instead have the default isolation levels as static constants on the interface (using an inner class to implement the interface). That way, the interface can still be extended by third parties, while discoverability/autocomplete is improved.

PR #1788, opened 2019-12-28
GH-1785 isolation levels

hmottestad:
GitHub issue resolved: #1785  

* switched over from using the interface IsolationLevel to using the class IsolationLevels
* code completion for IsolationLevels now works in Intelij
*


hmottestad:
I propose to use IsolationLevels directly instead of using the interface. This means that ctrl-space code completion works as expected in Intellij. 

This is backwards compatible for everyone who uses RDF4J as-is, but not backwards compatible for those who extend RDF4J. Those who extend the code with their own isolation levels will need to now extend IsolationLevels instead of implementing the IsolationLevel interface.

hmottestad:
@jeenbroekstra also propose to fix the tests with maven dep cycles by extracting them into a separate module. This is a lot less work them trying to mock stuff.

jeenbroekstra:
> I propose to use IsolationLevels directly instead of using the interface. This means that ctrl-space code completion works as expected in Intellij.

I'm not sure I understand what you mean with this, to be honest. What exact problem are you solving here? If I type `Isola<Ctrl-space>` in IntelliJ, it currently offers me both `IsolationLevel` and `IsolationLevels`. So how does this change help? 

> This is backwards compatible for everyone who uses RDF4J as-is, but not backwards compatible for those who extend RDF4J. Those who extend the code with their own isolation levels will need to now extend IsolationLevels instead of implementing the IsolationLevel interface. 

The problem with that is that `IsolationLevels` is an enum. It can't be extended. This is why we have the separation between the generic interface and the pre-provided isolation level types. 
 
> Is there a way to involve those parties here on github?

We have to assume that we will not able to reach all third parties that do this. But before we do that I would like a bit more discussion about the pros and cons of this change, to be honest. I'm not yet sure I understand what problem this solves.

hmottestad:
Try typing “SNAP” + ctrl+space

jeenbroekstra:
> Try typing “SNAP” + ctrl+space

Right. I understand the convenience of that. I don't think the current solution is quite right though. We can't give up a design feature like extensibility of supported levels for the sake of code completion convenience.  

> It’s really a pity that enum classes can’t be extended.
 
It's kind of by design: the whole point of an enum is that you have a closed (fully enumerated) set :)

> I wonder if we instead can use some generics instead.

An alternative approach would be that we get rid of the IsolationLevels enum, and instead declare the "standard" isolation levels as static constants on the IsolationLevel interface, using an inner class, e.g. something like this:

```
      public interface IsolationLevel {

	boolean isCompatibleWith(IsolationLevel otherLevel);
	IRI getURI();


	IsolationLevel NONE = new StandardIsolationLevel();

	IsolationLevel READ_UNCOMMITTED = new StandardIsolationLevel(NONE);

	class StandardIsolationLevel implements IsolationLevel {

		StandardIsolationLevel(IsolationLevel... compatibleLevels) {
                     ....
		}

		@Override public boolean isCompatibleWith(IsolationLevel otherLevel) {
		 ....
		}

		@Override public IRI getURI() {
			...
		}
	}
} 
```

hmottestad:
@jeenbroekstra seeing as how I've got quite a few open PRs I think this one could be closed and if still want to implement the `alternate approach where we get rid of the enum, rather than the interface` than that would be great. If not then maybe I'll look at it some other time, but no need to keep a PR open in the mean time.

PR #1796, opened 2019-12-31
print test results

hmottestad:
Script for printing test results.


hmottestad:
@jeenbroekstra thought i would have the github ci print a report at the end if there are test failures. What do you think?

jeenbroekstra:
@hmottestad should we just merge what you've got and see about the remarks for improvement in a followup? I'm rather interested to see this in action :)

Issue #1784, opened 2019-12-26
ExtensibleSail evaluation statistics


hmottestad:

jeenbroekstra:
CQ http://dev.eclipse.org/ipzilla/show_bug.cgi?id=21359

Issue #1795, opened 2019-12-31
UnionSailSource causes isolation leakage for MemoryStore


hmottestad:
All isolation levels leak data between transactions.

The base issue is that UnionSailSource will call flush against both underlying sources, causing the first flush to make all the changes from primary become visible to other transactions. So other transactions risk seeing a partially committed state. The other anomaly is that the second flush will be visible while it's running...so each statement flushed from `additional` will be visible as it gets applied.

```
	@Override
	public void flush() throws SailException {
		primary.flush();
		additional.flush();
	}
```

This is easiest to observe in SchemaCachingRDFSInferencerIsolationLevelTest but is also sometimes visible in ShaclIsolationLevelTest.

hmottestad:
<img width="1778" alt="Screenshot 2019-12-31 at 12 05 21" src="https://user-images.githubusercontent.com/797185/71619686-e2180c80-2bc5-11ea-805e-5d96191597b6.png">

hmottestad:
From the PR where this was first observed:

Here is the essence of the problem. Flushing either of additional or primary will flush the changes to the underlying memory store and also move to the next snapshot. The fact that it moves to the next snapshot will make the changes visible to other transactions, but this is problematic since additional and primary both operate on the same snapshot version.

The current algorithm can be flattened as something close to:

flush changes from primary
make current snapshot visible to other transactions
flush changes from additional to the visible snapshot
redundantly make the current snapshot visible again
It should be something along the lines of:

flush changes from primary
flush changes from additional
make current snapshot visible to other transactions
Swapping primary and additional like I did here doesn't fix the underlying issue, just masquerades it.

PR #1801, opened 2020-01-01
GH-1784 eval stats

hmottestad:
GitHub issue resolved: #1784   

* three stats implementations: constant, direct, dynamic
   * constant reuses existing implementation
   * direct queries the underlying data structure
   * dynamic uses arrays of HyperLogLog counters to keep estimates


hmottestad:
http://dev.eclipse.org/ipzilla/show_bug.cgi?id=21359

hmottestad:
Missing:
 - async & parallel dynamic stats
 - support remove operations from dynamic cache

hmottestad:
@jeenbroekstra what do you think of my approach here? I’m using HyperLogLog to count the number of statements for each pattern. 

I have arrays, single and two dimensional. I use parts of the pattern as a key. Eg. For the “subjectsIndex” array statement.getSubject().hashCode() % subjectsIndex.length gives the index into the array. I hash the statement with murmur and store the hash in the hyperloglog counter. 

The two dimensional arrays can index both a subject and predicate, for instance. 

I have two sets of arrays. One for added statements and one for removed. 

The biggest drawback is that it poorly handles removing a statement and the re-adding it. 

This is all still WIP.

core/sail/extensible-store/src/main/java/org/eclipse/rdf4j/sail/extensiblestore/evaluationstatistics/ExtensibleDynamicEvaluationStatistics.java
```diff
@@ -0,0 +1,337 @@
+package org.eclipse.rdf4j.sail.extensiblestore.evaluationstatistics;
```
jeenbroekstra: Copyright header is missing here, and a couple of other places as well.

core/sail/extensible-store/src/main/java/org/eclipse/rdf4j/sail/extensiblestore/evaluationstatistics/ExtensibleDynamicEvaluationStatistics.java
```diff
@@ -0,0 +1,337 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
+
+public class ExtensibleDynamicEvaluationStatistics extends ExtensibleEvaluationStatistics implements DynamicStatistics {
```
jeenbroekstra: I believe this class, and some of the other implementation classes as well, deserve some extensive javadoc that goes into the how and why of the algorithm used.

core/sail/extensible-store/src/main/java/org/eclipse/rdf4j/sail/extensiblestore/evaluationstatistics/ExtensibleDynamicEvaluationStatistics.java
```diff
@@ -0,0 +1,337 @@
...
+		}
+	};
+
+	private double getHllCardinality(HyperLogLogCollector[][] index, HyperLogLogCollector[][] index_removed,
+			Value value1, Value value2) {
+		return index[Math.abs(value1.hashCode() % index.length)][Math.abs(value2.hashCode() % index.length)]
+				.estimateCardinality()
+				- index_removed[Math.abs(value1.hashCode() % index_removed.length)][Math
+						.abs(value2.hashCode() % index_removed.length)]
+								.estimateCardinality();
```
jeenbroekstra: I'll have to believe you if you say that it does what it should do. Seriously though: I wonder if you could maybe split this out into more than one line of code to make it easier to read.

core/sail/extensible-store/src/main/java/org/eclipse/rdf4j/sail/extensiblestore/evaluationstatistics/ExtensibleDynamicEvaluationStatistics.java
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	@Override
+	public void add(Statement statement, boolean inferred) {
+
+		queue.add(new StatemetQueueItem(statement, inferred, true));
+
+		int size = queueSize.incrementAndGet();
+		if (size > QUEUE_LIMIT && queueThread == null) {
+			startQueueThread();
```
jeenbroekstra: What is the role of this thread? It updates the various indexes I gather, is that such an expensive operation that it should be handled asynchronously?

hmottestad:
> It looks like a solid enough approach - though I can't pretend to fully understand it. Got some inline editorial comments, including some questions.
> 
> On a more general note: I am curious why you are going with HyperLogLog, and an implementation that, if I'm not mistaken, comes from yet _another_ third-party database (Apache Druid). ElasticSearch itself has an API for retrieving index statistics - wouldn't it be simpler to base your evaluation statistics on that, directly?

I might end up writing something specific for the elasticsearch implementation, but I first wanted to make a generic and scaleable implementation that can be reused for other sails built on the ExtensibleStore. Might not end up making anything specific for the elasticsearch implementation since the network overhead might not be worth it. 

As for this particular HyperLogLog implementation, I'm not married to it. I could shop around a bit more, especially if the CQ drags out. HyperLogLog in general seems like a decent fit though, so all I need is a good-enough implementation.

core/sail/extensible-store/src/main/java/org/eclipse/rdf4j/sail/extensiblestore/evaluationstatistics/ExtensibleDynamicEvaluationStatistics.java
hmottestad: Will do :D - it got a lot worse when I added in the index_removed support.

core/sail/elasticsearch-store/src/test/java/org/eclipse/rdf4j/sail/elasticsearchstore/benchmark/DeleteBenchmark.java
```diff
@@ -60,7 +61,7 @@
 	public void beforeClass() throws IOException, InterruptedException {
 
 		embeddedElastic = TestHelpers.startElasticsearch(installLocation,
-				"/Library/Java/JavaVirtualMachines/jdk1.8.0_144.jdk/Contents/Home");
+				"/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home");
```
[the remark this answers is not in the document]
hmottestad: Yeah. Because of how benchmark forks the jvm, the JAVA_HOME env isn't being passed through as expected. This was a dirty hack to get it to function.

core/sail/extensible-store/src/main/java/org/eclipse/rdf4j/sail/extensiblestore/evaluationstatistics/ExtensibleDynamicEvaluationStatistics.java
hmottestad: It's to do async stats updates. The thread is started if the size of the queue is likely to be above the limit, then runs until the queue is empty. Since the stats are only estimates there is no guarantee for timeliness. Those guarantees are needed when running tests and benchmarking, so there is a poor mans latch that waits for the thread to finish.

jeenbroekstra:
Can this be postponed to release 3.2.0? I want to start creating the IP log for 3.1, and if we submit it for review without the Druid CQ, we can't merge this in until after the release.

hmottestad:
> Can this be postponed to release 3.2.0? I want to start creating the IP log for 3.1, and if we submit it for review without the Druid CQ, we can't merge this in until after the release.

Sure. This PR isn't finished anyway and might not be done for another few weeks.

hmottestad:
B38A58CA-68B4-42AB-A189-963D5A3DA6A9

That there doesn’t look particularly compatible with our license does it?

jeenbroekstra:
From the pic it looks like it's Apache licence, that's fine to use for us. As is BSD licence.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -657,17 +656,17 @@ public void clearInferred(Resource... contexts) throws SailException {
...
+				inferredOnlyDataset = branch.dataset(level);
+				inferredOnlySink = branch.sink(level);
+				explicitOnlyDataset = branch(IncludeInferred.explicitOnly).dataset(level);
 			}
 			if (this.hasConnectionListeners()) {
-				remove(null, null, null, inferredDataset, inferredSink, contexts);
+				remove(null, null, null, inferredOnlyDataset, inferredOnlySink, contexts);
 			}
-			inferredSink.clear(contexts);
+			inferredOnlySink.clear(contexts);
```
hmottestad: @jeenbroekstra 

I found out that the SailSourceConnection will get all statements and remove them from the inferred sink, instead of getting just the inferred statements. So I fixed that by introducing a tri-state enum, essentially you can now specify: all, inferredOnly or explicitOnly. This is only for this class....and `branch(...)` is a private method, so no breaking changes. 

I don't have any decent benchmarks for the rdfs sail that add and remove statements, but I do have some for the SPIN sail and that's seeing a 20% speed bump for some of the benchmarks.

jeenbroekstra: I'm lost. Is this related to #1795 ?

hmottestad: Not related no. This just makes clearing inferred statements faster.

core/sail/elasticsearch-store/src/test/java/org/eclipse/rdf4j/sail/elasticsearchstore/benchmark/DeleteBenchmark.java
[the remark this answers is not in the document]
hmottestad: I'm not going to try to fix this.

hmottestad:
@jeenbroekstra I'm getting ready to merge this now. Want to start testing it through work. I've cleaned up the things you pointed out, added some documentation to the classes, cleaned up and added some more tests. 

The biggest thing missing from this is support for storing and loading evaluation statistics between restarts. I'm going to consider that in a new PR, and I might switch out the HLL implementation when I look into that.

core/sail/elasticsearch-store/src/test/java/org/eclipse/rdf4j/sail/elasticsearchstore/benchmark/DeleteBenchmark.java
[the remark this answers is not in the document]
jeenbroekstra: Can you add a comment to explain why this is there?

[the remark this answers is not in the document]
hmottestad: Will do.

jeenbroekstra:
Other than adding the one code comment I'm fine with this being merged.

Issue #1802, opened 2020-01-02
RepositoryFederatedService has issues with vectored evaluation using VALUES clause


aschwarte10:
While doing some advanced federation tests I stumbled into some issues in the RepositoryFederatedService implementation, when it comes to the optimized vectored evaluation using the VALUES clause.

The optimization was introduced to reduce the number of remote requests to the service by providing the relevant bindings to the SERVICE using a VALUES clause, i.e. by pushing the join to the endpoint using blocks of join inputs.

In the current implementation the VALUES clauses is simply appendded to the SERVICE query at evaluation time, and we see queries like the following:

```
SELECT * WHERE {
    // service patterns here
} VALUES ?input { ... }
```

This has some disadvantages and issues due to the scope of variables when the VALUES clause is provided outside the WHERE block. 

These issues are summarized and illustrated in the following

1) Dependending on the data layout, the current implementation has significant influence on performance: the join / filtering on the relevant bindings is only done after the evaluation of the SERVICE WHERE pattern.

Scenario: We are interested in the names of a subset of persons for which we have computed the identifiers as input to the SERVICE. The (simplified) SERVICE query at evaluation time may look similar to the following

SELECT ?name WHERE { ?person foaf:name ?name } VALUES ?person { :person1 ?person2 ... }

Imagine the database has a high number of persons: this would result in a full table scan although we are only interested in a very small subset identified by the provided bindings. This is because of the scope of the provided VALUES (i.e. the join happens after evaluating the WHERE block)

2) Correctness

We currently run into query evaluation exceptions if the SERVICE block defines a full SELECT query. This is because the projections are defined by the query already, while the evaluation routine of the RepositoryFederatedService expects to add one of its own projection to map the rowIndex. 

This issue will be covered by a unit test.

3) Using bindings as input in the SERVICE routine

In some query engines the join order might be adjusted in a way to really using bindings as input in the SERVICE block. Such queries currently yield invalid results due to the scope of variables from the VALUES clause.

For illustration consider the following query:

```
SELECT * WHERE {
  ?s :p ?input .
  SERVICE urn:dummy { 
     BIND(CONCAT(?input, '_processed') AS ?output)
  }
}
```

Note that this example is a bit artificial, however, it shows that in certain engines it may be desirable to have the ?input bindings available in the the SERVICE block, i.e. users might intuitively think that bindings are available. Further note that above query strongly depends on the join order, i.e. the executing engine must be capable of producing a working join order (e.g. through query hints)

[...]

PR #1803, opened 2020-01-02
GH-1802 Fix vectored evaluation using VALUES in RepositoryFederatedService

aschwarte10:
GitHub issue resolved: #1802 

This change adjusts the optimized evaluation of SERVICE queries using
the VALUES clause.

In order to address performance and correctness issues, the VALUES
clause is now inserted in the beginning of the WHERE block (instead of
appending it to the generated query).

This change also does some cleanup to the RepositoryFederatedService
implementation (e.g. replacing deprecated code)

Change is covered by numerous integration unit tests

See #1802 for full details

### DIFF
--- a/compliance/repository/pom.xml
+++ b/compliance/repository/pom.xml
@@ -59,6 +59,20 @@
 			<artifactId>rdf4j-repository-manager</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-repository-sail</artifactId>
+			<version>${project.version}</version>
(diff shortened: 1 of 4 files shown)


aschwarte10:
@jeenbroekstra let me know if you are OK with the change in the RepositoryFederatedService. I am not entirely happy to use string manipulation logics on queries, but as far as I can see this is currently the only way. Happy to learn about better options.

The test cases illustrate different types of query patterns that I could come up with. They may look a bit strange, but this is the easiest way of reproducing integration tests. Using mocking is not easily possible in this scenario, as I explicitly want to rely on federation logics.

If you have questions we can also discuss this in more detail (also the rationals why I would like to do things like this)

Issue #1555, opened 2019-09-25
Documentation for FedX


jeenbroekstra:
Documentation should be included as part of the wider rdf4j documentation.

aschwarte10:
@jeenbroekstra can you point me to some developer help for this issue? I.e where an how can I create the documentation, haven't done this before

jeenbroekstra:
Ah I just posted about this on the developer list.

aschwarte10:
PR for the main documentation available in https://github.com/eclipse/rdf4j-doc/pull/118

Note: in addition to the above (which is the documentation for using the federation), I am planning to provide more documentation on the usage of FedX within the RDF4J workbench

PR #1804, opened 2020-01-02
GH-1555 Improve FedX Demo code and examples

aschwarte10:
GitHub issue resolved: #1555  

Improve the examples and demo snippets

### DIFF
--- a/tools/federation/src/test/java/demos/MonitorRequestsDemo.java
+++ b/tools/federation/src/test/java/demos/MonitorRequestsDemo.java
@@ -7,41 +7,50 @@
  *******************************************************************************/
 package demos;
 
-import java.io.File;
-
+import org.eclipse.rdf4j.federated.FedXConfig;
 import org.eclipse.rdf4j.federated.FedXFactory;
+import org.eclipse.rdf4j.federated.monitoring.MonitoringUtil;
+import org.eclipse.rdf4j.federated.repository.FedXRepository;
 import org.eclipse.rdf4j.query.QueryLanguage;
 import org.eclipse.rdf4j.query.TupleQuery;
 import org.eclipse.rdf4j.query.TupleQueryResult;
-import org.eclipse.rdf4j.repository.Repository;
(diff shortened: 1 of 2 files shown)


tools/federation/src/test/java/demos/MonitorRequestsDemo.java
```diff
@@ -0,0 +1,61 @@
...
+			try (TupleQueryResult res = query.evaluate()) {
+
+				int count = 0;
+				while (res.hasNext()) {
+					res.next();
+					count++;
+				}
+
+				System.out.println("# Done, " + count + " results");
+			}
```
hmottestad: Btw. Counting is nicer with 
```
try(.... stream = Iterations.stream(query.evaluate)){
   .... count = stream.count();
}
```

jeenbroekstra: Or, even simpler: 

```
 List<BindingSet> result = QueryResults.asList(query.evaluate());
 System.out.println("# Done, " + result.size() + " results");
```

But it depends on what you're trying to illustrate of course.

hmottestad: That keeps all your results in memory.

jeenbroekstra: Yes. Typically not a big deal unless you do monstrously big queries - but you're of course correct that streaming for this kind of thing is more efficient. 

I'm fine with either way, also fine if you keep the example as-is @aschwarte10 - the point of this doc is not really to explore the different ways to process RDF4J query results of course :)

aschwarte10: Thanks for the comments, very useful patterns for the examples, will keep them in mind for future changes

The purpose of this demo is is not the evaluation and consumption of the query, but rather the federation setup and specific configuration options. The system.out.println or in this example the count is just the simplest example. In here we in any case have something like 10 results only.

I would suggest to merge.

Issue #1787, opened 2019-12-28
More seamless integration of FedX Repository into RDF4J workbench


aschwarte10:
As of #1538 we integrated FedX into the RDF4J project. The code integration is looking quite fine, as validations using the current milestone reveal.

However, the integration into the RDF4J workbench is not as seamless as it could be:

* currently the rdf4j-tools-federation jar is not part of the rdf4j-server.war distribution (thus the binary repository implementation is not available in the workbench)
* the repository configuration template is not offered in the RDF4J workbench
* documentation is not provided for the integration

Currently FedX can be used in the RDF4J workbench with the following manual steps

1) Copy `rdf4j-tools-federation.jar` into the deployed rdf4j-server webapp (i.e. to `webapps/rdf4j-server/WEB-INF/lib`)
2) Manually create a repository folder in the RDF4J repositories dir with the federation config depicted below, e.g. to `/users/home/myuser/.RDF4J/server/repositories/my-federation/config.ttl`

```
#
# RDF4J configuration template for a FedX Repository
#
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix rep: <http://www.openrdf.org/config/repository#>.
@prefix fedx: <http://rdf4j.org/config/federation#>.

[] a rep:Repository ;
   rep:repositoryImpl [
      rep:repositoryType "fedx:FedXRepository" ;
      fedx:member [
         fedx:store "ResolvableRepository" ;
         fedx:repositoryName "repo1" 
      ] ,
      [
         fedx:store "ResolvableRepository" ;
         fedx:repositoryName "repo2" 
      ]
   ];
   rep:repositoryID "my-federation" ;
   rdfs:label "FedX Federation" .
```

This will create a federation repositoy in the RDF4J workbench (including all convenience like lifecycle management, SPARQL endpoint, ...). This example will have the two repositories "repo1" and "repo2" as members. The repositories are created and managed by the RDF4J Workbench (i.e. they can be created from the RDF4J workbench UI, for example as NativeStore)

aschwarte10:
@jeenbroekstra can you comment on whether you see this in for 3.1? Also can you provide me with some pointer on what needs to be done to get the jar into the webapp?

Regarding the template: from my understanding it would currently mean to add it as a resource to `/core/repository/api/src/main/resources/org/eclipse/rdf4j/repository/config`, as only these are exposed to the RDF4J workbench UI

jeenbroekstra:
[...]

PR #1811, opened 2020-01-04
GH-1787: integrate FedX federation support into the RDF4J workbench

aschwarte10:
GitHub issue resolved: #1787

This change integrates the 'rdf4j-tools-federation' module into
'rdf4j-server.war' and thus makes it available in the RDF4J workbench.

For convenience we reuse the existing UI facilities to create a
federation: instead of creating a federation sail, we now create a FedX
based Federation repository.

Note that dependencies on rdfj4-tools-federated have been explicitly
added to rdf4j-server and rdf4j-workbench. Ideally the dependency could
be added to rdf4j-storage, however this currently has cyclic dependency
issues.

Signed-off-by: Andreas Schwarte <aschwarte10@gmail.com>

### DIFF
--- a/tools/federation/pom.xml
+++ b/tools/federation/pom.xml
@@ -9,6 +9,7 @@
 
 	<name>RDF4J: Federation</name>
 	<description>A federation engine for virtually integrating SPARQL endpoints</description>
+	<packaging>jar</packaging>
 
 	<parent>
 		<groupId>org.eclipse.rdf4j</groupId>
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfigBuilder.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfigBuilder.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.federated.repository;
+
+import java.util.Collection;
+
+import org.eclipse.rdf4j.federated.util.Vocabulary.FEDX;
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.impl.TreeModel;
+import org.eclipse.rdf4j.repository.config.RepositoryConfig;
+
+/**
+ * Builder for {@link FedXRepositoryConfig}
+ * 
+ * @author Andreas Schwarte
+ */
+public class FedXRepositoryConfigBuilder {
+
+	public static FedXRepositoryConfigBuilder create() {
+		return new FedXRepositoryConfigBuilder();
+	}
+
+	private final Model members = new TreeModel();
+
+	private FedXRepositoryConfigBuilder() {
+	}
+
+	public FedXRepositoryConfigBuilder withResolvableEndpoint(String memberId) {
+		ValueFactory vf = SimpleValueFactory.getInstance();
+		members.add(vf.createIRI(FEDX.NAMESPACE, memberId), FEDX.STORE, vf.createLiteral("ResolvableRepository"));
+		members.add(vf.createIRI(FEDX.NAMESPACE, memberId), FEDX.REPOSITORY_NAME, vf.createLiteral(memberId));
+		return this;
+	}
+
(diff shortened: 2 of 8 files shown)


aschwarte10:
Hints for local testing (i.e. what I did): 

* run `mvn package` to produce rdf4j-server.war and rdf4j-workbench.war
* deploy the webapps in a web server (e.g tomcat)
* create two NativeStore repositories repo1 and repo2, and insert some test data
* create a federation through the UI (select "Federation" from the dropdown in the workbench), add repo1 and repo2 as members

Alternatively copy the following content to `.RDF4J/server/repositories/my-federation/config.ttl`

```
#
# RDF4J configuration template for a FedX Repository
#
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix rep: <http://www.openrdf.org/config/repository#>.
@prefix fedx: <http://rdf4j.org/config/federation#>.

[] a rep:Repository ;
   rep:repositoryImpl [
      rep:repositoryType "fedx:FedXRepository" ;
      fedx:member [
         fedx:store "ResolvableRepository" ;
         fedx:repositoryName "repo1" 
      ] ,
      [
         fedx:store "ResolvableRepository" ;
         fedx:repositoryName "repo2" 
      ]
   ];
   rep:repositoryID "my-federation" ;
   rdfs:label "FedX Federation" .
```

hmottestad:
I gave this a spin. 

I used a memory+rdfs and a memory, then created a federation between them. 

None of the inferred statements in the memory+rdfs are showing up. Is this a known limitation of the federated sail?

Btw. I used the docker setup. Which I prefer since I don't need to have tomcat running locally. 
```
cd assembly/src/main/dist/docker
./build.sh
docker-compose up -d
```
Then workbench will be running at: http://localhost:8080/rdf4j-workbench

hmottestad:
Maybe I'm doing this wrong.

I now tried to add data to one of my memory stores, but i still can't query it from the federated sail.

Any suggestions?

jeenbroekstra:
Also tested this out locally, and basic federation works as expected: I can create the federation through the UI and can query over it. If I add additional data to one of the federation members, it also shows up in the federation.

However, like Havard, I'm also noticing that any inferred triples in the federation members do not show up when querying the federation.

aschwarte10:
> 
> Btw. I used the docker setup. Which I prefer since I don't need to have tomcat running locally.
> 
> ```
> cd assembly/src/main/dist/docker
> ./build.sh
> docker-compose up -d
> ```
> 
> Then workbench will be running at: http://localhost:8080/rdf4j-workbench

Thanks for this hint, I wasn't aware of this script and setup.

I was actually also using a docker container, but used my own docker compose definition with mounted volumes where I put the wars. So quite a bit of additional effort for me

aschwarte10:
Thanks for evaluating this change and for your feedback.

So I understand that creating and using the federation is generally working for physical statements.

I have actually never tested with inference support, so I currently would state it is a limitation of the federation engine. Likely only that I have to set "Query#setIncludeInferred" for my subqueries. 

Maybe I have the chance to still address this for 3.1, should actually be only a smaller change. I tracked it in #1814  I also attached a unit test example which I used for reproducing. Thanks for providing the feedback.

Independent of the missing inference feature (which I see orthogonal to this change), what do you think about this PR?

hmottestad:
I think the UI is simple and intuitive. 

Should there be some way of federating over external endpoints too?

Btw. I didn’t get physical statements to work. What will have to test more.

aschwarte10:
> I think the UI is simple and intuitive.
> 
> Should there be some way of federating over external endpoints too?
> 

Hmm, you can already achieve this by registering a SPARQL Repository in the RDF4J workbench. For more complex cases I think a programmatic or factory approach should be taken.

> Btw. I didn’t get physical statements to work. What will have to test more.

Let me know if you get it working, for my examples it was working. If there are still issues, we should have a look before the release. Maybe it is just configuration?

hmottestad:
I did the following:

 - create two memory repos (mem1, mem2)
 - create one federation repo (fed)
 - query `select * where {?a ?b ?c}` from fed - no results
 - swap to mem2
 - add data by using a sparql update query:
```
PREFIX ex: <http://example.com/ns#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX sh: <http://www.w3.org/ns/shacl#>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

INSERT DATA {

ex:validPerson1 a ex:Person ;
	ex:age 123, "234"^^xsd:long.

}
```
 - query `select * where {?a ?b ?c}` from mem2 - 3 statemenets
 - swap to fed
 - query `select * where {?a ?b ?c}` from fed - no results

Am I not supposed to see results now?

aschwarte10:
Ok I see.

Explanation: for source selection I am using a pattern based cache. This means that the cache keeps information whether a source _potentially_ can provide data for a given pattern (including ?s ?p ?o).

I will look into this and see how I can fine tune this. Probably I switch the current cache implementation and replace it with a Guava based cache with some automatic eviction. Planned to do this anyways. 

Also as a background info: The initial design of the federation was meant for rather kind of static schema / data in the source. For this the current caching approach works fine.

I'll create an issue and propose to solve this also as an orthogonal change, also I'll write this down in the documentation. 

Thanks for the feedback, it's good to get this feedback before the release

hmottestad:
I agree with @jeenbroekstra

Issue #1812, opened 2020-01-05
Minor code cleanup


hmottestad:

jeenbroekstra:
It looks as if pr merge via rebase doesn't trigger automatic closing of the related issue. 😔

PR #1813, opened 2020-01-05
#1812 minor code cleanup

hmottestad:
GitHub issue resolved: #1812   

* minor code cleanup
* 
*

### DIFF
--- a/core/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
+++ b/core/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
@@ -136,14 +136,12 @@ public class ParsedIRI implements Cloneable, Serializable {
 		for (Object set : sets) {
 			if (set instanceof int[][]) {
 				int[][] ar = (int[][]) set;
-				for (int i = 0; i < ar.length; i++) {
-					list.add(ar[i]);
-				}
+				list.addAll(Arrays.asList(ar));
 			} else if (set instanceof Character) {
 				char chr = (Character) set;
 				list.add(new int[] { chr, chr });
 			} else {
-				assert false;
+				throw new IllegalStateException();
 			}
 		}
 		int[][] dest = list.toArray(new int[][] {});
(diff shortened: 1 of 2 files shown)


core/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
```diff
@@ -136,14 +136,12 @@
 		for (Object set : sets) {
 			if (set instanceof int[][]) {
 				int[][] ar = (int[][]) set;
-				for (int i = 0; i < ar.length; i++) {
-					list.add(ar[i]);
-				}
+				list.addAll(Arrays.asList(ar));
```
hmottestad: Arrays.asList(ar) is a free operation.

core/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
```diff
@@ -136,14 +136,12 @@
...
 			} else if (set instanceof Character) {
 				char chr = (Character) set;
 				list.add(new int[] { chr, chr });
 			} else {
-				assert false;
```
hmottestad: Asserts only execute if asserts are enabled. Throwing exceptions is better.

core/util/src/main/java/org/eclipse/rdf4j/common/net/ParsedIRI.java
```diff
@@ -237,8 +234,7 @@ public static ParsedIRI create(String str) {
 					}
 				}
 			} catch (UnsupportedEncodingException ex) {
-				assert false;
-				return null;
+				throw new IllegalStateException(ex);
```
jeenbroekstra: I'm happy to get rid of assertions, but I'd like to dig in a bit deeper before approving this, because it changes the production behavior: previously it would have silently returned here, now it throws an exception.

jeenbroekstra: Never mind, has a closer look, totally a "should never happen". Happy to merge.

Issue #1816, opened 2020-01-06
Improve and fix evaluation of SERVICE in FedX federation


aschwarte10:
For evaluation of (external) SERVICE clauses, i.e. federated services that are not explicitly member of the federation, FedX relies on the RepositoryFederatedService implementation.

Currently FedX is configured to use the simple evaluation (i.e. non-bound join using VALUES). This can already be changed in the configuration.

The goals of this task are

* activate bound join evaluation using the VALUES clause for external SERVICE clauses by default
* validate that there are no issues (no blocking, deadlocks)
* add unit tests covering this

A second orthogonal improvement of this issue is to make sure that any query is evaluated in the FedX engine, if it contains a SERVICE clause - even if the federation has only a single member. This is to guarantee consistent behavior in the evaluation of SERVICE clauses

aschwarte10:
PR is merged, closing issue

PR #1817, opened 2020-01-06
GH-1816: improve (external) SERVICE evaluation in FedX engine

aschwarte10:
GitHub issue resolved: #1816  

This change improves and fixes the evaluation of (external) SERVICE
clauses in FedX.

Before this change there have been issues with blocking behavior and and
hanging queries, when SERVICE queries were evaluated as bound joins. As
this was not the default, it was not yet discovered. The reason was that
HTTP slots in the background tuple results were not processed in a
suitable order.

As of this change we slightly adopted the evaluation logics: now we
basically consume the background tuple result and keep it in memory for
upcoming operators, thus process it in order.

Also the Bound Join evaluation is now activated by default.

Moreover, a slight follow up of #1802 is added: the boundJoinBlockSize
is no longer hard-coded in the RepositoryFederatedService, but now uses
the value from the field.

Always evaluate query in FedX if it contains SERVICE

If a query contains a SERVICE clause, it is now always evaluated in the
FedX engine (even if the federation onyl has a single member)

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
@@ -71,7 +71,9 @@ import org.eclipse.rdf4j.query.QueryEvaluationException;
 import org.eclipse.rdf4j.query.algebra.QueryRoot;
 import org.eclipse.rdf4j.query.algebra.TupleExpr;
 import org.eclipse.rdf4j.query.algebra.ValueExpr;
+import org.eclipse.rdf4j.query.algebra.Var;
 import org.eclipse.rdf4j.query.algebra.evaluation.ValueExprEvaluationException;
+import org.eclipse.rdf4j.query.algebra.evaluation.federation.FederatedService;
 import org.eclipse.rdf4j.query.algebra.evaluation.federation.ServiceJoinIterator;
 import org.eclipse.rdf4j.query.algebra.evaluation.impl.ConstantOptimizer;
 import org.eclipse.rdf4j.query.algebra.evaluation.impl.DisjunctiveConstraintOptimizer;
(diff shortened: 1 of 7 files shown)


aschwarte10:
The change still has some blocking issues - need to have a second look

aschwarte10:
> The change still has some blocking issues - need to have a second look

Found it. The working solution is to consume intermediate results from SERVICEs and keep them in memory. This guarantees that no HTTP slots are blocked. This is required because in FedX I execute services in parallel, thus potentially blocking HTTP slots depending on the order.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
```diff
@@ -516,8 +519,23 @@ protected void optimizeJoinOrder(TupleExpr query, QueryInfo queryInfo, GenericIn
 	public CloseableIteration<BindingSet, QueryEvaluationException> evaluateService(FedXService service,
 			final List<BindingSet> bindings) throws QueryEvaluationException {
 
-		return new ServiceJoinIterator(new CollectionIteration<>(bindings),
-				service.getService(), EmptyBindingSet.getInstance(), this);
+		Var serviceRef = service.getService().getServiceRef();
+		String serviceUri;
+		if (serviceRef.hasValue())
+			serviceUri = serviceRef.getValue().stringValue();
```
jeenbroekstra: This is nitpicky but we have a style agreement to always use curly braces for if-then-else clauses, even if they're single line.

aschwarte10: I fully agree, must have missed this. Will amend the change and add the brackets. Thanks

aschwarte10: Added this to the first commit. No other changes added

Issue #1814, opened 2020-01-06
Support inference in FedX federation (if members support it)


aschwarte10:
Currently FedX does not have support for inference if one of the configured federation members supports it. This means that a federation does not show any inferred triples.

Example scenario:

a) repo1 = memory store with rdfs
:p1 a foaf:Person .
:p2 a :Manager .
:Manager rdfs:subClassOf foaf:Person .

b) repo2 = memory store
:p3 a foaf:Person

Query: SELECT * WHERE { ?p a foaf:Person }

PR #1818, opened 2020-01-09
GH-1814: support to configure inference on federation queries

aschwarte10:
GitHub issue resolved: #1814 

As of this change it is possible to configure inference on federation
queries using Query#setIncludeInferred. By default the value is set to
'true', similar to the SailQuery.

The corresponding value is now passed to the TripleSource and is thus
considered for all sub-queries sent to the relevant endpoints.

Functionality for the query types is validated in a unit test.

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
@@ -97,7 +97,7 @@ public class FedXConnection extends AbstractSailConnection {
 			if (queryString == null)
 				log.warn("Query string is null. Please check your FedX setup.");
 			queryInfo = new QueryInfo(queryString, getOriginalQueryType(bindings),
-					getOriginalMaxExecutionTime(bindings), federationContext);
+					getOriginalMaxExecutionTime(bindings), includeInferred, federationContext);
 
 			if (log.isDebugEnabled()) {
 				log.debug("Optimization start (Query: " + queryInfo.getQueryID() + ")");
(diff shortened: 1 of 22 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
```diff
@@ -388,6 +389,21 @@ private static int getOriginalMaxExecutionTime(BindingSet b) {
 		return 0;
 	}
 
+	/**
+	 * Return the original explicit includeInferred {@link Operation#getIncludeInferred()} value.
+	 * 
+	 * @param b
+	 * @return the supplied value or {@link FedXConfig#getIncludeInferredDefault()}
+	 */
+	private boolean getOriginalIncludeInferred(BindingSet b) {
```
aschwarte10: We really should get a possibility inside the SAIL to access information on the Query object (e.g. in this case the includeInferred), I am using nasty workarounds by passing it as a binding

aschwarte10:
Unfortunately became a bit lengthy because I had to pass around the QueryInfo object (which keeps information about the general query, including includeInferred).

Most relevant parts of this change are the TripleSource implementation and the unit test

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
jeenbroekstra: I haven't looked at this long so perhaps I'm missing something obvious, but  I don't understand why you need this at all. The `includeInferred` parameter is an input for `evaluateInternal`. Why do you need this passing arround through a config object at all?

aschwarte10: @jeenbroekstra thanks for this. You are absolutetly right, includeInferred is available in evaluateInternal. Did not look properly or have missed it. This will make my change a bit simpler.
I will provide a revised version later.

Still for the query execution time there is no way to access it inside the sail.

aschwarte10: Done, I amended the change and could get rid of this piece of code

Issue #1809, opened 2020-01-04
Custom Collector class for Model objects


jeenbroekstra:
As an RDF4J user, I want to have a custom collector that collects the results of a Java stream in a `Model` object, so that I can do thing like this:

	Model result = QueryResults.stream(conn.getStatements(null, RDF.TYPE, null))
				.map(s -> vf.createStatement(s.getSubject(), RDFS.LABEL, vf.createLiteral("bar")))
				.collect(ModelCollectors.toModel());

See https://blog.indrek.io/articles/creating-a-collector-in-java-8/ for some background on how to create a custom collector.

jeenbroekstra:
I've just realized that you can already do something along these lines with standard java collector functionality:

```
Model result = QueryResults.stream(conn.getStatements(null, RDF.TYPE, null))
			.map(s -> vf.createStatement(s.getSubject(), RDFS.LABEL, vf.createLiteral("bar")))
			.collect(Collectors.toCollection(LinkedHashModel::new));
```

I think a custom collector class still makes sense to have though.

PR #1819, opened 2020-01-09
GH-1809: Added collector for Model + test

barthanssens:
GitHub issue resolved: #1809   

* Added ModelCollector class
* Added tests for serial and parallel stream

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.model.util;
+
+import java.util.EnumSet;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collector;
+
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.ModelFactory;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.impl.LinkedHashModelFactory;
+import org.eclipse.rdf4j.model.impl.TreeModelFactory;
+
+/**
+ * Collects a stream of Statements into a Model. By default a {@link org.eclipse.rdf4j.model.impl.LinkedHashModel
+ * LinkedHashModel} will be returned.
+ * 
+ * @author Bart Hanssens
+ */
+public class ModelCollector implements Collector<Statement, Model, Model> {
+	private final ModelFactory factory;
+
+	/**
+	 * Constructor
+	 */
+	public ModelCollector() {
+		this.factory = new LinkedHashModelFactory();
+	}
+
+	/**
+	 * Constructor
+	 * 
+	 * @param factory
+	 */
+	public ModelCollector(ModelFactory factory) {
+		this.factory = factory;
+	}
+
+	/**
+	 * Convenience method to obtain a ModelCollector.
+	 * 
+	 * @return a ModelCollector
+	 */
+	public static ModelCollector toModel() {
+		return new ModelCollector();
+	}
+
+	/**
+	 * Convenience method to obtain a ModelCollector, which will return a TreeModel.
+	 * 
+	 * @return a ModelCollector
+	 */
+	public static ModelCollector toTreeModel() {
+		return new ModelCollector(new TreeModelFactory());
+	}
+
+	@Override
+	public Supplier<Model> supplier() {
+		return () -> factory.createEmptyModel();
+	}
+
+	@Override
+	public BiConsumer<Model, Statement> accumulator() {
+		return (m, s) -> {
+			synchronized (m) {
+				m.add(s);
+			}
+		};
+	}
+
+	@Override
+	public BinaryOperator<Model> combiner() {
+		return (m1, m2) -> {
+			m1.addAll(m2);
+			return m1;
+		};
+	}
+
+	@Override
+	public Function<Model, Model> finisher() {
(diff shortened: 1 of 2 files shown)


barthanssens:
Never wrote a collector before, but I followed the excellent article.
Would this be more or less what was required ?

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
```diff
@@ -0,0 +1,74 @@
...
+			synchronized (lock) {
+				m.add(s);
+			}
+		};
+	}
+
+	@Override
+	public BinaryOperator<Model> combiner() {
+		return (m1, m2) -> {
+			m1.addAll(m2);
```
hmottestad: does this also need to be synchronized?

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Override
+	public Supplier<Model> supplier() {
+		return LinkedHashModel::new;
+	}
+
+	@Override
+	public BiConsumer<Model, Statement> accumulator() {
+		return (m, s) -> {
+			synchronized (lock) {
```
hmottestad: could this be synchronized on `m` instead of on `lock`?

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
```diff
@@ -0,0 +1,74 @@
...
+	 * 
+	 * @return a ModelCollector
+	 */
+	public static ModelCollector toModel() {
+		return new ModelCollector();
+	}
+
+	@Override
+	public Supplier<Model> supplier() {
+		return LinkedHashModel::new;
```
hmottestad: Could this use ModelFactory, so that we can add a toLinkedHashModel() and toTreeModel() method? I would like the default toModel() method to use the LinkedHashModel as this is typically the fastest.

hmottestad:
I think this looks great. Made a couple of comments, mostly questions and nothing that should stop this from getting merged.

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
barthanssens: Actually, if I understand the Collector correctly, this method could be removed or implemented returning just null....

Because the combiner does not get called in parallelStream() with the concurrent and unordered characteristics set.
Interestingly enough, with OpenJDK11 on my machine, the combiner _does_ get called in sequential stream(), but the result seems to be ignored anyway (just returning null does not break the test code)

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
barthanssens: thanks, seems to work

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
barthanssens: I've slightly modified the code to use ModelFactory, so a toTreeModel() could indeed be added

core/model/src/main/java/org/eclipse/rdf4j/model/util/ModelCollector.java
```diff
@@ -0,0 +1,92 @@
...
+		return Function.identity();
+	}
+
+	@Override
+	public Set<Characteristics> characteristics() {
+		Set<Characteristics> characteristics = new HashSet<>();
+		characteristics.add(Characteristics.CONCURRENT);
+		characteristics.add(Characteristics.IDENTITY_FINISH);
+		characteristics.add(Characteristics.UNORDERED);
+		return characteristics;
```
hmottestad: I took a look at some examples online and found a cool class called EnumSet:
```
return EnumSet.of(Characteristics.CONCURRENT, Characteristics.IDENTITY_FINISH, Characteristics.UNORDERED);
```

barthanssens: Aha, just what we need. Didn't know that one.

barthanssens:
FWIW I did some preliminary tests, on my quad-core the parallelstream is faster than the sequential stream until somewhere between 250K - 500K statements (I suppose the concurrent access of the Model creates more overhead)

But IMHO that's good enough...

Not using a concurrent Model (= separate Model instances per thread + running combiner) does not seem to help: then the paralleStream is slower from the beginning (iterating over the separate Models + adding all statements into 1)

hmottestad:
The combiner might work if we managed to create some sort of virtual model that wrapped a bunch of underlying models. Don't think it's worth it though.

I can make some jmh benchmarks for the current implementation after it's merged.

barthanssens:
@rdf4j-bot run tests

jeenbroekstra:
> @rdf4j-bot run tests

That's a jenkins CI thing, we no longer use that I'm afraid. If you want to re-run PR verification, click on 'Details' behind one of the "pull request verify" checks, and in the next screen hit "re-run checks" (top-right). 

Alternatively, push a new (empty) commit to trigger a new build.

jeenbroekstra:
The github build is failing, for some reason specifically on building the rdf4j-benchmark module. It is trying and failing to download `jakarta.annotation-api` but getting an error on that. 

Btw it's not just this PR that is failing with the error,  #1830 has the same problem. Is this just a glitch in the central repo or is something else going on? 

Fwiw I'm also unsure why we even need that dependency - looks like a Java 9-specific thing.

barthanssens:
Could be a coincidence, but see also https://www.alphabot.com/security/blog/2020/java/Your-Java-builds-might-break-starting-January-13th.html

barthanssens:
Seems to work, so I'll squash and merge it locally

barthanssens:
And now it doesn't work :-( Seems like a network glitch / caching issue, since I can download the pom and jar from maven ...
The annotation API is needed in JDK 9+ (because the benchmark uses some specific annotations that were removed in 9), though it is still strange that the Java8 check is also cancelled when the Java11 check fails.

hmottestad:
> And now it doesn't work :-( Seems like a network glitch / caching issue, since I can download the pom and jar from maven ...
> The annotation API is needed in JDK 9+ (because the benchmark uses some specific annotations that were removed in 9), though it is still strange that the Java8 check is also cancelled when the Java11 check fails.

The github ci cancels all other checks if one fails.

Issue #1820, opened 2020-01-09
RepositoryFederatedService reuses connections and does not always guarantee to close them


aschwarte10:
The current implementation of the RepositoryFederatedService reuses a RepositoryConnection, which is opened upon the first query evaluation.

There are two issues with this approach:

1) the retrieval of the "global" connection is not synchronized, i.e. if there are concurrent attempts to execute a SERVICE query, it may happen that two RepositoryConnections are opened. In addition, only a single reference is kept for closing. The other connection stays open and is thus a potential resource leak

2) the best practice in connection handling is to perform independent operations in isolated connections (leaving pooling etc to the underlying repository). This means that ideally a fresh connection should be openend for each service.

The goal of this task is to resolve the first issue by allowing synchronized access, and moreover introduce a new functionality to support fresh connections

aschwarte10:
While validating the M2 release I found that under certain conditions connections are still not properly closed: we are missing some catch clause in the simple SERVICE evaluation of RepositoryFederatedService. Patch request is coming up shortly

PR #1821, opened 2020-01-09
GH-1820: fix connection handling in RepositoryFederatedService

aschwarte10:
GitHub issue resolved: #1820 

This change introduces a mode in RepositoryFederatedService which allows
to always use a fresh RepositoryConnection for the actual query
invocation. In order to properly close the connection, the result
iteration is wrapped in a "CloseConnectionIteration"

The new mode is active by default, i.e. as of this change a fresh
connection is used for each SERVICE invocation.

For the re-use of the managed connection we now made sure to support
synchronized access, i.e. avoid to loose the references to created
connections.

The unit tests show that connections are properly closed. If it wasn't
the case, we would observe hanging tests for > 30s (i.e. until the
automatic timeout occurs)

### DIFF
--- a/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
+++ b/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
@@ -2,6 +2,9 @@ package org.eclipse.rdf4j.repository.sparql.federation;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.eclipse.rdf4j.common.iteration.Iterations;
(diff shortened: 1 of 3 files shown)


compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
```diff
@@ -262,6 +265,59 @@ public void test8a_subSelectAll() throws Exception {
...
+						.map(value -> vf.createStatement(iri("s1"), RDFS.LABEL, value))
+						.collect(Collectors.toList()));
+
+		String query = "SELECT ?var WHERE { SERVICE <urn:dummy> { ?s ?p ?var  } }";
+		try (RepositoryConnection conn = localRepo.getConnection()) {
+			try (TupleQueryResult tqr = conn.prepareTupleQuery(query).evaluate()) {
+
+				// consume only two items
+				tqr.next();
+				tqr.next();
```
jeenbroekstra: Could you add a comment here that explains why this test is a good indicator if things are properly closed? I know you explained in the PR description but I'm thinking it would be good to have it as a code comment.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedService.java
```diff
@@ -419,13 +503,28 @@ public void shutdown() throws QueryEvaluationException {
...
-		// (TODO add mechanism to unset/close connection)
-		if (conn == null) {
-			conn = rep.getConnection();
+	/**
+	 * Return a fresh {@link RepositoryConnection} from the configured repository.
+	 * 
+	 * @return
+	 * @throws RepositoryException
+	 */
+	protected RepositoryConnection freshConnection() throws RepositoryException {
```
jeenbroekstra: Same here: does this need to be protected? Or can it be made private?

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedService.java
```diff
@@ -128,10 +172,15 @@ protected void handleBindings() throws Exception {
 	 */
 	protected int boundJoinBlockSize = 15;
 
+	/**
+	 * Whether to use a fresh repository connection for individual queries
+	 */
+	protected boolean useFreshConnection = true;
```
jeenbroekstra: Can this be `private` instead?

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
aschwarte10: Added a comment to both tests.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedService.java
aschwarte10: done, made to private

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedService.java
aschwarte10: Done, see amended commit

aschwarte10:
@jeenbroekstra can you maybe provide the final approval? I addressed your minor comments (see above resolved conversations) in the amended commit. Would be good to have this change in the M2 build such that I can validate it in our application

PR #1826, opened 2020-01-11
deploy fix: upgrade dependency and add some timeout and reduce polling

hmottestad:

* set longer timeout 2 hours instead of 10 min (default)
* polling every 10 sec instead of every 3 (default)
*


hmottestad:
@jeenbroekstra no idea if this will actually make a difference, but worth a shot i reckon.

Issue #1831, opened 2020-01-12
Stop deploying compliance tests to Sonatype


jeenbroekstra:
Our current build process 'accidentally' deploys all compliance test modules to Sonatype (and the central repo). This is unintentional: compliance tests are internal only and not intended to be distributed as part of a release. 

The plugin setup needs to be tweaked.

PR #1832, opened 2020-01-12
GH-1831 configure nexus staging plugin to skip deployment for compliance modules

jeenbroekstra:
GitHub issue resolved: #1831  

* tweaked nexus-staging plugin to skip deployment for compliance modules
* removed obsolete reference to old maven-deploy-plugin (which we don't use as the nexus plugin replaces it)


hmottestad:
good catch

Issue #1031, opened 2018-05-08
NativeStore leaks memory with IsolationLevels.SERIALIZABLE


knoan:
# RDF4J Version

```
<groupId>org.eclipse.rdf4j</groupId>
<artifactId>rdf4j-bom</artifactId>
<version>2.3.1</version>
```

# Script

Execute the attached Java program ([Leak.java.txt](https://github.com/eclipse/rdf4j/files/1983809/Leak.java.txt))

# Expected

A steady memory usage, after the first warm-up iterations, like:

```
 1       4,419
 2       6,442
 3      11,069
 4      11,003
 5      10,965
 6      10,943
 7      10,932
 8      10,933
 9      15,615
10      15,650
11      15,639
12      15,612
13      15,588
14      15,572
15      15,568
```

# Actual

Memory usage steadily increases as performance drops, like:

```
 1      43,717
 2      84,822
 3     124,476
 4     166,430
 5     205,395
 6     245,391
 7     284,710
 8     328,641
 9     369,343
10     408,580
11     447,889
12     487,170
13     526,771
14     566,433
15     606,143
```

# Notes

- works as expected if connection isolation level is reduced to `SNAPSHOT` or not explicitly set;
- transactions are immaterial;
- a differential memory snapshot on a web application affected by the issue revels lots of objects locked in a Changeset referenced by SailSourceBranch at line 464.

![heap dump](https://user-images.githubusercontent.com/6105462/39756757-04baf228-52cb-11e8-98e2-b845244a5927.png)

![heap references](https://user-images.githubusercontent.com/6105462/39756759-0667e7ac-52cb-11e8-844d-8a1d5aae8bda.png)

![code](https://user-images.githubusercontent.com/6105462/39756761-07af8354-52cb-11e8-9a5b-8e63caf8a5c2.png)

barthanssens:
It is still an issue in 3.1.0-snapshot :-(

barthanssens:
There seems to be a huge regression in 3.1.0.
E.g.  running sample program with RDF4J 3.0.4 on OpenJDK 11 (Windows 10):
```
 1      24.342
 2      46.177
 3      67.632
...
13     283.837
14     305.435
15     327.034
```

But with 3.1.0:
```
 1      96.307
 2     207.978
 3     316.410
 ...
13   1.439.022
14   1.548.799
15   1.658.802
```

hmottestad:
[...]

PR #1840, opened 2020-01-16
#1031 allow flushing for SERIALIZABLE transactions

hmottestad:
GitHub issue resolved: #1031  

* 
* 
*

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
@@ -340,7 +340,7 @@ class SailSourceBranch implements SailSource {
 	void autoFlush() throws SailException {
 		if (autoFlush && semaphore.tryLock()) {
 			try {
-				if (serializable == null && observers.isEmpty()) {
+				if (observers.isEmpty()) {
 					flush();
 				}
 			} finally {


hmottestad:
@jeenbroekstra @barthanssens this is my proposed fix to the SERIALIZALE memory issue, hopefully it's old code that we don't need anymore....

barthanssens:
LGTM, there is good karma in solving old bugs with only a few keystrokes :-)

hmottestad:
I have no tests for this. Would appreciate if someone could figure out a decent one.

barthanssens:
Would this be a thing for the benchmarks, but perhaps for all stores and isolation levels ?
The java file attached to #1031 was already very helpful, so maybe adapt that one and just test if memory does not increase too much ?

jeenbroekstra:
Fwiw I ran the attached java scratch with this fix in place, and am getting much better results:

```
 1       6,236
 2      10,596
 3       9,403
 4       8,536
 5      10,425
 6       7,710
 7       7,136
 8       8,550
 9       8,067
10       9,837
11       6,766
12       7,042
13       7,158
14       7,287
15       6,700
finished in 22 s.
```

jeenbroekstra:
Fwiw I am taking a closer look at the compliance tests we have for txn isolation. I've noticed that most of the tests we have work with very small transactions - I'll do a probe to see if we can make those tests work with slightly larger batches.

Issue #1108, opened 2018-10-09
US-ASCII in TxnStatusFile is slower than using UTF-8


jeenbroekstra:
_From @hmottestad on February 16, 2018 16:13_

TxnStatusFile uses US-ASCII for encoding files. Switching to UTF-8 makes org.eclipse.rdf4j.sail.SailConcurrencyTest#testConcurrentAddLargeTxn 20% faster. StandardCharsets.UTF_8

Is there a reason for using US-ASCII?

Switching to UTF-8 will probably break existing NativeStores. So there needs to be some code for handling legacy stores. Options:
 - Autodetect encoding
 - Option for NativeStore (default: USC-ASCII)
 - Migrate existing stores automatically

_Copied from original issue: eclipse/rdf4j-storage#44_

hmottestad:
@jeenbroekstra I was wondering about picking this up again. I'm a bit worried if this would be considered non-backwards compatible? My worry is if someone tries to upgrade while a transaction is running, then I reckon it would crash when starting up again. However, if the nativestore was properly shutdown it shouldn't cause any issues.

PR #1848, opened 2020-01-19
#1108 transaction status now uses single byte for higher TPS

hmottestad:
GitHub issue resolved: #1108  

* use single bytes for transaction status for higher TPS
* 
*

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
@@ -24,30 +24,35 @@ class TxnStatusFile {
 		 * No active transaction. This occurs if no transaction has been started yet, or if all transactions have been
 		 * committed or rolled back.
 		 */
-		NONE,
+		NONE((byte) 1),
 
 		/**
 		 * A transaction has been started, but was not yet committed or rolled back.
 		 */
-		ACTIVE,
+		ACTIVE((byte) 2),
 
 		/**
 		 * A transaction is being committed.
 		 */
-		COMMITTING,
+		COMMITTING((byte) 3),
 
 		/**
 		 * A transaction is being rolled back.
 		 */
-		ROLLING_BACK,
+		ROLLING_BACK((byte) 4),
 
 		/**
 		 * The transaction status is unknown.
 		 */
-		UNKNOWN;
-	}
+		UNKNOWN((byte) 5);
 
-	private static final Charset US_ASCII = Charset.forName("US-ASCII");
+		byte[] onDisk;
+
+		TxnStatus(byte onDisk) {
+			this.onDisk = new byte[1];
+			this.onDisk[0] = onDisk;
+		}
+	}
 
 	/**
 	 * The name of the transaction status file.
@@ -58,7 +63,7 @@ class TxnStatusFile {
 
 	/**
 	 * Creates a new transaction status file. New files are initialized with {@link TxnStatus#NONE}.
-	 * 
+	 *
 	 * @param dataDir The directory for the transaction status file.
 	 * @throws IOException If the file did not yet exist and could not be written to.
 	 */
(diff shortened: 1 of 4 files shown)


core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
```diff
@@ -77,37 +82,48 @@ public void close() throws IOException {
...
+				status = TxnStatus.ROLLING_BACK;
+				break;
+			case 5:
+				status = TxnStatus.UNKNOWN;
+				break;
 			}
 		}
+
+		return status;
+
```
hmottestad: I should add a try catch or maybe a default case that uses the old strings based approach.

hmottestad:
## Before
```
Benchmark                                    (numberOfTransactions)  Mode  Cnt      Score     Error  Units
TransactionsPerSecondBenchmark.transactions                     100  avgt   10     37.630 ±   1.527  ms/op
TransactionsPerSecondBenchmark.transactions                    1000  avgt   10    303.066 ±   2.984  ms/op
TransactionsPerSecondBenchmark.transactions                   10000  avgt   10   3171.710 ±  46.521  ms/op
TransactionsPerSecondBenchmark.transactions                  100000  avgt   10  36035.554 ± 368.616  ms/op

```

hmottestad:
For transactions with 10 statements the difference is much much smaller. 1818 TPS vs 2028 TPS. 11% difference. I think that that's still significant enough to merge this.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
```diff
@@ -13,41 +13,70 @@
...
+	// check that none of the new TxnStatus conflict with the old ones, ie. if the first byte for the old values are the
+	// same as the first byte for the new values
+	{
+		boolean assertsEnabled = TxnStatus.class.desiredAssertionStatus();
+		if (assertsEnabled) {
+			for (TxnStatus value1 : TxnStatus.values()) {
+				for (TxnStatus value2 : TxnStatus.values()) {
+					byte firstByteInOldValue = value1.name().getBytes(US_ASCII)[0];
+					byte firstByteInNewValue = value2.onDisk[0];
+					assert firstByteInOldValue != firstByteInNewValue;
```
jeenbroekstra: I don't quite understand what you're checking here. It looks as if it's testing that the new byte representations have no overlaps with "legacy" txn statuses. Why do we need to run that test as part of this code? Surely you can make sure of that once (the status enums are constants, after all)?

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
```diff
@@ -13,41 +13,70 @@
...
+
 /**
  * Writes transaction statuses to a file.
  */
 class TxnStatusFile {
+	public static final byte NONE_BYTE = (byte) 0b00000001;
+	public static final byte ACTIVE_BYTE = (byte) 0b00000010;
+	public static final byte COMMITTING_BYTE = (byte) 0b00000100;
+	public static final byte ROLLING_BACK_BYTE = (byte) 0b00001000;
+	public static final byte UNKNOWN_BYTE = (byte) 0b00010000;
```
jeenbroekstra: Can these be moved into the enum? And can they be made private?

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
```diff
@@ -13,41 +13,70 @@
...
-		ROLLING_BACK,
+		ROLLING_BACK(ROLLING_BACK_BYTE),
 
 		/**
 		 * The transaction status is unknown.
 		 */
-		UNKNOWN;
+		UNKNOWN(UNKNOWN_BYTE);
+
+		byte[] onDisk;
```
jeenbroekstra: Can we make this a private (final) field with a separate accessor method?

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeStoreTxnTest.java
```diff
@@ -104,4 +109,35 @@ protected void removeStmt(Statement stmt) {
...
+		nioFile.truncate(bytes.length);
+		nioFile.writeBytes(bytes, 0);
+
+		repo.init();
+
+		try (RepositoryConnection con = repo.getConnection()) {
+
+			con.begin();
+			assertTrue(con.hasStatement(RDF.TYPE, RDF.TYPE, RDFS.RESOURCE, false));
+			con.commit();
```
jeenbroekstra: Do you want to put a further assertion here that after replaying the txn and getting everything in shape, the transaction file now uses the new format?

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
hmottestad: I moved it into a test instead. It checks that the first byte in the old statuses isn't the same as the first byte in the new ones.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
hmottestad: Done.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
hmottestad: Done.

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeStoreTxnTest.java
hmottestad: I found the TripleStoreRecoveryTest and based my test on that one, with the exception that I hardcoded the old "COMMITTING" string.

hmottestad:
Thanks for the great feedback @jeenbroekstra. Should be ready to roll now. What do you think? Should this be part of 3.1.0, or should we push it to 3.2.0?

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
```diff
@@ -77,25 +94,55 @@ public void close() throws IOException {
...
 	/**
 	 * Writes the specified transaction status to file.
-	 * 
+	 *
 	 * @param txnStatus The transaction status to write.
 	 * @throws IOException If the transaction status could not be written to file.
 	 */
 	public void setTxnStatus(TxnStatus txnStatus) throws IOException {
-		byte[] bytes = txnStatus.name().getBytes(US_ASCII);
+		byte[] bytes = txnStatus.onDisk;
```
jeenbroekstra: Does that work now that you've made it private? I would have expected you need to use `getOnDisk()` here instead.

jeenbroekstra:
Btw happy to put this in for 3.1.0.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TxnStatusFile.java
hmottestad: Was surprised too. Seems that this is legal. No compile issues.

PR #1854, opened 2020-01-20
GH-1820: fix RepositoryFederatedService to always close connection

aschwarte10:
GitHub issue resolved: #1820 

This change makes sure to always fresh connections in any kind of
RuntimeExceptions. Before we missed the case where the service
repository could have thrown a QueryEvaluationException.

Also we add some unit test in the federation repository which partially
cover this issue.


aschwarte10:
Unfortunately this issues is really hard to reproduce in an isolate unit tests. Not even the fedx tests fully cover the issue.
We have identified the issue in our application with Milestone 2, and validated this fix (by patching the binaries in our application)

Issue #1852, opened 2020-01-20
Using Repositories#tupleQuery helper function with FedXRepository yields UnsupportedOperationException


aschwarte10:
Whem using Repositories#tupleQuery with a FedX repository this results in an UnsupportedOperationException:

```
Exception in thread "main" java.lang.UnsupportedOperationException: Writing not supported to a federation: the federation is readonly.
	at org.eclipse.rdf4j.federated.write.ReadOnlyWriteStrategy.begin(ReadOnlyWriteStrategy.java:31)
	at org.eclipse.rdf4j.federated.FedXConnection.startTransactionInternal(FedXConnection.java:331)
	at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.begin(AbstractSailConnection.java:175)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.begin(SailRepositoryConnection.java:138)
	at org.eclipse.rdf4j.repository.util.Repositories.get(Repositories.java:161)
	at org.eclipse.rdf4j.repository.util.Repositories.tupleQuery(Repositories.java:304)
```

The reason is that FedX throws an UnsupportedOperationException  when invoking begin() (to start a transaction) (actually because FedX is meant to be readonly).

Proposal: just do not throw an exception when starting a transaction and basically do not offer any transactional behavior.

Note that Repositories#tupleQueryNoTransaction is working fine as expected

PR #1858, opened 2020-01-20
GH-1852: support Repositories#tupleQuery in FedX repository

aschwarte10:
GitHub issue resolved: #1852 

This change adjusts the implementations of #begin(), #commit() and
#rollback() to not throw an Exception, but rather do nothing.

Thus we now can use the Repositories#tupleQuery convenience function

### DIFF
--- a/tools/federation/src/test/java/demos/GettingStartedDemoMinimal.java
+++ b/tools/federation/src/test/java/demos/GettingStartedDemoMinimal.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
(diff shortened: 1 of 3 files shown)


tools/federation/src/test/java/demos/GettingStartedDemoMinimal.java
```diff
@@ -0,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Eclipse RDF4J contributors.
```
jeenbroekstra: No need to fix in this PR, but you may want to look at your template and update the year.

aschwarte10: Done, I adjusted my template. Thanks for paying attention

aschwarte10:
> Looks good me.
> 
> I wasn't aware FedX was currently read-only actually. The old federationsail supports distributed write I believe. Are you considering adding that functionality to FedX as well?

FedX also has a configuration option to activate a "write" mode, see org.eclipse.rdf4j.federated.write.WriteTest for demo.

However, this is a bit of an experimental feature and not yet documented. The current implementation is very basic and might desire a second look at some point in time. For this release I do not see it as a primary use-case.

PR #1859, opened 2020-01-20
GH-1555: add example on how to use FedX with RemoteRepositoryManager

aschwarte10:
GitHub issue resolved: #1555 

Just another demo snippet

### DIFF
--- a/tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
+++ b/tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
@@ -1,0 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package demos;
+
+import org.eclipse.rdf4j.federated.FedXFactory;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.vocabulary.FOAF;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.repository.Repository;
+import org.eclipse.rdf4j.repository.RepositoryConnection;
+import org.eclipse.rdf4j.repository.RepositoryResult;
+import org.eclipse.rdf4j.repository.manager.RemoteRepositoryManager;
+import org.eclipse.rdf4j.repository.manager.RepositoryManager;
+
+/**
+ * Demo which shows an example of a local federation which uses repositories from a remote RDF4J instance as federation
+ * members.
+ * 
+ * @author Andreas Schwarte
+ *
+ */
+public class FedXWithRemoteRepositoryManager {
+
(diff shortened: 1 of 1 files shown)


tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
```diff
@@ -0,0 +1,47 @@
+package demos;
```
jeenbroekstra: Missing copyright header.

tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
```diff
@@ -0,0 +1,47 @@
...
+		RepositoryManager repoManager = new RemoteRepositoryManager(serverUrl);
+
+		// assumes that remote repositories exist
+		Repository repo = FedXFactory.newFederation()
+				.withRepositoryResolver(repoManager)
+				.withResolvableEndpoint("my-repository-1")
+				.withResolvableEndpoint("my-repository-2")
+				.create();
+
+		repo.init();
```
jeenbroekstra: This reminds me: calling `init()` should not be mandatory. I'm not sure if the FedXRepository implements that (as it's a fairly new thing).

tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
```diff
@@ -0,0 +1,47 @@
...
+
+		try (RepositoryConnection conn = repo.getConnection()) {
+			try (RepositoryResult<Statement> repoResult = conn.getStatements(null, RDF.TYPE, FOAF.PERSON)) {
+				repoResult.forEach(st -> System.out.println(st));
+			}
+		}
+
+		repo.shutDown();
+		repoManager.shutDown();
```
jeenbroekstra: You can just do a `shutDown` on the `repoManager`, it will shut down the actual repository for you. And perhaps, if we're a bit paranoid, this should be in a `finally` clause.

tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
aschwarte10: Done, added in amended commit

tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
aschwarte10: Good point, just tested and it does not yet work. I'll track an issue and provide a fix for 3.1

tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
aschwarte10: The FedX repository is actually a local repository, while only the federation members are managed in the remote repository manager (i.e. in the HTTP server). Thus the shutdown is required. I tried to make this a bit more explicit in the example by naming the variable "localRepo"

The finally for the shutdown I typically omit in the very simplistic examples. I typically only add the try with resources for connections and tuple results.

aschwarte10:
@jeenbroekstra I amended the PR and addressed your comments

tools/federation/src/test/java/demos/FedXWithRemoteRepositoryManager.java
jeenbroekstra: Fair enough, makes sense.

Issue #1867, opened 2020-01-21
SPARQLRepository ignores connection parser config for graph queries


jeenbroekstra:
The `RepositoryConnection` interface has getters and setters for a `ParserConfig` object which is supposed to be used for any Rio-based activities the repository does. 

However, this config is not used when SPARQLRepository execute SPARQL graph queries. This makes it difficult to tweak the config for parsing query results (for example, to be less strict when querying a sparql endpoint that has sloppy syntax).

PR #1869, opened 2020-01-22
GH-1867 pass parserconfig to protocol session

jeenbroekstra:
GitHub issue resolved: #1867  

* SPARQLConnection passes its parser config on to client so it can be used for tweaking strictness of result processing.
* added simple unit test

### DIFF
--- a/core/repository/sparql/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnectionTest.java
+++ b/core/repository/sparql/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnectionTest.java
@@ -1,0 +1,38 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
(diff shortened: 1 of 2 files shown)


core/repository/sparql/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnectionTest.java
```diff
@@ -0,0 +1,38 @@
...
+		client = mock(SPARQLProtocolSession.class);
+		subject = new SPARQLConnection(null, client);
+	}
+
+	@Test
+	public void setParserConfigPassesToProtocolSession() throws Exception {
+		ParserConfig config = new ParserConfig();
+
+		subject.setParserConfig(config);
+		verify(client, times(1)).setParserConfig(config);
```
hmottestad: Haven’t seen this part of mockito. This tests that the setParserConfig method has been called exactly once with `config`?

jeenbroekstra: That's exactly right! The client is a mock, and we use verify to check that a particular method with a particular argument has been called on that mock.

Issue #1876, opened 2020-01-24
Refactor join and group optimizer in FedX evaluation strategy + introduce CostModel as data entity


aschwarte10:
Before we first release FedX I'd propose to use the opportunity to clean up the join and grouping optimizers in the federation evaluation strategy.

The main goal of this refactoring is to

a) have a clean definition of the cost model as its own entity
b) have the grouping and optimizing done in an optimizer and avoid static method invocations
c) provide extension points for the federation evaluation strategy to customize the cost model

PR #1877, opened 2020-01-24
GH-1876: refactor join order optimizer and federation cost model

aschwarte10:
GitHub issue resolved: #1876 

Before we first release FedX I would like to use the chance to do refactoring and cleanup of code, which will help fus for future maintenance.

**The main goal of this refactoring is to**

a) have a clean definition of the cost model
b) have the grouping and optimizing done in an optimizer and avoid
static method invocations
c) provide extension points for the federation evaluation strategy to
customize the cost model

**Performed Changes:**

* StatementGroupOptimizer has been renamed to
StatementGroupAndJoinOptimizer as it was taking care for grouping and
optimizing the join order
* The large methods in the StatementGroupAndJoinOptimizer are split up
into smaller parts
* the JoinOrderOptimizer has been removed: all cost estimation functions
have been moved unchanged to DefaultFedXCostModel, the actual join order
functions have been integrated (unchanged) into
StatementGroupAndJoinOptimizer
* NTuple has been fixed to correctly set the parent node when replacing
children

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/algebra/NJoin.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/algebra/NJoin.java
@@ -11,7 +11,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.eclipse.rdf4j.federated.optimizer.JoinOrderOptimizer;
+import org.eclipse.rdf4j.federated.optimizer.StatementGroupAndJoinOptimizer;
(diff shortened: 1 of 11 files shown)


aschwarte10:
@jeenbroekstra the change might look at bit big, but actually it is mostly moving of code and adjusting references. For the review I recommend to read the description of the PR which explains the rationals and where code is moved.

If you have questions regarding why to do this change, I can provide you more details. Especially the extension points of the cost model will help me to allow extensibility for our application.

Issue #1808, opened 2020-01-04
Add stream() method to QueryResult / RepositoryResult


jeenbroekstra:
As an RDF4J user, I want to directly stream query results, so that I can use Java stream processing directly without having to know about the appropriate utility class.

Folloup of #1747

Currently, RDF4J allows processing a query result as a [Java stream](https://www.geeksforgeeks.org/stream-in-java/) by means of a utility function: `QueryResults.stream`. This means that when you wish to use stream functionality (e.g. using it to map a stream of statements to something), you need to write code like the following:

	List<Statement> result = QueryResults.stream(conn.getStatements(null, RDF.TYPE, null))
				.map(s -> vf.createStatement(s.getSubject(), RDFS.LABEL, vf.createLiteral("bar")))
				.collect(Collectors.toList());

To increase discoverability we can add the `stream()` method directly on the QueryResult and RepositoryResult interface, so that the following is possible:

	List<Statement> result = conn.getStatements(null, RDF.TYPE, null).stream()
				.map(s -> vf.createStatement(s.getSubject(), RDFS.LABEL, vf.createLiteral("bar")))
				.collect(Collectors.toList());  

Although this is only marginally shorter code, it is far easier for an RDF4J end user to discover this method than the corresponding utility.

PR #1881, opened 2020-01-28
GH-1808 Iteration stream support

hmottestad:
GitHub issue resolved: #1808  

* created two .stream() methods on Iteration and CloseableIteration (because of different javadocs)
* changed all code in the repo to use .stream() instead of Iterations.stream(....)
*

### DIFF
--- a/core/util/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
+++ b/core/util/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
@@ -8,14 +8,16 @@
 
 package org.eclipse.rdf4j.common.iteration;
 
+import java.util.stream.Stream;
+
 /**
  * An {@link Iteration} that can be closed to free resources that it is holding. CloseableIterations automatically free
  * their resources when exhausted. If not read until exhaustion or if you want to make sure the iteration is properly
  * closed, any code using the iterator should be placed in a try-with-resources block, closing the iteration
  * automatically, e.g.:
- * 
+ *
  * <pre>
- * 
+ *
  * try (CloseableIteration&lt;Object, Exception&gt; iter = ...) {
  *    // read objects from the iterator
  * }
@@ -32,5 +34,16 @@ public interface CloseableIteration<E, X extends Exception> extends Iteration<E,
 	 */
 	@Override
 	public void close() throws X;
+
+	/**
+	 *
+	 * Convert the results to a Java 8 Stream. Stream should be closed by calling .close().
+	 *
+	 * @return stream
+	 */
+	@Override
+	default Stream<E> stream() {
+		return Iterations.stream(this);
+	}
 
 }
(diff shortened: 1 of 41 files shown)


hmottestad:
@jeenbroekstra I propose to add the .stream() methods all the way up to Iteration so that we can call .stream() on pretty much anything in RDF4J. Found this to be much more intuitive. Essentially now if it has a .next(), it'll have a .stream(). 

I think this will help to simplify a lot of the code.

core/util/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
```diff
@@ -33,4 +35,15 @@
 	@Override
 	public void close() throws X;
 
+	/**
+	 *
+	 * Convert the results to a Java 8 Stream. Stream should be closed by calling .close().
```
jeenbroekstra: Calling `close()` is only necessary if the stream is not fully consumed: the `Spliterator` implementation we use automatically closes the iteration under the hood when exhausted (or when an exception occurs). Perhaps we can reword to something like:

> The stream should be closed (by calling `Stream#close()`) if it is not fully consumed. 

Arguably, that is something that should be done with every Stream object anyway, so this could just be added to the javadoc of `Iteration.stream`, removing the need for the override here.

core/util/src/main/java/org/eclipse/rdf4j/common/iteration/Iteration.java
```diff
@@ -44,11 +45,22 @@
...
+	/**
+	 *
+	 * Convert the results to a Java 8 Stream. If this iteration implements CloseableIteration it should be closed (by
+	 * calling Stream#close() or using try-with-resource) if it is not fully consumed.
+	 *
+	 * @return stream
+	 */
+	default Stream<E> stream() {
+		return Iterations.stream(this);
+	}
```
hmottestad: @jeenbroekstra 

The javadocs say that most streams don't need to be closed:

> Streams have a BaseStream.close() method and implement AutoCloseable, but nearly all stream instances do not actually need to be closed after use. Generally, only streams whose source is an IO channel (such as those returned by Files.lines(Path, Charset)) will require closing. Most streams are backed by collections, arrays, or generating functions, which require no special resource management. (If a stream does require closing, it can be declared as a resource in a try-with-resources statement.)

I have removed the duplicate method and modified the docs (see above) to make it up to the user to decide what they want to do.

jeenbroekstra:
> > LGTM. Don't forget to squash before merging.
> 
> Thought we had stopped with squash and merge.

Yes, but we're still squashing before merging :) Sorry if this has been confusing. What we're stopping with is using Github's "squash and merge" button to merge PRs, instead just using rebase or merge commits to merge the PR - _but_ the idea is that you still squash your commits (yourself, manually) on your feature branch _before_ you merge - so that the end result is a branch that has only one or two meaningful commits. 

See https://rdf4j.org/documentation/developer/squashing/ for details. 

The reason we do all this, by the way, is that a slightly cleaner history makes doing bisecting or trawling through the history to figure out why something was changed a lot easier - after a feature is finished there is no need to have separate commits for _all_ the intermediate steps. And one reason we don't use Github's squash and merge is that it puts the burden of writing a meaningful commit message on whoever does the merge, rather than the author. 

Fwiw if, for PRs you authored, you still prefer using "squash and merge" over manual squashing for this purpose, that's fine with me - let me know and I'll hold off on disabling the option. REgardless of how you squash: just make sure that the commit message is descriptive of the overall change and contains the issue number (and a keyword like "resolved" or "fixed" if you want to automatically close the issue).

Issue #1882, opened 2020-01-28
Support basic property path expressions in FedX engine


aschwarte10:
In the current implementation the FedX federation engine throws different kinds of errors when trying to use property paths.

This task is about adding basic support for property paths:

* `?subClass rdfs:subClassOf/rdfs:label ?label`
* `?class rdfs:subClassOf+ ?class`
* `?concept rdfs:label|skos:altLabel ?label`

If time permits we can provide optimizations

PR #1884, opened 2020-01-28
GH-1882: basic support for property path in FedX engine

aschwarte10:
GitHub issue resolved: #1882

This change adds basic support for property paths in the federation
engine.

Example patterns:

* `?subClass rdfs:subClassOf/rdfs:label ?label`
* `?class rdfs:subClassOf+ ?class`
* `?concept rdfs:label|skos:altLabel ?label`

Note that some adjustments to queryalgebra PathIteration were required:

* as in FedX we generate SPARQL query strings, variables must not
contain the "-" character (which is invalid for SPARQL queries). This
character is replaced with _
* in the FedX engine we are not seeing "QueryBindingSet" instances, but
"MapBindingSet" instances. The addBinding method is made robust to
support both instance types
* in remote settings we even see ListBindingSet. In the check we make
sure to convert the BindingSet instance if required

Note also that the cost model estimation now only shows a warning if the
type of the node is unknown.

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/PathIteration.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/PathIteration.java
@@ -27,6 +27,7 @@ import org.eclipse.rdf4j.query.algebra.ZeroLengthPath;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryBindingSet;
 import org.eclipse.rdf4j.query.algebra.evaluation.impl.StrictEvaluationStrategy;
 import org.eclipse.rdf4j.query.algebra.helpers.AbstractQueryModelVisitor;
+import org.eclipse.rdf4j.query.impl.MapBindingSet;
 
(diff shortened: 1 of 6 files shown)


aschwarte10:
This change provides basic evaluation capabilities for property paths. If time permits I'll add some further optimizations

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/PathIteration.java
```diff
@@ -190,6 +195,16 @@ protected BindingSet getNextElement() throws QueryEvaluationException {
 		}
 	}
 
+	private void addBinding(BindingSet bs, String name, Value value) {
+		if (bs instanceof QueryBindingSet) {
+			((QueryBindingSet) bs).addBinding(name, value);
+		} else if (bs instanceof MapBindingSet) {
+			((MapBindingSet) bs).addBinding(name, value);
+		} else {
+			throw new IllegalStateException("Unexpected BindingSet implementation: " + bs.getClass());
```
jeenbroekstra: FWIW I have long thought that we are using the various bindingset implementations inconsistently throughout the framework. It could do with straightening out.

PR #1888, opened 2020-01-30
GH-1882: optimization for property paths with an exclusive endpoint

aschwarte10:
This change provides an optimization for property path expressions that
yield results at a single endpoint. Similarly to the concept of
ExclusiveStatement we now use a marker for them.

We introduce the interface ExclusiveTuplExpr which can now be used to
declare any exclusive TupleExpr. Note that the existing
ExclusiveStatement and ExclusiveGroup nodes have also become instances
of ExclusiveTupleExpr.

For rendering such nodes in sub-queries the ExclusiveTupleExprRenderer
interface is defined. Additionally, some rendering logic is introduced
in the Query utility classes.

Note that ExclusiveGroups are now generalized such that they can contain
any ExclusiveTupleExpr (and not only ExclusiveStatements).

Notes:

* the if/else branches in StatementGroupAndJoinOptimizer were shuffled
around such that the instanceOf check is working correctly
(ExclusiveGroup is now an instanceof ExclusiveTuplExpr)

GitHub issue resolved: #1882

### DIFF
--- a/tools/federation/src/test/java/org/eclipse/rdf4j/federated/PropertyPathTests.java
+++ b/tools/federation/src/test/java/org/eclipse/rdf4j/federated/PropertyPathTests.java
@@ -17,6 +17,7 @@ import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Literal;
 import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.FOAF;
 import org.eclipse.rdf4j.model.vocabulary.OWL;
 import org.eclipse.rdf4j.model.vocabulary.RDFS;
 import org.eclipse.rdf4j.model.vocabulary.SKOS;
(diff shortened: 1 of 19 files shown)


tools/federation/src/test/java/org/eclipse/rdf4j/federated/PropertyPathTests.java
```diff
@@ -85,6 +87,67 @@ public void testPropertyPath_Alternatives() throws Exception {
...
 
+	@Test
+	public void testPropertyPath_ExclusiveGroup() throws Exception {
+
+		prepareTest(Arrays.asList("/tests/propertypath/data1.ttl", "/tests/propertypath/data2.ttl"));
+
+		String query = "SELECT * WHERE { ?concept a skos:Concept . ?concept skos:broader+ :mammals . ?concept rdfs:label ?label}";
+
+		String actualQueryPlan = federationContext().getQueryManager().getQueryPlan(query);
+		assertQueryPlanEquals(readResourceAsString("/tests/propertypath/query_path_exclusiveGroup.qp"),
```
jeenbroekstra: I like this mechanism - simple but effective, although potentially brittle against future changes in the string representation of the query plan (it was never really designed to be a stable format, more for informing hoomans).

Issue #1865, opened 2020-01-21
Support lazy repository initialization without creating Repository#init in FedXRepository


aschwarte10:
Recently in RDF4J it was adjusted that users do not explicitly have to call Repository#init. This should implicitly happen when retrieving a connection

We should add support for such implicit initializatiion in the FedXRepository to be consistent with other RDF4J repositories.

PR #1894, opened 2020-02-01
Finalization of fedx for 3.1

aschwarte10:
GitHub issue resolved: #1538 and #1865 

This PR is organized in a set of isolated commits. Each commit by itself is consistent and consititutes to a final cleanup of the FedX code for the 3.1 releae. The main rational is to get those things clean before the release, which we are not able to change afterwards.

For the review I recommend to do this commit by commit. When following this commit by commit the review should be straight forward as the individual changes are small or simple refactoring. There is no logic change implied in this PR.

**GH-1538: minor fix to cost model, add additional OPTIONAL test**

* minor improvement to cost model for OPTIONAL
* add an additional test for the OPTIONAL clause
* refactor test utility methods to base class to make them re-usable
* activate one test which is working in the mean time
* small refactoring to move optimization to ExclusiveTupleExpr to
protected method as extension point

**GH-1865: support lazy initialization of FedXRepository**

**GH-1538: mark WriteStrategy as an experimental feature**

**GH-1538: remove experimental JMX exported MBeans**

JMX support was an experimental feature and was never really thoroughly
implemented nor documented. In the current form it does not make any
sense, thus we remove it prior to the first release.

**GH-1538: clean SparqlFederationEvalStrategy**

Deprecate the UNION based evaluation of bound joins, which was not
active for a long time, and instead integrate the VALUES based bound
join implementation directly into SparqlFederationEvalStrategy.

This change integrates the implementation of
SparqlFederationEvalStrategyWithValues directly into its super class and
thus we can remove this class.

The configuration is adjusted accordingly to use the
SparqlFederationEvalStrategy.


aschwarte10:
@jeenbroekstra I targeted this to the master branch (and based my commits on latest master), hope this is correct. If this is not the right base to start from let me know and I can adjust

jeenbroekstra:
All looks very good to me, great work @aschwarte10 , thanks!

PR #1903, opened 2020-02-04
ID-1538: restructure the FedX algebra interface hierarchy

aschwarte10:
This change restructures the FedX algebra interface hierarchy to share
common methods of all nodes.

The most top-level interface for any FedX specific algebra node is now a
FedXTupleExpr.

In addition, we have the new interface VariableExpr which can be used to
determine the free variables (e.g. during cost estimation).

Note that the logging severity for unknown nodes is reduced to DEBUG as
the fallback values are sufficient for most cases.


aschwarte10:
@jeenbroekstra one of the use-cases of this restructuring is to cleanly define extension points for the cost model computation and avoid warnings for unknown node types.
In addition the interface methods are not specified multiple times

Issue #1904, opened 2020-02-04
FedX ThreadPool size is initialized wrongly: always has a maximum size of 10


aschwarte10:
In FedX we use configurable threadpools for executing join and union sub queries, as well as other remote requests.

Also the pool size is meant to be configurable via FedXConfig setters.

However, due to a bug in the construction of the ThreadPoolExecutor the size is always set to a most 10 threads. This is because we have an unbound queue and set the minimumCoreSize to 10. In such setting new threads are only created if the core size of the pool is not reached.

See `org.eclipse.rdf4j.federated.evaluation.concurrent.ControlledWorkerScheduler.initWorkerThreads()` for details

PR #1905, opened 2020-02-04
GH-1904: fix ThredPool initialization in FedX

aschwarte10:
GitHub issue resolved: #1904 

In FedX we use configurable threadpools for executing join and union sub
queries, as well as other remote requests. Also the pool size is meant
to be configurable via FedXConfig setters.

However, due to a bug in the construction of the ThreadPoolExecutor the
size is always set to a most 10 threads. This is because we have an
unbound queue and set the minimumCoreSize to 10. In such setting new
threads are only created if the core size of the pool is not reached.

We now explicitly set the core pool size to the configured number of
workers, and in addition configure the pool in such a way that idle
threads can get destroyed. The idle timeout is set to 60s.

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
@@ -112,8 +112,10 @@ public class ControlledWorkerScheduler<T> implements Scheduler<T> {
 
 	protected void initWorkerThreads() {
 
-		executor = new ThreadPoolExecutor(Math.min(10, nWorkers / 2), nWorkers, 30L, TimeUnit.SECONDS, _taskQueue,
+		ThreadPoolExecutor _executor = new ThreadPoolExecutor(nWorkers, nWorkers, 60L, TimeUnit.SECONDS, _taskQueue,
 				new NamingThreadFactory(name));
+		_executor.allowCoreThreadTimeOut(true);
+		this.executor = _executor;
 	}
 
 	@Override
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
@@ -38,14 +38,14 @@ import org.slf4j.LoggerFactory;
  */
 public class ControlledWorkerScheduler<T> implements Scheduler<T> {
 
-	protected static final Logger log = LoggerFactory.getLogger(ControlledWorkerScheduler.class);
+	private static final Logger log = LoggerFactory.getLogger(ControlledWorkerScheduler.class);
 
-	protected ExecutorService executor;
+	private final ExecutorService executor;
 
-	protected LinkedBlockingQueue<Runnable> _taskQueue = new LinkedBlockingQueue<>();
+	private final LinkedBlockingQueue<Runnable> _taskQueue = new LinkedBlockingQueue<>();
 
-	protected int nWorkers;
-	protected String name;
+	private final int nWorkers;
+	private final String name;
 
 	/**
 	 * Construct a new instance with 20 workers.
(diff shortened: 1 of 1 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
```diff
@@ -112,8 +112,10 @@ public int getNumberOfTasks() {
 
 	protected void initWorkerThreads() {
 
-		executor = new ThreadPoolExecutor(Math.min(10, nWorkers / 2), nWorkers, 30L, TimeUnit.SECONDS, _taskQueue,
+		ThreadPoolExecutor _executor = new ThreadPoolExecutor(nWorkers, nWorkers, 60L, TimeUnit.SECONDS, _taskQueue,
 				new NamingThreadFactory(name));
+		_executor.allowCoreThreadTimeOut(true);
+		this.executor = _executor;
```
hmottestad: Btw. Unrelated to this particular PR. I noticed a lot of Protected variables in this class, do you need them to be available so broadly.

Also, I was wondering what happens if you call initWorkerThreads multiple times? Does it create a bunch of thread pools?

jeenbroekstra: @hmottestad I agree that some of these variables should be made private, but this close to release I'd rather have confirmation from @aschwarte10 that that won't break dependent code. 

As for your second remark: it looks as if the intent is that `initWorkerThreads` is only invoked from the constructor. So that would happen only once.

jeenbroekstra:
@aschwarte10 fwiw I did a quick change and made every protected field and method in this class private, which compiles without problems. I won't push this change though unless you can confirm that that won't give you headaches.

aschwarte10:
> @aschwarte10 fwiw I did a quick change and made every protected field and method in this class private, which compiles without problems. I won't push this change though unless you can confirm that that won't give you headaches.

Perfectly fine to make things private in this class, and I fully agree. The scheduler is not meant for extension. The current structure is merely for historical reasons and has grown over time - in quite some versions ago the extension hierarchy for the schedulers looked quite different which made it required to have certain methods protected.
Please push your change

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
aschwarte10: I agree that methods should be private for this scheduler (and also the initialization could be done in a better way). The current implementation structure is mostly for historical reasons (the initial code is almost 8 years old by now) and quite some versions ago the extension hierarchy for the schedulers looked different (e.g. for union and join) making it required to have certain things protected.
As suggested by Jeen we can have another push on this PR to make the methods private

jeenbroekstra:
Pushed. Have a final look please - if you're ok with this feel free to merge (or I'll do it tomorrow first thing).

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
```diff
@@ -38,14 +38,14 @@
  */
 public class ControlledWorkerScheduler<T> implements Scheduler<T> {
 
-	protected static final Logger log = LoggerFactory.getLogger(ControlledWorkerScheduler.class);
+	private static final Logger log = LoggerFactory.getLogger(ControlledWorkerScheduler.class);
 
-	protected ExecutorService executor;
+	private ExecutorService executor;
```
hmottestad: could executor be final?

aschwarte10: Sure, I'll add this to @jeenbroekstra's second commit. @hmottestad  Can you please approve once the PR is updated?

aschwarte10: @hmottestad PR is updated, please have a second look

PR #1919, opened 2020-02-14
GH-1795 MemoryStore computes next snapshot number on commit

jeenbroekstra:
This avoids transaction leak issues when two sinks flush to the same
store (for example as part of a UnionSailSource)

GitHub issue resolved: #1795   

* MemoryStoreConnection computes the next snapshot number on commit, instead of it happening as part of the sink flushing - this avoids it being incremented twice in the same txn.
* Tweaked SailSourceConnection with a couple of protected methods to allow customization/overriding of behavior of `commitInternal`. 
* Minor tweak in handling of restore of persisted data on Sail init to make sure the snapshot number is set correctly after restore.
* Re-enabled relevant integration tests to check this fixes the problem.

### DIFF
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
@@ -750,14 +750,9 @@ class MemorySailStore implements SailStore {
 	 * increments the current snapshot by 1 and schedules a snapshot cleanup if necessary.
 	 */
 	protected void incrementSnapshot() {
-		txnLockManager.lock();
-		try {
-			currentSnapshot++;
-			if (requireCleanup) {
-				scheduleSnapshotCleanup();
-			}
-		} finally {
-			txnLockManager.unlock();
+		currentSnapshot++;
+		if (requireCleanup) {
+			scheduleSnapshotCleanup();
 		}
 	}
 }
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
@@ -7,6 +7,12 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.memory;
 
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.concurrent.locks.ReentrantLock;
+
 import org.eclipse.rdf4j.IsolationLevel;
 import org.eclipse.rdf4j.IsolationLevels;
 import org.eclipse.rdf4j.common.concurrent.locks.Lock;
@@ -42,12 +48,6 @@ import org.eclipse.rdf4j.sail.memory.model.MemValue;
 import org.eclipse.rdf4j.sail.memory.model.MemValueFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.Set;
-import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * An implementation of {@link SailStore} that keeps committed statements in a {@link MemStatementList}.
@@ -360,8 +360,6 @@ class MemorySailStore implements SailStore {
 
 		private final Lock txnStLock;
 
-		private volatile int nextSnapshot;
-
 		private volatile Set<StatementPattern> observations;
 
 		private volatile boolean txnLock;
@@ -387,11 +385,15 @@ class MemorySailStore implements SailStore {
 				sb.append("inferred ");
 			}
 			if (txnLock) {
-				sb.append("snapshot ").append(nextSnapshot);
+				sb.append("snapshot ").append(getNextSnapshot());
 			} else {
 				sb.append(super.toString());
 			}
 			return sb.toString();
+		}
+
+		private int getNextSnapshot() {
+			return currentSnapshot + 1;
 		}
 
 		@Override
@@ -415,8 +417,9 @@ class MemorySailStore implements SailStore {
 							MemStatement st = iter.next();
 							int since = st.getSinceSnapshot();
 							int till = st.getTillSnapshot();
-							if (serializable < since && since < nextSnapshot
-									|| serializable < till && till < nextSnapshot) {
+							int next = getNextSnapshot();
+							if (serializable < since && since < next
+									|| serializable < till && till < next) {
 								throw new SailConflictException("Observed State has Changed");
 							}
 						}
@@ -428,7 +431,7 @@ class MemorySailStore implements SailStore {
 		@Override
 		public synchronized void flush() throws SailException {
 			if (txnLock) {
-				currentSnapshot = Math.max(currentSnapshot, nextSnapshot);
+				currentSnapshot = getNextSnapshot();
 				if (requireCleanup) {
 					scheduleSnapshotCleanup();
 				}
@@ -492,10 +495,10 @@ class MemorySailStore implements SailStore {
 			acquireExclusiveTransactionLock();
 			requireCleanup = true;
 			try (CloseableIteration<MemStatement, SailException> iter = createStatementIterator(null, null, null,
-					explicit, nextSnapshot, contexts);) {
+					explicit, getNextSnapshot(), contexts);) {
 				while (iter.hasNext()) {
 					MemStatement st = iter.next();
-					st.setTillSnapshot(nextSnapshot);
+					st.setTillSnapshot(getNextSnapshot());
 				}
 			}
 		}
@@ -509,25 +512,26 @@ class MemorySailStore implements SailStore {
 		@Override
 		public synchronized void deprecate(Statement statement) throws SailException {
 			acquireExclusiveTransactionLock();
+			int next = getNextSnapshot();
 			requireCleanup = true;
 			if (statement instanceof MemStatement) {
-				((MemStatement) statement).setTillSnapshot(nextSnapshot);
+				((MemStatement) statement).setTillSnapshot(next);
 			} else if (statement instanceof LinkedHashModel.ModelStatement
 					&& ((LinkedHashModel.ModelStatement) statement).getStatement() instanceof MemStatement) {
 				// The Changeset uses a LinkedHashModel to store it's changes. It still keeps a reference to the
 				// original statement that can be retrieved here.
 				MemStatement toDeprecate = (MemStatement) ((LinkedHashModel.ModelStatement) statement).getStatement();
-				if (toDeprecate.getTillSnapshot() > nextSnapshot && toDeprecate.isExplicit() == explicit) {
-					toDeprecate.setTillSnapshot(nextSnapshot);
+				if (toDeprecate.getTillSnapshot() > next && toDeprecate.isExplicit() == explicit) {
+					toDeprecate.setTillSnapshot(next);
 				}
 			} else {
 				try (CloseableIteration<MemStatement, SailException> iter = createStatementIterator(
 						statement.getSubject(),
 						statement.getPredicate(), statement.getObject(),
-						explicit, nextSnapshot, statement.getContext())) {
+						explicit, next, statement.getContext())) {
 					while (iter.hasNext()) {
 						MemStatement st = iter.next();
-						st.setTillSnapshot(nextSnapshot);
+						st.setTillSnapshot(next);
 					}
 				}
 			}
@@ -536,7 +540,6 @@ class MemorySailStore implements SailStore {
 		private void acquireExclusiveTransactionLock() throws SailException {
 			if (!txnLock) {
 				txnLockManager.lock();
-				nextSnapshot = currentSnapshot + 1;
 				txnLock = true;
 			}
 		}
@@ -548,6 +551,8 @@ class MemorySailStore implements SailStore {
 			MemIRI memPred = valueFactory.getOrCreateMemURI(pred);
 			MemValue memObj = valueFactory.getOrCreateMemValue(obj);
 			MemResource memContext = (context == null) ? null : valueFactory.getOrCreateMemResource(context);
+
+			int next = getNextSnapshot();
 
 			if (memSubj.hasStatements() && memPred.hasStatements() && memObj.hasStatements()
 					&& (memContext == null || memContext.hasStatements())) {
@@ -563,9 +568,9 @@ class MemorySailStore implements SailStore {
 
 						if (!st.isExplicit() && explicit) {
 							// Implicit statement is now added explicitly
-							st.setTillSnapshot(nextSnapshot);
-						} else if (!st.isInSnapshot(nextSnapshot)) {
-							st.setSinceSnapshot(nextSnapshot);
+							st.setTillSnapshot(next);
+						} else if (!st.isInSnapshot(next)) {
+							st.setSinceSnapshot(next);
 						} else {
 							// statement already exists
 							return null;
@@ -575,7 +580,7 @@ class MemorySailStore implements SailStore {
 			}
 
 			// completely new statement
-			MemStatement st = new MemStatement(memSubj, memPred, memObj, memContext, explicit, nextSnapshot);
+			MemStatement st = new MemStatement(memSubj, memPred, memObj, memContext, explicit, next);
 			statements.add(st);
 			st.addToComponentLists();
 			return st;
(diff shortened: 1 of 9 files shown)


hmottestad:
Hi Jeen,

I made a video to show an issue while debugging: https://youtu.be/Y6LMSWjEYv0

Your changes fixes the test, but you've introduced a new isolation issue.

Essentially your fix cause a single transaction to commit as two separate transactions if there are both explicit changes and inferred changes. The video shows how this happens when I debug on line 67 in UnionSailSource with the following criteria `!((SailSourceBranch) primary).changes.isEmpty() && !((SailSourceBranch) additional).changes.isEmpty()`.

Essentially the flushing the primary changes cause the snapshot to increase, so any transaction that starts at this point will forever see the primary changes but not the additional changes since those are flushed to the next snapshot.

hmottestad:
I create a test for this new isolation leak.

jeenbroekstra:
I'll have to look in more detail after the weekend, but kinda makes sense when I think about it. We've got to take a closer look at how snapshots are handled in the memorystore, because it's actually a bit weird that a flush would move to the next snapshot it in the first place.

hmottestad:
Seems like the native store uses autoflush to handle the equivalent of snapshot incrementing.

jeenbroekstra:
I've tried several tweaks to the snapshot update mechanism, but every time I fix one thing, something else breaks. From what I can tell this notion of using a snapshot counter in the memorystore is just fundamentally flawed: we can't reliably use a single number to check if a statement is "visible" in a particular transaction. The only robust fix I can think of is that we get of that mechanism altogether and instead introduce a set of flags to indicate status of each status (added, removed, committed), similar to how the native store uses persisted bit flags on each triple. We'd essentially copy over the nativestore mechanism to an in-memory context.

hmottestad:
I don't think the snapshot implementation is flawed. I just think that we are lacking a mechanism that allows us to say that the first flush should not increment the snapshot yet, and the second flush should. Maybe we should move the "increment snapshot" operation into the close() method in the UnionSailSource - that is essentially what the NativeStore achieves with its auto close functionality. 

Here is a diagram of the current implementation with two sail sources, one for inferred and one for explicit.

Screenshot 2020-02-15 at 10 01 45

Removing this abstraction would also solve the problem. Eg. supporting that statements carry their own "inferred" status and extending any get/add/remove method that take decompounded statementes (subject, predicate, object, context) to now also take an inferred flag.

jeenbroekstra:
> I don't think the snapshot implementation is flawed. I just think that we are lacking a mechanism that allows us to say that the first flush should not increment the snapshot yet, and the second flush should. Maybe we should move the "increment snapshot" operation into the close() method in the UnionSailSource - that is essentially what the NativeStore achieves with its auto close functionality.

The problem with that is that the notion of a snapshot increment is specific to the memorystore implementation, and has no place in the UnionSailSource (which is designed to be agnostic to the backing persistence implementation). 

> 
> Here is a diagram of the current implementation with two sail sources, one for inferred and one for explicit.
> 
> Screenshot 2020-02-15 at 10 01 45

I understand that, but I've tried a few approaches to fixing that without breaking the overall design of the SailSourceConnection and related base classes (in other words, trying to fix it by tweaking only memorystore-specific code) and I couldn't find a simple solution that didn't break the same or other tests. One problem is that the MemorySailSinks have no way to figure out if there are other sinks working on the same connection, and even if they did, there is no prescribed order they can depend on being executed (unless we specifically rely on the current implementation). 

> Removing this abstraction would also solve the problem. Eg. supporting that statements carry their own "inferred" status and extending any get/add/remove method that take decompounded statementes (subject, predicate, object, context) to now also take an inferred flag.

I'm actually starting to think this might be a more elegant approach - especially, since, as you mentioned before, it would also speed up querying. The only reason I'm hesitant is that I don't quite know why the distinction was introduced in the first place. It is not documented anywhere as far as I'm aware. Looking at the code I see no obvious reason for it either.

jeenbroekstra:
I'm doing a quick spike on removing the need for separate sources/sinks for explicit and inferrred triples, just to get a clearer understanding of what would be involved.

jeenbroekstra:
Refactoring is a _major_ pain.

However, I have had a bit of a think and I reckon I've got a possible fix in the MemoryStore that addresses the problem. Expect a push and a squash.

jeenbroekstra:
Added a fix. Notice that this now makes one test fail: `readPending` does not work in transaction isolation level NONE. This has to do with the fact that approved but not-yet-flushed changes are normally kept in a separate changeset object, to be included in any queries on the same transaction. However for this isolation level no such changeset is created.

I'm very tempted to just remove NONE as a supported level for the memory store. If we do that it will automatically upgrade to READ_COMMITTED when starting a transaction, and all will work as expected. The benefit of having a proper implementation of NONE would be that less locking and branching is necessary and therefore it would be more efficient for bulk uploads, but I'm not immediately clear on how I could support this without keeping a separate changeset - so we might as well just use READ_COMMITTED immediately.

jeenbroekstra:
Another alternative is of course just adding 1 to the snapshot number when reading on NONE isolation level. It might include changes from other transactions, but that's kinda the point of the level.

hmottestad:
I usually use the memory store with isolation level NONE when i use it in my code. One example is a micro service i built at my previous job where we passed around JSON-LD. We used the memory store wrapped by the RDFS reasoner to process our data. Switching to NONE for the isolation level gave us 2x throughput.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStoreConnection.java
```diff
@@ -57,7 +58,20 @@ protected void startTransactionInternal() throws SailException {
...
+			if (toCloseInferredBranch != null) {
+				toCloseInferredBranch.flush();
+			}
+			// after flushing, the MemorySailStore needs its internal snapshot counter updated
+			((MemorySailStore) sail.getSailStore()).incrementSnapshot();
+		} finally {
+			if (toCloseInferredBranch != null) {
+				toCloseInferredBranch.close();
+			}
+		}
```
hmottestad: That's an interesting solution. I don't see why it shouldn't work.

hmottestad:
Btw. The ShaclSail uses isolation level NONE to optimize one validation scenario. It relies on the side effect of reading uncommitted data.

jeenbroekstra:
> Btw. The ShaclSail uses isolation level NONE to optimize one validation scenario. It relies on the side effect of reading uncommitted data.

Works with the current implementation. However: you do realize that isolation levels are _minimal_ guarantees right? There is no absolute guarantee that at isolation level NONE (or READ_UNCOMMITTED, for that matter), you will _definitely_ get dirty reads. Just that you _may_ get them.

hmottestad:
> Works with the current implementation. However: you do realize that isolation levels are _minimal_ guarantees right? There is no absolute guarantee that at isolation level NONE (or READ_UNCOMMITTED, for that matter), you will _definitely_ get dirty reads. Just that you _may_ get them.

Yes,  I know. I might want to revisit how that's used in the ShaclSail.

hmottestad:
I've run the MemoryStore insert benchmarks and I don't see any difference between your branch and master, which is good! NONE is still typically 2x faster than any other isolation level, and there is little difference between the remaining levels.

jeenbroekstra:
The tests keep timing out. I think I may have introduced a deadlock somewhere. Probably in the incrementSnapshot method. I'll take a closer look.

jeenbroekstra:
For some reason the SHACL tests now take close to an hour by themselves, and several of them (over 100) fail. @hmottestad can you take a closer look please? Given that one of the failing tests is in the SailIsolationTest suite, and all other sails succeed on that suite, I'm suspecting there is a strange corner case in how the ShaclSail handles isolation.

hmottestad:
A lot of the failures are because of the assumption that a transaction with isolation level NONE will be able to read other open transactions. That's not the case anymore, even with your fix. This is though as you said not a guarantee, but at the moment it gives pretty decent performance benefits, so it would be a shame to break it.

As for the ShaclIsolationLevelTests, these do show that your fix leaks statements. I believe this could be due to you incrementing the snapshot number on ever transaction and not just on transactions that modify the memory store. I'm really struggling with debugging it. I've spent an hour now trying to figure out what's going on to no avail. Need to get back to work now.

I really don't believe that the ShaclSail is breaking anything wrt. the transaction isolation. It just uses the transaction isolation very actively, but doesn't modify or circumvent anything. I think it's just exacerbating the issue.

jeenbroekstra:
Thanks @hmottestad, that's helpful. I'll have another look at the shaclisolationleveltest myself as well.

jeenbroekstra:
>As for the ShaclIsolationLevelTests, these do show that your fix leaks statements. I believe this could be due to you incrementing the snapshot number on ever transaction and not just on transactions that modify the memory store. 

I've pushed a fix for that. It now only updates the snapshot number if there were actual updates in the transaction.

> A lot of the failures are because of the assumption that a transaction with isolation level NONE will be able to read other open transactions. That's not the case anymore, even with your fix.  This is though as you said not a guarantee, but at the moment it gives pretty decent performance benefits, so it would be a shame to break it.

How do you propose we deal with this? Clearly, GH-1795 is a fairly serious bug, and the changes we've made in this PR fix those bugs, and are conform specs. Performance is important of course, but it doesn't trump correctness. If these Shacl tests fail because internally it relies on being able to see what other parallel transactions do before it's been committed, then in my view we'll need to change that assumption - even if it does mean a drop in performance. 

Happy to hear alternatives, but tbh I can't think of anything else right now.

hmottestad:
We can just disable that part of the ShaclSail. It’s just a couple of lines that check if the transaction level is NONE and then directly uses the underlying sail instead of copying over all the statements into a new memory store. But it’s a bit early now to decide to do this. 

Btw. Another issue with your solution could be that the code uses three locked sections, one per branch and then once more for incrementing the snapshot number. Ideally all those three operations should happen within a single locked section.

jeenbroekstra:
Not sure I follow, can you clarify what you mean with a locked section?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
```diff
@@ -746,4 +745,19 @@ private boolean isContextResource(MemResource memResource, int snapshot) throws
 			}
 		}
 	}
+
+	/**
+	 * increments the current snapshot by 1 and schedules a snapshot cleanup if necessary.
+	 */
+	protected void incrementSnapshot() {
+		txnLockManager.lock();
```
hmottestad: A locked section like this.

jeenbroekstra: Ah right! Yeah, I may have been a bit overzealous there. I'll try and remove some redundant locks.

jeenbroekstra: As far as I'm aware this is the only new lock I introduced. I removed it because you're right it isn't necessary. But I'm not sure what you meant with your "one per branch" remark earlier.

jeenbroekstra:
> We can just disable that part of the ShaclSail. It’s just a couple of lines that check if the transaction level is NONE and then directly uses the underlying sail instead of copying over all the statements into a new memory store.  But it’s a bit early now to decide to do this.

I tried this locally, and while it fixes most failures, I am getting 8 tests that never terminate, presumably due to some kind of deadlock. All tests on NONE isolation level. This is why I introduced a timeout rule on the shacl tests btw, to make it a bit easier to deal with that.

jeenbroekstra:
I'm still open to an alternative approach for this, but the problem with refactoring the UnionSailSource and the handling of inferred statements is that it so deeply wired in, changing it is going to be a massive amount of work (and I can't really commit the time to it, right now).

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
hmottestad: No. That’s not what I meant. You definitely need a lock here. What I meant is that each branch also locks while flushing, and the incrementing of the snapshot used to be done within that locked section. Now we have three such sections, one for the explicit branch flush, one for the inferred and one for incrementing the snapshot.

hmottestad:
If it were just failing SHACL tests I think it would be fine. But since it’s hanging tests I think you’ve introduced either a deadlock or something similar. I also don’t think the tradeoffs are worth it with this fix. We don’t really have that good test coverage of the isolation levels and this diverges too much from the original design.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
jeenbroekstra: > You definitely need a lock here.

I should have looked better - I thought it could be removed because commitInternal already holds a lock. But that's not the case. I've reverted this change. 

> What I meant is that each branch also locks while flushing, and the incrementing of the snapshot used to be done within that locked section. Now we have three such sections, one for the explicit branch flush, one for the inferred and one for incrementing the snapshot.

I'm still not quite sure what problem you're seeing. There's a lock obtained in several places, yes. Why would that be a problem?

hmottestad: I’m worried that it would potentially allow two transactions to flush at the same time to the same snapshot, then one gets incremented (making the changes visible) and if the other gets preempted between flushing the inferred and explicit, then half that transaction will be visible.

jeenbroekstra:
> If it were just failing SHACL tests I think it would be fine. But since it’s hanging tests I think you’ve introduced either a deadlock or something similar.

It _is_ just failing SHACL tests. The 8 hanging tests are all SHACL tests. It's pretty clear that there's a deadlock of some sort going on but at this point in time I think that is specific to how the SHACLSail is implemented, rather than an internal bug in the MemoryStore. It might just be due to how I disabled the bit that was specific to the NONE isolation level. 

> I also don’t think the tradeoffs are worth it with this fix. We don’t really have that good test coverage of the isolation levels and this diverges too much from the original design.

I disagree. For one thing, although there's always room for improvement, we have quite extensive coverage of isolation levels with various test suites - and these tests show that almost all possible combinations of stores and inferencers behave as expected under concurrent transactions. All except the SHACL Sail. I also think that if the tradeoff is that we have to sacrifice some performance to have the ShaclSail interact with the base store in a manner that is conform the specs, and fixes these errors, then that's a tradeoff we have to make. 

The only alternative solution I've seen so far is a complete rewrite of the entire SailSourceConnection framework to get rid of UnionSailSource, and right now I think it's just not feasible with the resources we have. I'm happy to be proven wrong on that score though. 

As for how this specific change (I assume you're talking about the changes I made to the MemoryStore here) diverges from the original design - the actual change in the MemoryStore is fairly small and is an implementation detail that just slightly changes its internal behavior, without breaking any of its external contracts. 

Of course it's still possible that in doing so I overlooked something, which is why we need to investigate what is making these SHACL tests deadlock, but that doesn't invalidate the basic idea behind this change. Especially in the light of the fact that every non-SHACL isolation test runs as expected. 

There's also still the option of reassessing how the MemoryStore keeps track of what's in which snapshot. The snapshot counter thing it's using is very different from what the NativeStore uses. That might still be an alternative route to explore as well: an internal fix in the MemoryStore that "copies" the approach of the NativeStore to snapshot handling.

hmottestad:
The reason that I believe that this is a big change is that the initial design on the memory store is:
1. Take a transaction lock
1. Flush changes
1. Increment snapshot counter
1. Release lock

The issue with this approach is that we run this twice, once for inferred and once for explicit statements. 

Your fix changes this up quite a bit:

1. Take a transaction lock
1. Flush changes
1. Release lock
1. Take a transaction lock
1. Increment snapshot counter
1. Release lock

I believe that the reason that the snapshot counter works at all is because of that initial locking. It guarantees one transaction per snapshot ID.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
jeenbroekstra: Ah, I see what you mean now. Hm, yes. I need to think about this. It's possible we can do some magic with obtaining extra locks in the flush that we don't abandon until when increment is called. But it's getting a bit hairy, logic-wise, and moreover if we do that we effectively lose concurrent write access - _every_ transaction would block until the one that has the lock is committed or abandoned (which defeats the purpose of the whole SailSourceBranch model). 

_sigh_ I'll take a closer look, but we may need to abandon this approach after all. It's not been waste of time though, it gave me an opportunity to reacquaint myself with this part of the code.

jeenbroekstra:
Note to self (after reading back the history on this PR and our discussion): investigate the possibility to make a MemoryStore-specific subclass of UnionSailSource, where we can do nasty things to the snapshot counter on close. It will require that we make the branch method overrideable (that or introduce a dependency injection pattern for SailSource implementations).

jeenbroekstra:
I will have to park this for a bit - it's eating up too much time and I'm not making any headway with either approach.

jeenbroekstra:
Havard actually put up an alternative PR (#1937 ) that essentially implements the alternative discussed here before.

Issue #1484, opened 2019-07-17
Add support for nested triples storage and querying with RDF* and SPARQL* 


jeenbroekstra:
See  http://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/

RDF* is a proposal to extend RDF with "nested triples". 
SPARQL* is a proposal to extend SPARQL syntax with the ability to query such nested triples.

A basic form of support for this approach would involve a Rio parser that can read extended Turtle syntax and translate it to normal (reified / contextualized) RDF, as well as an extension to SPARQL that can use the extended syntax to query such reified structures.

Currently out of scope but possibly interesting for future work is to investigate lower-level support, e.g. at the level of the Java API (extending the Statement class to allow nested statements), or at the level of native storage extension to remove the need for verbose reification.

hartig:
For inspiration for how RDF*/SPARQL* support may be implemented, you may take a look at: https://github.com/RDFstar/RDFstarTools

hmottestad:
This is awesome. Property graphs are quite popular. Explaining to people that we can support it in the rdf world by using reification or graphs usually ends with someone asking why it’s so needlessly complex. 

I feel that performance is important if we implement something like this. It should be treated as a show stopper, eg. we won’t release it until fast enough to be as useful.

pulquero:
For the impatient, it is possible to create a poor-man's version that somewhat resembles SPARQL*. I simply created a magic property (TupleFunction) ':id' that maps (s p o) to an IRI (bijection). You can then write queries like
`(:bob foaf:age 23) :id/ex:certainty 0.9 .`
which superficially looks like
`<<:bob foaf:age 23>> ex:certainty 0.9 .`
I also created an equivalent inverse property `?id <- :fromid (?s ?p ?o)` so
`:john :said/:fromid (:bob foaf:age 23) .`

Further, you can create the following TupleFunctions as well:
`?stmtId rdf:subject -> ?s`
`?stmtId rdf:predicate -> ?p`
`?stmtId rdf:object -> ?o`

PR #1920, opened 2020-02-14
GH-1484 rdfstar and sparqlstar support

jeenbroekstra:
GitHub issue resolved: #1484   

Over-arching feature branch for RDF* and SPARQL* support.

To do:

- [x] RDF* parsers/writers for Turtle* and TriG*
- [x] model support
- [x] conversion utilities between Triple and RDF reification
- [x] update SPARQL parser (may split this out)


jeenbroekstra:
Looks like we forgot to squash some things earlier, commit history for this looks a bit odd. I'll clean that up once we're ready to go.

jeenbroekstra:
Rebasing (or indeed squashing) this branch turns out to be a bit of a nightmare. I hope that we can clean up at the end by using github squash and merge (good thing I hadn't disabled it yet ;))

jeenbroekstra:
At this point, I think the main open question is how to fit the RDF* support in with the Sail API.

With the current setup, loading an RDF* model directly into an in-memory or native store will fail, since these SAIL implementations do not know how to handle the new `Triple` object. 

Clearly, we can extend the specific SAIL implementations to support this (by introducing new `MemTriple` and `NativeTriple` objects), but I'm trying to think of some design that makes sure sail implementations that don't have dedicated support continue to work (for example by auto-converting to reified triples), or at least have an elegant way to signal that they can't deal with RDF* data. 

Apart from this: can I get opinions and tests from people to get a sense of how solid this is?

Another question: as far as I can tell, the changes as done sofar are backward-compatible. This gives us the option to publish this as an (experimental) new feature in a 3.2 release, rather than waiting for a major new release. Any thoughts on this? Anything I overlooked?

damyan-ognyanov:
It is stable - we use it internally for the upcomming release of GraphDB but our engine uses it's own evaluation strategy and also our sail implementation uses dedicated triple collections to index the triple entitites which are then uses to handle `evaluate(TripleRef)` requests more efficiently.

Currently there is no usable implementation of `StrictEvaluationStrategy.evaluate(TripleRef)` which intentionally was left to return empty iteration.

Initially, I've used a `naive` and way suboptimal implementation that enumerate and filter/match all subjects of statements with `rdf:type` predicate and `rdf:Statement` object (e.g. standard reification) but discarded it for the time being, since it `fixes` the default behaviour to standard reification and do not take into an account the (future) sail implemetations that may have native support for embeded triples which maintain dedicatd collections for more efficient handling of these requests.

One possible approach is to introduce a new interface that extends TripleSource and introduces a new method for the purpose of finding the triple entities in the dataset - nammed  `getTripleStatements()` or simmilar. 
Then, in the evaluatin strategy we can check, if the triple source provided by the sail implementation implements that interface and if so, use that in `evaluate(TripleRef)`, if not - the strategy reverts to use the defaut (suboptimal) implementation that uses standard reification.

jeenbroekstra:
I'm doing some work on this branch to get it in a better state. Some of this may involve rebasing and squashing. Please stay tuned, and be aware that you may be required to chuck away your local copy and check this branch out fresh after I'm done.

jeenbroekstra:
I have force-pushed a squashed/cleaned up commit history for this branch. There were a lot of merge conflicts I had to deal with. Local tests seem to indicate that I got it right, let's hope the CI build agrees. 

Please be aware that if you have a fork or a local copy of this branch, you should throw it away and check it out again. Apologies for the inconvenience.

jeenbroekstra:
After having taken a closer look I no longer believe that we can get away with marking this change as backward compatible. The code introduces several new methods in interfaces and abstract base classes that implementors will have to tweak. One particular example: the introduction of the abstract `consumeStatement` (originally called `handleStatementImpl`) method, in `AbstractRDFWriter`.

So our original plan stands: this will be targeted at the 4.0 release. Given that we are quite far along we can expedite that release I believe.

jeenbroekstra:
Support in the various SAIL implementations will be handled in followup issues, as well as the support for query evaluation that @damyan-ognyanov suggested above.

jeenbroekstra:
I believe this PR is now ready for final review and possibly merge. I would appreciate some additional review feedback (even if it's a simple "glanced at it, looks good to me"), ideally with a green tick of approval :)

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -753,4 +758,152 @@ private static Model toModel(Iterable<? extends Statement> iterable) {
...
+	 * Converts the supplied RDF* model to an RDF reification model.
+	 * <p>
+	 * The supplied value factory is used to create all new statements.
+	 *
+	 * @param vf    the {@link ValueFactory} to use for creating statements.
+	 * @param model the {@link Model} to convert.
+	 * @return the converted {@link Model}.
+	 */
+	public static Model convertRDFStarToReification(ValueFactory vf, Model model) {
+		Model reificationModel = new LinkedHashModel();
```
hmottestad: Could be use the new DynamicModel here?

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -753,4 +758,152 @@ private static Model toModel(Iterable<? extends Statement> iterable) {
...
+	 * The supplied value factory is used to create all new statements.
+	 *
+	 * @param vf       the {@link ValueFactory} to use for creating statements.
+	 * @param model    the {@link Model} to convert.
+	 * @param consumer the {@link Consumer} function for the produced statements.
+	 */
+	public static void convertReificationToRDFStar(ValueFactory vf, Model model, Consumer<Statement> consumer) {
+		Map<Resource, Triple> convertedStatements = new HashMap<>();
+		model.filter(null, RDF.TYPE, RDF.STATEMENT).forEach((s) -> {
+			Value subject = object(model.filter(s.getSubject(), RDF.SUBJECT, null)).orElse(null);
```
hmottestad: This kind of use of .filter(....) shows that we really need a .select(...) instead. I think .filter(....) is too heavyweight with all its guarantees about returning a model that reflects the underlying model...while all we really need here is at iterator.

hmottestad: I think it might be faster to have a java switch over `model.filter(s.getSubject(), null, null)` instead of doing a model.filter for each.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -753,4 +758,152 @@ private static Model toModel(Iterable<? extends Statement> iterable) {
...
+	 * Converts the supplied RDF reification model to an RDF* model.
+	 * <p>
+	 * The supplied value factory is used to create all new statements.
+	 *
+	 * @param vf    the {@link ValueFactory} to use for creating statements.
+	 * @param model the {@link Model} to convert.
+	 * @return the converted {@link Model}.
+	 */
+	public static Model convertReificationToRDFStar(ValueFactory vf, Model model) {
+		Model rdfStarModel = new LinkedHashModel();
```
hmottestad: DynamicModel

core/model/src/main/java/org/eclipse/rdf4j/model/util/Statements.java
```diff
@@ -26,6 +31,16 @@
...
+	/**
+	 * A {@link Function} that maps {@link Triple} to {@link org.eclipse.rdf4j.model.BNode} consistently. Multiple
+	 * invocations for the same {@link Triple} will return the same {@link org.eclipse.rdf4j.model.BNode}.
+	 *
+	 * The current implementation creates a {@link org.eclipse.rdf4j.model.BNode} by encoding the string representation
+	 * of the {@link Triple} using base64 URL-safe encoding.
+	 */
+	public static Function<Triple, Resource> TRIPLE_BNODE_MAPPER = (t) -> SimpleValueFactory.getInstance()
+			.createBNode(Base64.getUrlEncoder().encodeToString(t.stringValue().getBytes(StandardCharsets.UTF_8)));
+
```
hmottestad: Does this risk bnode identifiers being really long of the Triple contains a long literal?

core/model/src/test/java/org/eclipse/rdf4j/model/util/ModelsTest.java
```diff
@@ -445,4 +446,77 @@ public void testStripContextsSpecificContext() {
...
+
+		Model rdfStarModel3 = new TreeModel();
+		Models.convertReificationToRDFStar(VF, reificationModel, rdfStarModel3::add);
+		assertTrue("RDF reification conversion to RDF* with explicit VF, model-to-consumer",
+				Models.isomorphic(rdfStarModel3, referenceRDFStarModel));
+
+		Model rdfStarModel4 = new TreeModel();
+		Models.convertReificationToRDFStar(reificationModel, rdfStarModel4::add);
+		assertTrue("RDF reification conversion to RDF* with implicit VF, model-to-consumer",
+				Models.isomorphic(rdfStarModel4, referenceRDFStarModel));
```
hmottestad: I believe you can use .equals on models and it uses isomorphic underneath.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/IsTripleFunction.java
```diff
@@ -0,0 +1,36 @@
...
+public class IsTripleFunction implements Function {
+	@Override
+	public String getURI() {
+		return RDF.NAMESPACE + "isTriple";
+	}
+
+	@Override
+	public Value evaluate(ValueFactory vf, Value... args) throws ValueExprEvaluationException {
+		if (args.length != 1)
+			throw new ValueExprEvaluationException("expect exactly 1 argument");
```
hmottestad: { } is missing

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/StatementFunction.java
```diff
@@ -0,0 +1,42 @@
...
+public class StatementFunction implements Function {
+	@Override
+	public String getURI() {
+		return RDF.STATEMENT.toString();
+	}
+
+	@Override
+	public Value evaluate(ValueFactory vf, Value... args) throws ValueExprEvaluationException {
+		if (args.length != 3)
+			throw new ValueExprEvaluationException("expect exactly 3 arguments");
```
hmottestad: { }  is missing

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparator.java
```diff
@@ -69,8 +70,21 @@ public int compare(Value o1, Value o2) {
 			return 1;
 		}
 
-		// 4. RDF literals
-		return compareLiterals((Literal) o1, (Literal) o2);
+		// 4. Literals
```
hmottestad: @jeenbroekstra why do we have two ValueComapartors?

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
jeenbroekstra: > This kind of use of .filter(....) shows that we really need a .select(...) instead. I think .filter(....) is too heavyweight with all its guarantees about returning a model that reflects the underlying model...while all we really need here is at iterator.

Agreed. Issue logged: https://github.com/eclipse/rdf4j/issues/2030

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
jeenbroekstra: I'll have a look, and perhaps also create an overload that accepts a `ModelFactory`.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparator.java
jeenbroekstra: > @jeenbroekstra why do we have two ValueComapartors?

I'm probably overlooking something. What other one besides this class do you see?

core/model/src/test/java/org/eclipse/rdf4j/model/util/ModelsTest.java
jeenbroekstra: > I believe you can use .equals on models and it uses isomorphic underneath.

It does but I've just checked and we don't actually specify this in the API, as far as I can tell.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Statements.java
jeenbroekstra: I don't see that as a major issue, but if it becomes a problem we could switch to using a fixed-length message digest hash (MD5) as an alternative. I'd suggest it's YAGNI for now though.

hmottestad: Would collisions be an issue then?

jeenbroekstra: IIRC collision risk of MD5 is 1:2^128. For our purposes safe enough I'd say.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparator.java
hmottestad: Never mind, my fault. I had mistaken the LexicalValueComparator for being SPARQL compliant.

jeenbroekstra:
Oh drat, more merge conflicts :man_facepalming: . I'll fix and then immediately merge this PR.

PR #1923, opened 2020-02-14
GH-1484 sparqlstar

damyan-ognyanov:
Adds extensions to SparqlParser supporting SPARQL* 
* introduces two additional sparql algebra nodes `TripleRef` and `ValueExprTripleRef`
* allows use of SPARQL* embedded triple expressions in `BIND`, `VALUES`, and subject or object of a `StatementPattern`
* allows to use these also in `CONSTRUCT`, `INSERT` and `DELETE`
* added 5 new functions to work with `Triple` values - these reuse the rdf reification vocabulary
* added method to evaluate `ValueExprTripleRef` in `StrictEvaluationStrategy`. 
* currently there is no implementation of `evaluate(TripleRef)` because it is still unclear whether the default evaluation should lookup over the nodes introduced through standard reification or we need an extention to `TripleSource` that allow for lookup over the nodes of type `Triple` in the undelying Model/Sail

There is a basic test suite covering the extensions in `TestSparqlStarParser`. Each case there has the expected query model as comment above the testcase.


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/IsTripleFunction.java
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: This is not the correct copyright header I'm afraid. It should be this:
```

```

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderConstants.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 Eclipse RDF4J contributors, Aduna, and others.
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: No big dealm but there's no need to update copyright years in existing files. We never update these: they reflect the year the file was first created, not when it was last edited.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/IsTripleFunction.java
damyan-ognyanov: Updated the incorrect copyright headers

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderConstants.java
damyan-ognyanov: Updated the headers fro the existing files e.g. changed back to `2018` from `2020` ...

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/IsTripleFunctionTest.java
```diff
@@ -0,0 +1,84 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: You missed one I'm afraid.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/StatementFunctionTest.java
```diff
@@ -0,0 +1,239 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: This one too.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/TripleObjectFunctionTest.java
```diff
@@ -0,0 +1,82 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: And this one.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/function/triple/IsTripleFunctionTest.java
damyan-ognyanov: updated the rest of the headers and squashed all commits into one

Issue #1929, opened 2020-02-16
Making adding data faster by not always using a full model in Changeset


hmottestad:
Changeset is used to track which statements have been added or removed. It uses a model to track these changes.

Models are quite heavy, because they have indexes to facilitate fast statements retrieval for various patterns.

Use a regular Set if no such retrieval is required.

PR #1930, opened 2020-02-16
#1929 introduce a SimpleUpgradeableModel

hmottestad:
GitHub issue resolved: #1929  

* Use SimpleUpgradeableModel in Changeset which will upgrade from a HashSet to a Model if it encounters queries that it can't answer.
* Adding 1000 statements in SNAPSHOT takes ~6 ms instead of ~9ms

### DIFF
--- a/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
+++ b/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
@@ -33,7 +33,7 @@ import com.google.common.collect.Lists;
 
 /**
  * Integration tests for {@link RepositoryFederatedService}
- * 
+ *
  * @author Andreas Schwarte
  *
  */
@@ -101,7 +101,7 @@ public class RepositoryFederatedServiceIntegrationTest {
 				vf.createStatement(iri("s2"), RDFS.LABEL, l("val2")),
 				vf.createStatement(iri("s3"), RDFS.LABEL, l("val3"))));
 
-		String query = "SELECT ?var WHERE { VALUES ?var { 'val1' 'val2' } . SERVICE <urn:dummy> { SELECT ?var { ?s ?p ?var } LIMIT 1000  } }";
+		String query = "SELECT ?var WHERE { VALUES ?var { 'val1' 'val2' } . SERVICE <urn:dummy> { SELECT ?var { ?s ?p ?var } LIMIT 1000  } } order by ?var";
 
 		assertResultEquals(evaluateQuery(query), "var", Lists.newArrayList(l("val1"), l("val2")));
 	}
@@ -129,7 +129,7 @@ public class RepositoryFederatedServiceIntegrationTest {
 				vf.createStatement(iri("s2"), RDFS.LABEL, l("val2")),
 				vf.createStatement(iri("s3"), RDFS.LABEL, l("val3"))));
 
-		String query = "SELECT ?var WHERE { SERVICE <urn:dummy> { ?s ?p ?var } . SERVICE <urn:dummy> {  ?s ?p ?var  } }";
+		String query = "SELECT ?var WHERE { SERVICE <urn:dummy> { ?s ?p ?var } . SERVICE <urn:dummy> {  ?s ?p ?var  } } order by ?var";
 
 		assertResultEquals(evaluateQuery(query), "var", Lists.newArrayList(l("val1"), l("val2"), l("val3")));
 	}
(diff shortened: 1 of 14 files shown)


core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
```diff
@@ -511,7 +511,10 @@ public synchronized void deprecate(Statement statement) throws SailException {
 			acquireExclusiveTransactionLock();
 			requireCleanup = true;
 			if (statement instanceof MemStatement) {
-				((MemStatement) statement).setTillSnapshot(nextSnapshot);
+				MemStatement toDeprecate = (MemStatement) statement;
+				if (toDeprecate.getTillSnapshot() > nextSnapshot && toDeprecate.isExplicit() == explicit) {
+					toDeprecate.setTillSnapshot(nextSnapshot);
+				}
```
hmottestad: This fixes a bug I believe I've introduced previously. Not a problem until now though.

hmottestad:
Also tested with larger amounts of data and found a 30-40% speedup then too.

hmottestad:
Currently failing at: Repository compliance tests .................. FAILURE [01:44 min]

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
```diff
@@ -129,7 +129,7 @@ public void test4() throws Exception {
 				vf.createStatement(iri("s2"), RDFS.LABEL, l("val2")),
 				vf.createStatement(iri("s3"), RDFS.LABEL, l("val3"))));
 
-		String query = "SELECT ?var WHERE { SERVICE <urn:dummy> { ?s ?p ?var } . SERVICE <urn:dummy> {  ?s ?p ?var  } }";
+		String query = "SELECT ?var WHERE { SERVICE <urn:dummy> { ?s ?p ?var } . SERVICE <urn:dummy> {  ?s ?p ?var  } } order by ?var";
```
hmottestad: There were tests that assumed an order but didn't specify one. The new ConcurrentHashSet based model has reverse order from the LinkedHashSet based one. I think this is fine, unless there is some internal ordering requirement that I don't know about???

hmottestad:
@jeenbroekstra could you take a look at this branch and let me know what you think?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
jeenbroekstra: How is this fix related to the new SimpleUpgradeableModel class?

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/RepositoryFederatedServiceIntegrationTest.java
jeenbroekstra: A Model is supposed to have predictable iteration order but that doesn't translate to a fixed order on a SPARQL query result, so good find.

hmottestad:
Todo:

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
hmottestad: I've made some more changes here.

The concept of directly manipulating the statement was introduced with the ElasticsearchStore. Gives a decent performance improvement since it allows us to skip retrieving a statement we already have. Unfortunately it's common that we try to deprecate a statement that can't be deprecated (not in our snapshot or explicit/inferred). That's usually handled when retrieving the statement again, however since we are working on it directly we need to apply the same checks.

hmottestad: I've not copied this line from here: https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementIterator.java#L142

hmottestad:
Since this is going to be a complete new feature I’ve moved the SimpleUpgradeableModel to the same package as the other models and integrated it into those test.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
hmottestad: Btw. The reason this wasn't caught earlier was because the statements was never instanceof MemStatement directly until now. The LinkedHashModel wraps it's statement in a LinkedHashModelStatement (or something like that)

PR #1937, opened 2020-02-20
GH-1795 increment snapshot number when closing branch instead of when flushing

hmottestad:
GitHub issue resolved: #1795

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
@@ -260,13 +260,13 @@ class SailSourceBranch implements SailSource {
 				}
 				flush(prepared);
 				prepared.flush();
-				try {
-					if (prepared != serializable) {
-						prepared.close();
-					}
-				} finally {
-					prepared = null;
-				}
+//				try {
+//					if (prepared != serializable) {
+//						prepared.close();
+//					}
+//				} finally {
+//					prepared = null;
+//				}
 			}
 		} finally {
 			semaphore.unlock();
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
@@ -427,16 +427,18 @@ class MemorySailStore implements SailStore {
 
 		@Override
 		public synchronized void flush() throws SailException {
+
+		}
+
+		@Override
+		public void close() {
 			if (txnLock) {
 				currentSnapshot = Math.max(currentSnapshot, nextSnapshot);
 				if (requireCleanup) {
 					scheduleSnapshotCleanup();
 				}
 			}
-		}
 
-		@Override
-		public void close() {
 			try {
 				boolean toCloseTxnLock = txnLock;
 				txnLock = false;
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/UnionSailSource.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/UnionSailSource.java
@@ -75,7 +75,20 @@ class UnionSailSource implements SailSource {
 
 	@Override
 	public SailDataset dataset(IsolationLevel level) throws SailException {
-		return new UnionSailDataset(primary.dataset(level), additional.dataset(level));
+		SailDataset dataset1;
+		SailDataset dataset2;
+
+		// do {
+		dataset1 = primary.dataset(level);
+		try {
+			Thread.sleep(100);
+		} catch (InterruptedException e) {
+			e.printStackTrace();
+		}
+		dataset2 = additional.dataset(level);
+		// }while(!dataset1.sameTransactionAs(dataset2));
+
+		return new UnionSailDataset(dataset1, dataset2);
 	}
 
 }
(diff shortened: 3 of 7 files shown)


core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
```diff
@@ -260,13 +260,13 @@ public void flush() throws SailException {
...
-				} finally {
-					prepared = null;
-				}
+//				try {
+//					if (prepared != serializable) {
+//						prepared.close();
+//					}
+//				} finally {
+//					prepared = null;
+//				}
```
hmottestad: commenting out this means that the transaction write lock will remain locked

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
```diff
@@ -427,16 +427,18 @@ public synchronized void prepare() throws SailException {
...
+		}
+
+		@Override
+		public void close() {
 			if (txnLock) {
 				currentSnapshot = Math.max(currentSnapshot, nextSnapshot);
 				if (requireCleanup) {
 					scheduleSnapshotCleanup();
 				}
 			}
```
hmottestad: Moved the transaction incrementation into the closing of the branch instead of at the flushing or the branch.

hmottestad:
@jeenbroekstra this is an alternate direction

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemorySailStore.java
jeenbroekstra: I think that makes sense. Part of the problem with with this entire structure is that none of this is really documented - I keep finding myself trying to puzzle out the intent.

jeenbroekstra: I think that makes sense. Part of the problem with with this entire structure is that none of this is really documented - I keep finding myself trying to puzzle out the intent.

hmottestad:
Todo
 [ ] port the inference test from the other branch
 [ ] port any bug fixes from that branch
 [ ] run benchmarks
 [ ] consider creating benchmarks to test mixed parallel workloads

hmottestad:
that test failure isn't looking very promising :(

hmottestad:
```
2020-02-21T08:38:57.7857246Z [INFO] Running org.eclipse.rdf4j.sail.inferencer.fc.SchemaCachingRDFSInferencerIsolationLevelTest
2020-02-21T08:38:58.2094547Z 08:38:58.207 [Thread-426] ERROR Size was 1141. Expected 141 or 2141
2020-02-21T08:38:58.3857342Z 08:38:58.367 [Thread-426] ERROR 
```

So this again allows another transaction to peek between two flushed branches.

hmottestad:
Screenshot 2020-02-21 at 13 24 36

I've found a new issue. When creating two datasets they can be created from separate snapshots!

hmottestad:
I tried to fix this by adding a check for if the two datasets were in the same snapshot. This works for some of the tests, but others hang. It seems that when using regular SNAPSHOT isolation it stores the snapshot dataset in the SailSourceBranch, meaning that it will return a cached snapshot each time. My solution of "get a new dataset until we have two in the same snapshot" didn't work then.

Because we don't set the snapshot when beginning a transaction it's going to be hard to make sure that we don't create an inferredDataset at a different place than we create an explicitDataset.

hmottestad:
I'm very curious as to why the NativeStore doesn't have this kind of issue.

jeenbroekstra:
> I'm very curious as to why the NativeStore doesn't have this kind of issue.

The nativestore does not work with incrementing snapshot numbers at all. Instead, it keeps a bitmask persisted with each statement that indicates its transactional status: ADDED/REMOVED if added/removed by a currently active transaction (but not yet committed), and an EXPLICIT flag to indicate if the statement is inferred or explicit. This means that regardless of the isolation level, sinks can flush to disk, as filtering on these bit flags will ensure any 'dirty' statement will not be included in a concurrent transaction's read operations yet.

hmottestad:
I  believe the memory store uses something along the lines of MVCC.

hmottestad:
@jeenbroekstra To summarise the issue now:

 - T1: get inferred dataset => snapshot 0
 - T2: commit some changes => increment snapshot
 - T1: get explicit dataset => snapshot 1

T1 can now read old inferred data. This is correct. T1 can now read new explicit data, this is incorrect. 

This is incorrect for all levels starting at READ_COMMITTED since you are reading parts of a transaction. 

I don't know how to solve this. Any ideas @jeenbroekstra? I still think the best solution would be remove the inferred/explicit branch approach.

jeenbroekstra:
>  I still think the best solution would be remove the inferred/explicit branch approach.

If you want to start on that idea, go right ahead - but I did a bit of a spike on it earlier and it's a tremendous amount of work.

I'm personally thinking that the simpler fix is to see if we can copy over the bit flag mechanism the native store uses to track state (not literally copy perhaps, but use as inspiration). 

By the way there's of course no reason why we can't do both approaches in parallel. My idea is on the level of a bugfix, whereas yours is a more comprehensive redesign (and likely needs a minor or even a major new rleease).

jeenbroekstra:
@hmottestad I lost track of this - what's the state according to you? You want to rebase this to get it up to date with the current main branch? What needs to happen to get this in a working (and hopefully, mergable) state? 

Looking back at the discussion I think we have two general ideas about how to address the snapshot numbering issue. I like the idea you mentioned (getting rid of the separte explicit/inferred branches) a lot, and given that we are looking at a possible major release, now would be a good time to start planning for that. Do you think you will have capacity to look into that together with me? Is this something we want to add on top of this current branch, or should we just throw this away and start again?

hmottestad:
We were not able to fix the issue no. The only real fix I can think of is to add support for reasoning throughout the stack instead of the split approach we have today.

hmottestad:
Time wise I am very uncertain. At least not before august.

Issue #1939, opened 2020-02-21
Clearing statements from native store with inference does not work as expected


hmottestad:
Add `[] a rdfs:Resource` in a transaction. Call .clear(). Statement is still there.

hmottestad:
```
		long originalNumberOfStatements;
		try (RepositoryConnection connection = testRepository.getConnection()) {
			originalNumberOfStatements = connection.getStatements(null, null, null, true).stream().count();
		}

		try (RepositoryConnection connection = testRepository.getConnection()) {
			connection.begin(level);
			connection.add(vf.createBNode(), RDF.TYPE, RDFS.RESOURCE);
			connection.commit();
			connection.begin(level);
			connection.clear();
			connection.commit();
		}

		try (RepositoryConnection connection = testRepository.getConnection()) {
			List<Statement> collect = connection.getStatements(null, null, null, true)
					.stream()
					.collect(Collectors.toList());
			assertEquals(originalNumberOfStatements, collect.size());
		}
```

Issue #1932, opened 2020-02-17
Only attempt to clear inferred statements when calling clearInferred(...)


hmottestad:
SailSourceConnection has a clearInferred(...) method that actually gets all statements and then passes them onto the inferred branch to get removed. This is costly when there are more explicit statements than inferred statements.

PR #1940, opened 2020-02-21
#1939 bug when clearing inferred native store

hmottestad:
Test to show failure.

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/inferencer/InferencerConnectionWrapper.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/inferencer/InferencerConnectionWrapper.java
@@ -103,6 +103,7 @@ public class InferencerConnectionWrapper extends NotifyingSailConnectionWrapper 
 		long count2 = getWrappedConnection().getStatements(null, null, null, true).stream().distinct().count();
 		if (count != count2) {
 			System.out.println(count + " != " + count2);
+			assert count == count2;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
The test does not fail on RDFSchemaNativeRepositoryConnectionTest, which uses the deprecated ForwardChainingInferencer. On that basis I'd suspect the caching mechanism in the SchemaCachingInferencer somehow not being fully reset. It's weird that it doesn't fail on the memorystore though. Perhaps a discrepancy in how the notification mechanism works?

jeenbroekstra:
I've just noticed that if we revert the recent optimization fix from GH-1932, this test no longer fails.

jeenbroekstra:
What is interesting is that the extra statement is an inferred statement. If you adapt the test to only retrieve explicit statements at the end, the result is zero. So for some reason, the explicitly added statement is removed, but an inferred "copy" remains.

hmottestad:
I didn't look much into it. Just realised it when working on the snapshot isolation branch. It's great you've found out that it's only for the SchemaCaching... that should narrow down a fix. This issue only seems to apply when adding a statement that is otherwise inferred, so I wouldn't call it critical.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/inferencer/InferencerConnectionWrapper.java
```diff
@@ -93,7 +98,12 @@ public void prepare() throws SailException {
 	@Override
 	public void commit() throws SailException {
 		flushUpdates();
+		long count = getWrappedConnection().getStatements(null, null, null, true).stream().distinct().count();
 		super.commit();
+		long count2 = getWrappedConnection().getStatements(null, null, null, true).stream().distinct().count();
+		if (count != count2) {
+			System.out.println(count + " != " + count2);
+		}
 	}
```
hmottestad: What I've found out is that the count is correct before calling `super.commit()` but wrong after.

Issue #1944, opened 2020-02-24
FedX throws error when filters are pushed to ExclusiveGroups


aschwarte10:
On the mailing list it was reported that FedX throws an error in presence of a FILTER expression.

> 2020-02-24 13:26:48.024  WARN 30496 --- [           main] o.e.rdf4j.federated.FedXConnection       : Exception occured during optimization (Query: 1): Operation not supported for org.eclipse.rdf4j.federated.algebra.ExclusiveGroup, filters already to children during optimization.

An analysis revealed that indeed there is a bug in the current implementation as of the 3.1 release: 

the filter optimizer attempts to push a simple filter expression to an ExclusiveGroup, however, this method is not implemented correctly and throws an UnsupportedOperationException.

The implementation assumed to infer the filter expressions from the child algebra nodes. This was however not the case as the tree visitor did not visit the children.

Reproducible unit test in FilterTests

```
@Test
	public void testSimpleFilterExclusiveGroup() throws Exception {
		prepareTest(Arrays.asList("/tests/data/data1.ttl", "/tests/data/data2.ttl"));

		String query = "PREFIX foaf: <http://xmlns.com/foaf/0.1/>\n" +
				"PREFIX ns1: <http://namespace1.org/> "
				+ "SELECT * WHERE { "
				+ "?person a foaf:Person ."
				+ "?person foaf:name \"Person1\" .\n"
				+ "?person foaf:age 20 . "
				+ "FILTER (?person=ns1:Person_1)"
				+ "}";

		fedxRule.enableDebug();

		try (TupleQueryResult tqr = federationContext().getQueryManager().prepareTupleQuery(query).evaluate()) {

			List<BindingSet> res = Iterations.asList(tqr);

			System.out.println(res);
			assertContainsAll(res, "person",
					Sets.newHashSet(iri("http://namespace1.org/", "Person_1")));
		}
	}
```

PR #1945, opened 2020-02-24
GH-1944: fix filter evaluation for exclusive groups in FedX

aschwarte10:
The previous implementation did not correctly handle the optimization of
filter expressions in presence of ExclusiveGroups: if it was attempted,
an UnsupportedOperationException was thrown.

Now the appropriate methods are handled in the ExclusiveGroup and
filters as well as bound bindings are properly handled

GitHub issue resolved: #1944


aschwarte10:
@jeenbroekstra let me know if it is correct to target this PR for the master branch, when I want the fix to be part of 3.1.1

PR #1950, opened 2020-02-27
GH-1484: Compatibility serialization for RDF* in RDF and RDF* serialization for tuple query results

avataar:
This introduces:
* Compatibility serialization of RDF* triples in formats (both RDF and query results) that don't support RDF*.
  * The triples are serialized as special IRIs starting with urn:rdf4j:triple: followed by the base64-encoded (the URL-safe variant) of the N-Triples serialization of the triple. This is controlled by a boolean writer setting and it is ON by default. The setting is ignored by writers that support RDF* natively.
  * Such special IRIs are converted back to triples on parsing. This is controlled by a boolean parser setting and it is ON by default. It is respected by all parsers including those with native RDF* support.
* RDF* support for tuple query results
  * Binary query results supports it out of the box without introducing a new format, much like the binary RDF format.
  * Introduced SPARQL* JSON format based on SPARQL JSON (see `SPARQLStarResultsJSONConstants` for example of triple serialization).
  * Introduced SPARQL* TSV format based on SPARQL TSV. Triples are serialized in their N-Triples form (much like all other values in TSV).

Other changes:
* Added another MIME type for Turtle* to better reflect its Turtle origin
* Parser/writer (both RDF and query result) cleanup and contract enforcement, improved tests
  * `AbstractRDFWriter` defines a new abstract protected method `handleStatementImpl` that implementing classes must override instead of `handleStatement`.
  * The sequence `startRDF()` ... `handleXXX()` ... `endRDF()` is now enforced strictly (there were some cases where `handleXXX()` was called before `startRDF()`.
  * `AbstractQueryResultWriter` defines a new abstract protected method `handleSolutionImpl` that implementing classes must override instead of `handleSolution`.
* NTriplesUtil moved outside NTriples module (similar to some other NTriples related classes used in multiple places)


bom/pom.xml
```diff
@@ -83,6 +83,11 @@
 				<artifactId>rdf4j-queryresultio-sparqljson</artifactId>
 				<version>${project.version}</version>
 			</dependency>
+			<dependency>
+				<groupId>org.eclipse.rdf4j</groupId>
+				<artifactId>rdf4j-queryresultio-sparqlstarjson</artifactId>
```
jeenbroekstra: Could we add the functionality to the existing sparql-json module, instead of introducing a new module? Same for the text one.

core/queryresultio/api/src/main/java/org/eclipse/rdf4j/query/resultio/ValueMappingBindingSet.java
```diff
@@ -0,0 +1,93 @@
...
+import org.eclipse.rdf4j.query.AbstractBindingSet;
+import org.eclipse.rdf4j.query.Binding;
+import org.eclipse.rdf4j.query.BindingSet;
+
+import java.util.Iterator;
+import java.util.Set;
+import java.util.function.Function;
+
+/**
+ * A {@link BindingSet} that provides a mechanism to map values by using a custom {@link Function}.
```
jeenbroekstra: Clever.

bom/pom.xml
avataar: Sure, what about turtlestar and trigstag then? I'd be happy to add those to turtle and trig as well instead of having them as separate modules.

jeenbroekstra: Good point. Yes, if you can. I think the functionality is closely enough related that it makes more sense to have this in the same module. Plus we're already a bit module-heavy in RDF4J, I'd rather avoid creating even more unless we have to.

avataar:
Moved all new formats to the same modules as their respective non-RDF* versions (e.g. turtlestar to turtle, sparqlstar-json to sparql-json).

Added a fix for ordering Triple in ValueComparator.

Issue #1949, opened 2020-02-27
Running an INSERT DATA query fails with exception in writable FedX federation


aschwarte10:
When running an INSERT DATA query against a writable federation, this ends up in an error:

```
java.lang.UnsupportedOperationException: Not supported. the federation is readonly.
	at org.eclipse.rdf4j.federated.FedXConnection.setNamespaceInternal(FedXConnection.java:307)
	at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.setNamespace(AbstractSailConnection.java:629)
	at org.eclipse.rdf4j.repository.sail.helpers.RDFSailInserter.addNamespace(RDFSailInserter.java:65)
	at org.eclipse.rdf4j.repository.util.AbstractRDFInserter.endRDF(AbstractRDFInserter.java:138)
	at org.eclipse.rdf4j.rio.turtle.TurtleParser.parse(TurtleParser.java:186)
	at org.eclipse.rdf4j.repository.sail.helpers.SailUpdateExecutor.executeInsertData(SailUpdateExecutor.java:402)
	at org.eclipse.rdf4j.repository.sail.helpers.SailUpdateExecutor.executeUpdate(SailUpdateExecutor.java:122)
	at org.eclipse.rdf4j.repository.sail.SailUpdate.execute(SailUpdate.java:64)
	at 
```

This is because FedX currently throws an exception in the setNamespace method, also this method is not actually required for the evaluation of the query.

The simplest fix is to not throw an exception in the setNamespace method

Note that INSERT WHERE queries are working fine.

PR #1951, opened 2020-02-27
GH-1949: fix INSERT DATA query handling for writable FedX federation

aschwarte10:
* do not throw UnsupportedOperationException for internal namespace
handling methods
* pass information on the original update string to the engine
* UPDATE queries are always handled inside the federation engine to
adhere to the configured write strategy
* cover this by unit tests

GitHub issue resolved: #1949


aschwarte10:
> LGTM - couple of braces missing around if-blocks but that's not a blocker.

Where exactly do you want to see braces (just so that I know the next time), around `queryInfo.getQueryType() != QueryType.UPDATE`?

Issue #232, opened 2016-07-12
Support for HDT


barthanssens:
Would be nice to have a RIO reader/writer for HDT

ansell:
Some related background for a failed Google Summer of Code project hosted by Apache Marmotta that I was involved in can be found at:

https://issues.apache.org/jira/browse/MARMOTTA-593

The main sticking point was that Apache Marmotta and Eclipse RDF4J cannot reuse any of the core RDF/HDT library because they release it as LGPL. Looking at it again, they provide an "API" module that is Apache licensed, but haven't looked through it to know if it would actually be useful without the core module and whether it would limit the way HDT could be implemented in RDF4J:

https://github.com/rdfhdt/hdt-java

jeenbroekstra:
Just in case people are not aware: RDF4J does already have support for a binary RDF format, developed in-house. Still haven't included the documentation on the main site, but it can be found here: http://www.rivuli-development.com/2011/11/binary-rdf-in-sesame/ . 

Compared to HDT it lacks some important features (indexing/immediate search-browse in particular), but it does offer a compact transfer/storage format.

barthanssens:
Well, I'll give it a try and see how far I get (based upon the spec itself to avoid license issues). It can take a while though ;-)

ansell:
The other issue that we had for Marmotta was that the spec wasn't up to date with the current implementation, so you couldn't rely on it for interpreting any files generated by hdt-java. That situation may have improved since then if the spec has been updated recently.

ansell:
@jeenbroekstra The HDT-Java format indexes are being utilised for in-place searching/filtering by a few different groups in bioinformatics so they are proving to be valuable at least for that area.

barthanssens:
@ansell I happen to know the Belgian guys working on hdt-java/hdt-cpp, so I'll ask them if the spec on hdt.org has been updated accordingly (it is mentioned that the HDT spec was updated after submitting it to W3C, so hopefully the website is now documenting the latest version)

jeenbroekstra:
Great that you want to try and pick this up Bart! Before you dig in, please have a look at the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md), in particular the points on how to sign the Eclipse CLA and how pick the right branch etc. Let me know if you need any help with any of it.

barthanssens:
[...]

PR #1952, opened 2020-02-27
GH-232 RIO HDT Parser

barthanssens:
GitHub issue resolved: #232  

* Initial (experimental, non-optimized) HDT parser

The specification itself is slightly out of date and incomplete, the goal is to be able to read files generated by HDT-It 1.0 (see also http://www.rdfhdt.org/development/)

Basically this binary format tries to greatly reduce file size by creating dictionaries of subjects, predicates and objects, and using arrays of numeric references to encode triples. 
The assumptions are that a) the number of _different_ predicates are small but repeated very often, and b) objects of triples are often subject(s) of other triples (so these repetitive parts only need to be stored once)

Currently limited to in PlainFrontEncoded dictionaries, in SPO order, with Bitmap and Log64 arrays.

Number of entries probably limited to 512 M (due to reading unsigned int number of bytes, with max 4 bytes per entry), and limited by amount of memory.


barthanssens:
I'd be grateful if someone could give it a glance to see if there are obvious mistakes in the Rio part.

It's experimental, and I have to add more tests of course, though it already seems to work on a few larger test files (500 K triples) not included in this PR

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
```diff
@@ -0,0 +1,108 @@
...
+ * 
+ * <pre>
+ * +------+--------+---------+------+------...
+ * | type | nrbits | entries | CRC8 | data 
+ * +------+--------+---------+------+------...
+ * </pre>
+ * 
+ * @author Bart Hanssens
+ */
+public abstract class HDTArray extends HDTPart {
```
jeenbroekstra: Instead of sticking these components in a separate package and making them public, can we just stick them in the `rio.hdt` package and make them package-private instead? If I understand correctly, they're meant for internal use only, so I'd rather not have them be part of the public API if we can avoid it.

core/rio/pom.xml
```diff
@@ -24,5 +24,6 @@
 		<module>trix</module>
 		<module>turtle</module>
 		<module>trig</module>
+	 <module>hdt</module>
```
jeenbroekstra: Needs an extra indent :)

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+
+	/**
+	 * Get header data as raw byte stream. This often includes a series of triples (in N-Triple) with additional
+	 * metadata and statistics.
+	 * 
+	 * @return bytes
+	 */
+	public byte[] getHeaderData() {
+		return headerData;
+	}
```
jeenbroekstra: Do these three methods need to be public?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
```diff
@@ -0,0 +1,108 @@
...
+			return value;
+		}
+
+		private Type(int value) {
+			this.value = value;
+		}
+	}
+
+	protected int nrbits;
+	protected int entries;
```
jeenbroekstra: Can these be private?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
jeenbroekstra: Alternatively, we leave as-is and place our bets on eventually being able to upgrade to java 9, using modules to handle this kind of thing.

core/rio/hdt/src/test/java/org/eclipse/rdf4j/rio/hdt/util/CRC8Test.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ *
+ * @author Bart.Hanssens
+ */
+public class CRC8Test {
+	@Test
+	public void testHello() {
+		CRC8 crc = new CRC8();
+		crc.update("Hello world".getBytes(), 0, "Hello world".length());
```
ansell: `getBytes` should include the charset.

core/rio/hdt/src/test/java/org/eclipse/rdf4j/rio/hdt/util/CRC8Test.java
```diff
@@ -0,0 +1,33 @@
...
+		assertEquals("CRC hello world not correct", 0x41, crc.getValue());
+	}
+
+	@Test
+	public void testHelloPerByte() {
+		CRC8 crc = new CRC8();
+		for (byte b : "Hello world".getBytes()) {
```
ansell: `getBytes` should include the charset.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+		HDTDictionarySection objects = null;
+		HDTTriplesSection section = null;
+
+		// not using try-with-resources, since the counter is needed in the catch clause (JDK8)
+		CountingInputStream bis = new CountingInputStream(new BufferedInputStream(in, BUFLEN));
+		try {
+			reportLocation(0, -1);
+			HDTGlobal global = new HDTGlobal();
+			global.parse(bis);
+			globalProps = global.getProperties();
```
ansell: This seems to be the only reference to this field other than its unused accessor. Is it necessary to make this a field and not a local variable inside of this method?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+		return globalProps;
+	}
+
+	/**
+	 * Get header properties
+	 * 
+	 * @return key,value map
+	 */
+	public Map<String, String> getHeaderProperties() {
+		return headerProps;
```
ansell: This is the only reference, read or write, to this field.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+import org.eclipse.rdf4j.rio.hdt.part.HDTGlobal;
+import org.eclipse.rdf4j.rio.hdt.part.HDTHeader;
+import org.eclipse.rdf4j.rio.hdt.part.HDTTriples;
+import org.eclipse.rdf4j.rio.hdt.part.HDTTriplesSection;
+import org.eclipse.rdf4j.rio.hdt.part.HDTTriplesSectionFactory;
+
+import org.eclipse.rdf4j.rio.helpers.AbstractRDFParser;
+
+/**
+ * RDF parser for HDT v1.0 files. This parser is not thread-safe, therefore its public methods are synchronized.
```
ansell: It looks like the thread-safety issues (three publicised fields) are not necessary currently and this could infact be a thread-safe class with minimal changes.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+ * 
+ * Unfortunately the draft specification is not entirely clear and probably slightly out of date, since the open source
```
ansell: It would be nice to document what is expected in terms of file contents at a fairly low level at least where it is customised and we can't directly refer to a published or draft specification as the definitive documentation for the binary format.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTDictionary.java
```diff
@@ -0,0 +1,51 @@
...
+ * <pre>
+ * +------+------+-----+------+------------+------+-------+
+ * | $HDT | type | URI | NULL | key=value; | NULL | CRC16 |
+ * +------+------+-----+------+------------+------+-------+
+ * </pre>
+ * 
+ * @author Bart Hanssens
+ */
+public class HDTDictionary extends HDTPart {
+	public final static byte[] DICT_FORMAT = "<http://purl.org/HDT/hdt#dictionaryFour>".getBytes();
```
ansell: `getBytes` should have a charset specified.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTDictionarySectionFactory.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * HDT DictionarySection factory.
+ * 
+ * @author Bart Hanssens
+ */
+public class HDTDictionarySectionFactory {
+	public static HDTDictionarySection parse(InputStream is) throws IOException {
+		int dtype = is.read();
+		if (dtype != HDTDictionarySection.Type.FRONT.getValue()) {
+			throw new UnsupportedOperationException("Dictionary section: only front encoding is supported");
```
ansell: It would be useful to add to this message the `dtype` value that was found (for debugging purposes)

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTPart.java
```diff
@@ -0,0 +1,235 @@
...
+	 * 
+	 * @param is
+	 * @param format
+	 * @throws IOException
+	 */
+	protected static void checkFormat(InputStream is, byte[] format) throws IOException {
+		byte[] b = new byte[format.length];
+		is.read(b);
+		if (!Arrays.equals(b, format)) {
+			throw new IOException("Unknown format, expected " + new String(format));
```
ansell: The `String(bytes, Charset)` constructor should be used here to specify which Charset to use when interpreting the bytes.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTPart.java
```diff
@@ -0,0 +1,235 @@
...
+	 * @param props
+	 * @return
+	 */
+	protected static Map<String, String> mapProperties(byte[] props) {
+		Map<String, String> map = new HashMap<>();
+		if (props == null || props.length == 0) {
+			return map;
+		}
+
+		String strs[] = new String(props, 0, props.length).split(";");
```
ansell: The version of this String constructor with a Charset should be used.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTPart.java
```diff
@@ -0,0 +1,235 @@
...
+	/**
+	 * Writes the CRC to the output stream
+	 * 
+	 * @param cos checked output stream
+	 * @param os  output stream
+	 * @param len number of bytes of the checksum
+	 * @throws IOException
+	 */
+	/*
+	 * protected static void writeCRC(CheckedOutputStream cos, OutputStream os, int len) throws IOException { long calc
```
ansell: Would be best to leave this code out of the file rather than commented until the writer is implemented (if it is used at that point)

core/rio/hdt/src/test/java/org/eclipse/rdf4j/rio/hdt/HDTParserTest.java
```diff
@@ -0,0 +1,59 @@
...
+		try (InputStream is = HDTParserTest.class.getResourceAsStream("/test.hdt")) {
+			parser.setRDFHandler(new StatementCollector(m));
+			parser.parse(is, "");
+			assertEquals("Number of statements does not match", 43, m.size());
+		} catch (Exception e) {
+			fail(e.getMessage());
+		}
+	}
+
+	public void parseSimpleHeaderData() {
```
ansell: Was this intended to be a `@Test`?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTDictionarySectionPFC.java
```diff
@@ -0,0 +1,130 @@
...
+		}
+		bufsize = (int) val;
+
+		checkCRC(cis, is, 1);
+
+		HDTArray ha = HDTArrayFactory.parse(is);
+		ha.parse(is);
+
+		cis = new CheckedInputStream(is, new CRC32());
+		arr = new ArrayList(nrstrings);
```
ansell: The generics type or diamond-syntax is missing here. Can use `ArrayList<>(nrstrings)` rather than actually putting the type in.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
```diff
@@ -0,0 +1,108 @@
...
+
+		nrbits = cis.read();
+		long l = VByte.decode(cis);
+		if (l > Integer.MAX_VALUE) {
+			throw new UnsupportedOperationException("Maximum number of bytes in array exceeded: " + l);
+		}
+		entries = (int) l;
+
+		checkCRC(cis, is, 1);
+		// don't close CheckedInputStream, or it will close the underlying inputstream
```
ansell: You can fix this by wrapping the InputStream in a non-closeable wrapper. Not familiar with CheckedInputStream to know what the implications of not closing it are.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+	 * 
+	 * @param b byte buffer
+	 * @return literal, IRI or blank node
+	 */
+	private Value createObject(byte[] b) {
+		if (b[0] == '"') {
+			int i = b.length - 1;
+			for (; i > 1 && b[i] != '"'; i--) {
+				if (b[i] == '@') {
+					String lang = new String(b, i + 1, b.length - i - 1);
```
ansell: The Charset version of this constructor should be used here (even if lang tags are always assumed to be simple-ASCII, we shouldn't leave these things to the users locale)

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
```diff
@@ -0,0 +1,311 @@
...
+					valueFactory.createLiteral(new String(b, 1, i - 2, StandardCharsets.UTF_8), lang);
+				} else if (b[i] == '^') {
+					IRI datatype = valueFactory.createIRI(new String(b, i + i, b.length - i - 1));
```
ansell: The Charset version of this constructor should be used here.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTTriplesSectionFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * HDT Array factory.
+ * 
+ * @author Bart Hanssens
+ */
+public class HDTTriplesSectionFactory {
+
+	public static HDTTriplesSection parse(String str) throws IOException {
+		if (!str.equals(new String(HDTTriples.FORMAT_BITMAP))) {
```
ansell: Not clear why this is being round-tripped. There should be another variable with the String version to directly compare to.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTTriplesSectionFactory.java
```diff
@@ -0,0 +1,25 @@
...
+			throw new UnsupportedOperationException("Triples section: only bitmap encoding is supported");
```
ansell: The `str` variable should be included in this exception message for debugging.

jeenbroekstra:
Btw, could you please [squash  your commits](https://rdf4j.org/documentation/developer/squashing) when you're ready?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTPart.java
```diff
@@ -0,0 +1,235 @@
...
+	protected static void checkControl(InputStream is, HDTPart.Type ctype) throws IOException {
+		byte[] cookie = new byte[COOKIE.length];
+		is.read(cookie);
+		if (!Arrays.equals(cookie, COOKIE)) {
+			throw new IOException("$HDT marker not found");
+		}
+
+		byte b = (byte) is.read();
+		if (b != ctype.getValue()) {
+			throw new IOException("Expected different control info type");
```
ansell: The byte value that was found should be included in this message.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTBitmap.java
```diff
@@ -0,0 +1,79 @@
...
+		return bits;
+	}
+
+	@Override
+	public void parse(InputStream is) throws IOException {
+		CheckedInputStream cis = new CheckedInputStream(is, new CRC8());
+
+		int dtype = cis.read();
+		if (dtype != BITMAP1) {
+			throw new UnsupportedOperationException("Only bitmap v1 is supported");
```
ansell: `dtype` should be included in this message for debugging purposes.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTBitmap.java
```diff
@@ -0,0 +1,79 @@
...
+	private byte[] buffer;
+
+	/**
+	 * Get bit
+	 * 
+	 * @param i
+	 * @return 0 or 1
+	 */
+	public int get(int i) {
+		assert i < bits : "Bitmap out of range";
```
ansell: How important is this assertion? Not familiar with recent RDF4J coding standards, but using an always-on check and exception rather than a variable assertion is more generally accepted.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArrayLog64.java
```diff
@@ -0,0 +1,80 @@
...
+	private byte buffer[];
+
+	@Override
+	public int getType() {
+		return HDTArray.Type.LOG64.getValue();
+	}
+
+	@Override
+	public int get(int i) {
+		assert i < entries : "Entries out of range";
```
ansell: How important is this assertion? Should it be a full, always-on guard?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArrayFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * HDT Array factory.
+ * 
+ * @author Bart Hanssens
+ */
+public class HDTArrayFactory {
+
+	public static HDTArray parse(InputStream is) throws IOException {
+		int dtype = is.read();
+		if (dtype != HDTArray.Type.LOG64.getValue()) {
+			throw new UnsupportedOperationException("Array section: only Log64 encoding is supported");
```
ansell: Adding `dtype` to the exception message here would be useful for debugging.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTBitmap.java
jeenbroekstra: Good catch. I think we've stepped away from using assertions. They still lurk in the code base in places, but I'd prefer to gradually get rid of them in favor of exceptions.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
barthanssens: Sure, I can move them to rio.hdt.
This is actually a good example where JDK 9 modules are useful :-)

They *might* need to become visible again, if at some point there would be a HDT Sail (since the format is actually optimized for handling simple ?s ?p ?o queries), but until then it would indeed be better to make these classes private.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
barthanssens: To be honest I don't have a good use case just yet: the format actually allows one to embed metadata / statistics about the file (how many triples etc, original file size ...), pretty much like DOCX or PDF.
But the specification and the HDT-It are (once again) slightly different, so it looks like a nice to have.

I just remove the methods until a real use case pops up.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTGlobal.java
```diff
@@ -0,0 +1,52 @@
...
+ * <pre>
+ * +------+------+--------+------+------------+------+-------+
+ * | $HDT | type | format | NULL | key=value; | NULL | CRC16 |
+ * +------+------+--------+------+------------+------+-------+
+ * </pre>
+ * 
+ * @author Bart Hanssens
+ */
+public class HDTGlobal extends HDTPart {
+	public final static byte[] GLOBAL_FORMAT = "<http://purl.org/HDT/hdt#HDTv1>".getBytes();
```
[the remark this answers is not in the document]
hmottestad: FYI: https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTBitmap.java
hmottestad: Some background for assertions vs. exceptions. Assertions will only run if you enable assertions (like in a test), unlike exceptions. Assertions in small methods also hinder inlining (method becomes too big). 

That said, assertions are a helpful tool when debugging code. I find them especially useful while developing new data structures.

Another use case for assertions is when the check is semi important, but takes a lot of resources. I've once used it to discover resource leaks on a db connection where I had to use java reflection, which is very slow. The resource leak was important to catch, but I was fairly sure that one of my integration tests would catch it and didn't really want to crash in production either.

hmottestad:
I think this looks like a great idea :)

core/rio/hdt/src/test/java/org/eclipse/rdf4j/rio/hdt/util/CRC16Test.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ *
+ * @author Bart Hanssens
+ */
+public class CRC16Test {
+	@Test
+	public void testHello() {
+		CRC16 crc = new CRC16();
+		crc.update("Hello world".getBytes(), 0, "Hello world".length());
```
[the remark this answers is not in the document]
barthanssens: thanks, I've changed all the getbytes calls to use US_ASCII

core/rio/pom.xml
barthanssens: done

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
barthanssens: moved all classes to HDT package and made them package-private

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTArray.java
barthanssens: There should be no implications as it's just a simple filteredinputstream. 

But just to be sure, I've wrapped the inputstream in an UnclosedInputstream as suggested, which makes the code a bit cleaner (allows for try-with-resources). Thanks

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTPart.java
barthanssens: ok, done

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTBitmap.java
barthanssens: I'll just remove them, the main use was for catching some off-by-one errors in development

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
barthanssens: no use for it (yet), so I'll remove it

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTDictionarySectionFactory.java
barthanssens: good idea, added

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTDictionarySectionPFC.java
barthanssens: good catch

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/part/HDTPart.java
barthanssens: done

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTParser.java
barthanssens: I'm trying to add documentation (see also package-info.java) while still figuring out some of the quirks of the format :-/ But it wouldn't harm to repeat  / restructure that on the RDF4J doc website.

Obviously the best approach would be an updated draft from the HDT team :-)
In the past I've asked the researchers if they could do so, but they haven't done it just yet... I'll try contacting the developers of the HDT c++/java (Jena-based) tools, maybe they can help.

core/rio/hdt/src/test/java/org/eclipse/rdf4j/rio/hdt/HDTParserTest.java
barthanssens: removed it, since currently not used

Issue #1956, opened 2020-03-02
Use DynamicModel in more cases


hmottestad:
Consider if the SimpleUpgradeableModel should be used in more places.

jeenbroekstra:
Followup of #1929

PR #1957, opened 2020-03-02
#1956 use DynamicModel more

hmottestad:
GitHub issue resolved: #1956   

* use DynamicModel more
* fix .equals bug in DynamicModel
* rename SimpleUpgradeableModel to DynamicModel


hmottestad:
@jeenbroekstra @barthanssens I propose we use the new SimpleUpgradeableModel in Rio when loading data. The SimpleUpgradeableModel uses a LinkedHashSet for storing data (which is fast) and upgrades to a LinkedHashModel (which is slower) if the user tries to filter data or delete data based on a triple pattern. The tradeoff is that users who use this method and always query the data (always upgrades the SimpleUpgradeableModel) will have a slight performance degradation. I think this is worth it because of the number of users who don't query the data.

hmottestad:
I’m not proposing to use the SimpleUpgradeableModel in even more places. Most are assumed performance improvements. I did benchmark the SPIN one, and it’s a decent 15-20% boost and considerably more even due to less GC.

core/query/src/main/java/org/eclipse/rdf4j/query/QueryResults.java
```diff
@@ -66,7 +68,7 @@
 	 */
 	public static Model asModel(CloseableIteration<? extends Statement, ? extends RDF4JException> iteration)
 			throws QueryEvaluationException {
-		Model model = new LinkedHashModel();
+		Model model = new SimpleUpgradeableModel(new LinkedHashModelFactory());
```
jeenbroekstra: While we're at it, perhaps we could make the `ModelFactory` injectable here as well, by introducing a method overload. E.g. something like this:

    public static Model asModel(CloseableIteration<? extends Statement, ? extends RDF4JException> iteration) {
          return asModel(iteration, new SimpleUpgradeableModelFactory());
     }
     public static Model asModel(CloseableIteration<? extends Statement, ? extends RDF4JException> iteration, ModelFactory modelFactory) {
                     Model model = modelFactory.createEmptyModel();
                     ....

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/Rio.java
```diff
@@ -258,12 +259,12 @@ public static Model parse(Reader reader, String baseURI, RDFFormat dataFormat, R
 	public static Model parse(InputStream in, String baseURI, RDFFormat dataFormat, ParserConfig settings,
 			ValueFactory valueFactory, ParseErrorListener errors, Resource... contexts)
 			throws IOException, RDFParseException, UnsupportedRDFormatException {
-		Model result = new LinkedHashModel();
+		Model result = new SimpleUpgradeableModel(new LinkedHashModelFactory());
```
jeenbroekstra: Same here - could we make it injectable?

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/Rio.java
```diff
@@ -292,12 +293,12 @@ public static Model parse(InputStream in, String baseURI, RDFFormat dataFormat,
 	public static Model parse(Reader reader, String baseURI, RDFFormat dataFormat, ParserConfig settings,
 			ValueFactory valueFactory, ParseErrorListener errors, Resource... contexts)
 			throws IOException, RDFParseException, UnsupportedRDFormatException {
-		Model result = new LinkedHashModel();
+		Model result = new SimpleUpgradeableModel(new LinkedHashModelFactory());
```
jeenbroekstra: ..and here as well

jeenbroekstra:
Btw this just popped into my head but what do you think of renaming `SimpleUpgradeableModel` to something shorter, like just `UpgradeableModel`? The prefix `Simple` is really something we should only use for something that is the most naive default implementation of a particular interface. Clearly, this class is not the most naive kind of model. 

Given that we haven't done a release with this new feature yet we can easily rename it without worrying about backward compatibility.

hmottestad:
For naming I was also thinking about DynamicModel.

jeenbroekstra:
> For naming I was also thinking about DynamicModel or maybe LazyModel.

JITModel? :)

But seriously: DynamicModel sounds good to me.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
```diff
@@ -141,6 +142,35 @@ public boolean remove(Resource subj, IRI pred, Value obj, Resource... contexts)
...
+			return ret;
+		}
+
+		if (subj == null && pred == null && obj == null && contexts.length == 0) {
+			return this;
+		}
+
 		upgrade();
 		return model.filter(subj, pred, obj, contexts);
 	}
```
hmottestad: In order to get the most out of sails that use the notifying behaviour we need to implement simple support for filter. I read that the docs specify that filter(...) should return a view of the current model, rather than a new model or an unmodifiable model. Not sure how to implement this with the performance I want. Have you worked with the FilteredModel before @jeenbroekstra ?

jeenbroekstra: I'm not sure what problem you're trying to solve here, to be honest. Just filter on the underlying model when filter is used. It will use a FilteredModel internally.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
```diff
@@ -141,6 +142,35 @@ public boolean remove(Resource subj, IRI pred, Value obj, Resource... contexts)
 
 	@Override
 	public Model filter(Resource subj, IRI pred, Value obj, Resource... contexts) {
+		if (model == null && statements.size() == 0) {
+			return new EmptyModel(this);
```
jeenbroekstra: This isn't valid - you can't return a new Model object as the result of the filter. If someone were to add something to this empty model, that should be reflected in the original.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
```diff
@@ -141,6 +142,35 @@ public boolean remove(Resource subj, IRI pred, Value obj, Resource... contexts)
...
+		}
+
+		if (model == null && subj != null && pred != null && obj != null && contexts.length == 1) {
+			Statement needle = SimpleValueFactory.getInstance().createStatement(subj, pred, obj, contexts[0]);
+			if (statements.size() == 1) {
+				Statement s = (Statement) statements.toArray()[0];
+
+				DynamicModel ret = new DynamicModel(modelFactory);
```
jeenbroekstra: Same here.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
```diff
@@ -141,6 +142,35 @@ public boolean remove(Resource subj, IRI pred, Value obj, Resource... contexts)
...
+
+				if (s.equals(needle)) {
+					ret.add(s);
+				}
+
+				return ret;
+			}
+
+			DynamicModel ret = new DynamicModel(modelFactory);
```
jeenbroekstra: ...and here as well.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
jeenbroekstra: I haven't thought this through fully, but if your idea is to somehow postpone upgrading when filtering by just using the original DynamicModel in corner cases, you could just do this by returning a new FilteredModel that internally creates an iterator, and put your filtering / upgrade logic in that iterator. As long as you make sure that what you return is a filtered view on the original Model, so that any future modifications of that result object get reflected in the original model, you're good.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
jeenbroekstra: I'll push a couple of tests for this to this PR to make it easier.

jeenbroekstra:
On a more general note: if there are cases where the Sail code currently internally uses `Model.filter` where it doesn't really need the result be an internal view, but a copy would be fine, then perhaps that is a case where we should refactor it to just use something like `Model.stream().filter()` instead.

jeenbroekstra:
Oh and a tip: if you want to get your feature branch in sync with changes on master/develop, I'd use rebase instead of merge. It saves you a lot of headaches when you later need to squash your branch.

hmottestad:
Model.stream().filter doesn’t utilize any of the indexes. Was thinking we could implement at .select(sub,press,obj,ctx ...) method in Model that returns an iteration instead. 

I’m currently using the GitHub squash and merge button. I hope it’s still ok to do so. 

As for this actual PR, I think I’m going to remove the “filter” code for now so it can be merged as is. 

There is one bigger change in this PR that I need to point out to you @jeenbroekstra . The old RDFS reasoner adds all inferred statements to the null context. When adding something like :havard a :Person to context1, it would also add that exact same statement to the null context. To prevent this there is a check that stops adding inferred statements to the null context if they exist in other contexts. This check forces an upgrade of the DynamicModel, which I would rather be without. I’ve essentially removed that check. Now it checks that for every inferred statement that it’s not in the context that you are trying to add it to. This also works much better for the schema caching reasoner since it adds inferred statements to the “correct” context. I don’t consider this a breaking change. What do you think?

jeenbroekstra:
> Model.stream().filter doesn’t utilize any of the indexes. 

Fair point.

> Was thinking we could implement at .select(sub,press,obj,ctx ...) method in Model that returns an iteration instead.

I had thought of something similar (I assume you mean Iterator rather than Iteration, btw). It's worth at least investigating if there's a significant performance benefit. 

> I’m currently using the GitHub squash and merge button. I hope it’s still ok to do so.

Well like I said earlier on the dev list, I was aiming to disable that option soon, for various reasons.  But we can wait until after this PR is done at least :) 

> As for this actual PR, I think I’m going to remove the “filter” code for now so it can be merged as is.

Good call I think :+1:

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -544,9 +547,9 @@ public boolean addInferredStatement(Resource subj, IRI pred, Value obj, Resource
 			}
 			boolean modified = false;
 			if (contexts.length == 0) {
-				if (!hasStatement(explicitOnlyDataset, subj, pred, obj)) {
+				if (!hasStatement(explicitOnlyDataset, subj, pred, obj, NULL_CTX)) {
 					// only add inferred statements that aren't already explicit
-					if (!hasStatement(inferredOnlyDataset, subj, pred, obj)) {
+					if (!hasStatement(inferredOnlyDataset, subj, pred, obj, NULL_CTX)) {
```
jeenbroekstra: Just had a look at this after your remark. I'm not sure I follow the logic here to be honest. You've modified the check to only look in the null context. But what if the statement already existed (either inferred or explicit) in some other context? This adaptation will now report a new inferred statement even if in reality it might already have existed.

jeenbroekstra: It would seem more logical to me if this check explicitly looked at the context you're trying to insert the new statement to.

hmottestad: > It would seem more logical to me if this check explicitly looked at the context you're trying to insert the new statement to.

It does. There are two branches. One for context.length == 0 and one for length == 1. This is for length == 0.

jeenbroekstra: Ah, right, I missed that in the diff!

Issue #1958, opened 2020-03-02
SilentIteration in federated services does not always properly close inner resources


aschwarte10:
When running a federated SERVICE query with the SILENT keyword in our application, I noticed that connections do not get closed properly, thus resulting in resource leaks. The most obvious case is that the repository shutdown takes 20s until a timeout occurs.

Hints for reproducing: requires some repository that keeps track of open connections (e.g. a memory store) used as federated service

PR #1959, opened 2020-03-02
GH-1958: fix resource closing in federated service SilentIteration

aschwarte10:
The SilenetIteration failed to close the inner iteration, thus resulting
in resource leaks (e.g. pending connections).

Without this fix the attached unit tests takes 20s before forcefully
closing the connection. Now it correctly terminates immediately after
test execution.

Note that additionally a TRACE logger has been introduced to give users
the chance to observe errors.

GitHub issue resolved: #1958

### DIFF
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/SilentIteration.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/SilentIteration.java
@@ -7,8 +7,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.repository.sparql.federation;
 
+import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
 import org.eclipse.rdf4j.common.iteration.CloseableIteration;
-import org.eclipse.rdf4j.common.iteration.LookAheadIteration;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.QueryEvaluationException;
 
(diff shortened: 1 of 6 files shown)


core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/SilentIteration.java
```diff
@@ -34,9 +39,20 @@ protected BindingSet getNextElement() throws QueryEvaluationException {
...
 		return null;
 	}
 
+	@Override
+	protected void handleClose() throws QueryEvaluationException {
+		try {
+			Iterations.closeCloseable(iter);
+		} finally {
+			super.handleClose();
+		}
```
hmottestad: Should maybe this also fail silently?

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/SilentIteration.java
```diff
@@ -34,9 +39,20 @@ protected BindingSet getNextElement() throws QueryEvaluationException {
 				return iter.next();
```
hmottestad: I know it's not required by our styleguide, but I do prefer using { } for if/else/while.

jeenbroekstra: Agreed. Actually it is a convention for the RDF4J, it's just not explicitly enforced by the formatter because in Eclipse, this option is not part of the formatting rules (so I can't make it part of the maven formatter plugin, which uses Eclipse formatter under the hood).

Btw if you're using Eclipse, you can configure it to automatically add braces on save, by going to 'Preferences' -> 'Editor' -> 'Save Actions'. Enable 'Additional actions', click 'Configure...' and in the 'Code Style' dialog enable the option 'Use blocks in if/while/for/do statements'.  

Or if you can prefer you can use an Eclipse clean up profile to do the same thing.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/federation/SilentIteration.java
```diff
@@ -19,6 +22,8 @@
  */
 public class SilentIteration extends LookAheadIteration<BindingSet, QueryEvaluationException> {
```
jeenbroekstra: I only just noticed this after pushing an editorial fix, but apparently we have two `SilentIteration` classes:

- `org.eclipse.rdf4j.repository.sparql.federation.SilentIteration`
- `org.eclipse.rdf4j.query.algebra.evaluation.iterator.SilentIteration`

They look identical (barring the fixes you just did) - do we need both or can we get rid of the federated one?

jeenbroekstra: Ah - to answer my own question, this is probably a deliberate copy to avoid a cyclic dependency. 

Let's apply the fix to both copies in this PR. I'll look into a better solution so that we don't have to maintain separate copies separately.

jeenbroekstra: Looking a bit further, it seems the copy in `evaluation.iterator` is not actually used anywhere. Let's mark it deprecated, so we can decide to remove it in a later major release.

jeenbroekstra:
I will be pushing some fixes to this branch from my end - I need the practice to work with PRs from forked repos from my command line, and if all goes well I can clean this up quickly.

aschwarte10:
> @aschwarte10 I've made some changes, basically by introducing a generic SilentIteration class in rdf4j-util and re-using that in the federation code. Let me know what you think.

LGTM, thanks for following up

aschwarte10:
The commits are now squashed into a single one

Issue #1970, opened 2020-03-05
SparqlProtocolSession uses this.getClass() as logger name


aschwarte10:
The SparqlProtocolSession uses this.getClass() as a name for the logger.

In our application we have an extension of the class, which leads to misleading log output. Particularly the logging levels are also harder to control (w.r.t namespacing and packaging).

The general recommendation for logger names should be to use the class name, at least it should adhere to the rdf4j package structure.

PR #1971, opened 2020-03-05
GH-1970: fix logger name in SPARQLProtocolSession

aschwarte10:
The SparqlProtocolSession uses this.getClass() as a name for the logger.

When having an extension of the class, this leads to misleading log
output. Particularly the logging levels are also harder to control
(w.r.t namespacing and packaging).

The general recommendation for logger names should be to use the class
name, at least it should adhere to the rdf4j package structure.

GitHub issue resolved: #1970

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
@@ -148,7 +148,7 @@ public class SPARQLProtocolSession implements HttpClientDependent, AutoCloseable
 	 */
 	private final int maximumUrlLength;
 
-	final Logger logger = LoggerFactory.getLogger(this.getClass());
+	final Logger logger = LoggerFactory.getLogger(SPARQLProtocolSession.class);
 
 	/*-----------*
 	 * Variables *


core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -148,7 +148,7 @@
 	 */
 	private final int maximumUrlLength;
 
-	final Logger logger = LoggerFactory.getLogger(this.getClass());
+	final Logger logger = LoggerFactory.getLogger(SPARQLProtocolSession.class);
```
hmottestad: Btw. I prefer final static and possibly private for loggers.

jeenbroekstra: I agree with preferring private: as this example shows, inheriting loggers is just a recipe for confusion. Unfortunately, this was designed for this purpose in the past, so we're stuck with it at least until the next major release.

I'm not sure why you prefer static though. We use a LoggerFactory so there is no overhead - SLF4J caches instances per class name, so it's a singleton. And there may be issues caused by static logger references in some applications (see https://cwiki.apache.org/confluence/display/COMMONS/Logging+StaticLog ).

hmottestad: I wasn't aware of that issue. I mostly like static loggers because you can reference them from static context too.

Issue #1978, opened 2020-03-07
SPARQL MAX aggregate should result in type error on empty set


jeenbroekstra:
Reported by @manuelfiorelli on the rdf4j-user list.

Let's assume to evaluate the following queries against an empty repository.

The query below returns no results as expected:

    select ?s {
       ?s ?p ?o .
    }
    group by ?s

However, this query returns one result, which doesn't contain any binding:

    select ?s (max(?o) as ?omax) {
       ?s ?p ?o .
     }
     group by ?s

This is incorrect, the result should be empty, not 1 result with an empty binding.

The cause appears to be that evaluation of MAX (as well as MIN and SAMPLE) hides type errors and replaces them with an empty binding. This is not conform the spec: it should result in a type error (which should then be converted into a  empty result by the processing of the extension).

jeenbroekstra:
I came across this W3C compliance test, which indicates that in fact the current behavior is correct: https://www.w3.org/2009/sparql/docs/tests/summary.html#aggregates-agg-empty-group. The result is a single, empty solution.

I need to dive a bit deeper into the algebra to see what I missed.

jeenbroekstra:
While it is true that the result of a MAX/MIN/SAMPLE aggregate on an empty solution set is a type error, this is hidden by how groups and extensions are evaluated: the type error is converted to an empty binding, which eventually result in a query with a single, empty, solution. This is conform spec and not a bug.

afs:
cc: @manuelfiorelli

There is another step that plays a part in the final outcome. 

When there are zero rows from the query pattern, the grouping is empty and the
aggregate function is not called. The result is zero rows. (FWIW: Jena has got this wrong
more than once over the years.) The SPARQL test can't be right (for a start, it
has the wrong variables).

With the example query, where there are no values for the grouping.

    Group(exprlist, Ω) = { }

then the aggregation becomes:

     Aggregation(exprlist, func, scalarvals, { } )
       =  { (key, F(Ω)) | key → Ω in { } }

which is `{ }` - there is no `key → Ω in { }`. This is where `max` would be
called but if there are zero results from the group step, `max` aggregation isn't called.

But I found a problem in the spec when it is aggregation without a `group by`, and the pattern having no rows.

I've tried to write out the details, having got it wrong before, in 
https://afs.github.io/sparql-agg-group-empty
as I read the spec, with a possible fix for the no `group by` case.

The intuition is 
* group+aggregate: same number of rows as the group
[...]

PR #1979, opened 2020-03-07
GH-1978 cleaned up aggregate processing

jeenbroekstra:
- original behavior was correct: GH-1978 is not a bug
- used opportunity to clean up implementation flow
- added additional unit and compliance tests

GitHub issue resolved: #1978


hmottestad:
lgtm :)

Issue #1960, opened 2020-03-02
Harden SPARQLConnection.commit() against empty transactions


knoan:
**How to Reproduce**

Execute the following snippet against a SPARQL 1.1 Update Endpoint.

```java
final SPARQLRepository repository=new SPARQLRepository("{query/update endpoint}");

repository.init();

try (final RepositoryConnection connection=repository.getConnection()) {

	connection.begin();

	connection.prepareUpdate(
			"prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>\n"
			+"\n"
			+"delete data { rdf:nil rdf:value rdf:nil }"
	).execute();

	connection.commit();

} finally {
	repository.shutDown();
}
```

**Expected Results**

Update silently executed.

**Actual Results**

Exception is possibly thrown.

**Notes**

- updates on SPARQL endpoints are immediately executed rather than queued to the current transaction (see `org.eclipse.rdf4j.repository.sparql.query.SPARQLUpdate:41`)
- on commit an empty update is submitted to the endpoint (see `org.eclipse.rdf4j.repository.sparql.SPARQLConnection:394`) and if the endpoint doesn't gracefully handle the event an exception is thrown
- works as expected when replacing the SPARQL update with the functionally equivalent `connection.remove(RDF.NIL, RDF.VALUE, RDF.NIL)`
- removing transaction begin/commit is not an option in a real world setting, as the offending code is not necessarily aware of the specific type of the target repository

knoan:
P.S. Possibly relevant also for other repository types…

hmottestad:
> Exception is possibly thrown.

Are exceptions always thrown, or just sometimes? Which exceptions have you seen?

knoan:
> Are exceptions always thrown, or just sometimes? Which exceptions have you seen?

It depends on the endpoint: some will simply ignore the empty update, while other will respond with something like `400 Bad Request`, causing an exception to be always thrown.

jeenbroekstra:
Good find, thanks. I think there are two separate (though obviously related) issues here: the first is the fact that an empty commit string should not be transmitted, the second being that a SPARQL update that is part of a transaction should not be immediately executed, but should be appended. I'll separate that second problem out into its own ticket.

PR #1985, opened 2020-03-09
GH-1960 treat commit as no-op if no updates in the txn

jeenbroekstra:
GitHub issue resolved: #1960  

* check if transaction string is not empty before sending update
* added regression test

### DIFF
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -89,7 +89,7 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 
 	private final SPARQLProtocolSession client;
 
-	private StringBuffer sparqlTransaction;
+	private StringBuilder sparqlTransaction;
 
 	private Object transactionLock = new Object();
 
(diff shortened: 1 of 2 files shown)


core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -391,11 +392,13 @@ public void commit() throws RepositoryException {
 		synchronized (transactionLock) {
 			if (isActive()) {
 				synchronized (transactionLock) {
-					SPARQLUpdate transaction = new SPARQLUpdate(client, null, sparqlTransaction.toString());
-					try {
-						transaction.execute();
-					} catch (UpdateExecutionException e) {
-						throw new RepositoryException("error executing transaction", e);
+					if (sparqlTransaction.length() > 0) {
```
hmottestad: Would be nice if you could add a comment here. 

// empty transactions commit as no-op, some SPARQL endpoints don't handle empty transactions

hmottestad:
Looks great, nice and simple fix.

Btw. StringBuilder is usually recommended today over StringBuffer, do we need the synchronization of the StringBuffer or could I make an issue and switch over?

jeenbroekstra:
> Btw. StringBuilder is usually recommended today over StringBuffer, do we need the synchronization of the StringBuffer or could I make an issue and switch over?

I don't think we do, it's a private field in a RepositoryConnection impl, which is itself not thread-safe. I'll take a quick look and just fix it as part of this. It's simple enough that we don't need another issue for it.

Issue #1995, opened 2020-03-12
Adapt release script to push release branch and create PR


jeenbroekstra:
The current release process (and the release script) create a local release branch, but do not push this branch upstream - instead it only pushes the tag. It then pushes a commit directly on master branch to update pom versions after the release has been tagged.

To improve visibility of changes and stability of the master branch, I want to get to a state where we never have to push commits _directly_ to master (instead always using PRs to merge/rebase). This means that the release process needs to be modified a little as well.

Proposed improvement:

1. release branch is pushed upstream
2. after version tag has been created and pushed, pom versions get set for the next iteration _on the release branch_
3. script creates a pull request (using `hub`) to merge the release branch back into master.

jeenbroekstra:
Difficulty: this approach makes creating a merge branch to bring `develop` into sync more difficult.

jeenbroekstra:
Got a solution. Gonna try this out for the 3.1.2 release. If I stuff it up I can easily adjust.

jeenbroekstra:
Needs a few minor adjustments.

PR #1997, opened 2020-03-12
GH-1995 push release branch and automate PR creation

jeenbroekstra:
- push release branch to avoid direct commits on master
- use hub command to create PR for release branch to master
- use hub command to create PR for updating develop branch

GitHub issue resolved: #1995

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -13,6 +13,12 @@ increment_version() {
  echo "$v" | perl -pe s/$rgx.*$'/${1}'`printf %0${#val}s $(($val+1))`/
 }
 
+echo ""
+echo "The release script requires several external command line tools:"
+echo " - git"
+echo " - mvn"
+echo " - hub (https://hub.github.com/)"
+echo " - xmlllint (http://xmlsoft.org/xmllint.html)"
 
 echo ""
 echo "This script will stop if an unhandled error occurs";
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
I will merge this immediately so I can use it for the 3.1.2 release. Only way to really debug, unfortunately, is to actually use it for a release.

hmottestad:
For testing i have previously forked the RDF4J repo.

release.sh
```diff
@@ -153,7 +165,7 @@ git pull
 
 MVN_VERSION_DEVELOP=$(xmllint --xpath "//*[local-name()='project']/*[local-name()='version']/text()" pom.xml)
 
-git checkout master
+git checkout ${BRANCH}
```
hmottestad: Isn’t this the branch we are on already?

release.sh
```diff
@@ -13,6 +13,12 @@ increment_version() {
  echo "$v" | perl -pe s/$rgx.*$'/${1}'`printf %0${#val}s $(($val+1))`/
 }
 
+echo ""
+echo "The release script requires several external command line tools:"
+echo " - git"
+echo " - mvn"
+echo " - hub (https://hub.github.com/)"
```
hmottestad: Can we use the official command line client instead? https://cli.github.com/

hmottestad: Or is the benefit of using hub that we don’t have to log in again?

hmottestad: Btw. We need to test that hub (or whatever we end up with) works as an early step in the script. So it doesn’t crash half way through.

jeenbroekstra: I wasn't aware there was another client, I thought `hub` _is_ the official client. Happy to consider alternatives if they're better.

release.sh
jeenbroekstra: Yes, good point

release.sh
jeenbroekstra: > Or is the benefit of using hub that we don’t have to log in again?

No and actually that brings up a point: you need to set this tool up first before it works.

Issue #2013, opened 2020-03-18
NullPointerException when parsing SPARQL update with multiple expressions one of hich has SERVICE


damyan-ognyanov:
If a SPARQL update have multiple update expressions (delimited with `';'`) and one of those have `SERVICE`, a `NullPointerException` is thrown during parsing.

for instance:
```
PREFIX family: <http://examples.ontotext.com/family#>
DROP ALL ;
INSERT {
    family:Alice family:knows family:Bob .
}
WHERE {
    SERVICE <repository:1> {
        family:Alice family:knows family:Bob .
    }
}
```

we got:
```
java.lang.NullPointerException: null
        at org.eclipse.rdf4j.query.parser.sparql.ast.ASTServiceGraphPattern.getPatternString(ASTServiceGraphPattern.java:67)
        at org.eclipse.rdf4j.query.parser.sparql.TupleExprBuilder.visit(TupleExprBuilder.java:1079)
        at org.eclipse.rdf4j.query.parser.sparql.ast.ASTServiceGraphPattern.jjtAccept(ASTServiceGraphPattern.java:43)
        at org.eclipse.rdf4j.query.parser.sparql.TupleExprBuilder.visit(TupleExprBuilder.java:1046)
        at org.eclipse.rdf4j.query.parser.sparql.ast.ASTGraphPatternGroup.jjtAccept(ASTGraphPatternGroup.java:24)
        at org.eclipse.rdf4j.query.parser.sparql.ast.SimpleNode.childrenAccept(SimpleNode.java:152)
        at org.eclipse.rdf4j.query.parser.sparql.AbstractASTVisitor.visit(AbstractASTVisitor.java:730)
        at org.eclipse.rdf4j.query.parser.sparql.ast.ASTWhereClause.jjtAccept(ASTWhereClause.java:24)
        at org.eclipse.rdf4j.query.parser.sparql.UpdateExprBuilder.visit(UpdateExprBuilder.java:298)
        at org.eclipse.rdf4j.query.parser.sparql.UpdateExprBuilder.visit(UpdateExprBuilder.java:66)
        at org.eclipse.rdf4j.query.parser.sparql.ast.ASTModify.jjtAccept(ASTModify.java:25)
        at org.eclipse.rdf4j.query.parser.sparql.SPARQLParser.parseUpdate(SPARQLParser.java:118)
        at org.eclipse.rdf4j.query.parser.QueryParserUtil.parseUpdate(QueryParserUtil.java:76)
```

PR #2014, opened 2020-03-18
Fixes GH-2013 by traversing parents until sourceString is found

damyan-ognyanov:
Fixes GH-2013 by traversing parents until sourceString is found

GitHub issue resolved: #2013  

* added a simplified test case reproducing the issue
* fix in `ASTServiceGraphPattern` that traverse the node parents untl a `sourceString` is found

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/ASTServiceGraphPattern.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/ASTServiceGraphPattern.java
@@ -57,11 +57,8 @@ public class ASTServiceGraphPattern extends SimpleNode {
 	public String getPatternString() {
 
 		if (patternString == null) {
-			ASTOperationContainer parentContainer = (ASTOperationContainer) getParentContainer(this);
-
-			String sourceString = parentContainer.getSourceString();
-
 			// snip away line until begin token line position
+			String sourceString = getSourceString();
 			String substring = sourceString;
 			for (int i = 1; i < getBeginTokenLinePos(); i++) {
 				substring = substring.substring(substring.indexOf('\n') + 1);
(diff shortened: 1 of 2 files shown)


core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/ASTServiceGraphPattern.java
```diff
@@ -84,6 +81,24 @@ public String getPatternString() {
 		return patternString;
 	}
 
+	private String getSourceString() {
+		Node theParent = getParentContainer(this);
+		String sourceString = null;
+		if (theParent instanceof ASTOperationContainer)
+			sourceString = ((ASTOperationContainer) theParent).getSourceString();
```
jeenbroekstra: Editorial: could you add open and close brackets here and in other places where you have if/else constructions?

damyan-ognyanov: done

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/ASTServiceGraphPattern.java
```diff
@@ -84,6 +81,27 @@ public String getPatternString() {
...
+		while (sourceString == null && theParent != null) {
+			theParent = theParent.jjtGetParent();
+			if (theParent == null) {
+				break;
+			}
+			if (theParent instanceof ASTUpdateSequence) {
+				sourceString = ((ASTUpdateSequence) theParent).getSourceString();
+			}
+		}
+		return sourceString;
```
hmottestad: I was wondering if maybe there could be a null check here, so that any exception is thrown when looking for the source string instead of when using the resulting string. Maybe `return Objects.requireNonNull(sourceString);`

Unless of course there is a case where returning null from getSourceString() is acceptable.

damyan-ognyanov: Thought of that for a while, but decided that if we did not found the `sourceString` at this point, something else was really messed up, so throwing a custom exception early, would not help much here since we do not have a resonable alternative to continue processing update further - a NPE will still be thrown in the code that uses it ...

jeenbroekstra:
Thanks @damyan-ognyanov . One more thing: could you please [squash  your commits](https://rdf4j.org/documentation/developer/squashing)?

damyan-ognyanov:
@jeenbroekstra squashed commits but renaming it to follow the naming convention is a bit tricky since we already have this pull request refering to it...

jeenbroekstra:
That's all good, thanks, the branch naming thing was more for future reference.

Issue #1900, opened 2020-02-03
SHACL validation report remote repository


barthanssens:
See forum https://groups.google.com/forum/#!topic/rdf4j-users/VDlWNjQ5eHw

On a local repository, one can easily retrieve the ShaclSail validation report when a SHACL validation exception is thrown.

It would be nice to have an easy way to get/parse the same validation report in a client application, while the validation is done on a remote repository  / server (HTTPRepository)

hmottestad:
Should we go for a streaming approach so that we can handle large validation reports?

barthanssens:
Would definitely be a plus. 

E.g. in the context of the EU ISA2 programme, there is an "interop testbed" for validating with SHACL, and the EU open data portal has a similar "metadata quality service" for validating large(r) files (100 K - 1 M triples) harvested from national/regional data portals (PSI/Open data Directive).

A similar approach might be useful for Single Digital Gateway, Multi-Modal Travel Information Services, .... so there are use cases :-)

jeenbroekstra:
Perhaps this should be a separate issue, but: I have been thinking if maybe we should decouple the reporting from the exception handling. I can imagine use cases where people might want to use validation and get a report about what's invalid, but still commit the data.

hmottestad:
There is the revalidate method which returns a validation report. It’s a bit hidden though, since it’s on the shacl sail connection and it’s common to wrap the sail in the sail repository. 

Btw. The assumption within the shaclsail is that the underlying store is valid. The fast and efficient validation only works because it can assume that there is no need to check the data already in the store. 

We could create a model validator instead. Preferably based on a new SHACL engine that generates SPARQL queries. That is what Stardog does, which makes it fast at validating large datasets but slow at validating changes.

jeenbroekstra:
> We could create a model validator instead. Preferably based on a new SHACL engine that generates SPARQL queries. That is what Stardog does, which makes it fast at validating large datasets but slow at validating changes.

Wouldn't that mean that we'd essentially have two completely separate SHACL implementations? It sounds like an awful maintenance burden to keep feature-parity between both.

Anyway, this is off-topic for this issue, sorry for hijacking. I'll do some thinking and open a new ticket to discuss further.

hmottestad:
I've been looking at this one today a bit, but the server is old style spring magic. Been trying to change things to try to see what happens, thought I had found a short-term fix but then it only works if you speak to the server directly and not when you use the workbench. 

[...]

Issue #2011, opened 2020-03-18
Consolidate our git branch/merge strategy


jeenbroekstra:
Three different merge strategies for branches / pull requests are currently available on the RDF4J repository:

![Screenshot from 2020-03-18 13-43-05](https://user-images.githubusercontent.com/728776/76920253-1d052300-691f-11ea-8587-436b49bb13ae.png)

Unfortunately, we are currently not consistent among ourselves what merge strategy should be used in what situation, and this makes our commit history a little messy. 

In particular, I've observed several patterns in our commit history. For PRs merged by myself , using rebase merge results in a clean linear history on the master branch, for example:

![Screenshot from 2020-03-18 13-55-30](https://user-images.githubusercontent.com/728776/76920675-25aa2900-6920-11ea-868e-98b7afd6aaef.png)

While cleanly linear, a downside is that the link with the original PR is lost (though that can be retrieved by visiting the issue on Github, if necessary).

However, for PRs by several others, 'create a merge commit' is used, which results in pair of commits (one of them a merge commit) on the master branch history, instead of a single commit. For example:

![Screenshot from 2020-03-18 13-57-46](https://user-images.githubusercontent.com/728776/76920783-76ba1d00-6920-11ea-84cc-1104530cf3e1.png)

In this approach the link with the original PR is retained in the commit history itself, but it does make the history 'noisy' and less straightforward to read. 

## Possible approaches to more consistency

While either approach to merging is workable in and of itself, having both used interchangeably breeds confusion, especially in trying to figure out where things came from, and what is included on which branch. I see two options to resolve this:

### Option 1: use 'create a merge commit' exclusively 

We could accept that a completely clean/linear history is not worth the hassle and switch back to using merge commit exclusively to merge PRs. 

Consequences of this approach would be that we disable both the 'squash and merge' as well as the 'rebase' option. It means a slightly messier history, but provided we still squash PRs before merging, it's manageable. 

This raises a potential problem: even if we among ourselves agree to "be good" and properly squash, third party contributors could be unaware of this and use merge commits on their branch before putting up a PR. I'd hate to get in a situation where someone has gone through the effort of providing a patch and we can't accept it because their PR is hard to reconcile. 

### Option 2: use 'rebase and merge' exclusively for feature PRs

[...]

Issue #1502, opened 2019-08-14
rdf4j server / workbench self-contained executable 


jeenbroekstra:
Instead of / in addition to the WAR file and the docker image, we should look into distributing Rdf4j Server / Workbench as a self-contained executable that can be downloaded and immediately spun up without having to install a servlet container separately.

barthanssens:
Several options I guess, e.g. 

- try to use Spring Boot to generate a fat jar
- or replace the current Spring thingies with e.g. something implementing Eclipse MicroProfile
- or Quarkus / Micronaut
- or use GraalVM to build a native app (probably not the best option, native app would be more useful for the console)

hmottestad:
Upvote for spring boot 2

jeenbroekstra:
Fwiw I've started looking at building a spring boot app in my spare time. My wife hates me but it might come together.

hmottestad:
I have a decent amount of experience with spring boot, if you have any questions :)

jeenbroekstra:
Ah yes, this dropped by the wayside a bit I'm afraid. I'll see where I left off.

jeenbroekstra:
Related: #2423

barthanssens:
(yet) another somewhat related idea from the gitter chat: using jetty instead of tomcat for the war (https://www.baeldung.com/deploy-to-jetty)

naturzukunft:
> Fwiw I've started looking at building a spring boot app in my spare time. My wife hates me but it might come together.

Jeen, are you still working on it?

jeenbroekstra:
> > Fwiw I've started looking at building a spring boot app in my spare time. My wife hates me but it might come together.
> 
> Jeen, are you still working on it?

Not really, as with so many good intentions, this one got overrun by the reality of there only being 24 hours in a day...

naturzukunft:
Ha, I know this problem. I'm itching to do it, but there's actually no time. However, would there be tests available that test the REST endpoints?

jeenbroekstra:
> Ha, I know this problem. I'm itching to do it, but there's actually no time. However, would there be tests available that test the REST endpoints?

There are some controller tests available in the rdf4j-http-server and rdf4j-http-server-spring modules, but the more comprehensive testing is probably done in the compliance/integration tests in rdfj4-repository-compliance, which test the HTTPRepository and the SPARQLRepository against a running RDF4J Server.

benherber:
[...]

PR #2025, opened 2020-03-20
GH-1900 Server responds with SHACL validation report if validation failed

hmottestad:
GitHub issue resolved: #1900  

* Use the current exception handling to extract the validation report and output it as NQUADS in the response
* Extend the error handling in the client to pick up this error and create an appropriate exception (preferably a ShaclSailValidationException, but that needs more work)
* Detect this exception in the Workbench and print a nice turtle report to the user


hmottestad:
This is a fairly hacky solution that gives us a fix that we can remain compatible with. By idea is that in the future we can use either a header or the http status line to inform the client that this should be treated as a ShaclValidationException. At the moment the data is just available as the message.

hmottestad:
This is how it looks in work bench at the moment.

Screenshot 2020-03-20 at 11 39 54

hmottestad:
Would be nice to do something about the exceptions being thrown though, so people don't rely on how the exception looks today....since that's something we would like to change in the future.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -1041,6 +1044,10 @@ protected HttpResponse execute(HttpUriRequest method) throws IOException, RDF4JE
 						throw new MalformedQueryException(errInfo.getErrorMessage());
 					} else if (errInfo.getErrorType() == ErrorType.UNSUPPORTED_QUERY_LANGUAGE) {
 						throw new UnsupportedQueryLanguageException(errInfo.getErrorMessage());
+					} else if (exceptionIs(response, "ShaclSailValidationException")) {
+						throw new RepositoryException(new RemoteShaclSailValidationException(
+								new StringReader(errInfo.toString()), "", RDFFormat.NQUADS));
+
```
hmottestad: I'm trying out throwing a custom exception based based on if the correct header is set to indicate that this is a shacl violation.

hmottestad: I would prefer to throw the real ShaclSailValidationException, but doing so requires a dependency on the shacl module which causes a dependency cycle. I need to extract some interfaces instead....which maven model should I use for those interfaces do you think @jeenbroekstra ?

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -1055,6 +1062,21 @@ protected HttpResponse execute(HttpUriRequest method) throws IOException, RDF4JE
 		}
 	}
 
+	private boolean exceptionIs(HttpResponse response, String exceptionName) {
+		Header[] headers = response.getHeaders("X-Eclipse-RDF4J-Exception");
```
hmottestad: Went with `X-Eclipse-RDF4J-Exception´as the header, and it should contain a class name.

hmottestad:
This the output in the Workbench that I'm proposing now.

Screenshot 2020-03-20 at 16 05 15

hmottestad:
Got my issues numbers crossed, correct one is #1900

hmottestad:
Will need to look into renaming the branch and the commits

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
```diff
@@ -127,7 +134,27 @@ private void handleRequest(final HttpServletRequest req, final HttpServletRespon
...
+
+				writer.println("SHACL validation failed with the following report:\n");
+				WriterConfig writerConfig = new WriterConfig();
+				writerConfig.set(BasicWriterSettings.PRETTY_PRINT, true);
+				writerConfig.set(BasicWriterSettings.INLINE_BLANK_NODES, true);
+				Rio.write(model, writer, RDFFormat.TURTLE, writerConfig);
+
+				writer.println(
+						"\n" +
+								"THIS ERROR MESSAGE IS EXPERIMENTAL AND IS SUBJECT TO CHANGE - " +
```
jeenbroekstra: Why are you sending the validation report with this tagged on, and as plain text? Wouldn't it make more sense to send just the turtle data, as text/turtle?

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
```diff
@@ -127,7 +134,27 @@ private void handleRequest(final HttpServletRequest req, final HttpServletRespon
 		} catch (UnauthorizedException e) {
 			handleUnauthorizedException(req, resp);
 		} catch (RepositoryConfigException | RepositoryException e) {
-			throw new ServletException(e);
+			if (e.getCause() instanceof RemoteShaclSailValidationException) {
+				Model model = ((RemoteShaclSailValidationException) e.getCause()).validationReportAsModel();
+
+				resp.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
jeenbroekstra: Perhaps a 409 (conflict) is more specific.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: If you want a place to have a shared exception class, rdf4j-util might be a good spot. Create a package org.eclipse.rdf4j.common.exception perhaps?

jeenbroekstra:
Also: could you rename your branch please? Thanks.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: If we must use a custom header, this seems like a good name for it. I'd remove the `X-` prefix though. That is no longer recommend practice for custom headers. See https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions

jeenbroekstra: Alternatively: instead of using a custom header for this (which on some servers/firewalls is known to cause problems), we could also just let the client look at the response status code, and if it's a 409 conflict, see if there is a payload with a validation report in it.

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
hmottestad: It’s just a simple way to display it to the user. In the future someone could extend the user interface to show the error properly, at which point it could be changed to work some other way.

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
hmottestad: 409 seems good. Hadn’t heard of that one before.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -1041,6 +1044,10 @@ protected HttpResponse execute(HttpUriRequest method) throws IOException, RDF4JE
 						throw new MalformedQueryException(errInfo.getErrorMessage());
 					} else if (errInfo.getErrorType() == ErrorType.UNSUPPORTED_QUERY_LANGUAGE) {
 						throw new UnsupportedQueryLanguageException(errInfo.getErrorMessage());
+					} else if (exceptionIs(response, "ShaclSailValidationException")) {
```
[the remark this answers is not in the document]
hmottestad: I can try to move it there and see if it works.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
hmottestad: We shouldn’t rely on having to look at the content, since that will stop us from doing a streaming based approach down the line. Custom status line was my first idea until I realized how hard it would be to do with spring.

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
hmottestad: To explain further. This is actually a view layer issue...we need to return this error to the view layer in some way, but I've not started looking at the view layer of the workbench to figure out how to interact with it. Returning the text/plain here is just bypassing the view layer, returning text/turtle is not correct because the view layer is supposed to be html. Text/plain is just a temporary hack to make the error visible to the user. And by temporary I mean, probably permanent unless someone who built the workbench can help out.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/ModelInterface.java
```diff
@@ -0,0 +1,25 @@
...
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+
+package org.eclipse.rdf4j.http.client;
+
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.impl.DynamicModelFactory;
+
+public interface ModelInterface {
```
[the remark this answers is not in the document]
hmottestad: This is part of the way the shacl exception is built in the shacl sail. It will be moved into the common module. When that's done we can revisit how it's implemented.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RemoteValidation.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.rio.RDFFormat;
+import org.eclipse.rdf4j.rio.Rio;
+
+import java.io.IOException;
+import java.io.StringReader;
+
+public class RemoteValidation implements ModelInterface {
```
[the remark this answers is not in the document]
hmottestad: Intention is to support streaming validation reports down the line. Hopefully I can build that as part of this branch, so that I can get the requirements down from the get go.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: Fwiw I'm fine with going with the header for now but I don't see how looking at the content would stop us streaming: based on content type you can just immediately parse in the background, this is no different from how query results are processed. Also: do these reports really get so large that this is a serious consideration?

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
jeenbroekstra: I, um, did. But it was a while back. Let me have a think about this, I don't want to make REST API decisions based on shortcomings of the client tool.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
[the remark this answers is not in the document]
hmottestad: Took a look now. This execute(....) method is what is used by RDF4JProtocolSession for its requests. RDF4JProtocolSession extends SPARQLProtocolSession. We should probably extract an even lower class that only implements execute(....) so that we can be sure that no one above can bypass it. This isn't in scope for this PR though.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RemoteValidation.java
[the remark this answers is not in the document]
jeenbroekstra: Can we make it package private? I'm a little wary of introducing public interfaces and classes for this.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
hmottestad: Can you point me to where this is handled for the query results?

My fear is that someone will modify a shape and that will cause a lot of data to become invalid. I've done this myself a few times while developing validation rules. It's usually not intentional, but it would be a pity if we bring down the server because we run out of memory.

jeenbroekstra: > Can you point me to where this is handled for the query results?

Sure thing. For construct queries, for example:
- server-side streaming: `org.eclipse.rdf4j.http.server.repository.GraphQueryResultView` (the `renderInternal` method)
- client-side streaming: `SPARQLProtocolSession.getRDFBackground`.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/ModelInterface.java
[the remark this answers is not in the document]
jeenbroekstra: I would _really_ prefer keeping this lean, and not introducing additional public interfaces unless there is a clear need for them. Unless I completely misunderstand this does not look like code that an end user should ever want to touch.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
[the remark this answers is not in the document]
jeenbroekstra: Fair enough, let's leave for now. It's not in the way.

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/proxy/WorkbenchServlet.java
jeenbroekstra: Hang on, apologies, I completely misread this - for some reason I assumed I was looking at the RDF4J Server code, not the workbench. So yes - this makes much more sense now!

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
hmottestad: SPARQLProtocolSession.getRDFBackground is based on headers and not content. It gets the content type from the headers and then passes the content of the response to the parser. It never checks anything from the content.

We could extend the content type, essentially like how application/json is extended to application/json+ld we could extend application/n-quads to be application/n-quads+shacl-validation-report or application/json+ld+shacl-validation-report.

I still think I prefer a custom header. I'll post this to the SHACL forum, maybe someone has come up with something already that we can consider implementing too.

jeenbroekstra: > SPARQLProtocolSession.getRDFBackground is based on headers and not content. It gets the content type from the headers and then passes the content of the response to the parser. It never checks anything from the content.

Yes, I know, that's what I meant. Sorry if this wasn't clear, my point was that based purely on the status code and the response content type you can process the response body (streaming). No need for an additional custom header.

hmottestad: Andy (Jena dev) proposed the same to my email on the SHACL public mailing list. application/shacl-validation-report+n-quads. I’m starting to appreciate that. Will see what others reply before we decide. Would be nice if at least Jena and RDF4J did it the same way.

jeenbroekstra: I hadn't even thought of a custom content type - I was just thinking of using plain `text/turtle` (or `application/n-quads`). That's all you need to parse the incoming data. Figuring out that it is a SHACL validation report can be done on the parsed model.

jeenbroekstra:
Content-type approach looks good! Fine with that for a first time around.

As for the exception class and where it should go: I just realized that you can't put it in rdf4j-util _and_ have it extend RDF4JException, because that is in rdf4j-model. Which is stupid, it shouldn't be there, it should be in util (which contains the common package, which is supposed to contain all this kind of generally reusable stuff). But, you know, legacy.  Stick it in rdf4j-model for now and we'll clean up the exception hierarchy and which module contains what as a separate issue.

hmottestad:
Screenshot 2020-03-27 at 14 18 26

While working on this I've created a separate module to see what I actually need. I've created an interface for the actual exception, and then implemented two classes for either of remote or local shacl exception. 

I'm thinking of removing the model interface.

I've also changed up how the java based report looks. To avoid dependencies. Might also turn that into interfaces instead.

This is a sort of breaking change. It's fairly small though. Users will have to import a different package, but can use the same class name for handling the shacl exception. Those relying on the java based report will have to make more changes.

What are your thoughts @jeenbroekstra ?

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/shacl/RemoteShaclSailValidationException.java
```diff
@@ -0,0 +1,38 @@
...
+	 */
+	@Override
+	public Model validationReportAsModel() {
+		Model model = remoteValidation.asModel();
+		model.setNamespace(SHACL.PREFIX, SHACL.NAMESPACE);
+		return model;
+
+	}
+
+}
```
hmottestad: @jeenbroekstra 

This is the exception that gets thrown by the client when the server responds with a 409 and the validation report. The ValidationException interface is now what's common between the regular ShaclSailValidationException and this exception. Users would then need to migrate to "catch" this interface instead of the ShaclSailValidationException if they only retrieve the model and want their code to be portable between a local ShaclSail and remote one.

Things still left to consider:
 - Should this exception extend the RDF4JException (which is deprecated) or SailException (which is what the ShaclSailValidationException does? Not sure why the RDF4JException isn't an interface instead of an abstract class.
 - Should we mark this and the remote validation exception as experimental, since this isn't the final location of the interface at least?
 - How to support streaming validation reports?
 - What are we going to do with the current java based validation report? Maybe deprecate it so we can move it and generalise it?

jeenbroekstra: My take on things: 
- RDF4JException isn't deprecated. It inherits from the old OpenRDFException (which _is_ deprecated, and is a class because we made a design decision that all RDF4J exceptions are unchecked (so it needs to extend RuntimeException). Actually SailException itself extends RDF4JException as well (_all_ RDF4J exceptions do).  I still find the naming not quite right. The word "SAIL" shouldn't appear on the client side at all. Sails are an internal server-side matter. So the server-side version of the ValidationException should be called ShaclValidationException or something like that, and probably extend SailException. But the client-side version should be called something like RemoteShaclValidationException or something like that, but _not_ have Sail in its name, and not extend SailException
- marking things experimental is always good for a new feature - it gives us leeway :)
- YAGNI (or at the very least, not in scope for this iteration, though I'm fairly confident we can do something along these lines when it becomes necessary)
- is this the report functionality as previously released you mean? If we can deprecate it, keep it functional but then copy and migrate, that'd be ideal.

core/http/client/pom.xml
```diff
@@ -60,6 +60,11 @@
 			<groupId>commons-codec</groupId>
 			<artifactId>commons-codec</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-sail-api</artifactId>
```
jeenbroekstra: You can not let the http client depend on the sail api.

hmottestad: I think this is only due to the sail exception, so should be fixable if we change the exception to the rdf4jexception.

hmottestad:
I’m wondering if this might be ready now. I’ve squashed some commits. Still need to rename the branch before I merge it.

jeenbroekstra:
> Still need to rename the branch before I merge it.

If it's easier, you can also just edit the commit message when you click merge:

Screenshot from 2020-03-31 09-37-50

Just change 'eclipse/issues/GH-1502_' to 'eclipse/GH-1502-' and then confirm.

The main reason I'm obsessed with the branch name is that I'm trying to make sure this commit message for the merge commit is easy to read and (in the github UI) actually makes the issue number a clickable link. To do that, it needs to be followed by a dash rather than an underscore.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -102,6 +86,26 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+
+import static org.eclipse.rdf4j.http.protocol.Protocol.ACCEPT_PARAM_NAME;
```
jeenbroekstra: Don't bother fixing it for this PR, but: could you fix your formatter to not reorganize imports in this fashion? Most of our code follows the Eclipse convention to have static imports first, then normal imports organized with anything from java.* at the top, then javax.*, then org.*, then com.*. 

Reason I ask is that I've had to resolve some merge conflicts in the past that were caused by imports being moved around, and it's a hassle.

This was never made explicit because in the past everyone on the project used Eclipse, so it was just the implicit default. Perhaps we should document this now that we have a more diverse toolset.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -1055,6 +1066,62 @@ protected HttpResponse execute(HttpUriRequest method) throws IOException, RDF4JE
 		}
 	}
 
+	static RDFFormat getContentTypeSerialisation(HttpResponse response) {
```
jeenbroekstra: Why is this a static? You're not using it in a static way, as far as I can tell, nor are there any other existing static methods in the ProtocolSession class. It think it can just be a private instance method.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -1055,6 +1066,62 @@ protected HttpResponse execute(HttpUriRequest method) throws IOException, RDF4JE
...
+						serialisation.append("+");
+					}
+				}
+
+				logger.debug("SHACL validation report is serialised as: " + serialisation.toString());
+
+				Optional<RDFFormat> rdfFormat = RDFFormat.matchMIMEType(serialisation.toString(), rdfFormats);
+
+				if (rdfFormat.isPresent())
+					return rdfFormat.get();
```
jeenbroekstra: Missing curly braces.

jeenbroekstra:
@hmottestad  as a heads up: after thinking about my nagging you about squashing, and branch names, and all of that, I've reopened the discussion around merge strategies (see #2011 ) with a compromise that I hope makes both our lives easier. The TL;DR is to go back to preferring the use of  "Squash and merge" for all feature PRs, and optionally allow rebase in cases where you explicitly want preserve multiple commits. 

This PR is probably a good example of where using 'rebase and merge' instead of 'squash and merge' makes sense: there's a limited number of commits, they're all meaningfully described, and because one of them is for a different issue than the main issue being solved, it makes sense to keep them separate.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
hmottestad: Doesn’t everyone almost everyone use IntelliJ now? Since this is the default with IntelliJ. Btw, I’m not a fan of having to adjust my editor for every project, there was once an attempt at making a standardized editor config, maybe we should look at that. Also I think the maven formatter plugin should handle this and if it doesn’t then we need to just live with the fact that the imports are going to change. I’ll try not to optimize imports for files I don’t touch. And also IntelliJ can fix import conflicts for you very easily.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
hmottestad: Doesn’t rely on any instance data. This means it can be static, and if it can be static it should be static. Also much easier to test when it’s static.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
hmottestad: Will fix.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: > Doesn’t everyone almost everyone use IntelliJ now?

No.

> Since this is the default with IntelliJ. Btw, I’m not a fan of having to adjust my editor for every project, there was once an attempt at making a standardized editor config, maybe we should look at that.

To some extent this is unavoidable in an open-source world I'm afraid. Every project has different conventions. 

Then again this one is not a huge deal. I'll cop the occasional merge conflict fallout (it doesn't happen very often anyway - mostly it can resolve it fine, and even if it doesn't it's easy to fix in the IDE).

> Also I think the maven formatter plugin should handle this and if it doesn’t then we need to just live with the fact that the imports are going to change. I’ll try not to optimize imports for files I don’t touch.

I'll take a look at the formatter plugin, see if that's possible. Like I said earlier, we haven't really enforced this guideline before, and I'm not even sure what Netbeans (which Bart uses) does, so don't worry too much about it - I merely asked on the off chance it was a simple tweak on your end.

Issue #1873, opened 2020-01-23
Incorrect/inconsistent cardinality calculation


pulquero:
EvaluationStatistics.getCardinality() should use Set<Var> vars = pattern.getVars(new HashSet<>()) else
?s ?p ?o and ?s ?s ?s are considered to have the same cardinality.

hmottestad:
@pulquero can you provide a link to exact line in the code in github?

pulquero:
Sry, it's exactly in the fed version not the main one. 

https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java#L84

hmottestad:
Thanks

reeshabhkumarranjan:
Can I work on this issue?

barthanssens:
Sure, it's an open source project, and any help is welcome and much appreciated.

If this is your first contribution to RDF4J, there is one small but necessary thing you have to do: register for a (free) Eclipse account and sign a form

See "Legal stuff" in https://github.com/eclipse/rdf4j-doc/blob/master/.github/CONTRIBUTING.md

hmottestad:
@aschwarte10 is this class `org.eclipse.rdf4j.sail.federation.optimizers.EvaluationStatistics` used by FedX?

reeshabhkumarranjan:
So as per I understood the code, and the solution suggested in this issue, I just need to make sure that only distinct elements are in the `vars` list in this [line](https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java#L84), right?

hmottestad:
> So as per I understood the code, and the solution suggested in this issue, I just need to make sure that only distinct elements are in the `vars` list in this [line](https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java#L84), right?

I believe that is what is required. A test would also be nice.

reeshabhkumarranjan:
Alright, I will link a pull-request for this shortly.

reeshabhkumarranjan:
[...]

PR #2029, opened 2020-03-21
GH-1873 fixing incorrect cardinality calculation

reeshabhkumarranjan:
GitHub issue resolved: #1873 

* reassigned the `vars` list (list -> set -> list) so that it contains only distinct elements
* added a new package `optimizers` under `core/sail/federation/src/test/java/org/eclipse/rdf4j/sail/federation` as suggested by @hmottestad 
* added tests in a new file, `CardinalityTest.java` in the package `optimizers`, also added the header comment in the new file

### DIFF
--- a/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java
+++ b/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java
@@ -84,9 +84,7 @@ public class EvaluationStatistics {
 		}
 
 		protected double getCardinality(StatementPattern pattern) {
-			List<Var> vars = pattern.getVarList();
-			Set<Var> varSet = new HashSet<Var>(vars);
-			vars = new ArrayList<Var>(varSet);
+			Set<Var> vars = new HashSet<>(pattern.getVarList());
 			int constantVarCount = countConstantVars(vars);
 			double unboundVarFactor = (double) (vars.size() - constantVarCount) / vars.size();
 			return Math.pow(1000.0, unboundVarFactor);
--- a/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java
+++ b/core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java
@@ -7,7 +7,10 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.federation.optimizers;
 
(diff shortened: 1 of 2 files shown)


core/sail/federation/src/test/java/org/eclipse/rdf4j/sail/federation/optimizers/CardinalityTest.java
```diff
@@ -0,0 +1,66 @@
+/*******************************************************************************
+ * Copyright (c) ${year} Eclipse RDF4J contributors.
```
jeenbroekstra: Looks as if your editor forgot to replace `{$year}` with the actual year!

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java
```diff
@@ -82,6 +85,8 @@ public void meet(Slice slice) {
 
 		protected double getCardinality(StatementPattern pattern) {
 			List<Var> vars = pattern.getVarList();
+			Set<Var> varSet = new HashSet<Var>(vars);
+			vars = new ArrayList<Var>(varSet);
```
jeenbroekstra: I think you can also just do:

    Set<Var> vars = new HashSet<>(pattern.getVarList());

I don't think the rest of the method requires that `vars` is an ArrayList (or even a list), so no need to create an arraylist back out of the set. Saves having to copy the elements over twice.

core/sail/federation/src/test/java/org/eclipse/rdf4j/sail/federation/optimizers/CardinalityTest.java
reeshabhkumarranjan: I saw other files, they have "2015" in place of `${year}`. Should I do the same or should I put "2020"?

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/EvaluationStatistics.java
reeshabhkumarranjan: Alright! I will make this change and update this pull request.

core/sail/federation/src/test/java/org/eclipse/rdf4j/sail/federation/optimizers/CardinalityTest.java
reeshabhkumarranjan: I replaced it with `2020` in the latest commit.

Issue #2030, opened 2020-03-22
Add getStatements method to Model class as a light-weight alternative to filter


jeenbroekstra:
The `Model.filter` method is guaranteed to be backed by the original model. This is useful for live views on the data where a Model is being manipulated as well as read. However, in many use cases where a user simples wishes to iterate over the result, it can be an unnecessary performance burden.

We will add a new method `getStatements(subject, predicate, object, context)` to the model class that returns an `Iterator<Statement>`, with the typical semantics of an iterator over a collection (including optionally throwing a `ConcurrentModificationException` if the model is manipulated while the iterator is active). This will allow certain implementations of the Model API to provide a  more performant way to iterate over (subsets of) the model.

PR #2034, opened 2020-03-22
GH-2030 add getStatements method to Model

jeenbroekstra:
- allows optimized retrieval of statements in model implementations
- default implementation just reuses filter internally
- added some basic regression tests

GitHub issue resolved: #2030

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
@@ -8,6 +8,7 @@
 package org.eclipse.rdf4j.model;
 
 import java.io.Serializable;
+import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
> I know I’ve advocated for Select, but now I got to wondering if maybe we should call it getStatements instead, to not confuse users who might think of sparql select and also to reuse the design from the repository interface.

It's a fair point. I still personally prefer `select` because it is more succinct, but I may be unconsciously influenced by Ruby conventions here (which has a `select` method on arrays). But I think you're right that `getStatements` will be easier to discover and recognize. I'll change it.

jeenbroekstra:
Given you've both approved and this is simply a renaming change, I'll go ahead and merge.

PR #2036, opened 2020-03-23
GH-1784 make druid-hll "provided" because I didn't do a CQ for transitive dependencies

hmottestad:
GitHub issue resolved: #1784  

* make druid-hll "provided" because I didn't do a CQ for transitive dependencies
* 
*

### DIFF
--- a/core/sail/elasticsearch-store/pom.xml
+++ b/core/sail/elasticsearch-store/pom.xml
@@ -88,12 +88,11 @@
 			<artifactId>rdf4j-sail-extensible-store</artifactId>
 			<version>${project.version}</version>
 		</dependency>
-        <dependency>
+		<dependency>
 			<groupId>${project.groupId}</groupId>
-            <artifactId>rdf4j-sail-inferencer</artifactId>
+			<artifactId>rdf4j-sail-inferencer</artifactId>
 			<version>${project.version}</version>
-            <scope>test</scope>
-        </dependency>
-
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 </project>
--- a/core/sail/elasticsearch-store/pom.xml
+++ b/core/sail/elasticsearch-store/pom.xml
@@ -94,34 +94,6 @@
 			<version>${project.version}</version>
             <scope>test</scope>
         </dependency>
-		<dependency>
-			<groupId>org.apache.druid</groupId>
-			<artifactId>druid-hll</artifactId>
-			<version>0.16.1-incubating</version>
-			<exclusions>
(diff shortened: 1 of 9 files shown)


hmottestad:
I'm already looking at alternatives to druid-hll, with fewer dependencies. So instead of logging new CQs I think I will leave as is and try to integrate a new HLL implementation instead.

If we get close to 4.2.0 I can switch the ElasticsearchStore to use the constant evaluation stats instead.

core/sail/elasticsearch-store/pom.xml
```diff
@@ -94,5 +94,34 @@
 			<version>${project.version}</version>
             <scope>test</scope>
         </dependency>
-    </dependencies>
+		<dependency>
```
jeenbroekstra: I think your xml indentation settings might be out of date (or perhaps it's the previous edit on this file that went wrong). See https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#xml-indentation . There's a maven command to fix it for you - though it might be a good idea to adjust your editor settings as well.

core/sail/extensible-store/pom.xml
```diff
@@ -46,7 +46,9 @@
                     <groupId>org.asynchttpclient</groupId>
                 </exclusion>
             </exclusions>
-        </dependency>
+			<scope>provided</scope>
```
jeenbroekstra: As a temporary fix this is fine, but perhaps it should be optional instead (see https://stackoverflow.com/questions/40393098/when-to-use-optionaltrue-optional-and-when-to-use-scopeprovided-scope) - the difference is mostly semantic, but optional is more intended to say "it's up to you as a user to decide if you need/want this thing", whereas provided scope is more for things that are just _expected_ to be present in the runtime environment (e.g. for a webapp, you could have the servlet-api jar as provided). 

Also: I haven't looked at the code, but can the extensible store still operate if this library is _not_ available? Or will it just crash with a ClassNotFoundException? Don't bother if this is a massive effort but if we can do a simple check and have a fallback that (no matter how ineffecient) at least _works_, it makes the "optional" story a bit more convincing :)

hmottestad:
Everything should run without the HLL dependency now. Also formatted the xml tiles.

core/sail/extensible-store/pom.xml
hmottestad: Changed it to optional.

hmottestad:
I would like to get this in for the 3.2.0 release. Could you approve these changes if you agree with them

Issue #2015, opened 2020-03-18
Update to Elasticsearch 6.8.x and Lucene to 7.7.x and Solr 7.7.x


hmottestad:
Elasticsearch 6.8.7 is the latests version compatible with our current version 6.5.4.

Works-with:
- [x] CQ22007 elasticsearch-transport 6.8.7
- [x] CQ22008 elasticsearch-testframework 6.8.7
- [x] CQ22009 lucene-backward-codecs 7.7.2
- [x] CQ22010 lucene-spatial 7.7.2
- [x] CQ22011 solr-core 7.7.2

Normal:
- [x] CQ22012 solr-solrj 7.7.2
  - [x] Apache Commons IO 2.6 (existing CQ18796, PB not needed anymore under new eclipse rules)
  - [x] CQ22025 Apache Commons Math3 3.6.1
  - [x] CQ11713 Apache HTTP Client 4.5.10
  - [x] CQ11718 Apache HTTP Mime 4.5.10
  - [x] Apache Zookeeper 3.4.14 (existing CQ6031, PB not needed anymore under new eclipse rules)
  - [x] CQ16907 Woodstox Stax2-api 3.1.4
  - [x] CQ16906 Woodstox Core ASL 4.4.1
  - [x] CQ22028 Noggit 0.8
  - [x] CQ10578 JCL over Slf4j 1.7.26
- [x] lucene-core 7.7.2 (existing CQ21945, PB not needed anymore under new eclipse rules)
- [x] CQ22013 lucene-analyzers-common 7.7.2
- [x] CQ22014 lucene-highlighter 7.7.2
  - [x] CQ22023 lucene join 7.7.2
  - [x] CQ22022 lucene memory 7.7.2
- [x] CQ22016 lucene-queries 7.7.2
- [x] CQ22017 lucene-queryparser 7.7.2
  - [x] CQ22024 lucene-sandbox 7.7.2
- [x]  CQ22018 lucene-spatial-extras 7.7.2
  - [x] CQ22021 lucene-spatial3d 7.7.2
  - [x] CQ22121 s2-geometry-library-java 1.0.0

hmottestad:
Find all original requests: https://dev.eclipse.org/ipzilla/buglist.cgi?component=technology.rdf4j

What we need:
- Elasticsearch
- Elasticsearch X-Content
 - elasticsearch transport
 - elasticsearch test framework

---
- Apache Solr SolrJ
 - Apache Solr Core

---
[...]

PR #2037, opened 2020-03-23
GH-2015 lucene-spatial-extras dep optional so we don't distribute it by default

jeenbroekstra:
This makes sure we don't distribute the lucene-spatial-extras library or its transitive dependencies by default, which makes maintaining our IP log easier (as we no longer have to log CQs for its transitive dependencies).

### DIFF
--- a/core/sail/lucene-spin/pom.xml
+++ b/core/sail/lucene-spin/pom.xml
@@ -20,6 +20,12 @@
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
I forgot about the lucene-spin-sail, which (apparantly) has a hard dependency on this particular library.

jeenbroekstra:
Grmbl. Turns out that we use this library in part of the initialization process of the LuceneSail. We can't really make it optional, unfortunately.

Issue #2043, opened 2020-03-26
runtime-osgi: make more imports optional


kenwenzel:
The RDF4J OSGi bundle is really greedy in regard to its package imports. This makes the bundle almost useless as it requires too many other bundles to resolve.

Currently it uses the following BND instructions:
`<Import-Package>org.slf4j.*, org.eclipse.rdf4j.*;version=0, javax.*;resolution:=optional, org.apache.*;resolution:=optional, org.springframework.*;resolution:=optional, *</Import-Package>`

I would suggest to also make the imports of `org.locationtech`, `org.elasticsearch` and `ch.qos.logback` optional.

kenwenzel:
OK, I'll prepare a PR.

Would it also be an option to simply add bnd-maven-plugin to the parent POM to make every module a valid OSGi bundle?

PR #2045, opened 2020-03-27
GH-2043 make more imports optional of runtime-osgi

kenwenzel:
GitHub issue resolved: #2043 

Changed BND instructions to also make following imports optional:
* org.locationtech
* org.elasticsearch
* ch.qos.logback

### DIFF
--- a/tools/runtime-osgi/pom.xml
+++ b/tools/runtime-osgi/pom.xml
@@ -35,7 +35,16 @@
 					<instructions>
 						<Bundle-SymbolicName>org.eclipse.rdf4j</Bundle-SymbolicName>
 						<Export-Package>!*text-base,!*prop-base,org.eclipse.rdf4j.*</Export-Package>
-						<Import-Package>org.slf4j.*, org.eclipse.rdf4j.*;version=0, javax.*;resolution:=optional, org.apache.*;resolution:=optional, org.springframework.*;resolution:=optional, *</Import-Package>
+						<Import-Package>
+							org.slf4j.*,
+							javax.*;resolution:=optional,
+							org.apache.*;resolution:=optional,
+							org.springframework.*;resolution:=optional,
+							org.locationtech;resolution:=optional,
+							org.elasticsearch;resolution:=optional,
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
I have no real way to judge this as I have little expertise on OSGi. @manuelfiorelli I know you've used the OSGi bundle in the past, can you take a quick look at this PR and tell us if it makes sense?

manuelfiorelli:
Hi @jeenbroekstra. I looked at this PR without building the code by myself.
The first thing that I noticed is that the PR should actually do what it claims: just remove the 3 mentioned package imports. In fact, @kenwenzel also removed the the import of the packages `org.eclipse.rdf4j.*`, which are also exported by the bundle. This change should only affect users that deployed another bundle that exports these packages into the same container: in this case, the OSGi framework might force the rdf4j-runtime bundle to use the packages exported by the other bundle instead of its own versions.

In general, the main implication of removing a package import is the generation of exceptions when the JVM needs a class from that package. Since the JVM adopts a lazy loading strategy, these exceptions should be thrown only when the user tries to execute a portion of RDF4J that needs those classes:
*  in case of `org.locationtech`, the portion should include the GeoSpatial modules
* in case of `org.elasticsearch`, the portion should be related to fulltext search
* regarding `ch.qos.logback`, I don't know which portion of RDF4J explicitly uses that logger library: in most cases, RDF4J should just use `slf4j,` completely the choice of a logging library to the user.

Summarizing, the consequences of this PR are mainly two:
* the user no longer needs bundles providing the packages `org.locationtech` and `org.elasticsearch`
* but, if the user actually hits portions of the runtime that exploit these packages, the JVM will throw some exceptions.

In other words, this PR goes into the direction of having an RDF4J runtime which can be configured (by supplying specific dependencies) to enable a subset of the modules.

Existing users shouldn't be affected by this PR.

jeenbroekstra:
Excellent, thanks for your review @manuelfiorelli ! And @kenwenzel , thanks again for providing this patch!

Issue #2044, opened 2020-03-27
User should be able to view the effective query plan for a query


hmottestad:
It is useful for the user to be able to inspect the effective query plan for a query. There should be some way for a user to view the query plan.

The query plan is usually not finalised until after the query has been executed, but before the first result is retrieved.

The query plan generated when the query is instantiated is not the optimised plan that actually gets run.

Options for exposing the query plan:
 - String explain() method on QueryResult
 - String explain(QueryResult qr) on QueryResults

It would also be useful if `StatementPattern` showed the estimated cardinality.

hmottestad:
## Example query and plan

### Query
```
SELECT ?language (count(?a) as ?count) where {
        ?b dcat:dataset ?a .
        ?a dct:language ?language .
        FILTER(?language != eu-lang:ENG)
} group by ?language

````
### Plan
```
Projection
   ProjectionElemList
      ProjectionElem "language"
      ProjectionElem "count"
   Extension
      ExtensionElem (count)
         Count
            Var (name=a)
      Group (language)
         Join
            Filter
               Compare (!=)
                  Var (name=language)
                  ValueConstant (value=http://publications.europa.eu/resource/authority/language/ENG)
               StatementPattern
                  Var (name=a)
                  Var (name=_const_2783f1e8_uri, value=http://purl.org/dc/terms/language, anonymous)
                  Var (name=language)
            StatementPattern
               Var (name=b)
               Var (name=_const_7420303d_uri, value=http://www.w3.org/ns/dcat#dataset, anonymous)
               Var (name=a)
         GroupElem
            Count
               Var (name=a)
```

**Comment by hmottestad** [comment]
@jeenbroekstra do you think this is a good idea?

**Comment by kenwenzel** [comment]
Maybe you could also use some kind of magic keyword like `EXPLAIN` in front of the actual SPARQL query as in:
```
EXPLAIN
SELECT ?language (count(?a) as ?count) where {
        ?b dcat:dataset ?a .
        ?a dct:language ?language .
        FILTER(?language != eu-lang:ENG)
} group by ?language
```
The result could be a JSON object that represents the query plan.
This would also work with remote servers that support the EXPLAIN feature.

kenwenzel:
Oracle provides also the possibility to fetch the last query plan via SQL:
[...]

PR #2047, opened 2020-03-28
GH-2044 store cardinality in query model node

hmottestad:
Related GitHub issue: #2044  

* introduce getter and setter for cardinality in QueryModelNode
* default implementation in AbstractQueryModelNode
* include cardinality in signature for StatementPattern
    * prints `StatementPattern (resultSizeEstimate=1)` instead of just `StatementPattern`

Eg:

```
Projection
   ProjectionElemList
      ProjectionElem "name"
      ProjectionElem "mbox"
   Filter
      Regex
         Str
            Var (name=mbox)
         Var (name=pattern)
         Var (name=flags)
      Join
         StatementPattern (resultSizeEstimate=1)
            Var (name=y)
            Var (name=_const_2c03e873_uri, value=http://example.org/ns#pattern, anonymous)
            Var (name=pattern)
         Join
            StatementPattern (resultSizeEstimate=1)
               Var (name=y)
               Var (name=_const_144ac82a_uri, value=http://example.org/ns#flags, anonymous)
               Var (name=flags)
            Join
               StatementPattern (resultSizeEstimate=3)
                  Var (name=x)
                  Var (name=_const_23b7c3b6_uri, value=http://xmlns.com/foaf/0.1/name, anonymous)
                  Var (name=name)
               StatementPattern (resultSizeEstimate=3)
                  Var (name=x)
                  Var (name=_const_23b75369_uri, value=http://xmlns.com/foaf/0.1/mbox, anonymous)
                  Var (name=mbox)
```

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
@@ -7,13 +7,6 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.impl;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.Dataset;
 import org.eclipse.rdf4j.query.algebra.Extension;
@@ -27,6 +20,13 @@ import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizer;
 import org.eclipse.rdf4j.query.algebra.helpers.AbstractQueryModelVisitor;
 import org.eclipse.rdf4j.query.algebra.helpers.StatementPatternCollector;
 import org.eclipse.rdf4j.query.algebra.helpers.TupleExprs;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * A query optimizer that re-orders nested Joins.
@@ -109,7 +109,9 @@ public class QueryJoinOptimizer implements QueryOptimizer {
 					Map<TupleExpr, List<Var>> varsMap = new HashMap<>();
 
 					for (TupleExpr tupleExpr : joinArgs) {
-						cardinalityMap.put(tupleExpr, statistics.getCardinality(tupleExpr));
+						double cardinality = statistics.getCardinality(tupleExpr);
+						tupleExpr.setCardinality(cardinality);
+						cardinalityMap.put(tupleExpr, cardinality);
 						if (tupleExpr instanceof ZeroLengthPath) {
 							varsMap.put(tupleExpr, ((ZeroLengthPath) tupleExpr).getVarList());
 						} else {
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
@@ -110,7 +110,7 @@ public class QueryJoinOptimizer implements QueryOptimizer {
 
 					for (TupleExpr tupleExpr : joinArgs) {
 						double cardinality = statistics.getCardinality(tupleExpr);
-						tupleExpr.setCardinality(cardinality);
+						tupleExpr.setEstimatedRows(cardinality);
 						cardinalityMap.put(tupleExpr, cardinality);
 						if (tupleExpr instanceof ZeroLengthPath) {
 							varsMap.put(tupleExpr, ((ZeroLengthPath) tupleExpr).getVarList());
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
@@ -110,7 +110,7 @@ public class QueryJoinOptimizer implements QueryOptimizer {
 
 					for (TupleExpr tupleExpr : joinArgs) {
 						double cardinality = statistics.getCardinality(tupleExpr);
-						tupleExpr.setEstimatedRows(cardinality);
+						tupleExpr.setResultSizeEstimate(cardinality);
 						cardinalityMap.put(tupleExpr, cardinality);
 						if (tupleExpr instanceof ZeroLengthPath) {
 							varsMap.put(tupleExpr, ((ZeroLengthPath) tupleExpr).getVarList());
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
@@ -26,21 +26,21 @@ public interface QueryModelNode extends Cloneable, Serializable {
 
 	/**
 	 * Gets the node's parent.
-	 * 
+	 *
 	 * @return The parent node, if any.
 	 */
 	public QueryModelNode getParentNode();
 
 	/**
 	 * Sets the node's parent.
-	 * 
+	 *
 	 * @param parent The parent node for this node.
 	 */
 	public void setParentNode(QueryModelNode parent);
 
 	/**
 	 * Replaces one of the child nodes with a new node.
-	 * 
+	 *
 	 * @param current     The current child node.
 	 * @param replacement The new child node.
 	 * @throws IllegalArgumentException If <tt>current</tt> is not one of node's children.
@@ -50,7 +50,7 @@ public interface QueryModelNode extends Cloneable, Serializable {
 
 	/**
 	 * Substitutes this node with a new node in the query model tree.
-	 * 
+	 *
 	 * @param replacement The new node.
 	 * @throws IllegalStateException If this node does not have a parent node.
 	 * @throws ClassCastException    If <tt>replacement</tt> is of an incompatible type.
@@ -73,7 +73,7 @@ public interface QueryModelNode extends Cloneable, Serializable {
 	/**
 	 * Returns the signature of this query model node. Signatures normally include the node's name and any parameters,
 	 * but not parent or child nodes. This method is used by {@link #toString()}.
-	 * 
+	 *
 	 * @return The node's signature, e.g. <tt>SLICE (offset=10, limit=10)</tt>.
 	 */
 	public String getSignature();
@@ -81,8 +81,19 @@ public interface QueryModelNode extends Cloneable, Serializable {
 	/**
 	 * Returns a (deep) clone of this query model node. This method recursively clones the entire node tree, starting
 	 * from this nodes.
-	 * 
+	 *
 	 * @return A deep clone of this query model node.
 	 */
 	public QueryModelNode clone();
+
+	/**
+	 * Returns the number of outputs that this QueryNode predicts will be outputted. For a StatementPattern this would
+	 * be the estimated cardinality provided by the EvaluationStatistics. For a Join the would be the resulting number
+	 * of joined tuples.
+	 * 
+	 * @return cardinality
+	 */
+	public double getCardinality();
+
+	public void setCardinality(double cardinality);
 }
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
@@ -7,6 +7,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra;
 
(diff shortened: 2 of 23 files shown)


hmottestad:
It seems that we never explicitly generate a cardinality for the join operations.

hmottestad:
Going to fix the tests if you agree with this PR in general @jeenbroekstra .

jeenbroekstra:
I'm not really sure I do, to be honest. Why would you want to make the cost estimation part of the algebra model itself?

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
```diff
@@ -73,16 +73,27 @@
...
+	/**
+	 * Returns the number of outputs that this QueryNode predicts will be outputted. For a StatementPattern this would
+	 * be the estimated cardinality provided by the EvaluationStatistics. For a Join the would be the resulting number
+	 * of joined tuples.
+	 * 
+	 * @return cardinality
+	 */
+	public double getCardinality();
+
+	public void setCardinality(double cardinality);
```
jeenbroekstra: Strictly speaking this is not backward-compatible. I'm on the fence whether we can get away with by saying that any implementor who does not use the abstract base class has only themselves to blame.

jeenbroekstra:
Also: I hate to have to keep bringing this up, but could you please fix your branch name? No `issue/` prefix, and use dashes instead of underscores (at least for the first one after the issue number, otherwise github doesn't recognize/hotlink it in the UI).

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
```diff
@@ -73,16 +73,27 @@
...
 	public QueryModelNode clone();
+
+	/**
+	 * Returns the number of outputs that this QueryNode predicts will be outputted. For a StatementPattern this would
+	 * be the estimated cardinality provided by the EvaluationStatistics. For a Join the would be the resulting number
+	 * of joined tuples.
+	 * 
+	 * @return cardinality
+	 */
+	public double getCardinality();
```
jeenbroekstra: Should we call it 'getCostEstimate` as a more generic method? Cost estimation _can_ be done using cardinality, but it's not always the sole factor (for example a SERVICE clause has a significantly higher cost due to network latency).

jeenbroekstra:
I still have trouble with this - the fact that it is not fully backward compatible doesn't sit well with me. I'm also not sure how useful this information really is to end users. The point of the cardinality is really to have an internal mechanism to make some sort of 'best guess' ordering of the expression tree. In and of themselves the numbers don't have any meaning, and the result of the numbers (that is, the ordering) can already be seen just by looking at the difference between the original tree and the optimized tree.

So the question is: what problem is this solving?

hmottestad:
Pretty much every other query plan out there that I have seen exposes the cost model. It’s a natural thing for an advanced user to look at when trying to optimize a query.

Consider it an annotation. It’s a lot simpler and smoother adding an annotation inside the java object rather than having a hashmap or something else to store the annotation externally.

jeenbroekstra:
> Pretty much every other query plan out there that I have seen exposes the cost model. It’s a natural thing for an advanced user to look at when trying to optimize a query.

It does mean we'll need to do some documentation to explain this. As I said: the numbers in themselves are meaningless. We'll need to provide guidance on how to interpret them.

> Consider it an annotation. It’s a lot simpler and smoother adding an annotation inside the java object rather than having a hashmap or something else to store the annotation externally.

I fully agree it's easier. But that doesn't make it any less backward incompatible, unfortunately.

Note that this is not insurmountable obstacle - if this is the best way to do it, then let's do it this way. I'm starting to come around to the idea that this is a useful thing to further develop (I'm getting to the age where everything new is initially just "obviously wrong" - it sometimes takes me a while to get over that ;)).  It just means we'll need to schedule it for 4.0 instead of 3.2.

hmottestad:
> (I'm getting to the age where everything new is initially just "obviously wrong" - it sometimes takes me a while to get over that ;)). It just means we'll need to schedule it for 4.0 instead of 3.2.

Could say the same for myself and branch naming :P 

If I give the interface methods a default implementation it should be backwards compatible since for anyone else implementing them it would be optional to implement those two methods.

Maybe we should go all in though and implement support for something along the lines of what PostgreSQL does.

#### Example from PostgreSQL
`Seq Scan on tenk1  (cost=0.00..458.00 rows=10000 width=244)`
 - 0.00: Estimated start-up cost. This is the time expended before the output phase can begin, e.g., time to do the sorting in a sort node.
 - 458.00: Estimated total cost. This is stated on the assumption that the plan node is run to completion, i.e., all available rows are retrieved. In practice a node's parent node might stop short of reading all available rows 
 - 10000: Estimated number of rows output by this plan node. Again, the node is assumed to be run to completion.
 - 244: Estimated average width of rows output by this plan node (in bytes).

The costs are measured in arbitrary units determined by the planner's cost parameters.

Copied from the PostgreSQL documentation: https://www.postgresql.org/docs/10/using-explain.html#USING-EXPLAIN-BASICS

jeenbroekstra:
> If I give the interface methods a default implementation it should be backwards compatible since for anyone else implementing them it would be optional to implement those two methods.

If that were possible it'd be great, but one of your methods is a setter. Bar throwing a NotImplementedException there's not really any way to provide a default for it.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
```diff
@@ -73,16 +75,33 @@
...
+	 */
+	@Experimental
+	default double getEstimatedRows() {
+		return -1;
+	}
+
+	@Experimental
+	default void setEstimatedRows(double rows) {
+		// no-op for backwards compatibility
+	}
```
hmottestad: Propose we use the PostgreSQL naming of "rows" for this cardinality. I've prefixed it with "estimated" to make it more obvious what we are talking about.

In order for this to be backwards compatible I propose that setEstimatedRows be a no-op. We can remove both default implementations in 4.0.0, I'm just a bit unsure now how we can mark these methods to make it obvious for anyone who stumbles upon them. I have marked them Experimental though, since we might want to change the name or something else in the future. I'm open for other names though, "getEstimatedTuples" comes to mind.

jeenbroekstra: Couple of things:

1. Rows is SQL/table-speak. SPARQL doesn't _have_ rows or tables (certainly construct queries don't). It has solutions, and solution sequences. 

2. You're interpreting cardinality/cost to always be the number of solutions. That's just not true. 

I'd stick with the more abstract 'costEstimate' for this. This gives us freedom to use any  relevant measure (including number of solutions) as a cost estimate.

jeenbroekstra: Happy with the noop solution though, that makes sense.

hmottestad: Rows could also be interpreted as rows of solutions. I think “tuples” is better. 

Cost is different from tuples. Cost is a calculated field based on a cost model. It could take into account the difference between a request to disk vs network or a sort+mergejoin vs hashjoin. 

Tuples is an estimated ceiling. It doesn’t account for what is likely to happen, for instance based on bound variables or a limit higher up in the algebra model. A join operation should account for bound variables. Consider the tuples estimate to be what this single operation would produce if only that node with its sub tree existed.

hmottestad: If we want to introduce the abstract concept of cost, we should add a new field for that.

jeenbroekstra: > Rows could also be interpreted as rows of solutions. I think “tuples” is better.
> 
> Cost is different from tuples. Cost is a calculated field based on a cost model. It could take into account the difference between a request to disk vs network or a sort+mergejoin vs hashjoin.

...which is precisely what the (misnamed) cardinality calculator currently does. For example it assigns a higher cardinality score to a SERVICE clause, not because there are more solutions expected, but because it tries to account for network latency. So what the cardinality calculator gives you back is an estimated cost, _not_ an estimated number of solutions.

> Tuples is an estimated ceiling. It doesn’t account for what is likely to happen, for instance based on bound variables or a limit higher up in the algebra model. A join operation should account for bound variables. Consider the tuples estimate to be what this single operation would produce if only that node with its sub tree existed.

I should point out that to some extent the cardinality calculator also does these things. I don't believe it looks at limit clauses, but it does look at things like "is this variable already bound higher up in the tree" and "is this variable really a constant or is it likely to change?". It uses all sorts of tricks like taking the power of a number or dividing by some factor to tweak the numbers.

Getting the naming right wasn't really important earlier because this was always purely an internal feature (and yes, _originally_ it did produce purely a cardinality as in "number of solutions" but then SPARQL got more complicated and it evolved). If we're going to expose it publicly, we need to be specific about what the numbers actually mean. The numbers as currently produced by our cardinality calculators can _not_ be reliably interpreted as meaning "number of solutions/rows/tuples". From my point of view, the only thing you can really say about them is that they are a (crude) best guess at how expensive a particular operation is going to be. Hence 'cost estimate'.

jeenbroekstra: (Scratch what I said earlier, I had where things happened mixed in my head a bit - I deleted a previous reply because it would just have added to the confusion)

...if the number you intend to produce here is _purely_ the outcome of `EvaluationStatics.getCardinality` then yes, using something like `result size estimate` or `estimated number of solutions` is fine. In fact, if we want to avoid further naming confusion, we even just go with cardinality. `getEstimatedCardinality`?

My impression (and perhaps I was wrong there) was that what you wanted to produce was _not_ just the outcome of that estimator, but the outcome of the internal cost calculation that the QueryJoinOptimizer does based (partially) on that number.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
```diff
@@ -109,7 +109,9 @@ public void meet(Join node) {
 					Map<TupleExpr, List<Var>> varsMap = new HashMap<>();
 
 					for (TupleExpr tupleExpr : joinArgs) {
-						cardinalityMap.put(tupleExpr, statistics.getCardinality(tupleExpr));
+						double cardinality = statistics.getCardinality(tupleExpr);
+						tupleExpr.setEstimatedRows(cardinality);
```
jeenbroekstra: Right, so here's where I went wrong: I somehow had it in my had that you were using the outcome of `QueryJoinOptimizer.getTupleExprCardinality` to set this number. Apologies for the confusion.

Btw that might be a good thing to fix as part of this PR: rename that method to `getTupleExprCost`.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/QueryModelNode.java
hmottestad: `Cardinality` has a fairly strict definition in mathematics of being the number of distinct elements in a set. We need a term to represent the number of elements in a **bag** because we want to be able to use the same term on other nodes in the plan. `Size` is already commonly used in java, so we could go with that. I've had a look at other SQL databases though and `rows` is very common. 

For other graph databases:
 - Neo4J: rows
 - Amazone Neptune: patternEstimate
 - Stardog: cardinality
 - Ontotext: size or iterations
 - Oracle: 10 minutes of googling and still couldn't figure it out
 - Allegrograph: cursor-estimate
 - Blazegraph: unitsOut
 - MarkLogic: maybe "result estimate"
 - Jena: couldn't find any examples

At this point we could pick "potato" and we wouldn't be any worse off. Was really hoping for a clear winner when I started down this path of googling.

jeenbroekstra: You say potato...

Id prefer sticking with cardinality because that's what is already used in the EvaluationStatistics, but I'm also fine with size (or more specifically sizeEstimate or perhaps resultSizeEstimate). Hard no on rows for me I'm afraid, really don't want to get trapped in false equivalences with relational model.

hmottestad: resultSizeEstimate it is :)

core/sail/federation/src/main/java/org/eclipse/rdf4j/sail/federation/optimizers/QueryMultiJoinOptimizer.java
```diff
@@ -177,26 +177,26 @@ public void meetJoin(TupleExpr node) {
...
+					lowestCost = cost;
 					result = tupleExpr;
 				}
 			}
 
 			return result;
 		}
 
-		protected double getTupleExprCardinality(TupleExpr tupleExpr, Map<TupleExpr, Double> cardinalityMap,
+		protected double getTupleExprCost(TupleExpr tupleExpr, Map<TupleExpr, Double> cardinalityMap,
```
jeenbroekstra, line 197: :+1:

jeenbroekstra:
ObNag: Don't forget to use squash and merge when you're merging this PR :)

jeenbroekstra:
I linked this PR with the corresponding issue but then I realized that there is no corresponding issue for this PR - it is _related_ to #2044 but doesn't solve the full issue. So took the link off again.

Issue #2024, opened 2020-03-20
Add memory store support for RDF*


jeenbroekstra:
Followup of GH-1484

The MemoryStore needs to support storing RDF* data internally.

PR #2051, opened 2020-03-30
GH-2024 rdfstar support in memorystore

jeenbroekstra:
GitHub issue resolved: #2024  

* adding and retrieving of Triple objects in the memory store added
* modified SimpleTriple.equals method to allow equality with other implementations of the Triple interface (cf. other simple model object implementations, e.g. SimpleIRI.equals)
* added basic unit and integration tests

### DIFF
--- a/testsuites/repository/src/main/java/org/eclipse/rdf4j/repository/RDFStarSupportTest.java
+++ b/testsuites/repository/src/main/java/org/eclipse/rdf4j/repository/RDFStarSupportTest.java
@@ -1,0 +1,131 @@
+/******************************************************************************* 
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.repository;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
+
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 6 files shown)


jeenbroekstra:
@avataar I would appreciate your eyes on this as well.

testsuites/repository/src/main/java/org/eclipse/rdf4j/repository/RDFStarSupportTest.java
```diff
@@ -0,0 +1,131 @@
...
+		Triple rdfStarTriple = vf.createTriple(bob, FOAF.NAME, nameBob);
+
+		testCon.add(RDF.ALT, RDF.TYPE, rdfStarTriple);
+
+		assertThat(testCon.hasStatement(RDF.ALT, RDF.TYPE, rdfStarTriple, false)).isTrue();
+	}
+
+	@Test
+	public void testAddRDFStarContext() throws Exception {
+		Triple rdfStarTriple = vf.createTriple(bob, FOAF.NAME, nameBob);
```
jeenbroekstra: Just looking at this, I'm wondering how (non-)obvious the difference between `ValueFactory.createStatement` and `ValueFactory.createTriple` will be to users. Apart from making sure it's clearly explained in the javadoc, any ideas around how to make this more clear?

avataar: If you are a complete newbie it might be unclear but if that is the case there will lots of other confusing things too. Once you realize you can't pass a Triple where a Statement is expected it should be obvious they aren't interchangeable. The multitude of methods that have "statement" in their name (hasStatement, getStatements, exportStatements, handleStatement etc.) are also a clear indication of what is expected or returned there.

If you aren't new to RDF4J it shouldn't be a problem at all.

damyan-ognyanov:
Looks good to me ...

PR #2052, opened 2020-03-30
GH-232 rio hdt writer

barthanssens:
GitHub issue resolved: #232  

* Initial experimental support for writing binary HDT files

It is an experimental, non-optimized writer, though it at least was able to produce a test file that could be read by HDT-It exe.

It is limited in the sense that it
* only supports most commonly used encoding (S,P,O order, PlainFrontCoding)
* limited to input that fits into memory (which roughly means that the entire unencoded triples must fit into memory), which is arguably not very useful because HDT shines in storing large datasets
* more tests / test files are needed


core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTArrayLog64.java
```diff
@@ -26,20 +29,32 @@
...
 class HDTArrayLog64 extends HDTArray {
 	private byte buffer[];
 
+	@Override
+	protected void setMaxValue(int maxval) {
+		// ceil(log2(maxval))
+		int i = 0;
+		while (++i < 32 && (maxval >> i) > 0)
+			;
+		setNrBits(i);
```
hmottestad: I'm not sure what this loop does, but it looks like it both increments `i` and does some bitshifting. Could this be written more explicitly.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTArrayLog64.java
```diff
@@ -67,7 +82,7 @@ protected void parse(InputStream is) throws IOException {
 		try (UncloseableInputStream uis = new UncloseableInputStream(is);
 				CheckedInputStream cis = new CheckedInputStream(uis, new CRC32())) {
 			// read bytes, minimum 1
-			long bytes = (nrbits * entries + 7) / 8;
+			long bytes = (getNrBits() * size() + 7) / 8;
```
hmottestad: It looks like you have two methods called `size` where one returns a size and the other sets the size. Could the setter be called setSize()?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTBitmap.java
```diff
@@ -9,24 +9,47 @@
...
- *
+ * Bitmap
+ * 
+ * Just a series of 0 and 1 bits.
+ * 
+ * This is used in combination with a {@link org.eclipse.rdf4j.rio.hdt.HDTArray HDTArray} to encode triple parts.
+ * 
  * @author Bart Hanssens
  */
 class HDTBitmap extends HDTPart {
```
hmottestad: Did you consider using the java bitset? https://docs.oracle.com/javase/8/docs/api/java/util/BitSet.html

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
```diff
@@ -52,7 +57,9 @@
  * @author Bart Hanssens
  */
 class HDTDictionarySectionPFC extends HDTDictionarySection {
-	private byte[] buffer;
+	private byte[] readBuffer;
+	private byte[][] writeBuffers;
+	private int wbpos;
```
hmottestad: Could you type this variable out?

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
```diff
@@ -181,4 +252,76 @@ protected void parse(InputStream is) throws IOException {
...
+		// copy base string into first entry, with a trailing NULL
+		tmp[i] = new byte[base.length + 1];
+		System.arraycopy(base, 0, tmp[i], 0, base.length);
+
+		byte[] prev = base;
+
+		// encode a block, with a maximum of strings per block (minus the first base string)
+		for (i = 1; iter.hasNext() && i < stringsBlock; i++) {
+			byte[] str = iter.next().getBytes(StandardCharsets.UTF_8);
+			iter.remove();
```
hmottestad: Do we have to remove from the iterator? Just wondering if it can be skipped for performance.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
```diff
@@ -181,4 +252,76 @@ protected void parse(InputStream is) throws IOException {
...
+
+			// remove the common part, and only store the offset and the suffix and trailing NULL
+			int common = mismatch(prev, str);
+			byte[] c = VByte.encode(common);
+			tmp[i] = new byte[c.length + str.length - common + 1];
+			System.arraycopy(c, 0, tmp[i], 0, c.length);
+			System.arraycopy(str, common, tmp[i], c.length, str.length - common);
```
hmottestad: I'm glad you're using System.arraycopy, it's much better performance than a loop :)

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
```diff
@@ -181,4 +252,76 @@ protected void parse(InputStream is) throws IOException {
...
+			System.arraycopy(tmp[j], 0, ret, idx, tmp[j].length);
+			idx += tmp[j].length;
+		}
+		return ret;
+	}
+
+	/**
+	 * Compare 2 byte arrays and return the position of the first different byte.
+	 * 
+	 * Could be replaced with Arrays.mismatch in JDK9+
```
hmottestad: That would probably be a lot faster too. Since it would probably use intrinsics and SSE instructions.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
barthanssens: Indeed :-) One of the few Java-methods for low-level bit manipulations. Sometimes I do miss coding in c ...

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
barthanssens: Well, I'm trying to somewhat reduce the memory consumption, so I'd rather keep it in here.

(I'll need to rewrite it anyway, since the actual goal is to compress very large datasets...)

hmottestad: If your underlying data structure is an array list I think this causes the array list to be copied and the previous one to be garbage collected. 

Would be nice to actually benchmark this to see if it has any effect though. You are introducing a side effect that should be made obvious to the next developer that starts to wonder where all their data went.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTWriter.java
```diff
@@ -0,0 +1,369 @@
...
+	private long cnt;
+
+	// TODO: rewrite this to cater for larger input
+	// create dictionaries and triples, with some size estimations
+	private static int SIZE = 1_048_576 * 4;
+	private Map<String, Integer> dictShared = new HashMap<>(SIZE / 4);
+	private Map<String, Integer> dictS = new HashMap<>(SIZE / 8);
+	private Map<String, Integer> dictP = new HashMap<>(SIZE / 1024);
+	private Map<String, Integer> dictO = new HashMap<>(SIZE / 2);
+	private List<int[]> tripleRefs = new ArrayList<>(SIZE);
```
jeenbroekstra: I gather these lists and maps are the main reason this writer currently only works for input that fits in memory?  If so it might be worth looking at using MapDB to overflow to disk - we use it in several other places for this purpose, and it works quite well.

jeenbroekstra: Also: I'm not sufficiently familiar with HDT to understand how this works fully, but I assume these lists are kept to allow compression (using a short reference if a particular value reoccurs instead of serializing the entire value again)? If so, you could take a leaf out of the BinaryRDFWriter's approach, which has a configurable buffer for this kind of thing: it keeps references for compression in a fifo queue so that often re-occuring values can be referenced in shorthand, but it allows for dropping references from the queue and re-creating a full value (with a _new_ reference hook) later on in the file.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTBitmap.java
barthanssens: I considered it, but it does not make much of a difference at the moment. 
Bitset is a bit more sophisticated than what's needed here, and the HDTBitmap class moved in another direction by also parsing/writing and CRC-checking the bitmap block.

Though I should rephrase the javadoc to make this clear...

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTArrayLog64.java
barthanssens: Basically it checks the position of the highest bit set counting from 1

E.g. for storing value 7, or `0000 0000 0000 0111` b, only 3 bits are needed

Or ceil(log2(value)), hence the original HDT name "arraylog64" I guess, with a max of 64 bits in the c version)

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTWriter.java
barthanssens: Thanks, very useful ! I'll look into it.

Indeed, HDT
- first replaces the parts  of triples with numerical references, so it has to keep track of what string is already assigned a number
- writes out some statistics on the total number of triples, different subjects etc
- then orders the string (well, bytes actually) representation of these parts and compresses them into "blocks"
- renumbers the numerical references and encodes them as a sort of layers
- writes out these compressed blocks
- and only then writes out the encoded references

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
barthanssens: I see your point.

IIRC removing from an arraylist involves a memmove, not a copy, but I'm not entirely sure.
It would not require more memory, but involves (much) more memory shifting, so performance won't be great on larger sets :-/

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTArrayLog64.java
barthanssens: Good point, I'll change it

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTArrayLog64.java
hmottestad: That would be nice as a comment in the code. I could google things based on what you wrote now and understand the code a lot better :)

hmottestad: Btw. Could this method help you in any way? https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#highestOneBit-int- 

But maybe not if it works on two-complement....

hmottestad: Or maybe: `ceil(log2(x)) = 32 - Integer.numberOfLeadingZeros(x - 1)`

https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#numberOfLeadingZeros-int-

barthanssens: Indeed, the latter works as expected, thanks

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
barthanssens: This would actually be a candidate for a multi Java version JAR file... but not something we can do in JDK8

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTWriter.java
jeenbroekstra: Hm, if it requires that you write out statistics at the start of the file, and also requires a full mapping dictionary at the start that can not be changed later on in the file, I'm not sure how you would ever implement this in a streaming fashion.

hmottestad: Maybe if you split it into three files and then merge them later?

barthanssens: Full streaming is indeed not possible AFAIK....

I've already left out a few statistics that didn't seem to be used by HDT-It, and maybe I can leave out a few others...
But the "commonly" used encoding still seem to require e.g. total number of triples somewhat upfront, some parts need the encoded length... and there is also the sorting/renumbering of all triples.
AFAIK the HDT format does not have "chunks" (i.e. to encode per X million triples)

hmottestad: Benefit though is that we can write a read-only sail wrapper that works with an in-place HDT file (ExtensibleSail to the rescue). We could also look into writing a `dump` tool for the NativeStore that dumps the data in HDT format, which should be considerably more performant since i believe the NativeStore is already sorted (B-tree).

barthanssens: Indeed that is my long-term plan :-)

HDT is meant for read-only, fast lookup for linked data fragments / triple fragments applications

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTDictionarySectionPFC.java
barthanssens: Removed the iter.remove, and using clear() in other parts of the code

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTWriter.java
barthanssens: I've noticed the MapDB in RDF4J is quite old (1.0.8) and not supported anymore. 
Newer versions have more sophisticated settings (e.g. overflow to disk), is there a specific reason to sticking to 1.0.8 ?

jeenbroekstra: Nothing beyond "we haven't had the time or resources to upgrade", I think. I vaguely remember there may have been some non-compatible changes that makes upgrading non-trivial. But I haven't looked at it in a while.

jeenbroekstra: If it has built-in support for overflow now that would be good - I believe we currently implement our own overflow by juggling normal collections and then at some point switching to MapDB.

core/rio/hdt/src/main/java/org/eclipse/rdf4j/rio/hdt/HDTBitmap.java
```diff
@@ -49,12 +55,18 @@
 	 * 
 	 * @param number of bits
 	 */
+<<<<<<< HEAD
+	protected void size(int bits) {
+=======
```
hmottestad, line 60: Git merge conflict?

jeenbroekstra:
There's still a conflict on this PR unfortunately. 

Are you intending to merge this feature in time for 3.2, or is there more you want to do? I'm happy either way, it's a functional PoC so we can advertise it as experimental new feature, and improve as we go.

barthanssens:
Thanks for the heads-up Jeen.

I guess it is better to postpone the HDT writer to the next release (3.3)
 
While I'm quite happy with the parser (already in dev-tree, other PR) since it seems to be able to parse larger files (few M triples), the writer really isn't very useful at the moment, even for my own purposes :-/

I'll have to rewrite some parts to make it "good enough"

PR #2053, opened 2020-04-01
GH-2011 merge strategy and checklist added to PR template

jeenbroekstra:
GitHub issue resolved: #2011  

* added PR author checklist (not required for people to check things off, more a reminder)
* added a note about our default merge strategy being squash and merge
* added a link to the developer documentation that goes into more detail about our merge strategy.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,15 @@ GitHub issue resolved: # <!-- add a Github issue number here, e.g #123. This lin
 
 Briefly describe the changes proposed in this PR:
 
-* 
-* 
-* 
+<!-- short description of your change goes here -->
+
+---- 
+PR Author Checklist: 
+
+ - [ ] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
+ - [ ] I've added tests for the changes I made
+ - [ ] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change
+ - [ ] every commit has been [signed off](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,6 +4,15 @@ GitHub issue resolved: # <!-- add a Github issue number here, e.g #123. This lin
...
-* 
+<!-- short description of your change goes here -->
+
+---- 
+PR Author Checklist: 
+
+ - [ ] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
+ - [ ] I've added tests for the changes I made
+ - [ ] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change
+ - [ ] every commit has been [signed off](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)
```
hmottestad, line 15: From having added a checklist like this in one of my previous projects I don't belive that it actually works as a checklist. We can always try, but I think it would work just as well as a regular list.

jeenbroekstra: Yeah, agreed, I've used it before as well and it never really works as a proper list to check things off. My intent here is not so much to force people to actually tick the boxes, more as a list of reminders - using checkboxes to render to make it (hopefully) a little more obvious that these are actionable items.

If people tick them, great. If not, no harm done either.

jeenbroekstra:
I'll leave this open for a few more hours, but if I hear no further concerns/protests, I will assume we are all ok with the proposed new strategy, and merge and close the issue.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,9 +1,17 @@
 
-GitHub issue resolved: # <!-- add a Github issue number here, e.g #123. This line 
-                              automatically closes the issue when the PR is merged --> 
+GitHub issue resolved: # <!-- add a Github issue number here, e.g #123. -->
```
jeenbroekstra, line 2: I'm not 100% sure but I believe this may no longer work when using squash and merge because the PR description won't end up in the commit message. This is how auto-closing of issues worked in the past. It _might_ still work though based on whether the PR is actually linked to an issue (which is a feature Github introduced very recently). In any case, I'm taking it out for brevity - if it still works that's great, if not, we're not misleading people.

Issue #2056, opened 2020-04-01
More efficient use of  DynamicModel through the new getStatements method in Model


hmottestad:
The new getStatements method in Model opens up for efficient usage of the DynamicModel by reducing the need for it to upgrade to a full blown model. 

This requires both implementing getStatements in DynamicModel and also using getStatements more throughout RDF4J.

PR #2057, opened 2020-04-01
GH-2056 more use of getStatement instead of filter

hmottestad:
GitHub issue resolved: #2056  

* Change from Iterator to Iterable
* Use Iterable where possible in helper methods in Models class
* Use .getStatements where possible in the RDF4J code base

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
@@ -169,7 +169,7 @@ public interface Model extends Set<Statement>, Serializable, NamespaceAware {
 	}
 
 	/**
-	 * Returns an {@link Iterator} over all {@link Statement}s in this Model that match the supplied criteria.
+	 * Returns an {@link Iterable} over all {@link Statement}s in this Model that match the supplied criteria.
 	 * <p>
 	 * Examples:
 	 * <ul>
@@ -190,7 +190,7 @@ public interface Model extends Set<Statement>, Serializable, NamespaceAware {
 	 *                  disregarding their context. If one or more contexts are specified, statements with a context
 	 *                  matching any one of these will match. To match statements without an associated context, specify
 	 *                  the value {@code null} and explicitly cast it to type {@code Resource}.
-	 * @return an {@link Iterator} over the statements in this Model that match the specified pattern.
+	 * @return an {@link Iterable} over the statements in this Model that match the specified pattern.
 	 *
 	 * @since 3.2.0
 	 *
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
@@ -20,7 +20,7 @@ import org.eclipse.rdf4j.model.impl.SimpleNamespace;
  * <p>
  * Additional utility functionality for working with Model objects is available in the
  * {@link org.eclipse.rdf4j.model.util.Models Models} utility class.
- * 
+ *
  * @author James Leigh
  * @author Jeen Broekstra
  * @see org.eclipse.rdf4j.model.util.Models the Models utility class
@@ -32,14 +32,14 @@ public interface Model extends Set<Statement>, Serializable, NamespaceAware {
 	 * operations on the returned model "read through" to this model, and attempts to modify the returned model, whether
 	 * direct or via its iterator, result in an {@link UnsupportedOperationException}.
 	 * <p>
-	 * 
+	 *
 	 * @return an unmodifiable view of the specified set.
 	 */
 	public Model unmodifiable();
 
 	/**
 	 * Sets the prefix for a namespace. This will replace any existing namespace associated to the prefix.
-	 * 
+	 *
 	 * @param prefix The new prefix.
 	 * @param name   The namespace name that the prefix maps to.
 	 * @return The {@link Namespace} object for the given namespace.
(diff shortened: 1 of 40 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
```diff
@@ -281,12 +296,32 @@ public void clear() {
 		}
 	}
 
+	@Override
+	public Iterator<Statement> getStatements(Resource subject, IRI predicate, Value object, Resource... contexts) {
+		if (model == null && subject != null && predicate != null && object != null && contexts != null
+				&& contexts.length == 1) {
+			Statement statement = SimpleValueFactory.getInstance()
+					.createStatement(subject, predicate, object, contexts[0]);
+			Statement foundStatement = statements.get(statement);
```
jeenbroekstra, line 305: Ah. Using a HashMap to be able to match and retrieve the original statement, instead of the newly created copy. Clever.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/DistinctModelReducingUnionIteration.java
```diff
@@ -34,15 +34,15 @@
 public class DistinctModelReducingUnionIteration extends LookAheadIteration<Statement, SailException> {
 
 	DistinctModelReducingUnionIteration(CloseableIteration<? extends Statement, SailException> iterator, Model model,
-			Function<Model, Model> filterable) {
+			Function<Model, Iterator<Statement>> filterable) {
```
jeenbroekstra, line 37: Can we keep the old constructor signature as a (deprecated) overload? Just to keep things backward compatible.

jeenbroekstra:
Btw are you aware that instead of labeling something WIP you can actually create a Draft pull request as well? 

Screenshot from 2020-04-02 09-48-58

I'm fine with using WIP label too btw - as long as the intent is clear.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/DistinctModelReducingUnionIteration.java
jeenbroekstra: Actually never mind, I overlooked that this constructor is package-private anyway.

hmottestad:
I know. I did. But realized that it had the wrong target branch. Changed the target branch and then clicked the create button, but it had changed from draft to normal pr because I had changed the target branch. 

Unless you know how to make a branch draft after it’s created....

hmottestad:
Added benchmark is down from 19.2ms to 16.8ms, ~10% decrease. Not massive, but nice enough.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -267,15 +267,16 @@ public void handleClose() throws SailException {
 		}
 		Model deprecated = changes.getDeprecated();
 		if (deprecated != null && iter != null) {
-			iter = difference(iter, deprecated.filter(subj, pred, obj, contexts));
+			iter = difference(iter, deprecated);
```
hmottestad, line 270: I don't think this change makes a difference for the results. The set of statements that are in `iter` but not in `deprecated` is the same unless the pattern used to filter `deprecated` is different from the pattern used by `iter`.

The difference is for performance. Now later on in the FilterIteration we can work directly with the hashmap without having to upgrade to a full LinkedHashModel because of a complicated `filter` operation.

jeenbroekstra: The test suite appears to agree with you!

jeenbroekstra:
> I know. I did. But realized that it had the wrong target branch. Changed the target branch and then clicked the create button, but it had changed from draft to normal pr because I had changed the target branch.
> 
> Unless you know how to make a branch draft after it’s created....

No afraid not, this is where it falls down - I think you'd have to close the PR and create a completely new one. Labeling WIP is probably easier in those cases.

core/model/src/main/java/org/eclipse/rdf4j/model/Model.java
```diff
@@ -191,14 +191,14 @@ public default boolean remove(Resource subj, URI pred, Value obj, Resource... co
...
 	 * @since 3.2.0
-	 * 
+	 *
 	 * @see #filter(Resource, IRI, Value, Resource...)
 	 */
-	public default Iterator<Statement> getStatements(Resource subject, IRI predicate, Value object,
+	public default Iterable<Statement> getStatements(Resource subject, IRI predicate, Value object,
 			Resource... contexts) {
-		return filter(subject, predicate, object, contexts).iterator();
+		return () -> filter(subject, predicate, object, contexts).iterator();
```
hmottestad, line 201: @jeenbroekstra I'm proposing we move from iterator to iterable because iterable supports for-each loops and is also simpler to convert into a stream.

hmottestad: It's a pity that Java didn't introduce a combo interface with both iterable and stream. We can always create our own interface in the future though with extends both and still remain backwards compatible.

jeenbroekstra: > @jeenbroekstra I'm proposing we move from iterator to iterable because iterable supports for-each loops and is also simpler to convert into a stream.

I'm fine with that, but can you make sure you update the original issue's (#2030) description, and also the javadoc in Model?

hmottestad:
I feel happy with this now and ready to merge.

PR #2058, opened 2020-04-02
GH-99 allow SailConnection to bypass Query Parser

jeenbroekstra:
GitHub issue resolved: #99  

- adds new method 'prepareQuery' to SailConnection that gets the raw
  query string
- SailConnection can optionally pass back a custom implementation of
  TupleExpr. If this happens the SailRepository skips parsing the query
  and will just pass back in the custom TuplExpr when
  SailConnection.evaluate is called.
- by default the method does nothing (returns empty optional)

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
@@ -7,6 +7,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query;
 
+import org.eclipse.rdf4j.model.Statement;
+
 /**
  * A query on a repository that can be formulated in one of the supported query languages (for example SeRQL or SPARQL).
  * It allows one to predefine bindings in the query to be able to reuse the same query with different bindings.
@@ -21,9 +23,21 @@ public interface Query extends Operation {
 	 * 
 	 * @since 3.2.0
 	 */
-	public enum Type {
+	public enum QueryType {
+		/**
+		 * Boolean queries (such as the SPARQL ASK query form) return either {@code true} or {@code false} as the
+		 * result.
+		 */
 		BOOLEAN,
+		/**
+		 * Graph queries (such as the SPARQL CONSTRUCT and DESCRIBE query form) return a sequence of RDF
+		 * {@link Statement statements} as the result.
+		 */
 		GRAPH,
+		/**
+		 * Tuple queries (such as the SPARQL SELECT query form) return a sequence of {@link BindingSet sets of variable
+		 * bindings} as the result.
+		 */
 		TUPLE
 	}
 
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
@@ -17,6 +17,17 @@ package org.eclipse.rdf4j.query;
 public interface Query extends Operation {
 
 	/**
+	 * The different types of queries that RDF4J recognizes: boolean queries, graph queries, and tuple queries.
+	 * 
+	 * @since 3.2.0
+	 */
+	public enum Type {
+		BOOLEAN,
+		GRAPH,
+		TUPLE
+	}
+
+	/**
 	 * Specifies the maximum time that a query is allowed to run. The query will be interrupted when it exceeds the time
 	 * limit. Any consecutive requests to fetch query results will result in {@link QueryInterruptedException}s.
 	 * 
--- a/core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
+++ b/core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
@@ -226,9 +226,10 @@ public class SailRepositoryConnection extends AbstractRepositoryConnection imple
 	public SailTupleQuery prepareTupleQuery(QueryLanguage ql, String queryString, String baseURI)
 			throws MalformedQueryException {
 		Optional<TupleExpr> sailTupleExpr = sailConnection.prepareQuery(ql, Query.Type.TUPLE, queryString, baseURI);
-		ParsedTupleQuery parsedQuery = sailTupleExpr.isPresent()
-				? new ParsedTupleQuery(queryString, sailTupleExpr.get())
-				: QueryParserUtil.parseTupleQuery(ql, queryString, baseURI);
+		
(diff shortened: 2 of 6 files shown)


jeenbroekstra:
We may need more specific bypass operations for the various types of query (graph, tuple, boolean).

core/repository/sail/src/test/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+	@Before
+	public void setUp() throws Exception {
+		sailConnection = mock(SailConnection.class);
+		sailRepository = mock(SailRepository.class);
+
+		subject = new SailRepositoryConnection(sailRepository, sailConnection);
+	}
+
+	@Test
+	public void testPrepareQuery_not_bypassed() throws Exception {
```
jeenbroekstra, line 50: Fwiw I still haven't come up with a decent convention for this kind of test naming. Using CamelCase just quickly becomes unreadable.

jeenbroekstra:
@aschwarte10 @hmottestad If you can find the time, I would appreciate some eyes on this over the weekend. I'm thinking this might be particularly relevant for possible future improvements to FedX.

core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
```diff
@@ -16,6 +16,17 @@
  */
 public interface Query extends Operation {
 
+	/**
+	 * The different types of queries that RDF4J recognizes: boolean queries, graph queries, and tuple queries.
+	 * 
+	 * @since 3.2.0
+	 */
+	public enum Type {
```
jeenbroekstra, line 24: note to self: rename to `QueryType` for greater clarity.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
```diff
@@ -195,10 +199,23 @@ public SailQuery prepareQuery(QueryLanguage ql, String queryString, String baseU
 		ParsedQuery parsedQuery = QueryParserUtil.parseQuery(ql, queryString, baseURI);
 
 		if (parsedQuery instanceof ParsedTupleQuery) {
+			Optional<TupleExpr> sailTupleExpr = sailConnection.prepareQuery(ql, Query.Type.TUPLE, queryString, baseURI);
+			if (sailTupleExpr.isPresent()) {
+				parsedQuery = new ParsedTupleQuery(queryString, sailTupleExpr.get());
+			}
 			return new SailTupleQuery((ParsedTupleQuery) parsedQuery, this);
```
hmottestad, line 206: Does this end up potentially parsing twice? Parsing queries is one of the slowest parts of the SHACL sail, so I think we should avoid parsing twice at all costs.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
```diff
@@ -208,21 +225,30 @@ public SailQuery prepareQuery(QueryLanguage ql, String queryString, String baseU
 	@Override
 	public SailTupleQuery prepareTupleQuery(QueryLanguage ql, String queryString, String baseURI)
 			throws MalformedQueryException {
-		ParsedTupleQuery parsedQuery = QueryParserUtil.parseTupleQuery(ql, queryString, baseURI);
+		Optional<TupleExpr> sailTupleExpr = sailConnection.prepareQuery(ql, Query.Type.TUPLE, queryString, baseURI);
+		ParsedTupleQuery parsedQuery = sailTupleExpr.isPresent()
+				? new ParsedTupleQuery(queryString, sailTupleExpr.get())
+				: QueryParserUtil.parseTupleQuery(ql, queryString, baseURI);
 		return new SailTupleQuery(parsedQuery, this);
```
hmottestad, line 232: This looks like it ends up parsing only once.  Thats good :)

Btw. Optionals have some rather nice chaining potential in java 8.

```
		ParsedTupleQuery parsedQuery = sailTupleExpr
			.map(expr -> new ParsedTupleQuery(queryString, expr))
			.orElse(QueryParserUtil.parseTupleQuery(ql, queryString, baseURI));
```

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
hmottestad: I guess this is more of a legacy method, so we could probably live with having to parse the query twice if we maybe add some documentation to let the user know to prefer the more explicit methods.

core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
```diff
@@ -16,6 +16,17 @@
  */
 public interface Query extends Operation {
 
+	/**
+	 * The different types of queries that RDF4J recognizes: boolean queries, graph queries, and tuple queries.
+	 * 
+	 * @since 3.2.0
```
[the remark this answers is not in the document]
jeenbroekstra: We've been doing that for a while actually - just not very consistently. But yes, it seems like a good (low-cost) idea.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
jeenbroekstra: You're right that it parses twice, and of course that is a penalty. We could potentially improve it a bit by using some regex matching to determine the type (instead of a full parse) but at this point I didn't want to spend the effort.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
jeenbroekstra: I always forget about Optional.map for some reason. Good call, I'll fix that.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
jeenbroekstra: I'll add something to the javadoc to point to the more explicit methods though - it's a good thing to use those if you can, regardless.

Issue #2028, opened 2020-03-21
Add basic implementation of evaluationstrategy for TripleRef objects


jeenbroekstra:
From @damyan-ognyanov:

> Currently, there is no usable implementation of StrictEvaluationStrategy.evaluate(TripleRef) which intentionally was made to return empty iteration until the backward/forward compatibility is resolved.
> 
> Initially, I've used a naive and way sub-optimal implementation for evaluate(), that simply enumerated and subsequently, filter/match all subjects of statements with rdf:type predicate and rdf:Statement object (e.g. standard reification), but discarded it for the time being, since it --fixes-- the default evaluation strategy behaviour to support only standard reification and do not takes into an account the (future) sail implemetations that may have a more efficient native support for embeded triples , which maintain a dedicated collections (or other means) for more efficient handling of these requests.
> 
> One possible approach is to introduce a new interface that extends TripleSource and introduces a new method for the purpose of finding the triple entities in the dataset - named getTripleStatements() or simmilar.
> Then, in the evaluation strategy, we can check, and if the triple source provided by the sail implementation implements that interface, use it in evaluate(TripleRef), if not - the strategy can revert back to use the defaut (sub-optimal) implementation that uses standard reification.

Followup of GH-1484

PR #2062, opened 2020-04-02
GH-2028 TripleRef evaluation in StrictEvaluationStrategy

damyan-ognyanov:
GitHub issue resolved: #2028 

 - Introduced a new interface `RDFStarTripleSource` that  extends `TripleSource` interface with `getRdfStarTriples(s, p, o)`.
 - added implementation of `evaluate(TripleRef)` that uses the new interface, if possible or checks the nodes from standard reification for solutions
 - added a synthetic test cases to evaluate a `TripleRef` node through both types of sources

---- 
PR Author Checklist: 

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/RDFStarTripleSource.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/RDFStarTripleSource.java
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.query.algebra.evaluation;
+
+import org.eclipse.rdf4j.common.iteration.CloseableIteration;
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.Triple;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.query.QueryEvaluationException;
+
+public interface RDFStarTripleSource extends TripleSource {
+	/**
+	 * Gets all Triple nodes that have a specific subject, predicate and/or object. All three parameters may be null to
+	 * indicate wildcards.
+	 * 
+	 * @param subj A Resource specifying the triple's subject, or <tt>null</tt> for a wildcard.
+	 * @param pred A URI specifying the triple's predicate, or <tt>null</tt> for a wildcard.
+	 * @param obj  A Value specifying the triple's object, or <tt>null</tt> for a wildcard.
+	 * @return An iterator over the relevant triples.
+	 * @throws QueryEvaluationException If the rdf star triple source failed to get the statements.
+	 */
+	public CloseableIteration<? extends Triple, QueryEvaluationException> getRdfStarTriples(Resource subj, IRI pred,
+			Value obj) throws QueryEvaluationException;
+
+}
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
@@ -116,6 +116,7 @@ import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategy;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryBindingSet;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizer;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizerPipeline;
+import org.eclipse.rdf4j.query.algebra.evaluation.RDFStarTripleSource;
 import org.eclipse.rdf4j.query.algebra.evaluation.TripleSource;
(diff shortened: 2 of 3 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
```diff
@@ -1823,15 +1824,158 @@ public Value evaluate(ValueExprTripleRef node, BindingSet bindings)
...
+						Triple match = (Triple) iter.next();
+						if (subjValue != null && !subjValue.equals(match.getSubject())) {
+							continue;
+						}
+						if (predValue != null && !predValue.equals(match.getPredicate())) {
+							continue;
+						}
+						if (objValue != null && !objValue.equals(match.getObject())) {
+							continue;
+						}
```
jeenbroekstra, line 1882: Why are these checks necessary? Isn't this check already supposed to be done by `RDFStarTripleSource.getRdfStarTriples`?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/RDFStarTripleSource.java
```diff
@@ -0,0 +1,31 @@
...
+package org.eclipse.rdf4j.query.algebra.evaluation;
+
+import org.eclipse.rdf4j.common.iteration.CloseableIteration;
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.Triple;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.query.QueryEvaluationException;
+
+public interface RDFStarTripleSource extends TripleSource {
```
jeenbroekstra, line 17: This is minor, but since this is a behavioral interface it doesn't have to extend TripleSource - just thinking ahead of a possible future design where TripleSource and RDFStarTripleSource methods are not provided by the same object.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
```diff
@@ -1823,15 +1824,158 @@ public Value evaluate(ValueExprTripleRef node, BindingSet bindings)
...
 	 */
 	public CloseableIteration<BindingSet, QueryEvaluationException> evaluate(TripleRef ref, BindingSet bindings) {
-		// not supported
-		return new EmptyIteration<BindingSet, QueryEvaluationException>();
+		// Naive implementation that walks over all statements matching (x rdf:type rdf:Statement)
+		// and filter those that do not match the bindings for subject, predicate and object vars (if bound)
+		final org.eclipse.rdf4j.query.algebra.Var subjVar = ref.getSubjectVar();
+		final org.eclipse.rdf4j.query.algebra.Var predVar = ref.getPredicateVar();
+		final org.eclipse.rdf4j.query.algebra.Var objVar = ref.getObjectVar();
+		final org.eclipse.rdf4j.query.algebra.Var extVar = ref.getExprVar();
```
jeenbroekstra, line 1839: I'm not sure I follow where this `extVar` comes from. Is this intended to encapsulate use of RDF* triples in things like BIND clauses? Why not use a BindingSetAssignment for that case?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
damyan-ognyanov: It is like that to cover the case in which the implementation does not provide results that match the arguments but dump everything that is a `Triple` in the result iteration. Just a double check which does not put much overhead but ensure a correctness of the results.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/RDFStarTripleSource.java
damyan-ognyanov: The interface as is, does not provide any other meaningful uses for the classes that implement it. It makes sence only for those classes that also implement TripleSource that's why it extends TripleSource.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
damyan-ognyanov: Not sure how BindingSetAssignment can be re-used. 

Each reference to a Triple in the SPARQL* syntax is translated (recursively) to a TripleRef that provides bindings (instances of Triple) through it's `extVar` and that ext var is used at the place where the reference appear. E.g.:
 - the Statement pattern `<< ?s ?p ?o>> ?predicate ?object .` is translated to a Join between [TripleRef(?s ?p?o) to ?anon_ext] and [?anon_ext ?predicate ?object]
or in the case of bind:
 - `BIND(<<?s ?p ?o>> as ?t)` is translated to [TripleRef(?s ?p ?o) to ?anon_ext] and [BIND(?anon_ext as ?t)]

jeenbroekstra: Of course, that makes sense, thanks for clarifying!

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/RDFStarTripleSource.java
jeenbroekstra: I'm not sure I agree with that, but as said it's minor, fine to keep as-is, my "possible future design" may well be a YAGNI.

Issue #2006, opened 2020-03-16
FedX Federation fails to correctly set external bindings for SingleSourceQuery


aschwarte10:
If the FedX optimizer determines that a query can be executed a single relevant source, the entire query string is evaluated as-is at the respective source and the results are passed through.

In such case FedX fails to pass on bindings set via `TupleQuery#setBinding()` to the query evaluated at the single source.

Thus potentially this yields incorrect results (e.g. a superset)

Example test case for reproducing (assumes only endpoint 1 has data for the query):

```
	@Test
	public void testSingleSource_SetBinding() throws Exception {
		fedxRule.enableDebug();
		/* test a single source select query where we set a binding */
		prepareTest(Arrays.asList("/tests/basic/data01endpoint1.ttl", "/tests/basic/data01endpoint2.ttl"));

		try (RepositoryConnection conn = fedxRule.getRepository().getConnection()) {
			TupleQuery tq = conn
					.prepareTupleQuery("SELECT ?person WHERE { ?person <http://xmlns.com/foaf/0.1/name> ?name }");
			tq.setBinding("name", l("Alan"));
			TupleQueryResult tqr = tq.evaluate();
			List<BindingSet> res = Iterations.asList(tqr);
			assertContainsAll(res, "person", Sets.newHashSet(iri("http://example.org/", "a")));
		}
	}
```

Expecte behavior is that externally provided bindings are properly handled

PR #2067, opened 2020-04-03
GH-2006: fix handling of external bindings in FedX SingleSourceQuery

aschwarte10:
This change correctly set externally passed binding to single source
queries, i.e. where the query is relevant at a single owner (and thus
the original query string is passed as-is)

GitHub issue resolved: #2006

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/TripleSource.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/TripleSource.java
@@ -80,8 +80,30 @@ public interface TripleSource {
 	 * @throws RepositoryException
 	 * @throws MalformedQueryException
 	 * @throws QueryEvaluationException
+	 * @Deprecated will be removed in 4.0. Replaced with
+	 *             {@link #getStatements(String, BindingSet, QueryType, QueryInfo)}
+	 */
+	@Deprecated
+	public default CloseableIteration<BindingSet, QueryEvaluationException> getStatements(String preparedQuery,
+			QueryType queryType, QueryInfo queryInfo)
+			throws RepositoryException, MalformedQueryException, QueryEvaluationException {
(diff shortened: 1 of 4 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/TripleSource.java
```diff
@@ -82,6 +83,7 @@
 	 * @throws QueryEvaluationException
 	 */
 	public CloseableIteration<BindingSet, QueryEvaluationException> getStatements(String preparedQuery,
+			BindingSet queryBindings,
```
aschwarte10, line 86: Note that this is a slight change in the interface. Let me know if this is OK for a patch release.

I do consider this as an internal interface only.

As an alternitive I would need to duplicate the method (which is currently only called in a single location for single source queries)

hmottestad: Usually for a public interface/class and a public method the method signature would have to stay the same except for between major version. In that case I would propose to introduce a new method, mark this one as deprecated and then we can remove it in 4.0.0 release. 

However, if it's an actual bug where the queryBindings should have been provided and are required for this method to function correctly, then I'm not sure.

@jeenbroekstra should decide if that is the case.

jeenbroekstra: I very much doubt that in practice, third parties will have started relying directly on this method. But of course, strictly speaking, this is a public method and therefore this is a breaking change. I'd rather err on the side of being overly strict than catching anyone who relies on us observing semver to be caught out. 

It would be fairly simple to keep the old method signature in deprecated form and fill in `null` for the BindingSet argument. Let's do that, just to be sure, and then we can comfortably get away with doing this as part of a patch.

aschwarte10:
Done, I deprecated the original method and implemented it as default method in the interface which calls the new method. This is now fully backwards compatible

Issue #2085, opened 2020-04-13
SPARQL* query support for MemoryStore


jeenbroekstra:
followup of #2028

PR #2088, opened 2020-04-13
GH-2085 naive implementation of RDFStarTripleSource for memory store

jeenbroekstra:
GitHub issue resolved: #2085 

This enables SPARQL* querying on in-memory stores.

- added `getTriples` to SailDataset and related classes
- added implementation to SailDatasetTripleSource (which is used by memory store and native store)
- added memory store specific triple iterator

---- 
PR Author Checklist: 

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
@@ -333,5 +333,4 @@ class SailDatasetImpl implements SailDataset {
 		};
 	}
 
-
 }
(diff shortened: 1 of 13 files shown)


core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -285,6 +286,39 @@ public void handleClose() throws SailException {
...
+			iter = null;
+		} else {
+			iter = derivedFrom.getTriples(subj, pred, obj);
+		}
+
+		if (iter == null) {
+			return new EmptyIteration<>();
+		}
+		return iter; // TODO we will need to figure out a way to handle transaction isolation with deprecated and
+						// approved data
```
jeenbroekstra, line 303: FWIW I need to take a closer look at this to make sure, but it's possible that we don't need to handle this here, because it will be filtered out higher up the chain (when statements that use these triples as either subject or object are retrieved).

PR #2089, opened 2020-04-14
Extended instructions on how to build the project

jeenbroekstra:

Some editorial changes to the README to give clearer build/test instructions. 

---- 
PR Author Checklist: 

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,22 @@ For installation and usage instructions of the RDF4J Java libaries, see [Program
 
 ### Building from source
 
-To build the RDF4J project, including onejar and SDK files, from source, run:
+RDF4J is a multi-module [maven](https://maven.apache.org/index.html) project. It can be compiled, tested, and installed with the [usual maven lifecycle phases](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html) from the command line, for example:
 
-     `mvn -Passembly package`
+- `mvn verify` - compiles and runs all tests
+- `mvn package` - compiles, tests, and packages all modules
+- `mvn install` - compiles, tests, packages, and installs all artifacts in the local maven repository
+- `mvn -Pquick install` - compiles, packages and installs everything (skipping test execution)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,13 +15,22 @@ For installation and usage instructions of the RDF4J Java libaries, see [Program
 
 ### Building from source
 
-To build the RDF4J project, including onejar and SDK files, from source, run:
+RDF4J is a multi-module [maven](https://maven.apache.org/index.html) project. It can be compiled, tested, and installed with the [usual maven lifecycle phases](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html) from the command line, for example:
 
-     `mvn -Passembly package`
+- `mvn verify` - compiles and runs all tests
```
hmottestad, line 20: I have a question about this. I notice that when I build the project it downloads rdf4j dependencies off of the eclipse snapshot server, does this mean that mvn verify could run a test that fixes an issue that doesn't get built but instead gets pulled from the remote server and doesn't contain the fix, thus making the test fail?

jeenbroekstra: No, that shouldn't happen - the maven reactor always gives modules included in the multi-module build preference over (snapshot) dependencies in the repo. 

Of course it _would_ download snapshots if you are building only a part of the project (e.g. one module). Example: if you are running `mvn verify` while in `core/rio/binary`, it will download (a.o.) the `rio-api` snapshot dependency to make sure it has the most recent version. If you have made local changes to rio-api, those changes will _only_ be included in the build if you have previously run `mvn install` on rio-api, so that the snapshot dependency in your local repo is the most recent version, and maven will pick that over the (older) downloaded snapshot.

Are you sure you are seeing these downloads when building the entire project? If so can you open a ticket and attach the exact command you do and the output (you can use `-l <filename>` to log all maven's output to a file)? It's possible we have a misconfiguration somewhere of course, but I can't reproduce the problem locally.

hmottestad: You are right of course. I've been seeing the downloads from the japicmp tool and not looked at them too closely.

Issue #1982, opened 2020-03-07
Maven duplicate declaration warning


barthanssens:
Minor issue, warning when building:

```
[WARNING] Some problems were encountered while building the effective model for org.eclipse.rdf4j:rdf4j-repository-compliance:war:3.2.0-SNAPSHOT
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: ${project.groupId}:rdf4j-repository-sail:jar -> duplicate declaration of version ${project.version} @ line 59, column 15
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: ${project.groupId}:rdf4j-sail-memory:jar -> duplicate declaration of version ${project.version} @ line 65, column 15
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
```

barthanssens:
Seems to be fixed (or at least less noticeable)

PR #2095, opened 2020-04-15
GH-1982 fix duplicate dependency and several IntelliJ-specific build issues

hmottestad:
GitHub issue resolved: #1982  

Various small fixes to maven and the build.

---- 
PR Author Checklist: 

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
@@ -600,7 +600,7 @@ public class Models {
 	@Experimental
 	public static Model convertRDFStarToReification(ValueFactory vf, Model model) {
 		Model reificationModel = new LinkedHashModel();
-		convertRDFStarToReification(vf, model, reificationModel::add);
+		convertRDFStarToReification(vf, model, (Consumer<Statement>) reificationModel::add);
 		return reificationModel;
 	}
 
@@ -617,7 +617,7 @@ public class Models {
 	@Experimental
 	public static Model convertRDFStarToReification(ValueFactory vf, Model model, ModelFactory modelFactory) {
 		Model reificationModel = modelFactory.createEmptyModel();
-		convertRDFStarToReification(vf, model, reificationModel::add);
+		convertRDFStarToReification(vf, model, (Consumer<Statement>) reificationModel::add);
 		return reificationModel;
 	}
 
@@ -720,7 +720,7 @@ public class Models {
 	@Experimental
 	public static Model convertReificationToRDFStar(ValueFactory vf, Model model, ModelFactory modelFactory) {
 		Model rdfStarModel = modelFactory.createEmptyModel();
-		convertReificationToRDFStar(vf, model, rdfStarModel::add);
+		convertReificationToRDFStar(vf, model, (Consumer<Statement>) rdfStarModel::add);
 		return rdfStarModel;
 	}
 
--- a/compliance/repository/pom.xml
+++ b/compliance/repository/pom.xml
@@ -57,18 +57,6 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
-			<groupId>${project.groupId}</groupId>
-			<artifactId>rdf4j-repository-sail</artifactId>
-			<version>${project.version}</version>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>${project.groupId}</groupId>
-			<artifactId>rdf4j-sail-memory</artifactId>
-			<version>${project.version}</version>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
 			<groupId>org.eclipse.jetty</groupId>
 			<artifactId>jetty-server</artifactId>
 			<version>${jetty.version}</version>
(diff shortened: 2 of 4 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
```diff
@@ -600,7 +600,7 @@ public static void convertRDFStarToReification(Model model, Consumer<Statement>
 	@Experimental
 	public static Model convertRDFStarToReification(ValueFactory vf, Model model) {
 		Model reificationModel = new LinkedHashModel();
-		convertRDFStarToReification(vf, model, reificationModel::add);
+		convertRDFStarToReification(vf, model, (Consumer<Statement>) reificationModel::add);
```
jeenbroekstra, line 603: This shouldn't be necessary, as far as I can tell the code is legal without the cast. Are you getting an actual error in IntelliJ, or is it just giving a warning of some sort?

I'm not really in favor of changing legal code just to make one particular editor behave (for one thing it becomes tricky to maintain), but if it's really unworkable, I'd rather go with a a proper lambda (`st -> reificationModel.add(st)`) instead of this explicit cast.

jeenbroekstra:
Also looks as if this is targeted against the wrong branch.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
hmottestad: Screenshot 2020-04-15 at 15 19 18

hmottestad: I prefer the explicit cast to creating a lambda.

What are you seeing for this code in your editor?

hmottestad:
The rdf* is only in develop. And we are close to merging develop into master I hope with the release of 3.2.0.

hmottestad:
> Can you be a bit more specific about what IntelliJ complains about that bumping the plugin version solves?
> 
> I'm fine with bumping the version (though to be honest I would have preferred if you had created a separate issue and PR for it), but we'll need to run some full build and deployment tests to ensure that everything still works. The source plugin is mostly used as part of the release deployment cycle so it could all look fine now, only to bite us when it's release time.

Screenshot 2020-04-15 at 15 23 23

Documentation for that plugin isn't very good. No mention of what the source tag is supposed to do.

hmottestad:
I didn't really want to make 3 branches, 3 commits and 3 issues for minor non-functional fixes.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
jeenbroekstra: > I prefer the explicit cast to creating a lambda.
 
Why? The lambda is arguably clearer to read, and definitely shorter.  I doubt the fact that it's a delegate has a serious impact (I would imagine the compiler optimizes that away). 

> What are you seeing for this code in your editor?

Nothing. Neither Eclipse nor the OpenJDK compiler report any issue with it (using OpenJDK 8 to build). I'm really surprised IntelliJ would complain about this. Is this a known issue? Perhaps it is having difficulty with the fact that `add` is overloaded.

jeenbroekstra:
> I didn't really want to make 3 branches, 3 commits and 3 issues for minor non-functional fixes.

Fair enough, it's always a tradeoff to some extent. I'd prefer erring on the side of caution though. It makes it easier to separate things again later on if we find we need to roll back part of this.

hmottestad:
Merge commit will help with that.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
hmottestad: The lambda gets marked by the inspector as unnecessary while the casting gets ignored. 

It's most definitely at bug. Don't know if it's reported or not. It's very annoying though, since I can't do a full build from intellij anymore (which is needed to mitigate another bug I have).

jeenbroekstra: > The lambda gets marked by the inspector as unnecessary while the casting gets ignored.

That to me sounds more as if this inspector tool is configured a bit over-eager. But whatever, I don't like it, but if this is the only case then it's fairly trivial and not worth arguing over (at least not longer than we've alredy done ;)) - either way is legal Java.

jeenbroekstra:
> Documentation for that plugin isn't very good. No mention of what the source tag is supposed to do.

I can't find it either, but this was introduced into our pom about 8 months ago by @barthanssens (see https://github.com/eclipse/rdf4j/commit/4baf5084920f0b092e8f93e24088d724c301766d) apparently having something to do with getting SolR/ElasticSearch deps included in the javadoc. From the look of it I wonder if it's simply a copy-paste error (since the same config option was also added to the javadoc plugin, where it _is_ a documented option). @barthanssens  I know it's a while back but do you think you remember why you added this?

hmottestad:
Emailed Bart. Removed the erroneous tag and downgraded the version back to the original one. So I think this is ready to merge now. Nothing really needs to be tested now. I can squash and merge since we don't need to keep the history anymore.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Models.java
hmottestad: Reported the bug to the intellij people.

Issue #2101, opened 2020-04-15
Make WriteStrategy configurable


jeenbroekstra:
As a fedx user, I want to manually configure which federation member is the write target, so that I can control where data updates get sent as part of configuration.

Currently, the FedX WriteStrategy is determined automatically by inspecting the federation members and just picking any member that has the isWritable flag set to true (non-configurable, determined by the repository type). We should add an optional attribute to the configuration template for FedX so that we can explicitly identify the member intended to be the write target.

PR #2103, opened 2020-04-16
GH-2101 set fedx:writable property for explicit writable member endpoint

jeenbroekstra:
GitHub issue resolved: #2101  

Added `fedx:writable` property which can optionally be set to true for a member endpoint service (or a member repository when using "classic config").

Added simple unit tests to verify property is communicated. 

---- 
PR Author Checklist: 

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/write/WriteStrategy.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/write/WriteStrategy.java
@@ -10,7 +10,6 @@ package org.eclipse.rdf4j.federated.write;
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.model.Value;
-import org.eclipse.rdf4j.repository.RepositoryConnection;
 import org.eclipse.rdf4j.repository.RepositoryException;
 
 /**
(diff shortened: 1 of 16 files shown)


jeenbroekstra:
> LGTM, nice and clean change to add writable configurability via the repository configuration

Thanks, good to hear. 
 
> The next step is probably to also allow influencing the write strategy to pick a user-configured write target

Do you mean a way to set up a federation with a hand-picked write target programmatically? It's technically already possible of course, but I'll see if I can make it a bit more convenient by extending the FedXFactory (as an aside: that should really be called a builder rather than a factory, any objections if I deprecate/rename?).

jeenbroekstra:
@aschwarte10 I'd like to hear your thoughts on the latest changes. Anything I overlooked in terms of making per-endpoint configuration of writability more easily accessible?

One thing to note is that we don't enforce anywhere that only a single repository can be writable: if a user configuresmore than one, fedx will just pick the first one it encounters (similar to how this already worked). This is deliberate as I don't want to put checks in at config time that we may need to remove again later, when we look at strategies involving multiple target repos.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/write/WriteStrategy.java
```diff
@@ -25,49 +24,35 @@
...
-	public void initialize() throws RepositoryException;
-
-	/**
-	 * Returns true if this instance is initialized
-	 * 
-	 * @return flag indicating the initialization resources
-	 */
-	public boolean isInitialized();
-
+public interface WriteStrategy extends AutoCloseable {
```
aschwarte10, line 27: we should mention here that an instance should be specifically bound to a RepositoryConnection, at least as internal comment

aschwarte10:
> @aschwarte10 I'd like to hear your thoughts on the latest changes. Anything I overlooked in terms of making per-endpoint configuration of writability more easily accessible?
> 
> One thing to note is that we don't enforce anywhere that only a single repository can be writable: if a user configuresmore than one, fedx will just pick the first one it encounters (similar to how this already worked). This is deliberate as I don't want to put checks in at config time that we may need to remove again later, when we look at strategies involving multiple target repos.

Change looks all good to me. We require a bit more testing in action though

aschwarte10:
> 
> Do you mean a way to set up a federation with a hand-picked write target programmatically? It's technically already possible of course, but I'll see if I can make it a bit more convenient by extending the FedXFactory.

Makes perfectly sense to switch to a builder pattern, might be a good time to do this now when we introduce new configuration settings anyways

jeenbroekstra:
I'll keep the builder pattern idea in reserve for now, priority on merging this change.

Issue #2104, opened 2020-04-17
API for retrieving query plan


hmottestad:
This issue is for the API implementation related to #2044

PR #2105, opened 2020-04-18
GH-2104 explain query (query plan) API design and initial implementation

hmottestad:
GitHub issue resolved: #2104 

---- 
PR Author Checklist: 

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
@@ -74,7 +74,7 @@ public interface Query extends Operation {
 	 */
 	enum QueryExplainLevel {
 		Unoptimized,
-		Optimizaed,
+		Optimized,
 		Executed
 	}
 
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
@@ -12,7 +12,7 @@ import org.eclipse.rdf4j.model.Statement;
 /**
  * A query on a repository that can be formulated in one of the supported query languages (for example SeRQL or SPARQL).
  * It allows one to predefine bindings in the query to be able to reuse the same query with different bindings.
- * 
+ *
  * @author Arjohn Kampman
  * @author jeen
  */
@@ -20,10 +20,10 @@ public interface Query extends Operation {
 
 	/**
 	 * The different types of queries that RDF4J recognizes: boolean queries, graph queries, and tuple queries.
-	 * 
+	 *
 	 * @since 3.2.0
 	 */
-	public enum QueryType {
+	enum QueryType {
 		/**
 		 * Boolean queries (such as the SPARQL ASK query form) return either {@code true} or {@code false} as the
 		 * result.
@@ -44,21 +44,44 @@ public interface Query extends Operation {
 	/**
 	 * Specifies the maximum time that a query is allowed to run. The query will be interrupted when it exceeds the time
 	 * limit. Any consecutive requests to fetch query results will result in {@link QueryInterruptedException}s.
-	 * 
+	 *
 	 * @param maxQueryTime The maximum query time, measured in seconds. A negative or zero value indicates an unlimited
 	 *                     query time (which is the default).
 	 * @deprecated since 2.0. Use {@link Operation#setMaxExecutionTime(int)} instead.
 	 */
 	@Deprecated
-	public void setMaxQueryTime(int maxQueryTime);
+	void setMaxQueryTime(int maxQueryTime);
 
 	/**
 	 * Returns the maximum query evaluation time.
-	 * 
+	 *
 	 * @return The maximum query evaluation time, measured in seconds.
 	 * @see #setMaxQueryTime(int)
 	 * @deprecated since 2.0. Use {@link Operation#getMaxExecutionTime()} instead.
 	 */
 	@Deprecated
-	public int getMaxQueryTime();
+	int getMaxQueryTime();
+
+	// TODO - make default with no-op before merging to develop for backwards compatibility
+	// Also wondering if it makes sense to move this up a level to the Operation interface? Since we will be using a
+	// default no-op implementation we could move it without affecting backwards compatibility.
+	QueryExplainWrapper explain(QueryExplainLevel queryExplainLevel);
+
+	/**
+	 * The different types of queries that RDF4J recognizes: boolean queries, graph queries, and tuple queries.
+	 *
+	 * @since 3.2.0
+	 */
+	enum QueryExplainLevel {
+		Unoptimized,
+		Optimizaed,
+		Executed
+	}
+
+	interface QueryExplainWrapper {
+
+		String asHumanReadbleString();
+		// TupleExpr asTupleExpr(); location in maven hierarchy prevents us from using TupleExpr here
+	}
+
 }
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
@@ -80,7 +80,7 @@ public interface Query extends Operation {
 
 	interface QueryExplainWrapper {
 
-		String asHumanReadbleString();
+		String toString();
 		// TupleExpr asTupleExpr(); location in maven hierarchy prevents us from using TupleExpr here
 	}
 
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
@@ -75,6 +75,13 @@ public interface Query extends Operation {
 	 * </p>
 	 *
 	 * <p>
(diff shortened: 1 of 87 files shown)


core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
```diff
@@ -12,18 +12,18 @@
...
  */
 public interface Query extends Operation {
 
 	/**
 	 * The different types of queries that RDF4J recognizes: boolean queries, graph queries, and tuple queries.
-	 * 
+	 *
 	 * @since 3.2.0
 	 */
-	public enum QueryType {
```
hmottestad, line 26: Is public by definition.

core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
```diff
@@ -44,21 +44,44 @@
...
 	 * limit. Any consecutive requests to fetch query results will result in {@link QueryInterruptedException}s.
-	 * 
+	 *
 	 * @param maxQueryTime The maximum query time, measured in seconds. A negative or zero value indicates an unlimited
 	 *                     query time (which is the default).
 	 * @deprecated since 2.0. Use {@link Operation#setMaxExecutionTime(int)} instead.
 	 */
 	@Deprecated
-	public void setMaxQueryTime(int maxQueryTime);
+	void setMaxQueryTime(int maxQueryTime);
```
hmottestad, line 53: all interface methods are public by definition

core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
```diff
@@ -44,21 +44,44 @@
...
+		Unoptimized,
+		Optimizaed,
+		Executed
+	}
+
+	interface QueryExplainWrapper {
+
+		String asHumanReadbleString();
+		// TupleExpr asTupleExpr(); location in maven hierarchy prevents us from using TupleExpr here
+	}
```
hmottestad, line 85: New explain method, enums and returned interface.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/AbstractQueryPreparer.java
```diff
@@ -126,6 +126,11 @@ public boolean evaluate() throws QueryEvaluationException {
 				}
 			}
 		}
+
+		@Override
+		public QueryExplainWrapper explain(QueryExplainLevel queryExplainLevel) {
+			return null;
+		}
```
hmottestad, line 133: These are just placeholders for the time being.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailTupleQuery.java
```diff
@@ -84,4 +85,23 @@ public void evaluate(TupleQueryResultHandler handler)
...
+				getIncludeInferred());
+
+		return new QueryExplainWrapper() {
+			@Override
+			public String asHumanReadbleString() {
+				return explainedTupleExpr.toString();
+			}
+		};
+
+	}
```
hmottestad, line 106: This pretty much delegates to the sail connection in the same way as `eveluate()` does.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/SailConnection.java
```diff
@@ -441,4 +441,7 @@ default void removeStatement(UpdateContext op, Resource subj, URI pred, Value ob
 	@Deprecated
 	boolean pendingRemovals();
 
+	// TODO - make this a default no-op for backwards compatibility
+	TupleExpr explain(Query.QueryExplainLevel queryExplainLevel, TupleExpr tupleExpr, Dataset activeDataset,
+			BindingSet bindings, boolean includeInferred);
```
hmottestad, line 446: interface method for sail connections

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -255,6 +253,35 @@ protected EvaluationStrategy getEvaluationStrategy(Dataset dataset, TripleSource
...
+
+			return strategy.optimize(tupleExpr, store.getEvaluationStatistics(), bindings);
+
+		case Unoptimized:
+			return tupleExpr;
+		}
+
+		throw new UnsupportedOperationException("Unsupported queryExplainLevel: " + queryExplainLevel);
+	}
+
```
hmottestad, line 284: Implementation of getting the query plan.

A few things to note:
 - we use the evaluate() method to run the query instead of duplicating the code. We might need to duplicate it though since we need access to the evaluation strategy to enable tracking of result size when that gets merged in.
 - when using evaluate() it would clone tupleExpr, that had to be removed if we don't want to duplicate a lot of code
 - the code in `Optimized` is copied from evaluate, not sure if this is want

hmottestad:
I would appreciate your input on this guys. I'm not ready for a full review, but more would like you to consider the general approach.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
```diff
@@ -0,0 +1,73 @@
...
+		SailRepository sailRepository = new SailRepository(new MemoryStore());
+		try (SailRepositoryConnection connection = sailRepository.getConnection()) {
+			TupleQuery tupleQuery = connection.prepareTupleQuery(TUPLE_QUERY);
+
+			String queryPlan1 = tupleQuery.explain(Query.QueryExplainLevel.Unoptimized).asHumanReadbleString();
+
+			System.out.println(queryPlan1);
+
+		}
+	}
```
hmottestad, line 34: This is how it would be used.

core/query/src/main/java/org/eclipse/rdf4j/query/Query.java
```diff
@@ -44,21 +44,44 @@
...
+	enum QueryExplainLevel {
+		Unoptimized,
+		Optimized,
+		Executed
+	}
+
+	interface QueryExplainWrapper {
+
+		String asHumanReadbleString();
+		// TupleExpr asTupleExpr(); location in maven hierarchy prevents us from using TupleExpr here
```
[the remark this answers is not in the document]
hmottestad: I’m trying to consider what we could do in the future to give advanced users more control of their queries. 

An example is a query I have which takes hours to complete, but it’ll easily run in milliseconds if certain query optimizers are turned off. 

Other databases have query hints that users can add to their sparql queries for this purpose. 

I was playing with the idea of exposing the tupleexpr to facilitate some similar, but on a lower level. 

Also been thinking about using the resultSizeActual for optimizing a query that gets run frequently.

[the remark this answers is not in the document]
jeenbroekstra: Sounds interesting, but I think we're getting a bit far afield here from the scope. However, I can see that potentially tying it in more closely with TupleExpr is something we might want to keep our options open about. A simple solution would be to move this QueryExplanation interface into the query.algebra package (as part of the rdf4j-queryalgebra-model module). It's defensible to put it there anyway I think, and that way we will have the option in future to expose relevant bits of the query algebra model via this interface.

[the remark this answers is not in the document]
hmottestad: Screenshot 2020-04-25 at 13 36 29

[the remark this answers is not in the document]
hmottestad: I cant move the interface from the `query` package to the `queryalgebra` package because that would cause this circular dependency. I haven't looked into it yet though.

hmottestad:
This is coming along nicely. 

I have introduced a GenericPlanNode which can be used to represent any and all plan nodes. It has toString support equivalent to the current TupleExpr toString format. It also supports json (using Jackson).

Also supports ASK and CONSTRUCT. 

There is still one issue. The use of a mutable TupleExpr instead of an immutable one. I think I will look closer at this and maybe create a new evaluate method in SailConnection to support more options.

hmottestad:
I feel that this might start getting ready to merge.

hmottestad:
I’ve also added time tracking, so the user can see how much time each sub part takes. I find this particularly useful. 

I’ve quickly tested the overhead, and it was non-existent on my MacBook. I know that nanoTime used to be super slow on windows back in the XP time, but seems that all modern hardware and operating systems now support very low overhead nanosecond timing.

hmottestad:
I don’t want to implement support for remote explain() in this PR. Would be great if this initial implementation could be released as part of 3.2 and maybe I’ll have time to implement the remote feature too.

hmottestad:
TODO:
- [ ] check that methods that implement interface method with `@Experimental` show this in their javadcos

hmottestad:
Just waiting for Github Actions to merge this.

Issue #2107, opened 2020-04-19
NativeStore DataFile should predict the size of a read


hmottestad:
In DataFile the method `byte[] getData(long offset)` reads data from disk in the following format:

```
Data:  7   abcdefg
Disk: |....|.......|
```

The first 4 bytes represent the size (7) of the values (abcdefg) to be read.

This means that we first need to read 4 bytes to determine how much more to read.

If we instead predict the size, we can read both the size and the data in one operation and then extract the relevant data.

Say we in the example above predict size=20, we then read more data than needed. This isn't a problem because we can see that we only needed to read 4+7 bytes and can easily extract 7 bytes started on index +4.

PR #2108, opened 2020-04-19
GH-2107 predict the size of the data that needs to be read in DataFile

hmottestad:
GitHub issue resolved: #2107 

In DataFile try to predict the size of the data that needs to be read so that we can get away with a single read. This improves performance because even though the operating system might be smart, it's still cheaper to get 50 bytes in one read than to get 4+20 in two reads.

---- 
PR Author Checklist: 

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/DataFile.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/DataFile.java
@@ -7,6 +7,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.nativerdf.datastore;
 
+import org.eclipse.rdf4j.common.io.NioFile;
+
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


hmottestad:
I have three benchmark queries:
 - group by query: 25% faster
 - complex query: 6% faster
 - distinct predicates query: 35% faster

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/DataFile.java
```diff
@@ -121,29 +121,61 @@ public long storeData(byte[] data) throws IOException {
 		return offset;
 	}
 
+	// This variable is used for predicting the number of bytes to read in getData(long offset). This helps us to only
+	// need to execute a single IO read instead of first one read to find the length and then one read to read the data.
+	int dataLengthApproximateAverage = 25;
```
jeenbroekstra, line 126: Very clever. Is 25 as a starting point based on any sort of observation of what is "typical"? Or is it more a "we need to start somewhere"?

jeenbroekstra:
I'm curious: what made you decide to pick this up? Are you just trawling through TODO comments in the code base?

hmottestad:
Wanting to do a bigger SHACL benchmark. Just loading 250 million triples took all night. Then querying for all distinct predicates took 1 1/2 minutes.

At least with the distinct predicates workload 50 seemed to be close to max. Since we load in 2x the estimated average I started it at 25. For the complex workload I played around with the max a bit. 200 seemed to be the sweet spot. 

I’m also looking into caching the IDFile.

jeenbroekstra:
It's been a while since I looked at the guts of the native store but IIRC the ids are cached at the individual resource level - the NativeStoreValueFactory creates instances of IRIs and literal that cache their own  ids internally.

Issue #2111, opened 2020-04-20
Cache IDFile in NativeStore as much as possible


hmottestad:
A large portion of the time used to retrieve data from the NativeStore is used to lookup IDs to get offsets in the IDFile. The values.id file in a default NativeStore if typically quite small, ~390 MB for a NativeStore with ~250 000 000 triples. This compared to the triples indexes that are ~6 GB. 

Caching reads from this file will help performance, as will proactive read aheads.

Complete now with results of 40% speed improvement for queries that read large amounts of data.

PR #2112, opened 2020-04-20
GH-2111 cache IDFile

hmottestad:
GitHub issue resolved: #2111 

Cache and read ahead in the IDFile to reduce number of IO opertions.

---- 
PR Author Checklist: 

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/IDFile.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/IDFile.java
@@ -7,18 +7,20 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.nativerdf.datastore;
 
+import org.apache.commons.collections4.map.ReferenceMap;
+import org.eclipse.rdf4j.common.io.NioFile;
+
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
-
-import org.eclipse.rdf4j.common.io.NioFile;
+import java.util.Map;
 
 /**
  * Class supplying access to an ID file. An ID file maps IDs (integers &gt;= 1) to file pointers (long integers). There
  * is a direct correlation between IDs and the position at which the file pointers are stored; the file pointer for ID X
  * is stored at position 8*X.
- * 
+ *
  * @author Arjohn Kampman
  */
 public class IDFile implements Closeable {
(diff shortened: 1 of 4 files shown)


hmottestad:
## Benchmarks

### Before
Screenshot 2020-04-20 at 13 17 17

### After
Screenshot 2020-04-20 at 12 25 55

Fairly large benefit for the `distinctPredicateQuery` and `groupByQuery`.

hmottestad:
One fairly important question though is if the NativeStore concurrently reads or write to the IDFile, or if that is controlled by locks. In which case we might need to look at introducing synchronization to IDFile.

jeenbroekstra:
> One fairly important question though is if the NativeStore concurrently reads or writes to the IDFile, or if that is controlled by locks. In which case we might need to look at introducing synchronization to IDFile.

Reads are concurrent on a nativestore, and that goes right down to the ID file I think.

On write operations, the NativeSailStore (or rather, NativeSailSink) obtains an exclusive lock before calling `DataStore.sync()`- which is normally the point at which changes are flushed to the actual file on disk.

hmottestad:
The first thing this does is reduce the number of IO operations. Previously you would need one per ID, now it’s one per 4000 or so (need to test more and adjust that number for performance). The number of IO operations matters greatly. I’m on SSD and it even matters then. For spinning disks this should have a massive impact. 

The cache also scales with the amount of memory you have. Caching everything if you have the space. 

This makes the biggest difference when you are accessing large amounts of data. That’s why the distinct predicates query and the group by query are affected. 

I’ll have to look at the current caching scheme more to find out what kind of affect it is actually having compared to this.

hmottestad:
Here are two flamegraphs. I've bumped up the ID cache to 16 million, same for values cache and also bumped the namespace cache to 4 000. I think there is something funky with the namespace cache. I'll look into that later.

Here you can see that even with the massive ID cache we are still spending 21% of all our time looking up ID offsets.

Screenshot 2020-04-21 at 08 50 40

With the low lever caching in this branch that number is down to 4.5%.

Screenshot 2020-04-21 at 08 50 26

hmottestad:
I ready to merge this now.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/IDFile.java
```diff
@@ -105,52 +132,111 @@ public final File getFile() {
...
+
+			synchronized (this) {
+				// we try not to overwrite an existing cache line
+				if (!cache.containsKey(cacheLineLookupIndex)) {
+					cache.put(cacheLookupIndex, cacheLine);
+				}
+			}
+
+			gcReducingCache = cacheLine;
+			gcReducingCacheIndex = cacheLookupIndex;
```
hmottestad, line 222: Should probably synchronise this too.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/IDFile.java
```diff
@@ -105,52 +132,111 @@ public final File getFile() {
...
+		int cacheLineLookupIndex = id % cacheLineSize;
+
+		// the cache line which is of size cacheLineSize
+		Long[] cacheLine = getCacheLine(cacheLookupIndex);
+
+		if (cacheLine != null) {
+			return cacheLine[cacheLineLookupIndex];
+		}
+
+		// We only cache complete lines og size cacheLineSize. This means that the last line in the file will almost
```
hmottestad, line 204: Typo og=of

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/IDFile.java
```diff
@@ -105,52 +132,111 @@ public final File getFile() {
...
 
 	/**
 	 * Discards all stored data.
-	 * 
+	 *
 	 * @throws IOException If an I/O error occurred.
 	 */
 	public void clear() throws IOException {
 		nioFile.truncate(HEADER_LENGTH);
+		nioFileSize = nioFile.size();
```
hmottestad, line 239: Clear Cache!

hmottestad:
I’m getting ready to merge this soon.

Issue #2121, opened 2020-04-23
Group statement adds together in single INSERT DATA in SPARQLConnection


jeenbroekstra:
SPARQLConnection handles transaction by serializing updates to a sequence of SPARQL update operations. However, it treats each statement add/remove as a separate update, resulting in update sequences with a separate INSERT DATA clause for each triple. This is highly inefficient. 

A possible fix is to cache pending added/removed statements and only flush them to a serialized INSERT DATA operation when necessary.

hmottestad:
On this note I think adding triple "A" in a transaction, then removing triple "A" again in the same transaction should be simplified to a no-op in the cases were there are no prior or subsequent updates or read queries.

jeenbroekstra:
> On this note I think adding triple "A" in a transaction, then removing triple "A" again in the same transaction should be simplified to a no-op in the cases were there are no prior or subsequent updates or read queries.

I see what you're getting at, but given that the approach is to flush all added statements as soon as a remove comes in, we gain very little: only if the first statement in the remove batch happens to match one of the pending adds could we optimize, and even then we only optimize by only scrapping _one_ triple. I think the additional complexity in this case is not really worth it.

hmottestad:
Ok

PR #2126, opened 2020-04-24
GH-2121 group adds/removes together in sparql update sequence

jeenbroekstra:
GitHub issue resolved: #2121 

- SPARQLConnection internally keeps tracks of pending adds / removals. When either commit is called or an add is interleaved with remove (vice versa) the pendings adds/removals get flushed into a single INSERT/DELETE DATA operation
- added one or two unit tests to check content and sequence of update operations in the produced SPARQL string.

---- 
PR Author Checklist: 

### DIFF
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -89,18 +89,16 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 
 	private final SPARQLProtocolSession client;
 
+	private ModelFactory modelFactory = new DynamicModelFactory();
+
 	private StringBuilder sparqlTransaction;
 
 	private Object transactionLock = new Object();
 
-	private Model pendingAdds = this.getModelFactory().createEmptyModel();
-	private Model pendingRemoves = this.getModelFactory().createEmptyModel();
+	private final Model pendingAdds;
+	private final Model pendingRemoves;
 
 	private final boolean quadMode;
-
-	private ModelFactory getModelFactory() {
-		return new DynamicModelFactory();
-	}
 
 	public SPARQLConnection(SPARQLRepository repository, SPARQLProtocolSession client) {
 		this(repository, client, false); // in triple mode by default
@@ -110,6 +108,8 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 		super(repository);
 		this.client = client;
 		this.quadMode = quadMode;
+		this.pendingAdds = getModelFactory().createEmptyModel();
+		this.pendingRemoves = getModelFactory().createEmptyModel();
 	}
 
 	@Override
@@ -179,9 +179,7 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 			});
 			allGood = true;
 			return result;
-		} catch (MalformedQueryException |
-
-				QueryEvaluationException e) {
+		} catch (MalformedQueryException | QueryEvaluationException e) {
 			throw new RepositoryException(e);
 		} finally {
 			if (!allGood) {
@@ -996,6 +994,10 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 			}
 
 		};
+	}
+
+	private ModelFactory getModelFactory() {
+		return modelFactory;
 	}
 
 }
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -87,6 +87,8 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 
 	private static final String NAMEDGRAPHS = "SELECT DISTINCT ?_ WHERE { GRAPH ?_ { ?s ?p ?o } }";
 
+	private static final int DEFAULT_MAX_PENDING_SIZE = 1000000;
+
 	private final SPARQLProtocolSession client;
 
 	private ModelFactory modelFactory = new DynamicModelFactory();
@@ -95,8 +97,10 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 
 	private Object transactionLock = new Object();
 
-	private final Model pendingAdds;
-	private final Model pendingRemoves;
+	private Model pendingAdds;
+	private Model pendingRemoves;
+
+	private int maxPendingSize = DEFAULT_MAX_PENDING_SIZE;
 
 	private final boolean quadMode;
 
@@ -108,8 +112,6 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 		super(repository);
 		this.client = client;
 		this.quadMode = quadMode;
-		this.pendingAdds = getModelFactory().createEmptyModel();
-		this.pendingRemoves = getModelFactory().createEmptyModel();
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -827,39 +819,71 @@ public boolean isActive() throws UnknownTransactionStateException, RepositoryExc
...
-		String sparqlCommand = createInsertDataCommand(list, contexts);
+	private void flushPendingAdds() {
+		if (!pendingAdds.isEmpty()) {
+			for (Resource context : pendingAdds.contexts()) {
+				String sparqlCommand = createInsertDataCommand(pendingAdds.getStatements(null, null, null, context),
+						context);
+				sparqlTransaction.append(sparqlCommand);
+				sparqlTransaction.append("; ");
+			}
+			pendingAdds.clear();
```
hmottestad, line 860: I think .contexts() upgrades the DynamicModel to a HashModel. This is probably not going to be what is performance critical in this situation though. I'm also unsure if .clear() downgrades the model again.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -827,39 +819,71 @@ public boolean isActive() throws UnknownTransactionStateException, RepositoryExc
...
-			ValueFactory f = getValueFactory();
-
-			Statement st = f.createStatement(subject, predicate, object);
+		flushPendingAdds();
 
-			List<Statement> list = new ArrayList<>(1);
-			list.add(st);
-			sparqlCommand = createDeleteDataCommand(list, contexts);
+		if (subject != null && predicate != null && object != null) {
+			pendingRemoves.add(subject, predicate, object, contexts);
```
hmottestad, line 880: Bit hazy in the morning but I wonder how you handle removing from all contexts (implicitly).

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -972,4 +996,8 @@ protected Statement convert(BindingSet b) throws QueryEvaluationException {
 		};
 	}
 
+	private ModelFactory getModelFactory() {
+		return modelFactory;
+	}
```
hmottestad, line 1001: Do we need a private getter?

jeenbroekstra:
> Change generally looks good: this is a good and useful performance improvement.
> 
> Two comments for discussion and consideration
> 
>     * this change will make the SPARQLConnection less thread-safe. I think this kind of worked so far for the SPARQLRepository. Not sure if people used connections accross threads, but do we need to state this more explicitly in the release notes?

We actually are pretty explicit about RepositoryConnections in general not being thread-safe, by design. See https://rdf4j.org/documentation/programming/repository/#multithreaded-repository-access and also the javadoc (https://rdf4j.org/javadoc/latest/index.html?org/eclipse/rdf4j/repository/RepositoryConnection.html ). But if there's an easy way to make it even more explicit, it can't hurt of course.
 
>     * as of this change you are potentially buffering the entire update operation of a transaction in a model in memory (at least when this add-only). There won't be any issues with reasonably sized transactions. But how about larger chunks (if anybody tried to use this). Should we maybe add some logic to auto-flush after N additions (e.g. after 1M triples being added)? Maybe this barrier can be made configurable.

Good point. Shouldn't be hard, I'll look into that.

>       I cdo not have enough insights right now to fully understand the impact on the generated HTTP communication, i.e. if this supports large requests. In the end everything is converted to a SPARQL string which needs to be processed by the endpoint. So some kind of barrier may be useful.

At the end of the day, we are hampered by the fact that SPARQL protocol does not support transactions. This approach of serializing everything in a single update sequence is the closest we can come to make each transaction atomic.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
jeenbroekstra: Not necesarily, I think I originally made it public and then changed my mind :)

core/repository/sparql/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnectionTest.java
```diff
@@ -48,4 +53,52 @@ public void commitOnEmptyTxnDoesNothing() throws Exception {
...
+
+		String expectedSequence = "INSERT DATA[^{]*\\{[^}]*\\}[^D]+DELETE DATA[^{]*\\{[^}]*\\}[^I]+INSERT DATA.*";
+
+		assertThat(sparqlUpdate).containsPattern(expectedSequence);
+		assertThat(sparqlUpdate).contains(expectedAddedTriple1)
+				.contains(expectedAddedTriple2)
+				.contains(expectedAddedTriple3)
+				.contains(expectedRemovedTriple1);
+
+	}
```
hmottestad, line 103: Good we have some tests. Do we have any integration tests that would cover this too? Benchmarks would also be great, so we don't have a regression of this some time in the future.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
jeenbroekstra: > I'm also unsure if .clear() downgrades the model again.

That might be an interesting improvement to DynamicModel. I mostly just used a model here because it's a simple way to keep track of statements per context. I could perhaps also use a Multimap of some sort if this turns out to be heavy.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
jeenbroekstra: Ah - good call. I'll write a test case for that.

core/repository/sparql/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnectionTest.java
jeenbroekstra: Yes we have quite extensive integration tests for this, see rdf4j-repository-compliance. 

I don't think we have benchmarks though. I'll see what I can do this weekend. But I must admit I suck at benchmarks.

hmottestad:
Does the remote repo for the RDF4J server support transactions?

jeenbroekstra:
> Does the remote repo for the RDF4J server support transactions?

Yes. See https://rdf4j.org/documentation/rest-api/#transaction-operations . The HTTPRepository fully implements this REST API from the client side.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
jeenbroekstra: Just to be clear: this is in fact handled by just mapping to a DELETE DATA operation without a GRAPH clause. Funnily enough the integration tests didn't cover this specific case, so I've added a few unit tests as well as a new integration test.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
jeenbroekstra: For now I've changed it slightly to just drop the model and create a new empty one when flushing.

jeenbroekstra:
I realize I'm breaking the "wait for approval" rule but I want to get our milestone build out today and want this in, so I'll be merging this. I've followed your suggestions for improvements and everything currently greenlines.

Issue #2140, opened 2020-04-27
Execution of SPARQL query (as part of a reasoning task) does not finish on version 3.1.3 (rdf4j)


swc-kdzekov:
Hello,
We updated our system rdf4j library from version 2.3.3 to version 3.1.3. Now, we are facing an issue running this query (which is part of reasoning process)

```
PREFIX h:<urn:pp:internalHelper/>
PREFIX owl:<http://www.w3.org/2002/07/owl#>
PREFIX rdf:<http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
PREFIX skos:<http://www.w3.org/2004/02/skos/core#>
PREFIX skosxl:<http://www.w3.org/2008/05/skos-xl#>
PREFIX swcs:<http://schema.semantic-web.at/ppt/>
CONSTRUCT {?c a ?t. ?c a ?t2. } 
FROM <https://performance.sqa.semantic-web.at/MeSH2016-SQAGPS_ABC-localMemory/thesaurus> 
FROM <https://performance.sqa.semantic-web.at/MeSH2016-SQAGPS_ABC-localMemory/thesaurus/reasoning#7a9e2803-4ad7-4346-a9aa-e9fb4192b71b> 
WHERE {
    ?c a ?t.
    FILTER (?t IN (<https://pp-remotestore-test.semantic-web.at/SQA-GPO/GenericClass>))
    FILTER NOT EXISTS {
        ?c (swcs:appliedType|swcs:propagateType)/rdfs:subClassOf* ?t
    }
    FILTER NOT EXISTS {
        ?c ^skos:member ?root. ?root swcs:appliedType/rdfs:subClassOf* ?t
    }
    FILTER NOT EXISTS {
        ?root a skos:OrderedCollection; 
                skos:memberList ?member; 
                <http://schema.semantic-web.at/ppt/appliedType>/rdfs:subClassOf* ?t. 
        ?member rdf:rest*/rdf:first ?c.     }
    FILTER NOT EXISTS {
        ?c (skos:broader|^skos:narrower)+ ?root. ?root swcs:propagateType/rdfs:subClassOf* ?t
    }
    FILTER NOT EXISTS {
        ?c (skos:broader|^skos:narrower)*/(skos:topConceptOf|^skos:hasTopConcept) ?root. ?root swcs:appliedType/rdfs:subClassOf* ?t
    }
[...]

PR #2141, opened 2020-04-28
GH-2140 use HashJoinIteration instead of doing loop join

jeenbroekstra:
GitHub issue resolved: #2140 

- moved decision on join strategy to single place
- hope/expectation is that using HashJoinIteration will be more efficient than the previous loop join strategy.

This is a work in progress as there are likely corner cases that will fail. We'll also need to run some performance tests to see if it actually makes a difference (although I'm in favor of this change regardless as it's a cleaner approach).

---- 
PR Author Checklist: 

### DIFF
--- a/compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/federation/FederationSPARQL10QueryTest.java
+++ b/compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/federation/FederationSPARQL10QueryTest.java
@@ -22,12 +22,14 @@ public class FederationSPARQL10QueryTest extends SPARQLQueryTest {
 	public static Test suite() throws Exception {
 		return SPARQL10ManifestTest.suite(new Factory() {
 
+			@Override
 			public FederationSPARQL10QueryTest createSPARQLQueryTest(String testURI, String name, String queryFileURL,
 					String resultFileURL, Dataset dataSet, boolean laxCardinality) {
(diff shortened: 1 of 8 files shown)


hmottestad:
Memory could be an issue. If we have enough memory we should use hash join, if not we should use loop join.

jeenbroekstra:
I may re-introduce the loop join option, but I would make it a separate class that lives next to JoinIterator and HashJoinIteration.

hmottestad:
Feel free to merge or rebase develop and bring in the test I created on my branch. It made it very easy to see what was going on.

jeenbroekstra:
That's tricky, as this PR is against master, not develop.

jeenbroekstra:
Sigh - there's always one compliance test that I miss.

hmottestad:
> That's tricky, as this PR is against master, not develop, and I think your test relies on changes you made in the develop branch.

Yes, my test relies on the new explain api. I would recommend using the test and targeting develop instead, makes testing a lot easier. 3.2.0 is also very close to release, and you want to get it out the door before then you can instead do a new milestone release.

If you want it as part of 3.1.x I would suggest having two commits, first one is the fix and the second one is the test. Then you can cherry pick the first commit into master. Or you could always do it the other way around, you fix it here on master and I merge only the test on develop.

jeenbroekstra:
I am at a loss why that one test fails for the old federation sail. I can't even reproduce it when I try the same query (with same data) in a seperately set up unit test running on a federation. 

I'll disable that test for now and merge.

Issue #2145, opened 2020-04-28
Named Graph Support in FedX federation engine


aschwarte10:
Currently the federation engine operates in triple mode, and to large parts ignores any named graph handling.

As a user I would like to have proper quad mode support in the federation engine, i.e.

* retrieval of statements from a given context retrieves only the expected data;
* adding/removing statement from a given context modifies only the specified context;
* SPARQL queries/updates involving GRAPH or FROM (NAMED) clauses work as expected

aschwarte10:
From my point of view we are good for the 3.2 release. Closing this issue

PR #2146, opened 2020-04-28
GH-2145: named graph support in federation engine

aschwarte10:
GH-2145: fix context retrieval in federation engine

* context IDs retrieved from the endpoints have to be materialized as
they are only available while the connection is open
* add unit test infrastructure for named graph tests and define some
test data

GH-2145: add named graph handling for #getStatements API

Add named graph support to the RepositoryConnection API using the
getStatements method. Also covered with various unit tests.

Note that this change provides already the infrastructure to further add
support for named graph queries.

GH-2145: add named graph handling for simple federated SELECT patterns

* support GRAPH clause around single basic graph pattern
* support FROM clause in queries

* add named graph handling for simple federated SELECT patterns
** support GRAPH clause around single basic graph pattern
** support FROM clause in queries

* fix handling of base URIs for query parsing in FedX

* add named graph support for bound joins

GH-2149 GH-2145 use new compliance test suite for testing of FedX

GitHub issue resolved: #2145

---- 
PR Author Checklist: 

### DIFF
--- a/compliance/sparql/pom.xml
+++ b/compliance/sparql/pom.xml
@@ -120,6 +120,11 @@
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-tools-federation</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
--- a/compliance/sparql/src/test/java/org/eclipse/rdf4j/query/parser/sparql/manifest/W3CApprovedSPARQL11QueryTest.java
+++ b/compliance/sparql/src/test/java/org/eclipse/rdf4j/query/parser/sparql/manifest/W3CApprovedSPARQL11QueryTest.java
@@ -14,9 +14,11 @@ import org.eclipse.rdf4j.repository.Repository;
 import org.eclipse.rdf4j.repository.dataset.DatasetRepository;
 import org.eclipse.rdf4j.repository.sail.SailRepository;
 import org.eclipse.rdf4j.sail.memory.MemoryStore;
+import org.junit.Ignore;
 
 import junit.framework.Test;
 
+@Ignore("replaced by org.eclipse.rdf4j.sail.memory.MemorySPARQL11QueryComplianceTest")
 public class W3CApprovedSPARQL11QueryTest extends SPARQLQueryTest {
 
 	public static Test suite() throws Exception {
--- a/compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/federation/FedXSPARQL11QueryComplianceTest.java
+++ b/compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/federation/FedXSPARQL11QueryComplianceTest.java
@@ -1,0 +1,81 @@
+/******************************************************************************* 
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.federation;
+
+import java.io.IOException;
+
+import org.eclipse.rdf4j.federated.FedXFactory;
+import org.eclipse.rdf4j.federated.repository.FedXRepository;
+import org.eclipse.rdf4j.query.Dataset;
+import org.eclipse.rdf4j.query.parser.sparql.manifest.SPARQL11QueryComplianceTest;
+import org.eclipse.rdf4j.repository.Repository;
+import org.eclipse.rdf4j.repository.RepositoryException;
+import org.eclipse.rdf4j.repository.config.RepositoryConfig;
+import org.eclipse.rdf4j.repository.config.RepositoryConfigException;
+import org.eclipse.rdf4j.repository.config.RepositoryImplConfig;
+import org.eclipse.rdf4j.repository.dataset.DatasetRepository;
+import org.eclipse.rdf4j.repository.manager.RepositoryManager;
+import org.eclipse.rdf4j.repository.manager.RepositoryProvider;
+import org.eclipse.rdf4j.repository.sail.config.SailRepositoryConfig;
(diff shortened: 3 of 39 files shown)


jeenbroekstra:
I've also added the generic W3C SPARQL 1.1 compliance tests for FedX. We may end up taking that out again if it gets in the way, but we're currently only failing 5 tests, which is pretty good going.

aschwarte10:
> I've also added the generic W3C SPARQL 1.1 compliance tests for FedX. We may end up taking that out again if it gets in the way, but we're currently only failing 5 tests, which is pretty good going.

Can you shortly comment on how the test data is distributed amongst the federation members?

jeenbroekstra:
> > I've also added the generic W3C SPARQL 1.1 compliance tests for FedX. We may end up taking that out again if it gets in the way, but we're currently only failing 5 tests, which is pretty good going.
> 
> Can you shortly comment on how the test data is distributed amongst the federation members?

It currently just uploads all data into the first federation member, which admittedly is not very exciting. But it shouldn't be too hard to tweak the test rig a little so that we can distribute by named graph, so that at least the tests that involve named graphs in their dataset will end up querying multiple members of the federation.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FederationManager.java
```diff
@@ -332,7 +342,10 @@ public void updateStrategy() {
 		}
 
 		if (updated) {
-			strategy = FederationEvaluationStrategyFactory.getEvaluationStrategy(type, federationContext);
+			// TODO problem here: we need to (optionally) inject a Dataset object into the EvaluationStrategy, which is
+			// query-specific and determines what parts of the store are relevant for the query. However this part of
+			// the code assumes that we can instantiate an EvaluationStrategy once and then reuse it for each query.
```
jeenbroekstra, line 347: @aschwarte10 I could use your eyes on this problem. FedX is currently designed to instantiate the EvaluationStrategy once (unless the config changes). But in RDF4J Sails, it is generally assumed that each query gets its own EvaluationStategy - that is why the `Dataset` object is passed in on the constructor. I believe that this is caused some of the named graph queries to fail on FedX. I'd like to hear your opinion on the best way forward: can we just remove the assumption that the strategy is defined once and let FedX instantiate a new strategy for each query? 

I'm a little way of doing something like just setting/overwriting the Dataset  assigned to the strategy on each query, as I'm not sure what the consequences would be in a multithreaded environment.

aschwarte10: Hmm, I see. I wasn't aware of this assumption when initially implementing FedX.

Generally, I do not see a bigger problem in instantiating the strategy for each query (except for the amount of objects). However, quickly scanning the code and the references to where the strategy is passed around, such a change would be a huge refactoring.

I am not sure if at this time right before the release it is such a good idea to do this refactoring. If at all, I would suggest to do it outside of this PR to have a cleaner and easier change.

However, I think we can cope without it: the query specific QueryInfo object always has a reference to the externally passed Dataset. And this QueryInfo is available basically everywhere at evaluation time. This also how all the other named graph queries are working.

Let's synch offline on this and come to the best conclusion.

Note: I will probably re-order your last commit on top (i.e. a force-push), to have some other must changes integrated. The we have the possibility to decide on this special thing

jeenbroekstra: I agree a full refactoring is out of scope (and fwiw, the assumption I'm talking about isn't particularly well documented or obvious, so not blaming your design here). I may have an idea for a quick workaround fix. If I can get that to work I'll push it, otherwise I'll revert the WIP changes I made and log this as a known bug.

jeenbroekstra:
Github is reporting test failures but it greenlines when I run it locally.

aschwarte10:
> Github is reporting test failures but it greenlines when I run it locally.
> 
> **edit** I just realized that the tests in FedX are apparently set up to run several times, using different repository configurations. The test succeeds when run on a memory store, but fails on a native store.

Your understanding is right (but the issue is probably unrelated to NativeStore): it is more about the chosen strategy.
Tests are executed three times in different environments

1) Federation members as SPARQL endpoints => SPARQL strategy
2) Federatuin members as RDF4J  HTTP repository => SPARQL strategy
3) Federation members as native store => Sail strategy

The SailStrategy has currently an issue (already identified), the pointer is in https://github.com/eclipse/rdf4j/pull/2146/commits/ed19bbb4e4662535fe682bb2335961ccc0e49019

Once this is addressed the test should also be green.

Note: you can also locally in your IDE switch the environment, see `org.eclipse.rdf4j.federated.SPARQLServerBaseTest.REPOSITORY_TYPE`(i.e. temporarily set it to another value using the enum)

aschwarte10:
We are now in a good shape to get this PR merged. I will clean the commit history and force-push a locally squashed commit containing the entire change.

aschwarte10:
I have cleaned the branch history (squashed and rebased on latest develop). Note that I have split the content into two commits (=> changes to FedX, and changes to compliance test). A backup of the change before the squash is available in https://github.com/aschwarte10/rdf4j/commits/issues/%232145-named-graph-support-backup

The federation tests are now green locally.

Note: the CI check is currently failing due to an independent issue in the MemoryStore module: the issue is available on the develop branch already (maybe introduced to some backmerge from master). Once this is fixed and available in this branch through a rebase, this PR should also be greenlined. For reference: the test is `org.eclipse.rdf4j.sail.memory.QueryPlanRetrievalTest`

@jeenbroekstra maybe you can double check this PR and review.

Open issues for follow ups (which we explicitly do outside this PR):

* fix FROM NAMED handling for SailStrategy (=> #getStatements implementation)
* add support for GRAPH ?g { } (i..e a variable graph)
* Optionally: consider support to deactivate GRAPH handling (i.e. only triple mode)

jeenbroekstra:
> I have cleaned the branch history (squashed and rebased on latest develop). 

While I don't mind that you do this, it's not really necessary, as we squash-merge pull requests anyway. 

> Note: the CI check is currently failing due to an independent issue in the MemoryStore module: the issue is available on the develop branch already (maybe introduced to some backmerge from master). Once this is fixed and available in this branch through a rebase, this PR should also be greenlined. For reference: the test is org.eclipse.rdf4j.sail.memory.QueryPlanRetrievalTest

Ah thanks. I have a fix for that lined up, will merge into this branch as soon as it's available on develop.

jeenbroekstra:
I've logged issues for the two non-optional things you mentioned: #2185 and #2186. I'd like to sleep a little on whether making graph support optional is something that adds value.

PR #2147, opened 2020-04-28
GH-2140 Unions can have specified new scope

hmottestad:
GitHub issue resolved: #2140 

Unions are introduced when parsing queries with alternate paths. These are again optimized by the QueryModelNormalizer with new Unions. None of these Unions need to introduce a new scope, as they were not introduced by the user.

Added an option for specifying scope to Unions.

---- 
PR Author Checklist: 

### DIFF
--- a/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
+++ b/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
@@ -18,7 +18,6 @@ import java.nio.charset.StandardCharsets;
 
 import org.apache.commons.io.IOUtils;
 import org.eclipse.rdf4j.IsolationLevels;
-import org.eclipse.rdf4j.common.concurrent.locks.Properties;
 import org.eclipse.rdf4j.model.ValueFactory;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.model.vocabulary.FOAF;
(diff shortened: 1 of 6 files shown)


hmottestad:
What do you think about this approach @jeenbroekstra ?

hmottestad:
Can change it to use this instead: https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/GraphPatternGroupable.java

Does isGraphPatternGroup:true mean that we variables enter a new scope?

jeenbroekstra:
I'd prefer to use existing information (such as GraphPatternGroupable) for this, rather than introduce a new flag.

Please note that I have a draft PR for this same issue up (see #2141) , which I intend to continue work on.

jeenbroekstra:
I've got an alternative approach for this pushed to #2141, I think we can close this.

hmottestad:
Ok. I propose we wait until you've merged your branch, then we can delete this PR and branch.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
```diff
@@ -619,4 +618,94 @@ public void testTimeout() {
...
+					"select * where {",
+					"	?a (a|^a)* ?c. ?c (a|^a) ?b .",
+					"	?b rdfs:label ?type .",
+					"}"));
+
+			query.setMaxExecutionTime(1);
+
+			String actual = query.explain(Explanation.Level.Executed).toString();
+
+			assertEquals("Projection (resultSizeActual=84)\n" +
```
jeenbroekstra, line 651: I don't particularly like the notion of using the string representation of the query plan as the thing to test on. Far too brittle in my opinion - this test would easily break if we made any changes to query processing, even if those changes have nothing to do with alternative paths.

hmottestad: Fair enough. Let me just use this test to verify your fix, and we can delete both this PR and the branch after.

jeenbroekstra:
I've got my PR merged now, so if you rebase you should be able to run this test after my changes.

hmottestad:
Looks good!

Issue #2143, opened 2020-04-28
Ensure binary compatibility of upcoming release with previous minor releases


jeenbroekstra:
Some of the changes to the Models utility introduced by #2056 are not binary compatible with RDF4J 3.1. In particular, the change of method param types from `Model` to `Iterable<Statement>` is not. This makes it impossible to run existing projects against a new version of RDF4J without being forced to recompile.

Although we have not made an explicit commitment to binary compatibility before it has been somewhat implied by our adherence to semantic versioning and claiming that minor versions are "fully backward compatible". I'll look at reactivating the japicmp plugin with potentially letting it fail the build. As a short term fix for this particular problem I will re-introduce the old method signatures with suitable javadoc annotations.

jetztgradnet:
This change would be source compatible, but not binary compatible as the signature is part of the compiled byte code. Any re-compile will pick up the new signature, but anything already compiled will fail with a `NoSuchMethodError`...

jeenbroekstra:
Generalizing this task a bit because I've detected some other compatibility issues.

PR #2148, opened 2020-04-29
GH-2143 ensure project is binary compatible with 3.1 releases

jeenbroekstra:
GitHub issue resolved: #2143  

- modified several RDF* related changes to ensure reasonable defaults are in place
- modified several recent public and protected method changes to ensure binary compatibility
- configured [japicmp](https://siom79.github.io/japicmp/MavenPlugin.html) to check (and fail!) build on both source and binary compatibility with previous minor release. It ignore items annotated with InternalUseOnly or Experimental.  

---- 
PR Author Checklist: 

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
@@ -56,6 +56,12 @@ public class QueryJoinOptimizer implements QueryOptimizer {
 		tupleExpr.visit(new JoinVisitor());
 	}
 
+	/**
+	 * 
+	 * @deprecated This class is protected for historic reasons only, and will be made private in a future major
+	 *             release.
+	 */
+	@Deprecated
 	protected class JoinVisitor extends AbstractQueryModelVisitor<RuntimeException> {
 
 		Set<String> boundVars = new HashSet<>();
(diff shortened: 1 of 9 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
```diff
@@ -393,6 +393,12 @@ protected TupleExpr selectNextTupleExpr(List<TupleExpr> expressions, Map<TupleEx
 			return result;
 		}
 
+		@Deprecated
+		protected double getTupleExprCardinality(TupleExpr tupleExpr, Map<TupleExpr, Double> cardinalityMap,
```
jeenbroekstra, line 397: This is one of the many reasons why I want stop using `protected` as a modifier, as much as possible.

jeenbroekstra:
> I do not have a good understanding of the Maven plugin, can you maybe shortly comment how it works and how it influences the build?

Sure thing. It ties in at the `verify` phase of the maven build cycle, and what it does is a binary comparison of each jar artifact, comparing the current version against a specified old version (in this case, 3.1.4) for compatibility. I've configured it to detect both binary and source incompatibilities and to fail the build on both (so we get a definite warning) - this will effectively mean that pull requests will be blocked until compatibility issues are resolved. 

It currently checks compatibility for everything that is public or protected, with the exception of anything that is annotated as `@Experimental` or `@InternalUseOnly`. I had to tweak a bit to not flag certain corner cases in our code base, and I don't claim that the current setup is 100% correct and guarantees backward compatibilty, but it definitely flags most common issues and that can only be helpful.

hmottestad:
I think this looks good :) It will probably cause some hassle in the beginning as we discover if it manages to exclude everything we need. Could you create a separate build stage in the Github Actions just for this part. Something like this:

1. verify formatting
1. verify that code compiles
1. verify that code is source and binary compatible
1. verify that tests pass

Also: Do you need to bump the version number up for every release?

jeenbroekstra:
Good point, I'll have a quick look at configuring github actions. 

> Also: Do you need to bump the version number up for every release?

Not necessarily - if we commit to all minor releases being binary compatible, we can keep comparing with the currently set version for every future minor release. Of course, we do need to bump it as soon as we start working on the next major.

Issue #2175, opened 2020-05-05
SharedHttpClientSessionManager#executor corePoolSize should be configurable


erikgb:
We were occasionally struggling with 100% CPU utilisation, with no particular reason, and stumbled upon https://github.com/eclipse/rdf4j/issues/1271. But even after setting the hotfix system property `org.eclipse.rdf4j.client.executors.jdkbug`, our load tests indicates that the executor core thread pool size is too small. We have currently implemented a workaround, using the `SharedHttpClientSessionManager` constructor accepting a `ScheduledExecutorService`.

I am suggesting to **replace** the hotfix system property with a new system property that can be used to set the executor core thread pool size. The name of the new system property could be `org.eclipse.rdf4j.client.executors.corePoolSize`, and the default value should be 1. I do not think the previous value (0), before https://github.com/eclipse/rdf4j/pull/1272, should be avoided.

I will prepare a PR.

PR #2177, opened 2020-05-05
GH-2175: Introduced new system property to configure SharedHttpClient…

erikgb:
GitHub issue resolved: #2175

- Introduces a new system property (`org.eclipse.rdf4j.client.executors.corePoolSize`) to configure the core thread pool size of `SharedHttpClientSessionManager#executor`.

- Changes the default value of the `SharedHttpClientSessionManager#executor` core thread pool size from 0 to 1.

- This new system property **replaces** the hotfix system property (`org.eclipse.rdf4j.client.executors.jdkbug`) introduced in https://github.com/eclipse/rdf4j/pull/1272.

---- 
PR Author Checklist: 

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
@@ -33,13 +33,10 @@ import org.slf4j.LoggerFactory;
  */
 public class SharedHttpClientSessionManager implements HttpClientSessionManager, HttpClientDependent {
 	/**
-	 * FIXME: issue #1271, workaround for OpenJDK 8 bug. ScheduledThreadPoolExecutor with 0 core threads may cause 100%
-	 * CPU usage. Using 1 core thread instead of 0 (default) fixes the problem but wastes some resources.
-	 * 
-	 * @see <a href="https://bugs.openjdk.java.net/browse/JDK-8129861">JDK-8129861</a>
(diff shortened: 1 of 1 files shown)


core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
```diff
@@ -66,7 +63,8 @@
 
 	public SharedHttpClientSessionManager() {
 		final ThreadFactory backingThreadFactory = Executors.defaultThreadFactory();
-		this.executor = Executors.newScheduledThreadPool(cores, (Runnable runnable) -> {
+		final Integer corePoolSize = Integer.getInteger(CORE_POOL_SIZE_PROPERTY, 1);
```
jeenbroekstra, line 66: Shouldn't this use `System.getProperty` to get the _value_ of the property?

erikgb: It does, behind the scenes. Just have a look at the `Integer#getInteger` method! Using this approach avoids the boilerplate code required to cast a `String` to an `Integer`. But I see now that the `corePoolSize` variable should be `int`, and not the nullable `Integer`. I'll fix.

jeenbroekstra:
@erikgb thanks for submitting this! As indicated it looks fine to me, and I see no real issue with just removing the old system property (as the new default is what that property was supposed to tweak anyway). 

This will need an update in the documentation, I think, I'll look into that before I merge.

Issue #2171, opened 2020-05-05
Handle scoping issues in SPARQL 


jeenbroekstra:
Related to #1405 and #2119

Currently, the SPARQL parser and engine use a check whether something was parsed as a group graph pattern to determine if a variable scope change is needed (which influences the chosen join strategy).

This is a bit of a workaround approach and there are many exceptions. Filter conditions as well as named graph patterns are parsed as group patterns, but do not necessarily mean a scope change. In addition, the entire where clause is _by definition_ a group pattern, but a values clause defined outside the where clause is intended to be scoped to that where-clause. We are continually bumping into other corner cases and adapting the workaround as we go. 

To fix the issue more robustly, we should introduce a `isScopeChange` field in the SPARQL AST tree, which defaults to false. In the case of ASTGraphPatternGroups we can inspect the node parent to determine if the group is part of a named graph or a filter condition before returning true or false. The external values clause will need to be treated as a special case in the TupleExprBuilder (in fact it already is, we need to modify it slightly). 

Fix and regression tests incoming.

PR #2178, opened 2020-05-05
GH-2171 group pattern as direct child of where clause is not a scope change

jeenbroekstra:
GitHub issue resolved: #2171 

This covers the case where the group graph pattern has the where clause as its direct parent - in this case we shouldn't count it as a scope change.

There are still several failing tests in QueryPlanRetrievalTest but I'm not immediately sure if these are problems in the new approach or simply a consequence of slight brittleness. 

---- 
PR Author Checklist: 

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
@@ -51,7 +51,6 @@ import org.eclipse.rdf4j.query.algebra.Extension;
 import org.eclipse.rdf4j.query.algebra.ExtensionElem;
 import org.eclipse.rdf4j.query.algebra.Filter;
 import org.eclipse.rdf4j.query.algebra.FunctionCall;
-import org.eclipse.rdf4j.query.algebra.GraphPatternGroupable;
 import org.eclipse.rdf4j.query.algebra.Group;
 import org.eclipse.rdf4j.query.algebra.GroupConcat;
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Part of the problem seems to be that the same optional pattern appears twice in the test query, once as part of a scope change, once not.

jeenbroekstra:
From a closer look I think most of the other ones are failing because the engine now (correctly) sees that one of the optionals (the first one, in braces) involves a scope change and therefore applies a HashJoinIteration. I'll adapt the tests accorrdingly.

hmottestad:
This looks much better :)

I did a quick search and there are still two usages of setGraphPatternGroup in the code that you should take a look at to see if they should be migrated to the new method.

Screenshot 2020-05-05 at 10 56 26

hmottestad:
Btw. I noticed that you are using the "parent" node to determine if we are in a new scope. I hit on a few cases where parent was not actually the parent. The main problem was that we were reusing nodes instead of cloning them, and then updating the parent.

jeenbroekstra:
> Btw. I noticed that you are using the "parent" node to determine if we are in a new scope. I hit on a few cases where parent was not actually the parent. The main problem was that we were reusing nodes instead of cloning them, and then updating the parent ref.

I think that this is not an issue here. If I read your PR #2180  correctly, the problem you encountered is in the algebra tree, where indeed nodes are sometimes cloned and shared. Worth investigating if we can improve on that, but note that I only look at parent relations in the AST, _not_ the algebra.

jeenbroekstra:
Given that develop branch is currently unstable and this PR is intended to fix that, I'll merge this as soon as verification completes.

PR #2180, opened 2020-05-05
GH-2187 parent link in QueryModelNode

hmottestad:
@jeenbroekstra this is just to see the parent issue I encountered. If you want to take a look and consider if it'll have any implications for your new scope code. Problem for me is that there might be some obscure query that could trigger a different case that we don't know about yet.

### DIFF
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/AbstractQueryModelNode.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/AbstractQueryModelNode.java
@@ -43,6 +43,10 @@ public abstract class AbstractQueryModelNode implements QueryModelNode, Variable
 
 	@Override
 	public void setParentNode(QueryModelNode parent) {
+		if (this.parent != null && this.parent != parent) {
+			// debug here
+			System.out.println();
+		}
 		this.parent = parent;
 	}
 
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
@@ -24,6 +24,7 @@ public class QueryModelTreeToGenericPlanNode extends AbstractQueryModelVisitor<R
 	GenericPlanNode top = null;
 	QueryModelNode topTupleExpr;
 	ArrayDeque<GenericPlanNode> deque = new ArrayDeque<>();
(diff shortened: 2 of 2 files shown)


jeenbroekstra:
@hmottestad thanks for putting this together, will be interesting to look at more closely. Notice, however, that I only use the parent lookup in the AST, not in the algebra tree (which is what this example is looking at). So unless you've encountered a similar problem in the AST, I don't think this has direct consequences for the new scope approach.

hmottestad:
We can just remove the parent code from the algebra tree then. No need to spend time fixing if we don’t need the feature.

jeenbroekstra:
I don't think it's unneeded, exactly. There's plenty of code in various optimizers and other algebra tree visitors that uses the `getParentNode` method (for example, to travel back up to the root of the tree).

hmottestad:
I'll just leave this PR here then :)

jeenbroekstra:
I just had a closer look at this PR to try and understand what's going on, and I'm having a hard time following. You say "nodes are reused" and "parents are not maintained" but I don't see an example of this in this PR's code to be honest. What am I missing?

hmottestad:
You have to run the QueryPlanRetrievalTest in org.eclipse.rdf4j.sail.memory. The asserts will fire where there is a mismatch with the true parent in the tree and the one from getParent().

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/AbstractQueryModelNode.java
```diff
@@ -43,6 +43,10 @@ public QueryModelNode getParentNode() {
 
 	@Override
 	public void setParentNode(QueryModelNode parent) {
+		if (this.parent != null && this.parent != parent) {
+			// debug here
+			System.out.println();
```
hmottestad, line 48: @jeenbroekstra this is a good point to debug when running the tests in QueryPlanRetrievalTest

hmottestad:
@jeenbroekstra I was thinking about what we should do about my branches that reproduce various bugs. They make fixing the bugs much simpler because you essentially have a test for the bug already. I would like to keep them around in some way, but I also see the issue of trying to keep them up to date with the parent branch as a significant hurdle.

jeenbroekstra:
Ah, sorry, I closed and deleted this because I thought it was a stable branch for an issue that was already fixed. I think I just misread. Obviously if the case it's about is still open it makes sense to keep the branch open, at least (though perhaps close the PR until we're ready to pick it up again).

jeenbroekstra:
The naming of the branch also didn't help :)

jeenbroekstra:
replaced by PR #2403 (branch name fixed)

Issue #2206, opened 2020-05-09
Prettier query plans


hmottestad:
Use "Box-drawing characters" for making it more obvious which nodes belong together.

Example:

```
Projection (resultSizeActual=2)
╠══ProjectionElemList
║     ProjectionElem "a"
╚══LeftJoin (LeftJoinIterator) (resultSizeActual=2)
   ├──Join (JoinIterator) (resultSizeActual=2)
   │  ╠══StatementPattern (costEstimate=1, resultSizeEstimate=4, resultSizeActual=4)
   │  ║     Var (name=a)
   │  ║     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │  ║     Var (name=d)
   │  ╚══Filter (resultSizeActual=2)
   │     ├──Compare (!=)
   │     │     Var (name=c)
   │     │     Var (name=d)
   │     └──Join (HashJoinIteration) (resultSizeActual=6)
   │        ╠══StatementPattern (costEstimate=2, resultSizeEstimate=4, resultSizeActual=6)
   │        ║     Var (name=a)
   │        ║     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │        ║     Var (name=c)
   │        ╚══LeftJoin (new scope) (BadlyDesignedLeftJoinIterator) (costEstimate=5, resultSizeEstimate=12, resultSizeActual=4)
   │           ├──SingletonSet (resultSizeActual=4)
   │           └──StatementPattern (resultSizeEstimate=12, resultSizeActual=48)
   │                 Var (name=d)
   │                 Var (name=e)
   │                 Var (name=f)
   └──StatementPattern (resultSizeEstimate=12, resultSizeActual=2)
         Var (name=d)
         Var (name=e)
         Var (name=f)

```

barthanssens:
Looks nice, though I'm wondering the bars are displayed correctly in 'dumb' terminals (e.g. putty to remote server)

barthanssens:
Just tested it, it works 👍

PR #2207, opened 2020-05-09
GH-2206 prettier plans by using box-drawing characters to group joins

hmottestad:
GitHub issue closing #2206 

Use box-drawing characters (https://en.wikipedia.org/wiki/Box-drawing_character) for making query plans easier to read.

---- 
PR Author Checklist: 

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/explanation/GenericPlanNode.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/explanation/GenericPlanNode.java
@@ -217,6 +217,8 @@ public class GenericPlanNode {
 		this.algorithm = algorithm;
 	}
 
+	private static int prettyBoxDrawingType = 0;
+
 	/**
 	 * Human readable string. Do not attempt to parse this.
 	 *
@@ -224,7 +226,16 @@ public class GenericPlanNode {
 	 */
 	@Override
 	public String toString() {
+		return getHumanReadable(0);
+	}
 
+	/**
+	 *
+	 * @param prettyBoxDrawingType for deciding if we should use single or double walled character for drawing the
+	 *                             connectors between nodes in the query plan. Eg. ├ or ╠ and ─ o
+	 * @return
+	 */
+	private String getHumanReadable(int prettyBoxDrawingType) {
 		StringBuilder sb = new StringBuilder();
 
 		String newLine = System.getProperty("line.separator");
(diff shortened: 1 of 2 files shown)


hmottestad:
Example:

```
Projection (resultSizeActual=0)
╠══ProjectionElemList
║     ProjectionElem "a"
╚══Join (HashJoinIteration) (resultSizeActual=0)
   ├──StatementPattern (costEstimate=1, resultSizeEstimate=4, resultSizeActual=4)
   │     Var (name=a)
   │     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │     Var (name=type)
   └──Union (new scope) (resultSizeActual=24)
      ╠══Join (HashJoinIteration) (resultSizeActual=12)
      ║  ├──StatementPattern (costEstimate=2, resultSizeEstimate=12, resultSizeActual=12)
      ║  │     Var (name=a)
      ║  │     Var (name=b)
      ║  │     Var (name=c2)
      ║  └──Union (new scope) (resultSizeActual=96)
      ║     ╠══Join (JoinIterator) (resultSizeActual=48)
      ║     ║  ├──StatementPattern (new scope) (costEstimate=2, resultSizeEstimate=4, resultSizeActual=4)
      ║     ║  │     Var (name=c2)
      ║     ║  │     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
      ║     ║  │     Var (name=type1)
      ║     ║  └──StatementPattern (costEstimate=2, resultSizeEstimate=12, resultSizeActual=48)
      ║     ║        Var (name=a)
      ║     ║        Var (name=b)
      ║     ║        Var (name=c)
      ║     ╚══Join (JoinIterator) (resultSizeActual=48)
      ║        ├──StatementPattern (new scope) (costEstimate=2, resultSizeEstimate=4, resultSizeActual=4)
      ║        │     Var (name=c2)
      ║        │     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
      ║        │     Var (name=type2)
      ║        └──StatementPattern (costEstimate=2, resultSizeEstimate=12, resultSizeActual=48)
      ║              Var (name=a)
      ║              Var (name=b)
      ║              Var (name=c)
      ╚══StatementPattern (new scope) (costEstimate=5, resultSizeEstimate=12, resultSizeActual=12)
            Var (name=type)
            Var (name=d)
            Var (name=c)
```

core/query/src/main/java/org/eclipse/rdf4j/query/explanation/GenericPlanNode.java
```diff
@@ -217,14 +217,25 @@ public void setAlgorithm(String algorithm) {
...
+	private static int prettyBoxDrawingType = 0;
+
 	/**
 	 * Human readable string. Do not attempt to parse this.
 	 *
 	 * @return
 	 */
 	@Override
 	public String toString() {
+		return getHumanReadable(0);
```
jeenbroekstra, line 229: Can you remind me - this is separate from what I would get if I did 

     TupleQuery query = conn.prepareTupleQuery(....);
     System.out.println(query);

correct? What I'm getting at is that this influences only the output of the `Explanation` serialization, not if we directly serialize a query algebra node to string. If so: that's good, I'd rather not change the original tupleExpr serialization too much (even if it's not, strictly speaking, a standardized format).

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
```diff
@@ -87,32 +87,32 @@ public void testTupleQuery() {
...
+					"╠══ProjectionElemList\n" +
+					"║     ProjectionElem \"a\"\n" +
+					"╚══Filter\n" +
+					"   ├──Compare (!=)\n" +
+					"   │     Var (name=c)\n" +
+					"   │     Var (name=d)\n" +
+					"   └──LeftJoin\n" +
+					"      ╠══Join\n" +
+					"      ║  ├──Join\n" +
+					"      ║  │  ╠══LeftJoin (new scope)\n" +
```
jeenbroekstra, line 99: If I understood correctly, the idea is to use the double lines for all nodes that have two (or more) children, single lines only for unary predicates. So why is the outer join here single lined, and the inner one double?

hmottestad: It just swaps every other node. So that it’s easier to follow the lines visually. Double vs single has no meaning.

core/query/src/main/java/org/eclipse/rdf4j/query/explanation/GenericPlanNode.java
hmottestad: Yes. Printing the query like that just prints the parsed version of the tuple expression. Using the explanation transforms the query into GenericPlanNode, where we have this bow-drawing toString addition.

hmottestad:
Planning on merging this. Unless you guys think the use of single and double pipes is going to be overly confusing.

barthanssens:
No strong feelings about mixing double and single pipes or not. Either way LGTM :-)

Issue #2211, opened 2020-05-10
Parallel upgrading of DynamicModel can cause NullPointerException


hmottestad:
```
2020-05-10T20:06:47.1141940Z ----------------------------------------------------------------------
2020-05-10T20:06:47.1142203Z Class: java.util.concurrent.ExecutionException
2020-05-10T20:06:47.1142435Z Message: java.lang.NullPointerException
2020-05-10T20:06:47.1142674Z java.util.concurrent.ExecutionException: java.lang.NullPointerException
2020-05-10T20:06:47.1143116Z 	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
2020-05-10T20:06:47.1144095Z 	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
2020-05-10T20:06:47.1145921Z 	at org.eclipse.rdf4j.sail.shacl.MultithreadedTest.lambda$runValidationFailuresTest$5(MultithreadedTest.java:503)
2020-05-10T20:06:47.1146692Z 	at java.util.ArrayList.forEach(ArrayList.java:1257)
2020-05-10T20:06:47.1148431Z 	at org.eclipse.rdf4j.sail.shacl.MultithreadedTest.runValidationFailuresTest(MultithreadedTest.java:501)
2020-05-10T20:06:47.1150328Z 	at org.eclipse.rdf4j.sail.shacl.MultithreadedTest.testLotsOfValidationFailuresSerializable(MultithreadedTest.java:372)
2020-05-10T20:06:47.1151265Z 	at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
2020-05-10T20:06:47.1152951Z 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2020-05-10T20:06:47.1153729Z 	at java.lang.reflect.Method.invoke(Method.java:498)
2020-05-10T20:06:47.1155184Z 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
2020-05-10T20:06:47.1156586Z 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
2020-05-10T20:06:47.1157921Z 	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #2214, opened 2020-05-11
GH-2211 more thread safe upgrading of the DynamicModel

hmottestad:
GitHub issue resolved: #221 

 - LinkedHashMap is not removed while upgrading, allowing current users to keep using it
 - Upgrading during a write will now wait for the write to complete, thanks to synchronized writes
 
---- 
PR Author Checklist: 

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/impl/DynamicModel.java
@@ -37,10 +37,9 @@ import org.eclipse.rdf4j.util.iterators.SingletonIterator;
  * predicate).
  *
  * DynamicModel is thread safe to the extent that the underlying LinkedHashMap or Model is. The upgrade path is
- * protected by the actual upgrade method being synchronized and any writes being synchronized too. This means that
- * writes can safely assume that their underlying storage will not change while writing. The LinkedHashMap storage is
- * not removed once upgraded, so concurrent reads that have started reading from the LinkedHashMap can continue to read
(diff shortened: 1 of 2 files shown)


hmottestad:
I don't have a test for this, the bug was reproduced in a ShaclSail test. It doesn't seem that we have any decent multi-threaded reader with single writer tests for model, even though we use them in that way in the code. I know that the TreeModel has a race condition for this, and now apparently so the does the LinkedHashModel.

jeenbroekstra:
The Model classes, like most Java Collection classes, are not thread-safe by design. I haven't looked in detail at the problem yet but if there are places in the code where we assume a Model class is thread-safe, surely the fix is to change that assumption and make sure the code accesses the model in a thread-safe manner, rather than try and make all model implementations themselves thread-safe.

hmottestad:
From ArrayList docs (for reference): 

> Note that this implementation is not synchronized. If multiple threads access an ArrayList instance concurrently, and at least one of the threads modifies the list structurally, it must be synchronized externally.

Current implementation of the DynamicModel isn’t thread safe for multiple readers if one of the readers forces the model to upgrade. This is fixed in the pr. 

If the cost of synchronizing writes is small I think it’s a worthwhile fix to remove the chance that we write to the wrong underlying storage during multithreaded reads and single-threaded writes.

hmottestad:
I’ve now removed the synchronization again. I think you are right that we don’t want to imply thread safety. Multiple reader thread safety is nice to have though, so that is fixed in this PR, and I also have test to check for best-effort detection of concurrent access.

hmottestad:
TODO: Edit the java docs

hmottestad:
I believe this is done now and ready for review.

PR #2228, opened 2020-05-16
GH-2211 add more latches and loop to reduce chance of flaky failure

hmottestad:
GitHub issue resolved: #2211 

---- 
PR Author Checklist: 

### DIFF
--- a/core/model/src/test/java/org/eclipse/rdf4j/model/DynamicModelConcurrentModificationAndUpgradeTest.java
+++ b/core/model/src/test/java/org/eclipse/rdf4j/model/DynamicModelConcurrentModificationAndUpgradeTest.java
@@ -9,6 +9,7 @@
 package org.eclipse.rdf4j.model;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
 
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


core/model/src/test/java/org/eclipse/rdf4j/model/DynamicModelConcurrentModificationAndUpgradeTest.java
```diff
@@ -75,20 +91,23 @@ public boolean contains(Object o) {
 							@Override
 							public boolean hasNext() {
 								try {
-									countDownLatch.await();
+									countDownLatch1.countDown();
+									countDownLatch2.await();
```
hmottestad, line 95: This latching here should help reduce the chance of this one thread getting executed first.

core/model/src/test/java/org/eclipse/rdf4j/model/DynamicModelConcurrentModificationAndUpgradeTest.java
```diff
@@ -145,8 +164,13 @@ public void clear() {
...
 
 		Runnable upgrade = () -> {
+			try {
+				countDownLatch1.await();
+			} catch (InterruptedException e) {
+				e.printStackTrace();
+			}
 			model.filter(null, RDF.TYPE, null);
-			countDownLatch.countDown();
+			countDownLatch2.countDown();
```
hmottestad, line 173: and here is the count down for that latch

hmottestad:
@jeenbroekstra I don't really want that Set to be volatile, for this test to be correct it really should be, but I think running the test several times and adding yield will sufficiently increase the odds of reading the correct variable at some point during the test. I've also fixed a bigger issue with the test. If run on a machine with only 1 core then the first thread would just finish executing and then the second would be scheduled. I've added a second latch to force switching from one thread to the other.

jeenbroekstra:
If this gets things stable again I'm fine with it, but isn't this just rigging the test rather than fixing the underlying issue? Or are you saying that it's the test itself that's flawed?

hmottestad:
yeah, test is to check for the dynamicmodel best-effort concurrent modification detection

Issue #2017, opened 2020-03-18
SHACL advanced targeting


rdstn:
At the moment, the SHACL targeting system has some limitations. Assume that we have the following data:

```
<http://example.org/resource/ObjA-1> example:type example:ObjA ;
<http://example.org/resource/ObjA-1> example:name "Test Object A1" ;
<http://example.org/resource/ObjA-2> example:type example:ObjA ;
<http://example.org/resource/ObjA-2> example:name "Test Object A2" ;
<http://example.org/resource/ObjB-1> example:type example:ObjB ;
<http://example.org/resource/ObjB-1> example:name "Test Object B1" ;
```

We want to be able to target `ObjA-1` and `ObjA-2`.

- `sh:targetNode` works for `ObjA-1` and `ObjA-2`, but suppose we want to insert `ObjA-3`.
- `sh:targetClass` is hardcoded to work with `rdf:type` and we are using `example:type` instead.
- `sh:targetSubjectsOf` for `example:type` will also select `ObjB-1`.
- Implicit class targeting wouldn't work since we are not using `rdf:Type`.

The simple workaround is to force `rdf:type` to always be the class differntiator. It is possible, however, that some users may not want to do that - for example, they may be using types for two different systems and have to differentiate between them.

There are two potential resolutions I see:

1. #1912 - [SHACL advanced features](https://www.w3.org/TR/shacl-af/#targets) have both SPARQL-based targets and SPARQL-based target types specified. This looks quite powerful, but might be less performant. However, so long a we are just looking for a predicate - object combination, I don't see it as a major issue, even with large repositories.
2. Introduce a new target type focus node which can take a custom predicate-object pair. For example: 
```
example:testShape
  a sh:NodeShape ;
  sh:predicateTarget example:Type ;
  sh:objectTarget example:ObjA .
```
This is less powerful, but looks like it will be more performant. There is the added drawback that it is not part of the specification.

Any way to have non-standard targeting for the use case would be good.

Are there any other options that I'm missing? Or is SHACL Advanced and/or the extneded spec workaround the way to go?

rdstn:
@hmottestad, I believe you are also interested in the Advanced features, what's your take on this one - do you think we'll be able to get decent performance for SPARQL targeting?

hmottestad:
I believe it will not be possible to get good performance from sparql based targets when there are many targets in the underlying database. Sparql queries are very advanced, and this complexity will likely mean that the query will need to run against the entire database for every transaction.

[...]

PR #2229, opened 2020-05-16
GH-2017 SHACL support for advanced targeting

hmottestad:

Support for:
 - sh:hasValue #2301
 - sh:targetShape #2017
 - dash:AllObjectsTarget and dash:AllSubjectsTarget #2302
 - dash:valueIn (missing github issue)

---- 
PR Author Checklist: 


core/sail/shacl/src/test/resources/shaclDatatypeSparqlTarget.ttl
```diff
@@ -0,0 +1,34 @@
...
+                			}
+                			""" ;
+                ] ;
+
+	sh:property [
+		sh:path foaf:age ;
+		sh:datatype xsd:int ;
+
+	] .
+
```
hmottestad, line 34: Example of the SPARQL target support.

Note: You have to include prefixes in the string. We don't support `sh:prefixes`!

core/sail/shacl/src/test/resources/test-cases/class/predicateObjectTarget/shacl.ttl
```diff
@@ -0,0 +1,27 @@
...
+        rdf4j-sh:compoundTarget [
+	        rdf4j-sh:targetPredicate rdf:type;
+	        rdf4j-sh:targetObject ex:SecondTarget
+	];
+
+	sh:property [
+		sh:path ex:knows ;
+		sh:and ( [sh:class ex:Person]  [sh:class ex:Human])
+	] .
+
```
hmottestad, line 27: An example of the new predicate+object target support.

You can have as many `rdf4j-sh:compoundTarget`s as you want, but each one can only have a single predicate and single object. For the time being at least.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/benchmark/Main.java
```diff
@@ -24,7 +24,7 @@
 	}
 
 	public static void main(String[] args) throws RunnerException {
-		Options opt = new OptionsBuilder().include("")
+		Options opt = new OptionsBuilder().include("TargetBenchmarkInitialData.*")
 
 				// .addProfiler("stack", "lines=20;period=1;top=20")
 				.build();
```
hmottestad, line 30: Use this main method to compare performance between SPARQL targets and predicate+object targets. The benchmarks test how the number of targets already present in the database affect the performance of a standard number (and size) of transactions.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
```diff
@@ -273,7 +295,7 @@ static void runTestCase(String shaclPath, String dataPath, ExpectedResult expect
 				connection.begin(isolationLevel);
 				ValueFactory vf = connection.getValueFactory();
 				connection.add(vf.createBNode(), vf.createIRI("http://example.com/jkhsdfiu3r2y9fjr3u0"),
-						vf.createLiteral("auto-generated!"), vf.createBNode());
+						vf.createLiteral("123", XMLSchema.INTEGER), vf.createBNode());
```
jeenbroekstra, line 298: I think this is what causes your current test failure - on the develop branch, the import for XMLSchema in this class has been replaced with XSD. CI tries to build the PR after merging, and fails because it misses the import. 

Easiest way to fix is to rebase and then replace this reference with XSD.

core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/DASH.java
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2020 Eclipse RDF4J contributors, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
```
jeenbroekstra, line 7: This copyright header looks a bit odd: it's missing the full lines of asterisks at beginning and end.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
hmottestad: Ahh. That makes sense. The import gets changed to XSD, but the reference XMLSchema.INTEGER can't resolve.

hmottestad:
Experimenting with support for sh:filterShape which was a part of an earlier SHACL draft, but removed due to time constraints:

```
@base <http://example.com/ns> .
@prefix ex: <http://example.com/ns#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix dash: <http://datashapes.org/dash#> .

ex:PersonShape
	a sh:NodeShape  ;
	sh:target [a dash:AllSubjectsTarget] ;
	sh:filterShape [
        		a sh:Shape ;
        		sh:and (
        		        [
        		                sh:path rdf:type;
        		                sh:hasValue ex:Person
        		        ]

        		        [
        		                sh:path rdfs:label;
        		                sh:or (
        		                        [ sh:hasValue "Peter" ]
        		                        [ sh:hasValue "Jennifer" ]
        		                        [ sh:hasValue "Sue" ]
        		                )
        		        ]
        		);
        	] ;

	sh:property [
		sh:path ex:knows ;
		sh:class ex:Person ;
	] .

```

This shape essentially means that all ex:Person with rdfs:label either of "Peter", "Jennifer", "Sue" are constrained to only have ex:knows something of type ex:Person.

core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/DASH.java
```diff
@@ -0,0 +1,37 @@
...
+
+	/**
+	 * An immutable {@link Namespace} constant that represents the namespace.
+	 */
+	public static final Namespace NS = new SimpleNamespace(PREFIX, NAMESPACE);
+
+	static private final ValueFactory vf = SimpleValueFactory.getInstance();
+
+	public static final IRI AllObjectsTarget = createIRI("AllObjectsTarget");
+	public static final IRI AllSubjectsTarget = createIRI("AllSubjectsTarget");
```
hmottestad, line 31: I have chosen to go with the more natural naming rather than the forced uppercase.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AbstractSimplePropertyShape.java
```diff
@@ -101,7 +101,7 @@ static public PlanNode getGenericSingleObjectPlan(ConnectionsGroup connectionsGr
 
 			PlanNode discardedRight = innerJoin.getDiscardedRight(BufferedPlanNode.class);
 
-			PlanNode typeFilterPlan = nodeShape.getTargetFilter(connectionsGroup.getBaseConnection(), discardedRight);
+			PlanNode typeFilterPlan = nodeShape.getTargetFilter(connectionsGroup, discardedRight);
```
hmottestad, line 104: This is to be able to make sure that the getTargetFilter on NodeShape only operates on the base connection.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AllObjectsTarget.java
```diff
@@ -0,0 +1,126 @@
...
+										connectionsGroup.getRemovedStatements(),
+										null,
+										null,
+										null,
+										UnorderedSelect.OutputPattern.ObjectPredicateSubject),
+								0,
+								1)));
+
+		return connectionsGroup.getCachedNodeFor(select);
+	}
```
hmottestad, line 96: Removed doesn't guarantee to only return this that have been removed, but must return everything that could potentially have been removed. So this implementation will be fine.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AllSubjectsTarget.java
```diff
@@ -0,0 +1,221 @@
...
+	NodeShape filterShape;
+
+	AllSubjectsTarget(Resource id, ShaclSail shaclSail, SailRepositoryConnection connection, boolean deactivated,
+			Resource filterShape) {
+		super(id, shaclSail, connection, deactivated);
+		if (shaclSail.isExperimentalFilterShapeSupport() && filterShape != null) {
+			this.filterShape = new NodeShape(filterShape, shaclSail, connection, false);
+		}
+
+	}
```
hmottestad, line 54: AllSubjectsTarget is the only place where we are support sh:filterShape currently.

hmottestad:
@jeenbroekstra I think I’m mostly ready for a review now. Biggest goal is to make sure most of the code is understandable, especially variable and method names.

hmottestad:
@jeenbroekstra this has become a massive PR. I believe I have good test coverage and after a "self-review" tomorrow I intend to merge this PR. 

Could you let me know if you think that that is ok or not?

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ExperimentalFeaturesTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+	@Test
+	public void testDash() throws Exception {
+
+		SailRepository shaclRepository = Utils.getInitializedShaclRepository("test-cases/class/allSubjects/shacl.ttl",
+				false);
+
+		try (SailRepositoryConnection connection = shaclRepository.getConnection()) {
+			connection.begin();
+			connection.add(vf.createBNode(), ex_knows, vf.createBNode());
```
jeenbroekstra, line 43: This test contains no assertions - what exactly are we testing here? Perhaps clarify with a comment.

hmottestad:
It’s the code for supporting sparql based targets. Which is probably something we will support in the future when we have full sparql based validation.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ExperimentalFeaturesTest.java
hmottestad: There are two sets of tests that check that the experimental features are disabled by default and that you can enable them after you've added you shapes. Essentially the tests are "without the experimental features enabled there should be no shacl exception".

hmottestad:
Another point for discussion @jeenbroekstra. 

We've added support for non-standard SHACL features. Some of them are from DASH (http://datashapes.org), some will be added to DASH and one will be added to the SHACL Advanced Note. 

To enable **all** these features the user needs to call:

```
		shaclSail.setExperimentalDashSupport(true);
		shaclSail.setExperimentalTargetShapeSupport(true);
```

Do you think this naming is decent? 

Another option would be a single setter for all experimental features.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ExperimentalFeaturesTest.java
hmottestad: I will add a comment to the test

hmottestad:
TODO:
 - [ ] ExperimentalFeaturesTest comments

jeenbroekstra:
> Another point for discussion @jeenbroekstra.
> 
> We've added support for non-standard SHACL features. Some of them are from DASH (http://datashapes.org), some will be added to DASH and one will be added to the SHACL Advanced Note.
> 
> To enable **all** these features the user needs to call:
> 
> ```
> 		shaclSail.setExperimentalDashSupport(true);
> 		shaclSail.setExperimentalTargetShapeSupport(true);
> ```
> 
> Do you think this naming is decent?

I'd call it "Extended" instead of "Experimental". If we want to indicate it's (currently) experimental, use  an annotation and comment in the javdadoc. 

Also: these new settings should also be added to the ShaclSailConfig class (if you haven't done so yet).

> Another option would be a single setter for all experimental features.

That would perhaps be simpler, but is there a use case where people may want one or the other but where activating all features would be a problem? 

For the two settings you have here, I think that people who want DASH support but are not interested in experimental target support are not inconvenienced by having it active - they simply won't use the other feature in their SHACL data. But I'm not sure if that holds for all experimental features you have in mind.

hmottestad:
Here is what I've ended up with:

```java

	/**
	 * Support for SHACL Advanced Features W3C Working Group Note (https://www.w3.org/TR/shacl-af/). Enabling this
	 * currently enables support for sh:targetShape.
	 *
	 * EXPERIMENTAL!
	 *
	 * @param shaclAdvancedFeatures true to enable (default: false)
	 */
	@Experimental
	public void setShaclAdvancedFeatures(boolean shaclAdvancedFeatures) {
		this.shaclAdvancedFeatures = shaclAdvancedFeatures;
		forceRefreshShapes();
	}

	/**
	 * Support for SHACL Advanced Features W3C Working Group Note (https://www.w3.org/TR/shacl-af/). Enabling this
	 * currently enables support for sh:targetShape.
	 *
	 * EXPERIMENTAL!
	 *
	 * @return true if enabled
	 */
	@Experimental
	public boolean isShaclAdvancedFeatures() {
		return shaclAdvancedFeatures;
	}

	/**
	 * Support for DASH Data Shapes Vocabulary Unofficial Draft (http://datashapes.org/dash). Currently this enables
	 * support for dash:valueIn, dash:AllObjectsTarget and and dash:AllSubjectsTarget.
	 *
	 * EXPERIMENTAL!
	 *
	 * @param dashDataShapes true to enable (default: false)
	 */
	@Experimental
	public void setDashDataShapes(boolean dashDataShapes) {
		this.dashDataShapes = dashDataShapes;
		forceRefreshShapes();
	}

	/**
	 * Support for DASH Data Shapes Vocabulary Unofficial Draft (http://datashapes.org/dash). Currently this enables
	 * support for dash:valueIn, dash:AllObjectsTarget and dash:AllSubjectsTarget.
	 *
	 * EXPERIMENTAL!
	 *
	 * @return true if enabled
	 */
	@Experimental
	public boolean isDashDataShapes() {
		return dashDataShapes;
	}
```

Issue #2192, opened 2020-05-06
Upgrade JSON-LD and Jackson


barthanssens:
Upgrade to 0.13.0, which uses jackson 2.10.

- [x]  CQ 22136, JSON-LD Java

Jackson core, databind and annotation already covered by  CQ21186, CQ21187 and CQ21280

As mentioned on the forum (https://groups.google.com/forum/#!topic/rdf4j-users/231Y18nIRLc), Jackson 2.9.x has a few open CVE issues (which may or may not have impact on RDF4J)

barthanssens:
Bit puzzled why the test sometimes work on jenkins and sometimes don't, looking into it

jeenbroekstra:
Given that this is merged, I think we can close it.

PR #2233, opened 2020-05-17
GH-2192 upgrade jsonld-java and jackson + fix for test

barthanssens:
GitHub issue resolved: #2192 

* updated jackson dependencies
* updated json-ld java
* small change in test
* include newer snake YAML dependency for ES compliance test
---- 
PR Author Checklist: 

### DIFF
--- a/compliance/elasticsearch/pom.xml
+++ b/compliance/elasticsearch/pom.xml
@@ -104,6 +104,12 @@
 			</exclusions>
 		</dependency>
 		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.25</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.apache.logging.log4j</groupId>
 			<artifactId>log4j-core</artifactId>
 			<version>2.11.1</version>
--- a/core/rio/jsonld/src/test/java/org/eclipse/rdf4j/rio/jsonld/JSONLDParserCustomTest.java
+++ b/core/rio/jsonld/src/test/java/org/eclipse/rdf4j/rio/jsonld/JSONLDParserCustomTest.java
@@ -359,7 +359,7 @@ public class JSONLDParserCustomTest {
 			assertTrue(e.getCause() instanceof JsonProcessingException);
 			JsonProcessingException cause = (JsonProcessingException) e.getCause();
 			assertEquals(2, cause.getLocation().getLineNr());
-			assertEquals(1, cause.getLocation().getColumnNr());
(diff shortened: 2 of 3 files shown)


barthanssens:
It seems to fail randomly on the Jenkins machine (or perhaps JDK13), I'll have to check again.

It's probably good idea to wait for the release of Jackson 2.10.5 anyway, since there is an annoying JSON allow-missing-value bug (https://github.com/FasterXML/jackson-core/issues/616)

barthanssens:
Issue in compliance test .ElasticsearchSailGeoSPARQLTest, snakeyaml dependency, could be due to different jackson versions used by elasticsearch-x-content

compliance/elasticsearch/pom.xml
```diff
@@ -101,8 +101,22 @@
...
+					<artifactId>jackson-core</artifactId>
+				</exclusion>
 			</exclusions>
 		</dependency>
+		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.26</version>
+			<scope>test</scope>
+		</dependency>
```
hmottestad, line 119: Maybe just exclude it from the dependency that has the wrong version?

barthanssens: Good point, I excluded the jackson-databind-yaml dependency (which caused the snakeyaml issue) instead of explicitly including the specific (newer) version of snakeyaml

jeenbroekstra: :man_juggling:

jeenbroekstra:
Unfortunately it looks as if excluding the older version does not work - at least locally for me it doesn't. I've pushed the original fix you had in mind, which is to include the newer version as an explicit (test) dependency.

jeenbroekstra:
Argh that's annoying. I've stared using the `gh` command to check out local copies of pulll request from forked repos (such as Fedict). But unlike the old `hub` command, when I push, it just creates a new branch on the origin rather than pushing back to the fork.

jeenbroekstra:
...and it also looks as if I don't have editing rights on your fork Bart.

jeenbroekstra:
I've created a pull request on your fork to resolve this. https://github.com/Fedict/rdf4j/pull/10. I'm sure there must an easier way but my Sunday-brain won't let me think of it.

barthanssens:
Thanks @jeenbroekstra , original idea appears indeed the only one that works. Other attempts did often work _once_ on my local machine, but re-running  clean install -P verify failed on a _second_ run ...

I'd suggest to use Snakeyaml 1.25, since there is already a CQ21086 for it

Issue #1301, opened 2019-02-20
Unable to control precedence of connective operators such as AND and OR expressions


reckart:
AND and OR expressions should be parenthesized by the SPARQL query builder to allow controlling their precedence.

Consider this code:

```
            GraphPattern pattern = VAR_SUBJECT.has(labelProperty, VAR_LABEL_CANDIDATES).filter(
                    or(
                            // Match without language
                            function(STRSTARTS, literalOf(aPrefixQuery)),
                            // Match with default language
                            and(function(STRSTARTS, literalOf(aPrefixQuery)), 
                                    function(LANGMATCHES, function(LANG, VAR_LABEL_CANDIDATES), 
                                            literalOf(language)))));
```

It should match either strings starting with a given prefix or strings in a given language starting with the given prefix, but the AND/OR precedence is lost when the query is rendered:

```
SELECT DISTINCT ?s
WHERE { { ?s <http://www.w3.org/2000/01/rdf-schema#label> ?lc .
FILTER ( STRSTARTS( "Barack" ) || STRSTARTS( "Barack" ) && LANGMATCHES( LANG( ?lc ), "en" ) ) } }
GROUP BY ?s
LIMIT 1000
```

It should be 

```
SELECT DISTINCT ?s
WHERE { { ?s <http://www.w3.org/2000/01/rdf-schema#label> ?lc .
FILTER ( STRSTARTS( "Barack" ) || ( STRSTARTS( "Barack" ) && LANGMATCHES( LANG( ?lc ), "en" ) ) ) } }
GROUP BY ?s
LIMIT 1000
```

reckart:
Just found out that I can call `Expression.parenthesize()` on the `and(...)` - IMHO would be more intuitive if this happened automatically for disjunctions and conjunctions.

abhishekag03:
Hi! We (Me, [@Kvatsx](https://github.com/Kvatsx/), [@ishaanbassi](https://github.com/ishaanbassi/))  would like to work on this issue. 
We have understood it and have implemented the solution suggested by [@reckart](https://github.com/reckart). We have parenthesized the expression in case of AND operator by calling the parenthesize function in the Operation method present in sparqlbuilder/constraint/Operation.java file. This is working for query mentioned by @reckart and for some other queries tested by us.
We have also attached a screenshot of the query and the output as shown in terminal on running the example file.
[...]

PR #2243, opened 2020-05-19
GH-1301 control connective operator precedence

abhishekag03:
GitHub issue resolved: #1301  

Used parenthesize function in case of AND operator by calling it in the Operation method present in sparqlbuilder/constraint/Operation.java file.

---- 
PR Author Checklist: 


barthanssens:
perhaps @anmolsinghal98 could take a look at this ?

anmolsinghal98:
Do you want me to work on the issue @barthanssens ?

barthanssens:
Hi @anmolsinghal98 

I was wondering if perhaps you could take a look at the pull request abhishek created.
(the code and test looks good to me, but I'm not familiar with the SparqlBuilder, so help is appreciated)

abhishekag03:
Hey, @anmolsinghal98 please let us know in case you need any clarifications related to the Pull Request submitted by us?

anmolsinghal98:
Hey. Apologies but it would not be possible for me to review it.  The contributions we are making in rdf4j are part of a course project. Since both Abhishek and I are enrolled in the same course, my review would act as a conflict of interest and the course instructor has advised against it. Hope you understand @barthanssens

barthanssens:
Ow, I didn't know that. 
Thank you all for choosing RDF4J as part of your course projects ! I'll try to look into it

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Operation.java
```diff
@@ -23,7 +23,9 @@
 	Operation(SparqlOperator operator, int operandLimit) {
 		super(operator);
 		this.operandLimit = operandLimit;
-//		parenthesize();
+		if (operator.toString().equals("AND")) {
```
hmottestad, line 26: Using toString() and then .equals() is typically very slow. Try to use the enums instead: ConnectiveOperator.AND

core/sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/examples/sparql11spec/Section6Test.java
```diff
@@ -61,4 +61,19 @@ public void example_6_3() {
...
+
+		GraphPatternNotTriples pricePattern = GraphPatterns.and(x.has(ns.iri("price"), price))
+				.filter(Expressions.or(Expressions.lt(price, 20),
+						Expressions.and(Expressions.lt(price, 50), Expressions.gt(price, 30))))
+				.optional();
+
+		query.prefix(dc, ns).select(title, price).where(x.has(dc.iri("title"), title), pricePattern);
+		p();
+	}
+
```
hmottestad, line 78: I don't see how this tests actually checks if the changes you've made are correct. The test still passes even if I comment out your change.

hmottestad: Btw. I'm sorry the test code in general here is so poor. `p()` isn't particularly useful here (for those reading this comment it actually means System.out.println(query) ), and all the tests are stateful, even though there is no need to keep state between tests.

hmottestad: Second btw. I've take another look at the tests within the `examples` package and realises that none of these test anything, ie. there are no asserts anywhere. When I remove these "tests" I end up with a 13% test coverage. This is very unfortunate, but in now way your fault @abhishekag03

hmottestad: I propose this assert:

```
Assert.assertThat(query.getQueryString(), CoreMatchers.containsString("( ?price < 50 && ?price > 30 )"));
```

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Operation.java
abhishekag03: Thanks for the suggestion! Done the necessary changes.

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Operation.java
```diff
@@ -23,7 +23,7 @@
 	Operation(SparqlOperator operator, int operandLimit) {
 		super(operator);
 		this.operandLimit = operandLimit;
-		if (operator == ConnectiveOperator.AND) {
+		if (operator == ConnectiveOperator.AND || operator == ConnectiveOperator.OR) {
```
abhishekag03, line 26: We noticed that the Paranthesize function needs to be called for the OR operator as well. This is because for queries which have an OR inside an AND, the query output is not how it should be. It does not take into consideration that the OR part must be evaluated first. 
However, calling the Paranthesize function as we have done here is giving us extra parenthesis inside the filter query whenever we have boolean binary operators (AND/OR). Is that fine? If not, kindly help us with how we should remove it. Although considering it evaluation wise, it will always give the correct output. Just that the brackets there are not essential.

hmottestad: I'm ok with the extra parenthesis. I think it would be hard to fix and it's not semantically incorrect, just superfluous.

hmottestad:
LGTM. 

@jeenbroekstra what do you think?

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Operation.java
hmottestad: I am wondering if all connective operators need parenthesises around them? Do you know @abhishekag03 ?

abhishekag03: Yeah, you're actually right! We need to include it for all Connective Operators. We will add the code for the same along with test cases for each operator. I'm just wondering if this should be done in this issue itself or shall we open a new issue and commit the changes there? Or can the issue heading be changed? Please advise what shall be done?

hmottestad: It’ll be easier to do it in this PR and just update the title and description of both the issue and this PR.

abhishekag03: Sure, we have done the same and pushed the code. Since the check had to be done for all connective operators, we have now used `instanceOf` instead of equality check. Kindly have a look and let us know if something needs to be changed.

core/sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/examples/sparql11spec/Section6Test.java
```diff
@@ -61,4 +64,65 @@ public void example_6_3() {
...
+		Variable title = SparqlBuilder.var("title"), price = SparqlBuilder.var("price"), x = SparqlBuilder.var("x");
+
+		GraphPatternNotTriples pricePattern = GraphPatterns.and(x.has(ns.iri("price"), price))
+				.filter(Expressions.or(Expressions.lt(price, Expressions.subtract(Rdf.literalOf(20),
+						Expressions.multiply(Rdf.literalOf(2), Rdf.literalOf(5)))),
+						Expressions.lt(price, 50)))
+				.optional();
+
+		query.prefix(dc, ns).select(title, price).where(x.has(dc.iri("title"), title), pricePattern);
+		Assert.assertThat(query.getQueryString(), CoreMatchers.containsString("( 20 - ( 2 * 5 ) )"));
```
hmottestad, line 110: 20-2*5 == 20-(2*5)

Could you instead test for (20-2)*5 since this is different from 20-2*5?

abhishekag03: Sure. Added the same and pushed it. Also, shall we add all these test cases to the file Section6Test.java only or should there be a separate file for these test cases? I couldn't understand how the Test cases are divided into the various SectionTest files.

hmottestad: As I stated earlier. There are actually almost no tests for the SparqlBuilder (ie. no tests will fail if I change random things in the code). I think adding the tests like you have done here should be fine. But @jeenbroekstra should take a look to see if he agrees. Jeen is the project manager for RDF4J.

hmottestad: Also, thank you @abhishekag03!

abhishekag03: Thank you for the valuable suggestions and prompt responses @hmottestad. 
Will wait for @jeenbroekstra to have a look at the changes made so that this issue can be closed.

jeenbroekstra:
@ahmadbanijamali with respect to the tests for this: as @hmottestad also mentioned, the tests in package `org.eclipse.rdf4j.sparqlbuilder.examples` are not actually real tests: all they do is create a query and print it out. The different sections refer to the sections of the [SPARQL 1.1 Query Language W3C Recommendation](https://www.w3.org/TR/sparql11-query/) - the examples recreate the queries used in the various sections there. 

I would suggest to not edit any of the code existing in `org.eclipse.rdf4j.sparqlbuilder.examples`, and instead introduce your tests as part of other test classes more closely linked to the actual structure and function of what you're testing. In this case, since you are testing the working of 'or' and 'and', which are both expressions, I'd add some a few very specific tests to `org.eclipse.rdf4j.sparqlbuilder.constraint.ExpressionsTest`, and the more general tests (which use the entire SparqlBuilder to create a full query) to `org.eclipse.rdf4j.sparqlbuilder.core.SparqlBuilderTest` (doesn't exist yet, feel free to create).

core/sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/examples/sparql11spec/Section6Test.java
jeenbroekstra: I added a comment on the PR about this, but just to repeat here: I suggest you move these tests out of these `SectionTest` classes completely, and instead introduce a test class `org.eclipse.rdf4j.sparqlbuilder.core.SparqlBuilderTest` where you can add these cases. I'd prefer not to edit the code in the package `org.eclipse.rdf4j.sparqlbuilder.examples` - they are meant to be examples rather than actual code (I don't know if we want to keep them that way, but for now I'd like our actual tests to be separate from it).

abhishekag03:
Thank you @jeenbroekstra @barthanssens @hmottestad. We have made the required changes and pushed the code. Kindly review it once. Please let us know if something is still remaining.
P.S: Kindly update the description and title of the issue to include precedence order of all the connective operators.

core/sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/core/SparqlBuilderTest.java
```diff
@@ -0,0 +1,130 @@
+/*******************************************************************************
+ * Copyright (c) ${year} Eclipse RDF4J contributors.
```
jeenbroekstra, line 2: Looks like your editor forgot to substitute `${year}` with 2020 :)

core/sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/core/SparqlBuilderTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf.iri;
+
+import org.eclipse.rdf4j.model.vocabulary.DC;
+import org.eclipse.rdf4j.sparqlbuilder.constraint.Expressions;
+import org.eclipse.rdf4j.sparqlbuilder.core.query.Queries;
+import org.eclipse.rdf4j.sparqlbuilder.core.query.SelectQuery;
+import org.eclipse.rdf4j.sparqlbuilder.graphpattern.GraphPatternNotTriples;
+import org.eclipse.rdf4j.sparqlbuilder.graphpattern.GraphPatterns;
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf;
+import org.hamcrest.CoreMatchers;
```
jeenbroekstra, line 20: We actually prefer using AssertJ rather than Hamcrest, but no big deal, you can leave as-is.

core/sparqlbuilder/src/test/java/org/eclipse/rdf4j/sparqlbuilder/core/SparqlBuilderTest.java
```diff
@@ -0,0 +1,130 @@
...
+	public void testLogicalOperatorPrecedence1() {
+		Prefix dc = SparqlBuilder.prefix("dc", iri(DC_NS)), ns = SparqlBuilder.prefix("ns", iri(EXAMPLE_ORG_NS));
+		Variable title = SparqlBuilder.var("title"), price = SparqlBuilder.var("price"), x = SparqlBuilder.var("x");
+
+		GraphPatternNotTriples pricePattern = GraphPatterns.and(x.has(ns.iri("price"), price))
+				.filter(Expressions.or(Expressions.lt(price, Rdf.literalOf(20)),
+						Expressions.and(Expressions.lt(price, Rdf.literalOf(50)),
+								Expressions.gt(price, Rdf.literalOf(30)))))
+				.optional();
+		System.out.println(query);
```
jeenbroekstra, line 49: Could you remove the System.out.println ? Just so there is less output to scroll through when running these tests.

abhishekag03:
We have made all the suggested changes and pushed the final code. 
It was great working with you @jeenbroekstra @hmottestad @barthanssens.
It was a great first experience for us. Looking forward to more open source contributions in the future.

Thanks a lot!

hmottestad:
LGTM

Issue #399, opened 2016-09-07
Invalid UPDATE queries are only flagged up when executing the query rather than at parse time


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-2041)

Invalid update queries seem to parse OK - they only fail when calling the execute method on them

{code}
public class InvalidUpdateQueryTest extends TestCase {

```
private Repository m_repository;

@Override
protected void setUp() throws Exception {
    super.setUp();
    m_repository = new SailRepository(new MemoryStore());
    m_repository.initialize();
}

@Override
protected void tearDown() throws Exception {
    super.tearDown();
    m_repository.shutDown();
}

public void testInvalidUpdate() throws Exception {
    RepositoryConnection connection = m_repository.getConnection();

    try {
        Update update = connection.prepareUpdate(QueryLanguage.SPARQL, "insert data { ?s ?p ?o }");
        // should have failed already - NOT in the call below
        update.execute();
    }
    finally {
        connection.close();
    }
}
```

}
{code}

I would have expected the query to fail on prepareUpdate rather than execute

NB this appears to be a regression issue as this worked in 2.7.7

ishaanbassi:
Hi! We (Me, [Kvatsx](https://github.com/Kvatsx), [abhishekag03](https://github.com/abhishekag03)) would like to work on this issue. 
As of now we have run the above test case and managed to successfully get RDFParseException when prepareUpdate is called. We have used the TurtleParser.java code which is called on datablocks only when executeInsertData is called during execution. We have done the same for delete data case.

It would be great if you could assign us the issue. Meanwhile, we will work on creating a pull request as soon as possible.

barthanssens:
Great, thank you !

Issue #420, opened 2016-09-07
HTTPRepositoryConnection.prepareGraphQuery() fails to throw IllegalArgumentException as documented in RepositoryConnection


abrokenjester:
(Migrated from https://openrdf.atlassian.net/browse/SES-1932)

In SES-1931, a developer was confused by an exception later in execution when they provided a tuple query text to {{HTTPRepositoryConnection.prepareGraphQuery(...)}}. The {{RepositoryConnection.prepareGraphQuery(...)}} javadoc states that an {{IllegalArgumentException}} will be thrown for malformed queries. Some client-side parsing, if only to validate the query text, should happen so that the code fails fast as expected.

barthanssens:
related to #399 ?

jeenbroekstra:
@barthanssens not really, in this case. The situation here is that queries do not get parsed/validated at the client side in the HTTPconnection - it just sends the unparsed query string to the server when evaluate() is called, and then the server parses it - so from the client perspective the error happens on execute rather than parse. 

The issue you link to is, in a way, the same sort of thing, but it happens only for specific kinds of updates (namely INSERT DATA updates), and happens regardless of whether you are working over HTTP or with a local store. The cause in this case is that these updates are not pushed through the query parser but instead are passed on, at execution time, to a modified Turtle parser instead.

PR #2248, opened 2020-05-20
GH-399 detect invalid INSERT/DELETE on parse

ishaanbassi:
GitHub issue resolved: #399

Added code for detecting invalid queries for both the insert and delete data expressions in prepareUpdate function. Also added test cases to check that RDFParseException is thrown in both cases.

---- 
PR Author Checklist: 

 - [x] every commit message starts with the issue number (GH-399) followed by a meaningful description of the change


ishaanbassi:
Hey! In addition to the pull request we require your valuable feedback on the following things:

1) We have added the test case in the file. We tried adding it in [SPARQLRepositorySparqlUpdateTest.java](https://github.com/eclipse/rdf4j/pull/2248/files#diff-f14b11ade3205543fa7dab3c41724b5d) but that was giving us a circular dependency error. We also tried making a new class inside the [package](https://github.com/eclipse/rdf4j/tree/master/core/repository/sail/src/test/java/org/eclipse/rdf4j/repository/sail) but that was not working. Please advise if this is fine or where shall we add it.

2) To assert that the test case raises an "RDFParseException" we have added a try-catch and asserted the line number of the exception to be 7 as per the insert query. We tried adding `@Test(exception = RDFParseException.class)`, but that did not work. Please suggest if this is fine or not?

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
```diff
@@ -262,7 +269,22 @@ public SailBooleanQuery prepareBooleanQuery(QueryLanguage ql, String queryString
...
 		ParsedUpdate parsedUpdate = QueryParserUtil.parseUpdate(ql, update, baseURI);
+		List<UpdateExpr> updateExprs = parsedUpdate.getUpdateExprs();
+		SPARQLUpdateDataBlockParser parser = new SPARQLUpdateDataBlockParser(this.getValueFactory());
+		for (UpdateExpr expr : updateExprs) {
+			String datablock = "";
+			if (expr instanceof InsertData) {
+				datablock = ((InsertData) expr).getDataBlock();
+			} else if (expr instanceof DeleteData) {
+				datablock = ((DeleteData) expr).getDataBlock();
+			}
```
jeenbroekstra, line 280: I was thinking that perhaps we should move all of this into the SPARQLParser itself. It seems the more logical place to handle it.Then again doing it here gives us a bit more flexibility. Fine either way.

When actually executing the insert/delete, the parser is reconfigured for specific types of errors. For INSERTs, the following config settings are done:

   		parser.setLineNumberOffset(insertDataExpr.getLineNumberOffset());
		parser.getParserConfig().addNonFatalError(BasicParserSettings.VERIFY_DATATYPE_VALUES);
		parser.getParserConfig().addNonFatalError(BasicParserSettings.FAIL_ON_UNKNOWN_DATATYPES);
		parser.getParserConfig().set(BasicParserSettings.SKOLEMIZE_ORIGIN, null);

and for DELETE, the following:

		parser.setLineNumberOffset(deleteDataExpr.getLineNumberOffset());
		parser.setAllowBlankNodes(false); // no blank nodes allowed in DELETE DATA.

I don't think we need the 'non-fatal error' settings (after all we _want_ a parse exception when an error occurs), but the other settings we should probably duplicate here. The line number offset will hopefully correct the odd line numbers you're currently seeing in the tests.

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLRepositorySparqlUpdateTest.java
```diff
@@ -52,3 +64,42 @@
 //		System.err.println("temporarily disabled testAutoCommitHandling() for HTTPRepository");
 //	}
 //}
+public class SPARQLRepositorySparqlUpdateTest extends TestCase {
+
+	private Repository m_repository;
+
+	@Override
+	protected void setUp() throws Exception {
```
jeenbroekstra, line 72: In JUnit 4, rather than `@Override`, this should have a `@Before` annotation.

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLRepositorySparqlUpdateTest.java
```diff
@@ -52,3 +64,42 @@
...
+		super.setUp();
+		m_repository = new SailRepository(new MemoryStore());
+	}
+
+	@Override
```
jeenbroekstra, line 77: Use JUnit 4 `@After`  instead of overrriding.

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLRepositorySparqlUpdateTest.java
```diff
@@ -52,3 +64,42 @@
...
+	}
+
+	// @Test (expected = org.eclipse.rdf4j.rio.RDFParseException.class)
+	@Test
+	public void testInvalidInsertUpdate() {
+		RepositoryConnection connection = m_repository.getConnection();
+		try {
+			Update update = connection.prepareUpdate(QueryLanguage.SPARQL, "insert data { ?s ?p ?o }");
+		} catch (RDFParseException rdfpe) {
+			Assert.assertEquals(7, rdfpe.getLineNumber());
```
jeenbroekstra, line 90: Why is it reporting line 7, on a one-line update? That can't be right.

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLRepositorySparqlUpdateTest.java
```diff
@@ -52,3 +64,42 @@
 //		System.err.println("temporarily disabled testAutoCommitHandling() for HTTPRepository");
 //	}
 //}
+public class SPARQLRepositorySparqlUpdateTest extends TestCase {
```
jeenbroekstra, line 67: Ah, I only just now noticed that this was an existing, commented-out test class that you added your test to. I don't think that's quite right, let me have a closer look and I'll try and give some advice on how to better organize this.

jeenbroekstra:
> Hey! In addition to the pull request we require your valuable feedback on the following things:
> 
>     1. We have added the test case in the file. We tried adding it in [SPARQLRepositorySparqlUpdateTest.java](https://github.com/eclipse/rdf4j/pull/2248/files#diff-f14b11ade3205543fa7dab3c41724b5d) but that was giving us a circular dependency error. We also tried making a new class inside the [package](https://github.com/eclipse/rdf4j/tree/master/core/repository/sail/src/test/java/org/eclipse/rdf4j/repository/sail) but that was not working. Please advise if this is fine or where shall we add it.

I think you should add these test cases to `org.eclipse.rdf4j.query.parser.sparql.SPARQLUpdateTest`, in the rdf4j-sparql-testsuite module. To then execute the test on a local repository, you can run `org.eclipse.rdf4j.sail.memory.MemorySPARQLUpdateTest` (in the rdf4j-sparql-compliance module). And to run it against the HTTP Repository, you can run `org.eclipse.rdf4j.query.parser.sparql.HTTPSparqlUpdateTest`. Both of these classes extend the abstract SPARQLUpdateTest class. 
 
>     2. To assert that the test case raises an "RDFParseException" we have added a try-catch and asserted the line number of the exception to be 7 as per the insert query. We tried adding `@Test(exception = RDFParseException.class)`, but that did not work. Please suggest if this is fine or not?

It's fine as-is (except for the fact that I think it should throw a MalformedQueryException, not an RDFParseException). If you still want to use a test annotation for this, that's fine too. The correct syntax is `@Test(expected = MalformedQueryException.class)` (so `expected`, not `exception`).

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLRepositorySparqlUpdateTest.java
jeenbroekstra: Yeah, I think you should undo this, leave this old class commented out, and move your tests to the test suite. See my other comment, below, for details.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
ishaanbassi: Hey @jeenbroekstra, thanks for the valuable feedback. We are facing the following error when adding SPARQLUpdateDataBlockParser in the SPARQLParser.java file - 
"Adding dependency on module 'rdf4j-repository-sail' will introduce **circular dependency** between modules 'rdf4j-queryalgebra-evaluation' and 'rdf4j-queryparser-sparql'" .

Let us know how to proceed. Meanwhile, we are trying to figure out some other way to parse datablocks.

jeenbroekstra: @ishaanbassi Ah good point, I hadn't thought of that. This will require a bit of refactoring. I suggest you move the SPARQLUpdateDatablockParser class to package `org.eclipse.rdf4j.query.parser.sparql`, in the `rdf4j-queryparser-sparql` module. That should solve the circular dependency issue. You will also need to add a new dependency for the trig parser to the pom.xml in the `rdf4j-queryparser-sparql` module:

		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>rdf4j-rio-trig</artifactId>
		</dependency>

To preserve backward compatibility, leave an empty class with the old name in the old location, marked `@deprecated`.  

You know, I was actually doing these change while writing this, to make sure I wasn't making a mistake. I might as well push to your branch, so you can see what I mean.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java
```diff
@@ -130,23 +130,23 @@ public ParsedUpdate parseUpdate(String updateStr, String baseURI) throws Malform
...
 					} else if (updateExpr instanceof DeleteData) {
-						DeleteData deleteDataExpr = (DeleteData)updateExpr;
+						DeleteData deleteDataExpr = (DeleteData) updateExpr;
 						parser.setLineNumberOffset(deleteDataExpr.getLineNumberOffset());
 						parser.setAllowBlankNodes(false);
 						datablock = deleteDataExpr.getDataBlock();
 					}
-					if(!datablock.equals("")){
+
+					if (!datablock.equals("")) {
```
ishaanbassi, line 144: Thanks for the help @jeenbroekstra . We have done all the changes suggested above.

We also noted one thing in the SPARQLParser.java file. In case the `updateNode` itself is not null, then do we need to check separately that the `datablock` is empty? We have kept this condition for now.

jeenbroekstra: Yeah, I think that how you currently have it is correct: you only want to process the datablock if it is not empty. Even if `updateNode` is not null, it can still be a different type of update, e.g. a  `Modify` operation, in which case there is no datablock to parse. 

I don't think it matters too much because the parser will return very quickly on an empty string, but I think doing a quick string equality check is marginally more efficient.

jeenbroekstra:
Failed tests: 
```
  HTTPSparqlUpdateTest.testInvalidDeleteUpdate Expected exception: org.eclipse.rdf4j.query.MalformedQueryException
  HTTPSparqlUpdateTest.testInvalidInsertUpdate Expected exception: org.eclipse.rdf4j.query.MalformedQueryException
  W3CApprovedSPARQL10SyntaxTest>SPARQL11SyntaxTest.runTest:166 Positive test case failed: Expected ':', found ' ' [line 2]
  W3CApprovedSPARQL10SyntaxTest>SPARQL11SyntaxTest.runTest:166 Positive test case failed: Expected ':', found ' ' [line 2]
  W3CApprovedSPARQL11SyntaxTest.runTest:44->SPARQL11SyntaxTest.runTest:166 Positive test case failed: Expected ':', found ' ' [line 2]
  W3CApprovedSPARQL11SyntaxTest.runTest:44->SPARQL11SyntaxTest.runTest:166 Positive test case failed: Expected ':', found ' ' [line 2]
```

The first two test failures are expected - this is because in the HTTPRepository, we don't parse queries or updates at the client side right now (this is a known issue, see #420). This is out of scope for this PR, so I suggest you just disable those two test cases in HTTPSparqlUpdateTest by overriding and marking them `@Ignored` - possible add a code comment linking to github issue #420. There's several tests already being ignored in HTTPSparqlUpdate, just follow that same pattern.

The other three test failures are trickier. I suspect they have something to do with the base URI not being passed in.

jeenbroekstra:
I had a closer look at one of the failing W3C syntax tests, and I believe your fix has uncovered an underlying bug in the SPARQLUpdateDataBlockParser. 

The following update fails to parse: 

```
INSERT DATA { 
  <s1> <p1> <o1>
  GRAPH <G> { <s> <p1> 'o1'; <p2> <o2> } 
  GRAPH <G1> { <s> <p1> 'o1'; <p2> <o2> } 
  <s1> <p1> <o1>
}
```

The reason it fails is that the parser is based on the TriGParser, which expects a '.' after the first triple pattern (so after the <o1> on the first line of the block). Now, in TriG syntax, that expectation is correct: the TriG grammar mandates that a triple pattern ends with a '.' (see https://www.w3.org/TR/trig/#grammar-production-triplesOrGraph). However, SPARQL 1.1 update syntax is more lenient (see https://www.w3.org/TR/sparql11-query/#rQuadData): the full stop at the end is optional only. So this update query is in fact syntactically correct, and our parser is wrong to throw an exception. We just never caught this before because the sparql parser previously didn't process the data blocks. 

The cause is in TriGParser, line 198, where it reads away the final character after parsing a graph pattern. We should override this behavior in the SPARQLUpdateDatablockParser, and instead of just reading the code point away, peek at it, and only read it away if it is either a fullstop or a closing curly brace.

Let me know if you'd like to try and fix this yourself, or if you want me to push a fix to your branch. It's really a bug that is independent from your fix, but it's probably easy enough to fix that we can just do it on your PR.

ishaanbassi:
Hey @jeenbroekstra, we were able to fix the 6 test cases that were failing before. However we now encountered new test cases in `ShaclTest` that are failing after the above changes. Kindly look into this.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLUpdateDataBlockParser.java
```diff
@@ -70,7 +72,90 @@ public RDFFormat getRDFFormat() {
...
+				parsePredicateObjectList();
+			}
+			// Or if we didn't recognise anything, just parse as Turtle
+			else {
+				unread(c);
+				parseTriples();
+			}
+		}
+
+		if (c == '.' || c == '}') {
```
ishaanbassi, line 156: @jeenbroekstra, we found that the failing testcases in ShaclTest, are due to this condition on readCodePoint function. Hence, we are not sure if these changes are sufficient. Kindly suggest an alternative solution or you can push a fix. Thanks for the help!

jeenbroekstra: The problem is that you are looking at the wrong character. You haven't advanced `c` yet, so it's still the last processed char - which will never be equal to '.' or '}', so the end char is never read. This is what causes the failures in the SHACL tests.

To fix, add this line before your if-condition:

 		c = peekCodePoint();

That sets the value of `c` to the next char to be read, and then we decide to actually read if it is indeed a full stop or a closign bracket.

jeenbroekstra:
Excellent, thanks again! I will merge this now, and it will either be part of the next patch release (3.2.2) or of the next minor release (3.3.0), whichever comes first.

As a last thing, could I ask you to comment on https://github.com/eclipse/rdf4j/issues/2269 and list your names so I can mention you in the release notes?

Issue #405, opened 2016-09-07
add rdf:Alt, rdf:Bag, rdf:Seq to org.eclipse.rdf4j.model


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-2008)

In many cases it is very convenient to create statements with java/scala in a typesafe controlled manner and without SPARQL string evaluation. But it is not possible with current API as there are no (rdf:Alt, rdf:Bag, rdf:Seq) That means that  in many cases instead of writing several lines of scala/java code users have to write SPARQL Update/Insert quries that are not only often inconvenient and potential source of bugs (as they are not typesafe) but also have SPARQL parsing overhead (esp. in such intensive operations like generating SPIN SPARQL rules from quires in embedded storages for instance).

guvi007:
Hi @jeenbroekstra, I would like to know if this issue is still open and if I could work on this.

jeenbroekstra:
@guvi007 it's certainly still open and we'd welcome your input. The requirements here are a little vague though. Do you have a particular design in mind so we can discuss the approach?

If you need inspiration, one approach might be something similar to what we did for RDF collections: a couple of utility methods to convert to/from Java collections.

reeshabhkumarranjan:
We are thinking to refer to RDFCollections.java and code a similar class for RDFContainers. We will create a base class RDFContainer, and will extend it from Bag, Sequence and Alt. We might need to copy some code from RDFCollection, if that's fine. This is because much of the basic implementation might be the same. Also, I would like to tell you that three people- me, @prince17080, @guvi007 ) will be working on this issue, and the pull request will be done by Gaurav

jeenbroekstra:
Sounds good, although I wonder if it is necessary to have separate classes for Bag, Sequence and Alt - the only structural difference is just the type, so a single RDFContainers util that has methods for producing the three different kinds of container (represented as Model objects) might be easier. Feel free to put up an early pull request to discuss what you've got, even if it's not complete yet. 

Also: before you get started, please have a look at the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md), in particular the points on how to sign the Eclipse ECA. Since you will be working with multiple authors, be aware that _every_ person who adds a commit to your change will need to sign this ECA. It should be straightforward, it's an online form, shouldn't take more than 10 minutes to sort out. Let me know if you need any help with any of it.

reeshabhkumarranjan:
[...]

Issue #2235, opened 2020-05-18
release notes for rdf4j 3.3.0


jeenbroekstra:
Task to keep track of what needs to go into next release notes. 

See release milestone at https://github.com/eclipse/rdf4j/milestone/51

jeenbroekstra:
PR up at https://github.com/eclipse/rdf4j-doc/pull/161

reeshabhkumarranjan:
Hey @jeenbroekstra , here are our names:
me: Reeshabh Kumar Ranjan
@guvi007: Gaurav Agarwal
@prince17080: Prince Sachdeva

SakshiSaini17092:
Hey @jeenbroekstra,
@SakshiSaini17092 : Sakshi Saini
@gauri-b : Gauri Bhatanagar
@raghav17083: Raghav Rathi

barthanssens:
Thanks for your help, I've added the names to https://github.com/eclipse/rdf4j-doc/pull/161/files

navdhaagarwal:
Hi @barthanssens .
@navdhaagarwal : Navdha Agarwal
@shubhi0108 : Shubhi Jain

barthanssens:
Thank you, I've added your names to the release notes

jeenbroekstra:
To do: add some notes about new transaction settings, new function interface, and other new features

hmottestad:
Also update the SHACL example that explains how to validate large transactions (bulk).

PR #2254, opened 2020-05-21
GH-405 add alt bag seq utility functions

reeshabhkumarranjan:
GitHub issue resolved: #405 

We have referred `RDFCollections.java` to implement the `consumeContainer()` (analogous to `consumeCollection()`) in the newly created `RDFContainers.java`, and same for `asRDF()` method. We are taking the type of collection from the input, and are checking that it should only be one out of `RDF.ALT`, `RDF.BAG` and `RDF.SEQ`.
In the last `consumeContainer()` method, we have iterated over the supplied `values` iterable, and added 2 types of predicates for every element:

1. `value rdfs:member container`
2. `container rdf:_nnn value`

In the above example, `container` is the actual `Resource` representing the container in RDF.


core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,65 @@
...
+		Objects.requireNonNull(values, "input collection may not be null");
+		Objects.requireNonNull(consumer, "consumer may not be null");
+		Objects.requireNonNull(vf, "injected value factory may not be null");
+
+		Resource current = container != null ? container : vf.createBNode();
+		boolean validType = Objects.equals(containerType, RDF.ALT) ||
+				Objects.equals(containerType, RDF.BAG) ||
+				Objects.equals(containerType, RDF.SEQ);
+
+		assert validType: "containerType should be one of ALT, BAG or SEQ";
```
jeenbroekstra, line 50: I'd do a check with an exception thrown here, rather than an assertion.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+
+package org.eclipse.rdf4j.model.util;
+
+import org.eclipse.rdf4j.model.*;
```
jeenbroekstra, line 11: Our formatting guidelines don't allow wildcard imports. Have a look at https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#code-formatting for some tips on how set up your environment and/or run a few quick maven commands to fix your formatting.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,65 @@
...
+
+		Iterator<?> iter = values.iterator();
+		int elementCounter = 1;
+		while (iter.hasNext()) {
+			Object o = iter.next();
+			Value v = o instanceof Value ? (Value) o : Literals.createLiteralOrFail(vf, o);
+			IRI elementCounterPredicate = vf.createIRI(RDF.NAMESPACE, "_" + elementCounter);
+			elementCounter++;
+			Statements.consume(vf, current, elementCounterPredicate, v, consumer, contexts);
+			Statements.consume(vf, (Resource) v, RDFS.MEMBER, current, consumer, contexts);
```
jeenbroekstra, line 62: Isn't this duplicating information? having `:foo rdf:_1 :bar` already implies that `:foo rdfs:member :bar`. I don't think you need to insert the `rdfs:member` property explicitly. Or do you have a specific use case in mind?

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+/**
+ * TODO add documentation
+ */
+public class RDFContainers {
+
+	public static <C extends Collection<Statement>> C asRDF(IRI containerType, Iterable<?> values, Resource container, C sink,
```
jeenbroekstra, line 27: Perhaps a nice addition would be to also have conversion methods that pick the container type, e.g `asAlt`, `asSeq`, and `asBag`.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
reeshabhkumarranjan: Alright, we will replace it with throwing an exception.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
reeshabhkumarranjan: Actually we were trying to implement the suggestion in the book Semantic Web Technologies by Pascal Hitzler, Markus Krötzsch and Sebastian Rudolph, page number 61.
The `:container rdf:_1 :element` is for the ordering in `rdf:Seq`.
The `:element rdfs:member :container` is to reduce the time taken to check the membership of an element in a container. If we remove this triple, we would need to iterate over all `:container rdf:_nnn :element` to see if it exists. Otherwise, a single check for `:element rdfs:member :container` will be sufficient.

reeshabhkumarranjan: The `:container rdf:_1 :element` is for the ordering in `rdf:Seq`.
The `:element rdfs:member :container` is to reduce the time taken to check the membership of an element in a container. If we remove this triple, we would need to iterate over all `:container rdf:_nnn :element` to see if it exists. Otherwise, a single check for `:element rdfs:member :container` will be sufficient.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
reeshabhkumarranjan: We will take care about this in our further commits/PRs.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
reeshabhkumarranjan: Actually the `containerType` passed into the `asRDF` method is for that purpose only. We will use the `containerType` parameter in further implementations. Should we have those separate methods as well?

reeshabhkumarranjan:
@jeenbroekstra I will format the file as per guidelines in the final commit, before that, please check my replies to your comments above. Also I have added all the necessary methods in the file. I will soon add the documentation.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
hmottestad: Just a comment on the use of “as”. Java has “toString” and that’s made me use “to” rather than “as” in other code.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
jeenbroekstra: You'll also need to run `mvn impsort:sort`.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
jeenbroekstra: @hmottestad I think in this case it's modeled off of the naming we use in the RDFCollections class for similar methods. I kind of agree with you that 'to' might have been a better choice ('as' to me suggests a view on an object rather than a new, converted, object), but to stay consistent I'm fine with sticking with this.

jeenbroekstra: @hmottestad  actually, coming back to it, I think you are right, and we might as well change this. @reeshabhkumarranjan do you think you could do a fix to change all the `as...` method names `asRDF`, `asValues` to `to...` method names (`toRDF`, `toValues`, etc.)?

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
jeenbroekstra: That makes sense, thanks, and well picked up on that mistake, I completely missed that!

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,397 @@
...
+	 *                      {@code null}.
+	 * @param vf            the {@link ValueFactory} to use for creation of new model objects. May not be {@code null}
+	 * @param contexts      the context(s) in which to add the RDF Container. This argument is an optional vararg and
+	 *                      can be left out.
+	 * @throws LiteralUtilException if one of the supplied values can not be converted to a Literal.
+	 * @see <a href="https://www.w3.org/TR/rdf-schema/#ch_container">RDF Schema 1.1 section on Collection
+	 *      vocabulary</a>.
+	 * @see Literals#createLiteralOrFail(ValueFactory, Object)
+	 *
+	 * @since 3.0
```
jeenbroekstra, line 212: I think that should be '3.3.0'.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,397 @@
...
+		Objects.requireNonNull(consumer, "consumer may not be null");
+		Objects.requireNonNull(vf, "injected value factory may not be null");
+
+		Resource current = container != null ? container : vf.createBNode();
+		boolean validType = Objects.equals(containerType, RDF.ALT) ||
+				Objects.equals(containerType, RDF.BAG) ||
+				Objects.equals(containerType, RDF.SEQ);
+
+		if (!validType) {
+			throw new RuntimeException("containerType should be one of ALT, BAG or SEQ");
```
jeenbroekstra, line 227: I think this should be a `ModelException`, rather than a generic `RuntimeException`.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,397 @@
...
+	 *                      RDF Collection statements will be added. May not be {@code null}.
+	 * @param contexts      the context(s) in which to add the RDF Containter. This argument is an optional vararg and
+	 *                      can be left out.
+	 * @return the supplied sink {@link Collection} of {@link Statement}s, with the new Statements forming the RDF
+	 *         Collection added.
+	 * @throws LiteralUtilException if one of the supplied values can not be converted to a Literal.
+	 * @see <a href="https://www.w3.org/TR/rdf-schema/#ch_container">RDF Schema 1.1 section on Collection
+	 *      vocabulary</a>.
+	 */
+	public static <C extends Collection<Statement>> C asRDF(IRI containerType, Iterable<?> values, Resource container,
```
jeenbroekstra, line 83: Like @hmottestad also suggested above, I think renaming this to `toRDF` (and similarly, also rename `asValues` to `toValues`) is better. I realize you followed the naming we used in RDFCollections for this but I think we got this wrong there, and we might as well do it right in this one (we'll fix it in the RDFCollections later as well, but that's a separate issue).

reeshabhkumarranjan: > Like @hmottestad also suggested above, I think renaming this to `toRDF` (and similarly, also rename `asValues` to `toValues`) is better. I realize you followed the naming we used in RDFCollections for this but I think we got this wrong there, and we might as well do it right in this one (we'll fix it in the RDFCollections later as well, but that's a separate issue).

@hmottestad @jeenbroekstra Unable to reply [here](https://github.com/eclipse/rdf4j/pull/2254#discussion_r429586815) because GitHub added the `outdated` tag to the conversation there. Yes, I can change the naming from `as*` to `to*`. However, I have to do this in `RDFContainers.java` only, right?

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
reeshabhkumarranjan: @jeenbroekstra Yes I can do this. Just to confirm, I have to do this only in `RDFContainers.java`, right?

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
prince17080: `mvn impsort:sort` has fixed the formatting successfully.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
jeenbroekstra: Yes, that's correct, only `RDFContainers`. We'll fix RDFCollections later, separately.

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFContainers.java
```diff
@@ -0,0 +1,160 @@
...
+				break;
+			}
+			Statement statement = statementSupplier.get(container, annotatedMembershipPredicate, null, contexts).get();
+
+			if (containerType.equals(RDF.ALT)) {
+				if (encountered.contains(statement.getObject())) {
+					throw exceptionSupplier.apply("rdf:alt cannot contain duplicate values").get();
+				}
+				encountered.add(statement.getObject());
+			}
```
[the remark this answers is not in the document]
jeenbroekstra: I'd actually leave this check out, completely. It may seem logical to test this, but there is nothing in the RDF specifications that says anything about the members of an `rdf:Alt` container having to be distinct.

reeshabhkumarranjan:
> Thanks for the latest update, looking good. See my previous response regarding the check you added for rdf:Alt containers.
> 
> Also, finally: can you add a few unit tests? Nothing too difficult, just some tests that create some simple containers from Java collections and vice versa. I think you can look at RDFCollectionsTest for inspiration.

Sure, we will add the tests soon.

jeenbroekstra:
One final request: could you give your names as you'd like them mentioned for our release notes? We make it a point to mention who contributed to a release, and I want to make sure I get spelling right and include everyone who contributed. Please add a comment to #2235.

Issue #2208, opened 2020-05-09
Graphviz DOT based query plans


hmottestad:
Output plans in Graphviz DOT format. Use colors to highlight slow nodes.

hmottestad:
Example.

![graphviz](https://user-images.githubusercontent.com/797185/81484347-1b801900-9245-11ea-9162-540cbd79fa38.png)

hmottestad:
@jeenbroekstra @barthanssens what do you guys think of this?

barthanssens:
Looks very, very nice !

PR #2257, opened 2020-05-23
GH-2208 dot format output for query plans

hmottestad:
GitHub issue resolved: #2208  

New output format for query plans based on Graphviz DOT format to visualise query plan better and also add some color to indicate timing

---- 
PR Author Checklist: 


hmottestad:
Here is an example:

 - Header color saturation indicates % of total time. Eg. more saturated red == higher % of total time.
 - Some table fields also have colors
 - Boxes drawn around subgraphs to indicate new scope

graphviz-6

hmottestad:
Would also be nice to have prefixes in the query plan, eg. `rdf:type` instead of `http://www.w3.org/1999/02/22-rdf-syntax-ns#type`. But that's for another day.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
```diff
@@ -30,6 +30,7 @@
 import org.eclipse.rdf4j.repository.sail.SailRepository;
 import org.eclipse.rdf4j.repository.sail.SailRepositoryConnection;
 import org.eclipse.rdf4j.rio.RDFFormat;
+import org.hamcrest.CoreMatchers;
```
jeenbroekstra, line 33: Can you replace with using assertJ matchers?

I'm actually kind of surprised this compiles - we switched from hamcrest to assertj a while back, and I thought we would have removed the hamcrest libs from the build path.

jeenbroekstra: Ah, of course, I keep forgetting that hamcrest core is actually a dependency of Junit 4... Still, I'd prefer if we stuck to assertJ instead of hamcrest, as a matter of consistency.

hmottestad: I've fixed this now. Maybe we should exclude hamcrest in our pom files?

Issue #2261, opened 2020-05-24
Rename XMLSchema to XSD


jeenbroekstra:
The vocabulary class that represents the `xsd` namespace is currently called `XMLSchema`. This is annoyingly inconsistent, hard to discover and also potentially confusing because there is a class of the same name in Apache Xerces.

To fix, I suggest we rename to `XSD`, keeping the old name in deprecated form for backward compatibility.

PR #2264, opened 2020-05-24
GH-2261 rename XMLSchema to XSD

jeenbroekstra:
GitHub issue resolved: #2261 

- renamed `XMLSchema` to `XSD` (because it has been irritating me for years)
- kept deprecated `XMLSchema` class for backward compatibility

---- 
PR Author Checklist: 


hmottestad:
God yes. I always have to spend a minute trying to find it anytime I want a standard datatype. 

FYI Jena is just as bad. Theirs is called XSDDatatype.

core/model/src/main/java/org/eclipse/rdf4j/model/vocabulary/XSD.java
```diff
@@ -0,0 +1,187 @@
...
+	public final static IRI NMTOKEN = create("NMTOKEN");
+
+	/** <tt>http://www.w3.org/2001/XMLSchema#NMTOKENS</tt> */
+	public final static IRI NMTOKENS = create("NMTOKENS");
+
+	/** <tt>http://www.w3.org/2001/XMLSchema#Name</tt> */
+	public final static IRI NAME = create("Name");
+
+	/** <tt>http://www.w3.org/2001/XMLSchema#NCName</tt> */
+	public final static IRI NCNAME = create("NCName");
```
hmottestad, line 125: I also prefer this structure over the one where we have a separate static block to instantiate the field.

Btw. I still think the use of uppercase for our fields is absolutely horrible. It is convention over common sense. This is particularly bad for the SHACL namespace, eg. sh:nodeKind and sh:NodeKind.

jeenbroekstra: I don't think we envisaged different case having different meanings when we originally adopted the convention for this. You're right though. I wouldn't mind changing it but we'll have to do them all in one go, and of course keep old all-uppercase variants around for backward compatibility.

Issue #1913, opened 2020-02-10
N3Parser ? just handling "=" sign


tfrancart:
Hello

Some files out there are N3 files, just using the "=" sign as an alias of owl:sameAs, not using any of the others N3 features. (e.g. https://cdstar.shh.mpg.de/bitstreams/EAEA0-3DAE-E27B-4692-0/glottolog_language.n3.gz from https://glottolog.org/meta/downloads).

```
<http://glottolog.org/resource/languoid/id/cani1243> a dcterms:LinguisticSystem,
        gold:Language ;
    rdfs:label "Canichana"@en ;
    void:inDataset <http://glottolog.org/> ;
    = <http://dbpedia.org/resource/ISO_639:caz> ;
    geo:lat "-14.3058"^^xsd:float ;
    geo:long "-64.944455"^^xsd:float ;
.
```

It would be nice to have a basic N3Parser that can handle this special case, probably without other features of N3.

Could be a simple addition to the TurtleParser, too.

jeenbroekstra:
Currently, Rio's N3Parser is just the TurtleParser, under the hood. Adding this feature would require introducing a subclass that extends the Turtle syntax (I don't think it's a good idea to add this to the TurtleParser itself - let's keep that standards-compliant as much as possible).

tfrancart:
Sounds good. I'll see if I can propose something, bandwidth permitting.

jeenbroekstra:
That'd be great @tfrancart! Before you get going, please have a look at the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md), in particular the points on how to sign the Eclipse Contributor Agreement. Shouldn't take more than 10 minutes to get set up. Let me know if you need any help with any of it.

tanishapandey:
Hello
I would like to work on this issue. I was hoping if the issue could be assigned to me. 
I have signed the Eclipse Contributor Agreement as mentioned in the contributor guidelines.

Thank you

barthanssens:
Ok, thank you for your help !

PR #2278, opened 2020-05-27
GH-1913: Added N3Parser to handle '=' sign 

tanishapandey:
Signed-off-by: tanishapandey tanishapandey99@gmail.com

GitHub issue resolved: #1913 

- Handled '=' sign in N3 format by creating a new parser which extends TurtleParser and overwrites the parsePredicate function to handle the '=' sign.
- Made corresponding changes in the N3ParserFactory file to return N3Parser object
- Added unit test in the new N3ParserTest file.

### DIFF
--- a/core/rio/n3/src/main/java/org/eclipse/rdf4j/rio/n3/N3Parser.java
+++ b/core/rio/n3/src/main/java/org/eclipse/rdf4j/rio/n3/N3Parser.java
@@ -1,0 +1,84 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.rio.n3;
+
+import java.io.IOException;
+
+import org.eclipse.rdf4j.model.IRI;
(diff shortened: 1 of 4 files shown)


core/rio/n3/src/main/java/org/eclipse/rdf4j/rio/n3/N3Parser.java
```diff
@@ -0,0 +1,84 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: This is not the correct copyright header I'm afraid. It should be this:

```
```

core/rio/n3/src/test/java/org/eclipse/rdf4j/rio/n3/N3ParserTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+		assertTrue(errorCollector.getWarnings().isEmpty());
+		assertTrue(errorCollector.getErrors().isEmpty());
+		assertTrue(errorCollector.getFatalErrors().isEmpty());
+
+		assertFalse(statementCollector.getStatements().isEmpty());
+		assertEquals(1, statementCollector.getStatements().size());
+
+		for (Statement st : statementCollector.getStatements()) {
+			System.out.println(st);
```
jeenbroekstra, line 70: Ah, only just now noticed this - instead of printing this to stdout can you use do another assertion here, to check that the statement is what we expect: `(ex:foo, owl:sameAs, ex:bar)`?

jeenbroekstra:
Replaced this PR with #2345

Issue #1079, opened 2018-09-10
TurtleWriter with INLINE_BLANK_NODES loses bnode id when bnode is referenced more than once


pmaria:
Writing Turtle using `BasicWriterSettings.INLINE_BLANK_NODES` where a blank node is referenced more than once leads to incorrect serialisation.

There is a warning in the Javadoc of this option that states

> Inline blanks nodes by their value and don't write any blank node labels when this setting is true.
This setting should only be used when blank nodes never appear in the context and there are no blank node cycles.

But I don't know if that covers this case. I don't understand what is meant by "never appear in the context".

Here is a snippet which reproduces the issue:

```java
ValueFactory vf = SimpleValueFactory.getInstance();
    
BNode bnode = vf.createBNode("bnode");

Model model = new ModelBuilder().subject("http://ex.com/subject")
	.add(vf.createIRI("http://ex.com/rel1"), bnode)
	.add(vf.createIRI("http://ex.com/rel2"), bnode)
	.subject(bnode)
	.add(RDFS.LABEL, "the bnode")
	.build();

model.setNamespace(RDFS.NS);
model.setNamespace("ex", "http://ex.com/");
StringWriter sw = new StringWriter();
BufferedWriter writer = new BufferedWriter(sw);
WriterConfig config = new WriterConfig();
config.set(BasicWriterSettings.INLINE_BLANK_NODES, true);
Rio.write(model, writer, RDFFormat.TURTLE, config);
System.out.println(sw.toString());
```

The result of this is:

```
@prefix ex: <http://ex.com/> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ex:subject ex:rel1 [
      rdfs:label "the bnode"
    ];
  ex:rel2 [] .
```

While I would expect the writer not to inline the blank node, which in this case is the object of two triples.

```
@prefix ex: <http://ex.com/> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ex:subject 
	ex:rel1 _:bnode ;
	ex:rel2 _:bnode .

_:bnode rdfs:label "the bnode" .

```

Mirr99:
Hi
I would like to work on this issue. Can it be assigned to me?

barthanssens:
Sure, any help is appreciated :-)

[...]

PR #2281, opened 2020-05-27
GH-1079 INLINE_BLANK_NODES-lose-bnode

Mirr99:
Signed-off-by: Mirr99 <mihir17166@iiitd.ac.in>

GitHub issue resolved: #1079

- Added a function call to `queueBlankStatements(SubjectInContext key)` to ensure that the statements related to a repeated BNode are also written.

Please let me know if a test case is required.

### DIFF
--- a/core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriterTest.java
+++ b/core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriterTest.java
@@ -44,12 +44,15 @@ public class ArrangedWriterTest {
 
 	private RDFWriterFactory writerFactory;
 
-	private BNode bnode;
+	private BNode bnode1;
+
+	private BNode bnode2;
 
 	public ArrangedWriterTest() {
 		vf = SimpleValueFactory.getInstance();
 
-		bnode = vf.createBNode("bnode");
+		bnode1 = vf.createBNode("bnode1");
+		bnode2 = vf.createBNode("bnode2");
 
 		exNs = "http://example.org/";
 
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Thanks for looking into this! A regression test would be great, yes. The most logical place, I think, is to add a test case to the `ArrangedWriterTest`.

Mirr99:
I have added a test case `testWriteRepeatedInlineBlankNode()`. Please let me know if this test case will be suitable.

core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriterTest.java
```diff
@@ -70,6 +81,38 @@ public void testWriteSingleStatementWithSubjectNamespace() {
...
+				"@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> ." + sep +
+				"@prefix xsd: <http://www.w3.org/2001/XMLSchema#> ." + sep + sep +
+				"ex:subject ex:rel1 [" + sep +
+				"      rdfs:label \"the bnode\"" + sep +
+				"    ];" + sep +
+				"  ex:rel2 [" + sep +
+				"      rdfs:label \"the bnode\"" + sep +
+				"    ] ." + sep;
+
+		assertEquals(expectedResult, stringWriter.toString());
```
jeenbroekstra, line 113: I like the structure of the test, but I'm not sure I agree that the expected output is correct - see also the comment I left on the issue.

Mirr99: Hi
I have made the changes according to the discussion. There is no inlining for a repeated blank node. The test case has also been updated.

Issue #1105, opened 2018-10-09
SPARQL query with regex escaping of special chars does not work as expected


abrokenjester:
_From @jeenbroekstra on August 30, 2018 3:42_

Example data:

     <foo:bar> <foo:prop> "foo.bar" 

Example query:

    select ?d ?r where {?s <foo:prop> ?d . BIND(REPLACE(?d, "\.", "FOO") as ?r) }

Result:
 
```
+-------------------------------------+-------------------------------------+
| d                                   | r                                   |
+-------------------------------------+-------------------------------------+
| "foo.bar"                           | "FOOFOOFOOFOOFOOFOOFOO"             |
+-------------------------------------+-------------------------------------+  
```

The expected result is `fooFOObar`.

Query 2: 

     select ?d ?r where {?s <foo:prop> ?d . BIND(REPLACE(?d, "\\.", "FOO") as ?r) }

Result: 

    Malformed query: Lexical error at line 1, column 1771.  Encountered: "." (46), after : "\"\\"

Query 3:

    select ?d ?r where {?s <foo:prop> ?d . BIND(REPLACE(?d, "\\\.", "FOO") as ?r) }

Result:

Malformed query: Lexical error at line 1, column 1771.  Encountered: "." (46), after : "\"\\"

Query 4:

    select ?d ?r where {?s <foo:prop> ?d . BIND(REPLACE(?d, "\\\\.", "FOO") as ?r) }

Result:

```
+-------------------------------------+-------------------------------------+
| d                                   | r                                   |
+-------------------------------------+-------------------------------------+
| "foo.bar"                           | "fooFOObar"                         |
+-------------------------------------+-------------------------------------+
```

_Copied from original issue: eclipse/rdf4j-tools#72_

barthanssens:
Could perhaps be solved (not tested) by setting some options on JLine parser in ConsoleIO

jeenbroekstra:
(copied from comment on PR): I had a closer look and I believe this is actually a problem in the SPARQL AST parser. It has explicit code inside it to "handle" escaped characters (specifically the UnicodeEscapeStream class), and I believe that is messing with things.

deepali17043:
I, @29kritika and @anmolsinghal98 would like to look into this problem, could you please guide us with it?

barthanssens:
As Jeen mentioned, it is not (only) the console, e.g. this fails

```
import org.eclipse.rdf4j.model.ValueFactory;
[...]

PR #2289, opened 2020-05-28
GH-1105 regex escaping special chars sparql query

deepali17043:
GitHub issue resolved: #1105 

We have added a code to handle an odd number of backslashes. The functions meant for reading the regex expect a backslash ('\\') before the backslash meant to treat the regex as a normal character. Thus the odd number of backslashes were leading to the "Malformed query" error.
---- 

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderTokenManager.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderTokenManager.java
@@ -3340,14 +3340,22 @@ public class SyntaxTreeBuilderTokenManager implements SyntaxTreeBuilderConstants
 		}
 	}
 
-	public String backSlashCorrection(String image){
-		String newImage = "";
-		if (image.equals("\"\\.\"")){
-			newImage = "\"\\\\.\"";
+	public String backSlashCorrection(String image) {
+		StringBuilder newimg = new StringBuilder();
+		int imglen = image.length(), cnt = 0;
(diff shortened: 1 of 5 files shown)


core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderTokenManager.java
```diff
@@ -3340,6 +3340,25 @@ public void SwitchTo(int lexState) {
...
+				cnt = 0;
+			}
+			newimg.append(image.charAt(i));
+		}
+		if (cnt % 2 == 1) {
+			newimg.append('\\');
+		}
+		String newImage = new String(newimg);
+		return newImage;
+	}
```
jeenbroekstra, line 3360: Unfortunately, this is the wrong place to fix the issue. The SyntaxTreeBuilderTokenManager is generated code, created/updated when we run [JavaCC](https://javacc.github.io/javacc/) on the sparql.jjt file. Next time we regenerate the parser from the jjt file, your changes would be overwritten. 

It's been a while since I last used JavaCC, but I think the way to fix this would be by adding a new method to the SyntaxTreeBuilder class definition you will find at the top of the sparql.jjt file, a method that takes a regular expression token and corrects the escaping, and then make sure in the actual grammar production rule for the regex (lines 1619-1623) use that method to produce the correct token. Have a look at why we use a similar technique to create a datablock for INSERT DATA updates, using the `readToMatchingBrace` method. 

By the way there is an Eclipse plugin for JavaCC that makes working with it a lot easier: https://marketplace.eclipse.org/content/javacc-eclipse-plug

Issue #1033, opened 2018-05-25
Aggregation functions don't use non-strict comparisons when using ExtendedEvaluationStrategy


pulquero:
ExtendedEvaluationStrategy doesn't pass-thru strict=false to QueryEvaluationUtil.compareLiterals() for aggregation functions (GroupIteration).

SakshiSaini17092:
Hey, @pulquero can you please explain the issue in detail? 
An example maybe?

pulquero:
Aggregate functions like min/max use compareLiterals. But the behaviour doesn't respect strict=true/false. So if I have some filter(?x>?y) involving dates and then try to aggregate min(?x) using extended evaluation strategy, it doesn't produce the answer you would expect.

SakshiSaini17092:
Hey,
Pardon me if I understand the problem incorrectly, but the compare literals function takes care of strict behaviour, 

```
public static boolean compareLiterals(Literal leftLit, Literal rightLit, CompareOp operator, boolean strict)
			throws ValueExprEvaluationException {
		// type precendence:
		// - simple literal
		// - numeric
		// - xsd:boolean
		// - xsd:dateTime
		// - xsd:string
		// - RDF term (equal and unequal only)
```
like this..
Also, by aggregate functions not taking care of strict rule in ExtendedEvaluationStrategy, you mean that this function passes the strict as false always ?
```
@Override
	public Value evaluate(Compare node, BindingSet bindings)
			throws ValueExprEvaluationException, QueryEvaluationException {
		Value leftVal = evaluate(node.getLeftArg(), bindings);
		Value rightVal = evaluate(node.getRightArg(), bindings);

		// return result of non-strict comparisson.
		return BooleanLiteral.valueOf(QueryEvaluationUtil.compare(leftVal, rightVal, node.getOperator(), false));
	}
```

SakshiSaini17092:
Or, 
the problem is with the FILTER method that it only accounts for the datatypes like string, integers, doubles?

pulquero:
Aggregate functions always use strict=true, even if extended strategy uses strict=false for expressions. If I am using extended strategy, both aggregates and expressions should use strict=false, not just expressions.

SakshiSaini17092:
Hey,
Please refer to my latest commit, I changed the compare method to 
`QueryEvaluationUtil.compare(v, max, Compare.CompareOp.GE, false)` 

please tell me if it is correct. 
And tell me how can I write a test for that ?

[...]

PR #2293, opened 2020-05-29
GH-1033: Changed the Compare Method of Aggregrate

SakshiSaini17092:
Signed-off-by: SakshiSaini17092 <sakshi17092@iiitd.ac.in>

GitHub issue resolved: #1033 

I changed the Compare method of the Aggregrate Method min/max

---- 

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
@@ -7,6 +7,18 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.iterator;
 
+import java.io.File;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Random;
+import java.util.Set;
+
 import org.eclipse.rdf4j.common.iteration.CloseableIteration;
 import org.eclipse.rdf4j.common.iteration.CloseableIteratorIteration;
 import org.eclipse.rdf4j.common.lang.ObjectUtil;
@@ -34,24 +46,13 @@ import org.eclipse.rdf4j.query.algebra.ValueExpr;
 import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategy;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryBindingSet;
 import org.eclipse.rdf4j.query.algebra.evaluation.ValueExprEvaluationException;
+import org.eclipse.rdf4j.query.algebra.evaluation.impl.StrictEvaluationStrategy;
 import org.eclipse.rdf4j.query.algebra.evaluation.util.MathUtil;
 import org.eclipse.rdf4j.query.algebra.evaluation.util.QueryEvaluationUtil;
 import org.eclipse.rdf4j.query.algebra.evaluation.util.ValueComparator;
 import org.eclipse.rdf4j.query.impl.EmptyBindingSet;
 import org.mapdb.DB;
 import org.mapdb.DBMaker;
-
-import java.io.File;
-import java.io.IOException;
-import java.io.Serializable;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Random;
-import java.util.Set;
 
 /**
  * @author David Huynh
@@ -476,10 +477,15 @@ public class GroupIterator extends CloseableIteratorIteration<BindingSet, QueryE
 		@Override
 		public void processAggregate(BindingSet s) throws QueryEvaluationException {
 			Value v = evaluate(s);
+			boolean strict = false;
+			if (strategy instanceof StrictEvaluationStrategy) {
+				strict = true;
+			}
+			boolean compareResult = QueryEvaluationUtil.compare(v, min, Compare.CompareOp.LE, strict);
 			if (v != null && distinctValue(v)) {
 				if (min == null) {
 					min = v;
-				} else if (QueryEvaluationUtil.compare(v, min, Compare.CompareOp.LE, false)) {
+				} else if (compareResult) {
 					min = v;
 				}
 			}
@@ -506,11 +512,16 @@ public class GroupIterator extends CloseableIteratorIteration<BindingSet, QueryE
 
 		@Override
 		public void processAggregate(BindingSet s) throws QueryEvaluationException {
+			boolean strict = false;
+			if (strategy instanceof StrictEvaluationStrategy) {
+				strict = true;
+			}
 			Value v = evaluate(s);
+			boolean compareResult = QueryEvaluationUtil.compare(v, max, Compare.CompareOp.GE, strict);
 			if (v != null && distinctValue(v)) {
 				if (max == null) {
 					max = v;
-				} else if (QueryEvaluationUtil.compare(v, max, Compare.CompareOp.GE, false)) {
+				} else if (compareResult) {
 					max = v;
 				}
 			}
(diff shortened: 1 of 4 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
```diff
@@ -474,10 +477,14 @@ public MinAggregate(Min operator) {
 		@Override
 		public void processAggregate(BindingSet s) throws QueryEvaluationException {
 			Value v = evaluate(s);
+			boolean strict = false;
+			if (strategy instanceof StrictEvaluationStrategy) {
+				strict = true;
+			}
```
jeenbroekstra, line 483: Unfortunately we also have subclasses of `StrictEvaluationStrategy` that are not 'non strict' - like the FederatedEvaluationStrategy. 

I think a simple way to fix is to turn this around: set `strict` to `true` by default, and only if the strategy is an instance of `ExtendedEvaluationStrategy`, set it to false.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
```diff
@@ -504,11 +511,15 @@ public MaxAggregate(Max operator) {
 
 		@Override
 		public void processAggregate(BindingSet s) throws QueryEvaluationException {
+			boolean strict = false;
+			if (strategy instanceof StrictEvaluationStrategy) {
+				strict = true;
```
jeenbroekstra, line 516: Same here.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
jeenbroekstra, line 483: This won't quite work. Pretty much _every_ EvaluationStrategy implementation in RDF4J is a subclass of `StrictEvaluationStrategy`. Even `ExtendedEvaluationStrategy`. 

I think you need to do this the other way around to make it work:

                        boolean strict = true;
			if (strategy instanceof ExtendedEvaluationStrategy) {
				strict = false;
			}

SakshiSaini17092: Okay I made the check to `ExtendedEvaluationStrategy`.

SakshiSaini17092: But It is still failing on some test cases in build.

jeenbroekstra:
@SakshiSaini17092 are you still available to finish this PR? No rush, but I noticed that after a lot of activity it has suddendly gone a bit quiet. Also if you don't have time to finish this, just let us know, and we will take over.

SakshiSaini17092:
@jeenbroekstra Thanks for the support, All the test cases are passing now? Guide me to write the test cases. Where I can check for the strict = false ?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparator.java
```diff
@@ -87,6 +89,14 @@ public int compare(Value o1, Value o2) {
 		return compareTriples((Triple) o1, (Triple) o2);
 	}
 
+	public void setStrict(boolean flag) {
+		this.strict = flag;
+	}
+
+	public boolean getStrict() {
```
jeenbroekstra, line 96: This is nitpicking a bit, but the convention is to name methods that return a boolean value `is...` rather than `get...`. So `isStrict()`.

jeenbroekstra:
> @jeenbroekstra Thanks for the support, All the test cases are passing now. Guide me to write the test cases. Where I can check for the strict = false ?

Great stuff! I think if you just add one or two tests to the already-existing `ValueComparatorTest` class, you're fine.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparator.java
SakshiSaini17092: this is done.

SakshiSaini17092:
@jeenbroekstra I have a doubt in the test case, in the thread of the issue, it is mentioned that the date type comparison is breaking because of strict=true restriction. But even if I set the strict= true the test case if passing. Am I missing anything?

jeenbroekstra:
> @jeenbroekstra I have a doubt in the test case, in the thread of the issue, it is mentioned that the date type comparison is breaking because of strict=true restriction. But even if I set the strict= true the test case if passing. Am I missing anything?

Ah. I hadn't thought of that. There's two reasons you're seeing this. 

Firstly, in your test, you are comparing two literals which are both of type xsd:Date. Even in strict comparison, the engine is always able to compare these, so we would expect this comparison to work in both strict and non-strict mode. 

Non-strict comparison is where it becomes possible to compare two literals that are _not_ exactly the same datatype, but which are both (different) calendar datatypes (for example one is an xsd:Date, the other is a xsd:gYear). 

The second reason is a bit more complicated, and it's this: in the QueryEvaluationUtil, when comparing in strict mode, a ValueExprEvaluationException will be thrown when you are comparing, for example, and xsd:Date and an xsd:gYear literal. However, the ValueComparator catches this exception and falls back to using a comparison of the actual datatype strings to force an ordering. 

This means that to detect that our code is _actually_ using the non-strict date comparison and not just looking at the datatype string,  we need to compare two literals where the first has a _value_ that is larger than the second one, but has a _datatype_ that is (lexicographically) smaller. For example, this:

		Literal date1 = vf.createLiteral("2018-09-02", XMLSchema.DATE);
		Literal date2 = vf.createLiteral("2016", XMLSchema.GYEAR);
		assertTrue(cmp.compare(date1, date2) > 0);

This test will fail in strict mode: the QueryEvaluationUtil will throw an error because the datatypes are not compatible in strict mode, and the ValueComparator will fall back to comparing the IRIs of the xsd:Date and xsd:gYear datatypes. Because the 'g' in 'gYear' comes after the 'D' from 'Date' in the alphabet, it will return that date1 is _smaller_. 

In non-strict mode, however, this test will succeed, because the QueryEvaluationUtil will not throw an exception but just give a comparison result based on the actual date values being before or after each other.

SakshiSaini17092:
@jeenbroekstra the ValueComparator was catching the exception raised by QueryEvaluationUtil. In order to test the functionallity correctly I commented out that try and catch. I hope it is correct?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparator.java
```diff
@@ -99,23 +109,23 @@ private int compareLiterals(Literal leftLit, Literal rightLit) {
...
 				}
-			} catch (ValueExprEvaluationException e) {
-				// literals cannot be compared using the '<' operator, continue
-				// below
+				return 1;
 			}
+//			} catch (ValueExprEvaluationException e) {
+//				// literals cannot be compared using the '<' operator, continue
+//				// below
+//			}
```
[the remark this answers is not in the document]
SakshiSaini17092: ohh Okay.. thanks for the info.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/util/ValueComparatorTest.java
```diff
@@ -152,4 +155,29 @@ public void testOrder3() throws Exception {
...
 		assertTrue(valueList.indexOf(int1000) < valueList.indexOf(float2000));
 	}
+
+	@Test
+	public void testNonStrictComparisons() throws Exception {
+		cmp.setStrict(false);
+		assertTrue(cmp.isStrict() == false);
+		Literal date1 = vf.createLiteral("2019-09-02", XMLSchema.DATE);
+		Literal date2 = vf.createLiteral("2020", XMLSchema.GYEAR);
+		assertTrue(cmp.compare(date1, date2) < 0);
```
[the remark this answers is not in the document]
SakshiSaini17092: okay, now I understood the problem, better. Thanks again. Updated the test cases in my latest commit.

Issue #1368, opened 2019-03-28
RDF/XML inline blank nodes, writing without rdf:nodeID ?


barthanssens:
See https://groups.google.com/forum/#!topic/rdf4j-users/RR1d8JvhChU, reported by Manuel Fiorelli

> I need to produce a compact RDF/XML document representing an OWL ontology containing anonymous class descriptions. In particular, I need to inline blank nodes as follows:
> 
> <owl:Class rdf:about="http://example.org/Italian">
>     <owl:equivalentClass>
>         <owl:Restriction rdf:nodeID="node1d6qktouex1">
>             <owl:hasValue rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Italian</owl:hasValue>
>             <owl:onProperty>
>                 <owl:DatatypeProperty rdf:about="http://example.org/nationality" />
>             </owl:onProperty>
>         </owl:Restriction>
>     </owl:equivalentClass>
> </owl:Class>
> 
> I was able to produce the fragment above using a combination of:
> org.eclipse.rdf4j.rio.rdfxml.util.RDFXMLPrettyWriterFactory
> org.eclipse.rdf4j.rio.turtle.ArrangedWriter
> I used the ArrangedWriter to overcome the limitation of RDFXMLPrettyWriterFactory that requires  the statements in the "right order".
> 
> As you can see (highlighted in yellow), the output include the identifier of the blank node associated with the restriction.
> 
> It is uncommon to see such identifiers in a compact RDF/XML document: is there any option to remove them? Or is there a simpler mechanism to achieve my goal

barthanssens:
Rio Turtle writer is wrapped in an ArrangedWriter to inline blank nodes (= duplicate them if needed, because they can be used/referenced in multiple places)

This could be moved to a higher / more abstract level + some changes in RdfxmlWriter to check whether the RioSetting INLINE_BLANK_NODES is active (and if that is indeed the case, to not write out the blank node id in RDF/XML)

barthanssens:
See also #1079 , looks like the ArrangedWriter might need some bug fixing first.

REVERB283:
Hi!
We (me, @iamaryananand & @Cool007Satyam) were working on the issue and facing some difficulties modelling the problem and want to confirm few things.

This is the test case (as mentioned in the issue #1079 ) we used to understand the working of library better.
`@prefix ex: <http://ex.com/> .
[...]

PR #2294, opened 2020-05-29
GH-1368 Added a RioSetting Compact_Xml.

iamaryananand:
GitHub issue resolved: #1368  

Added a <RioSetting> "COMPACT_XML" in BasicWriterSettings to present the XML output as proposed, without rdf:nodeID. 

We are checking If COMPACT_XML setting is set as **true** in RDFXMLPrettyWriter, the output is formatted accordingly. 

---- 

### DIFF
--- a/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLWriterSettings.java
+++ b/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLWriterSettings.java
@@ -58,17 +58,6 @@ public class XMLWriterSettings {
 			"org.eclipse.rdf4j.rio.use_single_quotes", "Use single quotes", Boolean.FALSE);
 
 	/**
-	 * Boolean setting for writer to determine whether it should remove the rdf:nodeID attribute from blank nodes.
-	 * <p>
-	 * Defaults to False.
-	 * <p>
-	 * Can be overridden by setting system property {@code org.eclipse.rdf4j.rio.compact_xml}.
-	 */
-	public static final RioSetting<Boolean> COMPACT_XML = new BooleanRioSetting(
-			"org.eclipse.rdf4j.rio.compact_xml",
-			"To generate a compact XML i.e. without Blank Node ID.", Boolean.FALSE);
-
-	/**
 	 * Boolean setting for RDF/XML Writer to determine if the character used to quote attribute values, (single quote or
 	 * double quote) is also replaced within text nodes by it's corresponding entity.
 	 *
--- a/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLWriterSettings.java
+++ b/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLWriterSettings.java
@@ -58,6 +58,17 @@ public class XMLWriterSettings {
 			"org.eclipse.rdf4j.rio.use_single_quotes", "Use single quotes", Boolean.FALSE);
 
 	/**
+	 * Boolean setting for writer to determine whether it should remove the rdf:nodeID attribute from blank nodes.
+	 * <p>
+	 * Defaults to False.
+	 * <p>
+	 * Can be overridden by setting system property {@code org.eclipse.rdf4j.rio.compact_xml}.
+	 */
+	public static final RioSetting<Boolean> COMPACT_XML = new BooleanRioSetting(
+			"org.eclipse.rdf4j.rio.compact_xml",
+			"To generate a compact XML i.e. without Blank Node ID.", Boolean.FALSE);
+
+	/**
 	 * Boolean setting for RDF/XML Writer to determine if the character used to quote attribute values, (single quote or
 	 * double quote) is also replaced within text nodes by it's corresponding entity.
 	 *
--- a/core/rio/rdfxml/pom.xml
+++ b/core/rio/rdfxml/pom.xml
@@ -61,11 +61,5 @@
 			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
-		<dependency>
-			<groupId>org.eclipse.rdf4j</groupId>
-			<artifactId>rdf4j-rio-turtle</artifactId>
-			<version>3.2.2-SNAPSHOT</version>
-			<scope>compile</scope>
-		</dependency>
 	</dependencies>
 </project>
--- a/core/rio/rdfxml/pom.xml
+++ b/core/rio/rdfxml/pom.xml
@@ -61,5 +61,11 @@
 			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.rdf4j</groupId>
+			<artifactId>rdf4j-rio-turtle</artifactId>
+			<version>3.2.2-SNAPSHOT</version>
+			<scope>compile</scope>
+		</dependency>
 	</dependencies>
 </project>
--- a/core/rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLPrettyWriterTest.java
+++ b/core/rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLPrettyWriterTest.java
@@ -8,25 +8,30 @@
 package org.eclipse.rdf4j.rio.rdfxml;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
-import java.io.IOException;
-import java.io.StringReader;
-import java.io.StringWriter;
+import java.io.*;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.regex.Pattern;
 
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
-import org.eclipse.rdf4j.model.Resource;
-import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.common.io.FileUtil;
+import org.eclipse.rdf4j.model.*;
+import org.eclipse.rdf4j.model.impl.LinkedHashModel;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.OWL;
 import org.eclipse.rdf4j.model.vocabulary.RDF;
-import org.eclipse.rdf4j.rio.RDFHandlerException;
-import org.eclipse.rdf4j.rio.RDFWriter;
-import org.eclipse.rdf4j.rio.RDFWriterTest;
-import org.eclipse.rdf4j.rio.WriterConfig;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.rio.*;
 import org.eclipse.rdf4j.rio.helpers.BasicWriterSettings;
+import org.eclipse.rdf4j.rio.helpers.XMLWriterSettings;
+import org.eclipse.rdf4j.rio.rdfxml.util.RDFXMLPrettyWriter;
 import org.eclipse.rdf4j.rio.rdfxml.util.RDFXMLPrettyWriterFactory;
 import org.junit.Test;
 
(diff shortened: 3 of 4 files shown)


core/rio/rdfxml/pom.xml
```diff
@@ -61,5 +61,11 @@
 			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>org.eclipse.rdf4j</groupId>
+            <artifactId>rdf4j-rio-turtle</artifactId>
```
barthanssens, line 66: Could you explain a little bit why the rio-turtle dependency is needed here ? I would assume the code also compiles without it...

iamaryananand: Yes, it is compiling without it, but this was added automatically, we did not change anything in the rio-turtle files. We only added our code in the rio-util and helper files (RDFXMLPrettyWriter and BasicWriterSettings) only.

iamaryananand:
> Thanks for contributing.
> 
> Could you also please format the code with mvn process-resources (see also the details in https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#code-formatting)
> The build server is quite strict about it, so even if your code is technically correct, it will refuse to build (that's why the checks above failed)

The commit is taking time because git is doing code analysis after reformatting.

core/rio/rdfxml/pom.xml
jeenbroekstra: Added automatically - that's really strange. Can I ask what editor you are using?

jeenbroekstra:
Unfortunately, another problem is that the commit has not been signed off, as described in the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md). 

Since your branch only has one commit, you could try to fix this by doing `git commit -s --amend` on your local branch, and then using `git push --force-with-lease` to force-push the (now correctly signed off) commit.

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BasicWriterSettings.java
```diff
@@ -113,6 +113,17 @@
...
+	 * Boolean setting for writer to determine whether it should remove the rdf:nodeID attribute from blank nodes.
+	 * <p>
+	 * Defaults to False.
+	 * <p>
+	 * Can be overridden by setting system property {@code org.eclipse.rdf4j.rio.compact_xml}.
+	 */
+	public static final RioSetting<Boolean> COMPACT_XML = new BooleanRioSetting(
+			"org.eclipse.rdf4j.rio.compact_xml",
+			"To generate a compact XML i.e. without Blank Node ID.", Boolean.FALSE);
+
```
jeenbroekstra, line 126: I like this, but perhaps a better place for this is `XMLWriterSettings`, since this configuration option is only relevant for XML.

core/rio/rdfxml/pom.xml
iamaryananand: IntelliJ IDEA Community Edition 2020.1.1 x64

REVERB283:
So we've reformatted the code according to the guidelines and signed it off as well. Still the Last check from this Eclipse Foundation is failing until now.

barthanssens:
Hi @REVERB283 

assuming you have signed the Eclipse CA, have you used the same email address in the git commit signoff ? If they don't match, the last test will fail.

iamaryananand:
The commit was done by @iamaryananand, and his Eclipse ID was made after
the commit because we were not aware about this Contributor Agreement.

barthanssens:
OK, thank you for the information

@iamaryananand have you used the same email address in the git commit signoff as in your Eclipse Contributor Agreement ? Because it looks like git is using an Anonymous / `noreply` address.

You can change this easily, either in the IDE itself or using the command line: https://stackoverflow.com/questions/30677775/how-do-i-set-user-email-and-user-name-in-intellij-14-with-git-integration

Then you probably have to do `git commit -s --amend`  and `git push --force-with-lease`  again, like @jeenbroekstra mentioned in https://github.com/eclipse/rdf4j/pull/2294#issuecomment-636248790

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BasicWriterSettings.java
REVERB283: sure.

iamaryananand:
@barthanssens @jeenbroekstra 
We can't seem to remove the Failure in our eclipsefdn/eca test, even after following all the necessary steps you sent us. 
Is there any way we can pass that test, or do we need to generate another pull request.
barthas

barthanssens:
No worries

When all the changes have been made (e.g. moving the COMPACT_XML from BasicWriterSettings to eclipse.rdf4j.rio.helpers.XMLWriterSettings) and you're happy with the code, all the commits can be  "squashed" into 1 single commit (see https://rdf4j.org/documentation/developer/squashing/)

And you can then signoff that squashed commit with the "correct" email address.

core/rio/rdfxml/pom.xml
barthanssens: Strange, could you perhaps manually remove this, and see what happens ?

iamaryananand: This dependency had to be added back because of the `Class<?> arrangedWriterCls = Class.forName("org.eclipse.rdf4j.rio.turtle.ArrangedWriter");` line, without adding the dependency the code was not running.

REVERB283:
So we've finally managed to successfully commit the changed made and signoff with proper author credentials. Kindly check for any discrepancy if it's ready for merge.

core/rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLPrettyWriterTest.java
```diff
@@ -104,6 +109,58 @@ public void outOfSequenceItemsAreNotAbbreviated() throws RDFHandlerException, IO
...
+		BNode bnode = vf.createBNode("bnode");
+
+		model.add(subject, relation, bnode);
+		model.add(bnode, RDFS.LABEL, vf.createLiteral("the bnode"));
+		model.setNamespace(RDFS.NS);
+		model.setNamespace("ex", "http://example.org/");
+
+		List<Statement> shuffledStatements = new ArrayList<>(model);
+		Collections.shuffle(shuffledStatements);
+		Rio.write(shuffledStatements, arrangedXmlWriter);
```
jeenbroekstra, line 145: Like I said above: I don't really think you need this shuffle on the order of statements here: your test should only be about the improvement to the RDFXMLPrettyWriter that you did. Making sure statements are always in the correct order is not part of that class' job. So by doing this shuffle, all you're really testing is that ArrangedWriter is doing its job - and that should not be part of _this_ unit test. 

The only thing you should be testing here is that _if_ statements are fed in in the correct order, and the COMPACT_XML setting is turned on, the output is as expected. So I'd suggest that instead of creating a Model and then shuffling that Model, you just create two Statement using the ValueFactory, then add those two statements to a simple ArrayList in the correct order, and then pass that Arraylist to the RDFXMLPrettyWriter.

core/rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLPrettyWriterTest.java
```diff
@@ -104,6 +109,58 @@ public void outOfSequenceItemsAreNotAbbreviated() throws RDFHandlerException, IO
 		assertEquals(Arrays.asList("<rdf:RDF", "<rdf:Bag", "<rdf:_0", "<rdf:_2"), rdfLines);
 	}
 
+	@Test
+	public void compactXMLPrintTest() throws RDFHandlerException, IOException, ClassNotFoundException,
+			NoSuchMethodException, IllegalAccessException, InvocationTargetException, InstantiationException {
+		OutputStream outputStream = new ByteArrayOutputStream();
+		PrintStream printStream = new PrintStream(outputStream);
+		System.setOut(printStream);
```
jeenbroekstra, line 117: Rather than redirecting System.out, why not just give the outputStream argument directly to the RDFXMlPrettyWriter?

core/rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLPrettyWriterTest.java
```diff
@@ -104,6 +109,58 @@ public void outOfSequenceItemsAreNotAbbreviated() throws RDFHandlerException, IO
...
+		RDFWriter rdfXmlWriter = new RDFXMLPrettyWriter(System.out);
+
+		Class<?> arrangedWriterCls = Class.forName("org.eclipse.rdf4j.rio.turtle.ArrangedWriter");
+		Constructor<?> constructor = arrangedWriterCls.getConstructor(RDFWriter.class);
+		constructor.setAccessible(true);
+		RDFWriter arrangedXmlWriter = (RDFWriter) constructor.newInstance(rdfXmlWriter);
+
+		arrangedXmlWriter.set(BasicWriterSettings.PRETTY_PRINT, false);
+		arrangedXmlWriter.set(BasicWriterSettings.INLINE_BLANK_NODES, true);
+		arrangedXmlWriter.set(XMLWriterSettings.COMPACT_XML, true);
```
jeenbroekstra, line 128: If you get rid of using the ArrangedWriter, you can also remove the first two settings, and just set the third one (but directly on the rdfXmlWriter, of course).

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLWriterSettings.java
```diff
@@ -57,6 +57,17 @@
...
+	 * Boolean setting for writer to determine whether it should remove the rdf:nodeID attribute from blank nodes.
+	 * <p>
+	 * Defaults to False.
+	 * <p>
+	 * Can be overridden by setting system property {@code org.eclipse.rdf4j.rio.compact_xml}.
+	 */
+	public static final RioSetting<Boolean> COMPACT_XML = new BooleanRioSetting(
+			"org.eclipse.rdf4j.rio.compact_xml",
+			"To generate a compact XML i.e. without Blank Node ID.", Boolean.FALSE);
+
```
jeenbroekstra, line 70: I just had another look at this, and at the issue, and I realized that this new setting is not really necessary, as there already exists a setting for this: `BasicWriterSettings.INLINE_BLANK_NODES`. 

Why not use that setting to check in RDFXMLPrettyWriter, and remove this setting?

REVERB283: Well I didn't quite understand the what this setting INLINE_BLANK_NODES is all about. While understanding the issue I went through the entire link that @barthanssens shared and that contained some test cases that the author was trying to model this problem and I took help from that. But If you're saying that this can be achieved using INLINE_BLANK_NODES settings only then I think I can do that.

Moreover I had a doubt regarding the checks: Is it necessary that the checks will succeed only if @iamaryananand (author of the pull request) will do the commit?

jeenbroekstra: > Well I didn't quite understand the what this setting INLINE_BLANK_NODES is all about. While understanding the issue I went through the entire link that @barthanssens shared and that contained some test cases that the author was trying to model this problem and I took help from that. But If you're saying that this can be achieved using INLINE_BLANK_NODES settings only then I think I can do that.

Not a problem, it took me a while to remember this as well. But yes, I believe we can reuse the existing setting.

> Moreover I had a doubt regarding the checks: Is it necessary that the checks will succeed only if @iamaryananand (author of the pull request) will do the commit?

If you mean the sign off check: it doesn't matter who created the pull request, it looks at the commits inside the pull request. Each individual commit that you add needs to be signed off by the author of that commit, and that author needs to be on record as having the signed the Eclipse Contributor Agreement. So if you add a commit, _you_ need to sign it off, if @iamaryananand adds a commit, they need to sign off that commit. 

If you are interested in learning why we need this: our project is part of the Eclipse Foundation, which requires a formal record of each contribution, who has created it, and that that person has given us license to use their contribution. As an open-source contributor you still own your own code, so we need written evidence that you given us permission to reuse it. In other words, we need to have an audit trail for [Intellectual Property (IP)](https://www.eclipse.org/projects/handbook/#ip).

jeenbroekstra:
@iamaryananand  @REVERB283 are you still available to finish this PR? No rush, but I noticed that after a lot of activity it has suddendly gone a bit quiet. Also if you don't have time to finish this, just let us know, and we can take over.

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/XMLWriterSettings.java
REVERB283: Okay! So we managed to delete the CompactXML setting as you requested and are using the existing INLINE_BLANK_NODES for our purpose.

core/rio/rdfxml/src/test/java/org/eclipse/rdf4j/rio/rdfxml/RDFXMLPrettyWriterTest.java
```diff
@@ -104,6 +109,58 @@ public void outOfSequenceItemsAreNotAbbreviated() throws RDFHandlerException, IO
...
+		OutputStream outputStream = new ByteArrayOutputStream();
+		PrintStream printStream = new PrintStream(outputStream);
+		System.setOut(printStream);
+
+		RDFWriter rdfXmlWriter = new RDFXMLPrettyWriter(System.out);
+
+		Class<?> arrangedWriterCls = Class.forName("org.eclipse.rdf4j.rio.turtle.ArrangedWriter");
+		Constructor<?> constructor = arrangedWriterCls.getConstructor(RDFWriter.class);
+		constructor.setAccessible(true);
+		RDFWriter arrangedXmlWriter = (RDFWriter) constructor.newInstance(rdfXmlWriter);
```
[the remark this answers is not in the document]
REVERB283: We've cleared some clutter (reflections) and are using just the normal instances but I don't understand why do you want us to remove the arrangedwriter part. I mean to check if the setting really works I believe we have to compare the output with the expected one. If you can clear this a bit more, we'll be really grateful.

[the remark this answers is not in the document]
jeenbroekstra: > We've cleared some clutter (reflections) and are using just the normal instances but I don't understand why do you want us to remove the arrangedwriter part. I mean to check if the setting really works I believe we have to compare the output with the expected one. If you can clear this a bit more, we'll be really grateful.

My point is that you don't need the ArrangedWriter to test that. The RDFXMLPrettyWriter is expected to work if the input statements are sent in the correct order. All you really need is a  List with statements in the correct order, feed that list into the RDFXMLPrettyWriter, and then check if the output is as you expect.

REVERB283:
Hi!
We've made the changes as you guys requested and would like to ask for another review.

Issue #2304, opened 2020-06-04
Major impact of using ValueExprEvaluationException for flow control


avataar:
One of the design decisions in Sesame and later RDF4J is using exceptions for flow control in some cases. Probably the most notable example is ValueExprEvaluationException thrown when a value is incompatible in a certain expression. Using exceptions for flow control is OK per se but throwing exception in general is very expensive because the stack trace needs to be collected. I have observed some simple queries (for example using LANG(?x) where ?x is bound to an IRI or ORDER BY ?x where there is no binding for ?x) throwing ~44 _million_ ValueExprEvaluationException instances. On the test machine the query completed in about 4,5 minutes. There is a way to disable stack trace collection for a given exception so I tried that with ValueExprEvaluationException and the same query completed in about a minute.

I propose a simple fix (PR coming) that disables stack trace collection for ValueExprEvaluationException. This will allow us to keep using the exception for flow control as it is now but not pay the stack trace collection penalty.

jeenbroekstra:
Good find, thanks.

If we could do things over I would have chosen to introduce a "type error" return type for all value operators, instead of using exceptions - but it's probably not worth the effort and pain to make that change now.

So disabling stack trace collection seems a decent compromise.

PR #2305, opened 2020-06-04
GH-2304: Disables stack trace collection for ValueExprEvaluationException

avataar:
GitHub issue resolved: #2304

Disable stack trace collection for ValueExprEvaluationException.
---- 


hmottestad:
In other places we have used a static instance of the exception. Would that be possible in this case? Would that also make this faster?

avataar:
> In other places we have used a static instance of the exception. Would that be possible in this case? Would that also make this faster?

That would make it faster too as stack trace will be collected only once but there are too many places where a new ValueExprEvaluationException would have to be replaced with the static instance and it won't work for third-party code (e.g. custom functions that throw ValueExprEvaluationException when they get an invalid argument).

hmottestad:
Ok. LGTM.

Issue #2306, opened 2020-06-05
FedXConnection fails to inject external binding if __originalQueryMaxExecutionTime is missing


jeenbroekstra:
In FedXConnection, the logic to determine if external bindings needs to supplied to query evaluator is based on counting the number of supplied bindings, and checking if it is larger than the (expected) number of "special" FedX bindings (such as `__originalQueryString` and `__originalQueryMaxExecutionTime`). See FedXConnection line 142. 

However, this logic is brittle against some special bindings missing. For example neither `__baseUri` nor `__originalQueryMaxExecutionTime` are always supplied. In these cases, the connection can fail to recognize that there is an external binding it needs to inject.

A fix is do a proper collection comparison to check if any supplied binding names do not occur in the set of predefined fedx bindings. 

A side consequence of this bug is that in merging / converting bindingsets, line 511-513:

    for (String bindingName : sourceBinding.getBindingNames()) {
          mergedSet.addBinding(sourceBinding.getBinding(bindingName));
    }

a NullPointerException can occur if, for a particular binding name, `sourceBinding` has no value.

jeenbroekstra:
Possible fix is to change line 512 to 
      
     mergedSet.addBinding(bindingName, sourceBinding.getValue(bindingName));

PR #2307, opened 2020-06-07
GH-2306 more robust check for external bindings in FedXConnection

jeenbroekstra:
GitHub issue resolved: #2306  

- in FedXConnection, when checking if supplied bindings contain anything other than the "special" FedX bindings, do not assume we know the exact number of supplied FedX bindings, as it can vary.
- in SailUpdateExecutor, when merging result bindings with preset bindings, ensure a binding is not null before merging.

---- 

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
@@ -138,10 +138,8 @@ public class FedXConnection extends AbstractSailConnection {
 
 		try {
 			// make sure to apply any external bindings
-			// Note: we use a conditional check for performance reasons.
-			// Further note that we subtract -1 from the size as the baseUri is optional
 			BindingSet queryBindings = EmptyBindingSet.getInstance();
-			if (bindings.size() > FedXRepositoryConnection.FEDX_BINDINGS.size() - 1) {
+			if (!FedXRepositoryConnection.FEDX_BINDINGS.containsAll(bindings.getBindingNames())) {
 				MapBindingSet actualQueryBindings = new MapBindingSet();
 				bindings.forEach(binding -> {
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
We get a nullpointerexception "in the wild" when using this in a strange corner case. I have yet to succeed in constructing a simple test case, but regardless: since getBinding is allowed to return `null` and addBinding is expected to throw an error if null is supplied, this check should be there.

jeenbroekstra:
The actual root cause of the issue was a bug in FedXConnection, which explains why it was  not reproducible on other Sails: it failed to inject an external binding. Fix added.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
```diff
@@ -138,10 +138,8 @@ public FedXConnection(FedX federation, FederationContext federationContext)
 
 		try {
 			// make sure to apply any external bindings
-			// Note: we use a conditional check for performance reasons.
-			// Further note that we subtract -1 from the size as the baseUri is optional
 			BindingSet queryBindings = EmptyBindingSet.getInstance();
-			if (bindings.size() > FedXRepositoryConnection.FEDX_BINDINGS.size() - 1) {
+			if (!FedXRepositoryConnection.FEDX_BINDINGS.containsAll(bindings.getBindingNames())) {
```
jeenbroekstra, line 142: Fwiw I believe this is still a check worth doing as in most cases it will skip creating a new MapBindingSet, and instead just pass in the `EmptyBindingSet` singleton - which other parts of the engine have some optimizations for.

aschwarte10: Good solution and way more robust. Should do the job as expected

PR #2330, opened 2020-06-23
Releases/3.3.0 m1

hmottestad:
Move to 3.3.0-M1-SNAPSHOT

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.rdf4j</groupId>
 		<artifactId>rdf4j</artifactId>
-		<version>3.3.0-M1</version>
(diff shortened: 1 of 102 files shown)


assembly/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.rdf4j</groupId>
 		<artifactId>rdf4j</artifactId>
-		<version>3.3.0-SNAPSHOT</version>
+		<version>3.3.0-M1-SNAPSHOT</version>
```
jeenbroekstra, line 7: I wouldn't use milestone markers in the SNAPSHOT versions in the pom if I were you. Stick with just 3.3.0-SNAPSHOT. Only use "3.3.0-M1" for the actual tagged version.

Issue #2322, opened 2020-06-19
Support for sail config on Repository level


hmottestad:
Users who talk to the ShaclSail through a `HttpRepository` and `HttpRepositoryConnection` should be able to:
 - enable or disable validation in the ShaclSail
 - call revalidate() in the ShaclSailConnection (split out into new issue #2325)

hmottestad:
@jeenbroekstra this is a use case that Ontotext has come with. Could you shed some light on how remote configuration is currently handled in RDF4J.

My natural instinct is to create a new class for `HttpShaclSail` and `HttpShaclSailConnection` that could be retrieved from the `HttpRepository` and `HttpRepositoryConnection` respectively. These would then expose the relevant methods to the user.

jeenbroekstra:
How do people who work locally currently do this?

jeenbroekstra:
What I'm getting at with that is that we should think about a generic mechanism to control such settings at the Repository API level, so that (either locally or remotely), people can control this without having to directly interact with the SAIL. 

Normally, remote configuration is handled by sending an RDF-serialized RepositoryConfig (with a SailImplConfig etc) through the RemoteRepositoryManager, but that mechanism is not really suitable for these kinds of "per-connection" or even "per-operation" config settings.

hmottestad:
I personally get the sail from the repo and cast it to a ShaclSail. Same for the connection. When working locally that is.

The RemoteRepositoryManager would work for disabling and enabling validation, since this is a global setting.

For interacting with the connection I really like the casting approach:

```java
HttpRepositoryConnection connection ... ;
HttpShaclEnabledRepositoryConnection shaclConnection = connection.as(HttpShaclEnabledRepositoryConnection.class);

// revalidation
ValidationReport = shaclConnection.revalidate();

// custom transaction support
shaclConnection.begin(Validation.disabled);
....
shaclConnection.commit();
```

One thing I'm wondering about. Does the HttpRepositoryConnection support transactions fully. Essentially, do we have a POST /startTransaction endpoint?

jeenbroekstra:
On my phone so brief, but I'm not a fan of solving this through inheritance. 

I'm thinking instead we add a new ConnectionSettings object that you can get/set on a RepositoryConnection.

jeenbroekstra:
> One thing I'm wondering about. Does the HttpRepositoryConnection support transactions fully. Essentially, do we have a POST /startTransaction endpoint?

[...]

PR #2340, opened 2020-06-28
GH-2322 repository settings for changing sail configuration

hmottestad:
GitHub issue resolved: # 

---- 

### DIFF
--- a/assembly/src/main/dist/docker/Dockerfile.amd64
+++ b/assembly/src/main/dist/docker/Dockerfile.amd64
@@ -1,30 +1,32 @@
 FROM amd64/tomcat:8.5-jre8-alpine
 MAINTAINER Bart Hanssens (bart.hanssens@bosa.fgov.be)
 
-ARG VERSION="2.5.1"
-
 ENV JAVA_OPTS="-Xmx2g"
 ENV CATALINA_OPTS="-Dorg.eclipse.rdf4j.appdata.basedir=/var/rdf4j"
 
-RUN adduser -S tomcat 
+RUN adduser -S tomcat
 
-RUN cd /tmp && \
-	wget -q "https://www.eclipse.org/downloads/download.php?file=/rdf4j/eclipse-rdf4j-${VERSION}-sdk.zip&r=1" -O /tmp/rdf4j.zip && \
-	unzip -q /tmp/rdf4j.zip && \
-	rm -rf /usr/local/tomcat/webapps/* && \
-	cp /tmp/eclipse-rdf4j-${VERSION}/war/*.war /usr/local/tomcat/webapps && \
-	rm -rf /tmp/eclipse && \
-	rm /tmp/rdf4j.zip && \
-	mkdir -p /var/rdf4j && \
-	chown -R tomcat /var/rdf4j /usr/local/tomcat && \
-	chmod a+x /usr/local/tomcat /usr/local/tomcat/bin /usr/local/tomcat/bin/catalina.sh
+COPY ignore/rdf4j.zip /tmp/rdf4j.zip
+
+WORKDIR /tmp
+
+RUN	unzip -q /tmp/rdf4j.zip
+RUN	rm -rf /usr/local/tomcat/webapps/*
+RUN	cp /tmp/eclipse-rdf4j*/war/*.war /usr/local/tomcat/webapps
+RUN	rm -rf /tmp/eclipse
+RUN	rm /tmp/rdf4j.zip
+RUN	mkdir -p /var/rdf4j
+RUN	chown -R tomcat /var/rdf4j /usr/local/tomcat
+RUN	chmod a+x /usr/local/tomcat /usr/local/tomcat/bin /usr/local/tomcat/bin/catalina.sh
 
 COPY web.xml /usr/local/tomcat/conf/web.xml
 
 USER tomcat
 
-VOLUME /var/rdf4j
-VOLUME /usr/local/tomcat/logs
+WORKDIR /usr/local/tomcat/
+
+# never got this syntax to work with docker-compose
+#VOLUME /var/rdf4j
+#VOLUME /usr/local/tomcat/logs
 
 EXPOSE 8080
-
--- a/core/repository/api/src/main/java/org/eclipse/rdf4j/repository/Repository.java
+++ b/core/repository/api/src/main/java/org/eclipse/rdf4j/repository/Repository.java
@@ -10,6 +10,7 @@ package org.eclipse.rdf4j.repository;
 import java.io.File;
 
 import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.repository.config.RepositorySettings;
 
 /**
  * An RDF4J repository that contains RDF data that can be queried and updated. Access to the repository can be acquired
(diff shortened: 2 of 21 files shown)


core/repository/api/src/main/java/org/eclipse/rdf4j/repository/config/RepositorySettings.java
```diff
@@ -0,0 +1,33 @@
...
+		this.settings = new HashMap<>(settings);
+	}
+
+	public RepositorySettings set(IRI iri, Literal value) {
+
+		RepositorySettings repositorySettings = new RepositorySettings(settings);
+		repositorySettings.settings.put(iri, value);
+
+		return repositorySettings;
+	}
```
hmottestad, line 23: immutable

core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPRepository.java
```diff
@@ -374,4 +376,16 @@ boolean useCompatibleMode() throws RepositoryException {
...
+	public RepositorySettings getSettings() {
+		try (RDF4JProtocolSession client = createHTTPClient()) {
+		}
+		return new RepositorySettings(new HashMap<>());
+	}
+
+	@Override
+	public void setSettings(RepositorySettings settings) {
+		super.setSettings(settings);
+	}
```
hmottestad, line 390: This is where I would retrieve the RepositorySettings from the http client.

core/repository/http/src/test/java/org/eclipse/rdf4j/repository/http/HttpRepositoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+					.iterator()
+					.next();
+
+			config.remove(null, vf.createIRI("http://rdf4j.org/config/sail/shacl#validationEnabled"), null);
+			config.add((Resource) next, vf.createIRI("http://rdf4j.org/config/sail/shacl#validationEnabled"),
+					BooleanLiteral.FALSE);
+
+			httpClient.updateRepository(new RepositoryConfig());
+
+		}
```
hmottestad, line 51: There is support for both retrieving and updating repository config, except you can't directly retrieve a RepositoryConfig object. I actually found it fairly hard to deal with, a lot of boiler plate code. And even though the returned statement "stream" contains the sail config, there doesn't seem to be any way of writing it back.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
```diff
@@ -366,6 +373,25 @@ public ValueFactory getValueFactory() {
...
+	public void setSettings(Map<IRI, Literal> settings) {
+		settings.forEach((key, value) -> {
+
+			if (key.equals(MemoryStoreSchema.SYNC_DELAY)) {
+				setSyncDelay(value.longValue());
+			}
+
+		});
+
+	}
```
hmottestad, line 393: Here is part of my solution proposal. The repository layer uses an immutable RepositorySettings object, but the sail layer just uses a Map<IRI, Literal>. The map is created by the lowest level sail, and each wrapping sail can add to the map. We reuse the vocabulary from the sail config.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -863,4 +884,79 @@ public void setDashDataShapes(boolean dashDataShapes) {
...
+			}
+			if (key.equals(SHACL_ADVANCED_FEATURES)) {
+				setShaclAdvancedFeatures(value.booleanValue());
+			}
+			if (key.equals(DASH_DATA_SHAPES)) {
+				setDashDataShapes(value.booleanValue());
+			}
+
+		});
+	}
```
hmottestad, line 961: Here is the ShailSail implementation aspect of the new settings approach. Looking at this I'm partly wondering if the old fashioned model based sail config could just delegate to these settings methods.

Also an important note. For settings that are unchangeable once the sail has been initialised we would either ignore them or throw an exception.

hmottestad:
And a final note. The discoverability on this is atrocious. A user that isn't already comfortable with the sail config approach will really struggle to use the new settings methods. We should guarantee that `getSettings()` returns all possible settings (eg. false for all disabled settings), but it's still a lot harder than just `.disableValidation()`.

hmottestad:
I would like your feedback on this approach @jeenbroekstra before I continue with the HTTP aspect (which I'm really dreading, wish we were on spring boot tbh).

assembly/src/main/dist/docker/Dockerfile.amd64
```diff
@@ -1,30 +1,32 @@
 FROM amd64/tomcat:8.5-jre8-alpine
 MAINTAINER Bart Hanssens (bart.hanssens@bosa.fgov.be)
 
-ARG VERSION="2.5.1"
-
 ENV JAVA_OPTS="-Xmx2g"
 ENV CATALINA_OPTS="-Dorg.eclipse.rdf4j.appdata.basedir=/var/rdf4j"
 
-RUN adduser -S tomcat 
+RUN adduser -S tomcat
```
jeenbroekstra, line 7: I'm not sure why this PR touches the Dockerfile. Accidental commit?

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/Repository.java
```diff
@@ -99,12 +100,21 @@ default void init() throws RepositoryException {
...
-	public ValueFactory getValueFactory();
+	ValueFactory getValueFactory();
+
+	default RepositorySettings getSettings() {
+		throw new UnsupportedOperationException();
+	}
+
+	default void setSettings(RepositorySettings settings) {
+		throw new UnsupportedOperationException();
+	}
```
jeenbroekstra, line 118: I don't understand why you're adding this to the Repository interface. Didn't we want a configuration option for a Repository_Connection_ ? We already have an existing configuration model for repositories and sail stacks - the use case was inspired by needing to tweak options on the fly, per connection (or even per request).

hmottestad:
I realized that we had talked past each other. All I want at this point is some way to configure the underlying sails from the repository (and the HttpRepository). If this is already possible then I don’t want to reinvent the wheel. 

The per-connection aspect was for extending the current methods. Eg. with ‘.begin(Validation.Disabled)’ and the revalidate method (or just validate).

jeenbroekstra:
> I realized that we had talked past each other. All I want at this point is some way to configure the underlying sails from the repository (and the HttpRepository). If this is already possible then I don’t want to reinvent the wheel.

It is, though admittedly it is not very easy to use purely from code. If we want to improve to make it easier to manipulate I'd be all for it, but ideally we'd do that as an improvement of the existing basic mechanism. 

> The per-connection aspect was for extending the current methods. Eg. with ‘.begin(Validation.Disabled)’ and the revalidate method (or just validate).

I'd still be in favor of enabling this kind of thing through a ConnectionSetting object.

If you have time, shall we try and schedule a call later this week to talk through scenarios, and get us all on the same page?

hmottestad:
Sent you and email to coordinate a meeting.

jeenbroekstra:
@hmottestad I believe this PR has now been made obsolete by the work on PR #2351. Or at the very least if something is still necessary here I think we agree that we need a different approach. Can this PR be closed?

hmottestad:
Yes. It's definitely obsolete for the problem it's trying to solve. It would still be great to allow some of the sail config a to be changed at runtime, but it's not something I intend to pursue.

Issue #2350, opened 2020-07-04
Transaction settings


hmottestad:
Allow the user to supply settings to a transaction. 

This is useful for:

 - controlling how the SHACL validation works
    - `con.begin(IsolationLevels.NONE, ShaclSail.Settings.BulkValidation)`
 - controlling indexing into Elasticsearch
    - `con.begin(IsolationLevels.NONE, ElasticearchSail.Settings.AsynIndexing)`
 - possibly for controlling bulk loading
   - `con.begin(NativeStore.Settings.BulkLoading)`

Or even multiple settings at once:
    - `con.begin(IsolationLevels.NONE, ShaclSail.Settings.BulkValidation, NativeStore.Settings.BulkLoading)`

PR #2351, opened 2020-07-04
GH-2350 transaction settings

hmottestad:
GitHub issue resolved: #2350  

 - Support for user supplied settings per transaction.
 - Support for bulk validation ShaclSail setting

---- 


core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -375,61 +397,70 @@ public final void prepare() throws SailException {
...
 				}
 			} finally {
-				updateLock.unlock();
+				connectionLock.readLock().unlock();
 			}
 		} finally {
-			connectionLock.readLock().unlock();
+			transactionSettings.clear();
 		}
 	}
```
hmottestad, line 465: This is just the git diff engine not understanding that I wrapped everything in try-finally to clear the transaction settings.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/SailConnectionWrapper.java
```diff
@@ -206,6 +207,11 @@ public Explanation explain(Explanation.Level level, TupleExpr tupleExpr, Dataset
...
 	}
 
+	@Override
+	public void begin(TransactionSetting... settings) {
+		wrappedCon.begin(settings);
+	}
+
 	@Override
 	public void begin() throws SailException {
 		wrappedCon.begin();
```
hmottestad, line 217: @jeenbroekstra 

The whole wrapped connection approach here makes it hard for me to add support for transaction settings. The problem is that code higher up will often override the `begin()` and `begin(IsolationLevel level)` methods to hook into the transaction handling approach. With the new `begin(TransactionSetting[] settings)` simply calling `wrapped.begin(settings)` the code higher up will have to override the new begin method in order to fully hook into the transaction flow.

I'm not sure right now how to fix this. I was initially hoping to hook in like I did in the AbstractSailConnection.

hmottestad:
PS: I'm currently failing the build due to not begin backwards compatible. I'll fix that later once I get the design down.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/SailConnectionWrapper.java
hmottestad: Maybe I want to try out just extending the current `begin(IsolationLevel level)` method with the transaction settings vararg, and in the next major release remove that in favour of just a simpler `begin(TransactionSetting... settings)`.

hmottestad: I actually tried this out now.

```
void begin(IsolationLevel level, TransactionSetting ... settings)
```

And it's not backwards compatible with any of the code that implements the current `begin` interface methods. I'm getting a bunch of compile time errors saying that various classes have `@Override` annotation for methods that are not inherited.

Also found another issue with introducing a new begin method that only takes a TransactionSetting vararg.

Screenshot 2020-07-04 at 17 54 29

This isn't exactly good code though, since it making the isolation level nullable means that all code actually needs to check for null. Instead of just having is non-null and instead passing in a default isolation level if none was defined.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -459,6 +460,11 @@ public void begin() throws RepositoryException {
 		}
 	}
 
+	@Override
+	public void begin(TransactionSetting... settings) {
+		throw new UnsupportedOperationException("SPARQLConnection does not support transaction settings");
```
jeenbroekstra, line 465: I think silently ignoring (possibly logging a warning) might be better here. Not the most important thing right now though.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/SailConnectionWrapper.java
jeenbroekstra: You can avoid this by explicitlying cast the null argument to a TransactionSetting for this to work I think:

    this.begin((TransactionSetting)null);

That way the vararg itself is never null, it just receives an array with a single element, and that element is null.

jeenbroekstra: You can avoid this by explicitlying cast the null argument to a TransactionSetting for this to work I think:

    this.begin((TransactionSetting)null);

That way the vararg itself is never null, it just receives an array with a single element, and that element is null.

jeenbroekstra: > Maybe I want to try out just extending the current `begin(IsolationLevel level)` method with the transaction settings vararg, and in the next major release remove that in favour of just a simpler `begin(TransactionSetting... settings)`.

That sound like a good idea to me.

hmottestad: I tried adding a vararg, but it's not source level compatible.

hmottestad:
Screenshot 2020-07-05 at 09 30 48

japicmp is complaining that I've added new interface methods to the IsolationLevel interface by having it extend TransactionSetting, even though I've implement defaults for those two methods. I don't think this a correct assessment.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/SailConnectionWrapper.java
jeenbroekstra: > @jeenbroekstra I tried adding a vararg, but it's not source level backwards compatible.

I'm sorry I'm not sure what exact variant we're talking about here.

jeenbroekstra:
> japicmp is complaining that I've added new interface methods to the IsolationLevel interface by having it extend TransactionSetting, even though I've implement defaults for those two methods. I don't think this a correct assessment.

I agree this is a false positive, but I think you can just get around that by adding the default methods to TransactionSettings itself instead:

```
public interface TransactionSetting {

	default String getName() {
		return this.getClass().getCanonicalName();
	}

	default String getValue() {
		return this.toString();
	}
}
```

The only other option is to reconfigure the plugin for this case. We can't generally configure it to consider METHOD_ADDED_TO_INTERFACE as compatible, because clearly it's a (source) compatibility issue _unless_ you also provide a default. I can't see an easy way to configure that generally, looks like a limitation of the plugin. You could just add an exclusion for these two particular methods in the plugin config in the root pom:

    <exclude>org.eclipse.rdf4j.IsolationLevel#getValue()</exclude>
    <exclude>org.eclipse.rdf4j.IsolationLevel#getName()</exclude>

Slight preference for moving the default methods up the hierarchy, myself, but either solution works.

core/model/src/main/java/org/eclipse/rdf4j/TransactionSetting.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.rdf4j;
```
jeenbroekstra, line 1: Can you move this to a new package `org.eclipse.rdf4j.common.transaction`? That's really where `IsolationLevel` should also be, but at least let's not add more inferfaces in a place we know we'll want to refactor later. 

Also means that it's probably better to move it to the rdf4j-util module btw, because that's where all the other `org.eclipse.rdf4j.common` packages are.

core/model/src/main/java/org/eclipse/rdf4j/TransactionSettingImpl.java
```diff
@@ -0,0 +1,10 @@
+package org.eclipse.rdf4j;
+
+public interface TransactionSettingImpl extends TransactionSetting {
```
jeenbroekstra, line 3: I'm not sure what we need this for (especially if we move the default methods to TransactionSetting) but I don't think  an interface should be called `....Impl`. If we need some sort of base implementation class it should be an abstract class, not an interface. It should also be called AbstractTransactionSetting to stay consistent with naming conventions elsewhere in the projbect.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/SailConnectionWrapper.java
hmottestad: begin(IsolationLevel level, TransactionSettings ... settings)

If I add the vararg to the current begin method like above it’s not source level compatible.

core/model/src/main/java/org/eclipse/rdf4j/TransactionSettingImpl.java
hmottestad: Was initially a class, but then I switched to enum in the ShaclSail and had to make it an interface. Might just remove it.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/SailConnectionWrapper.java
jeenbroekstra: Not even if you also keep 

    begin(IsolationLevel level) 

and just redirect it to:

    begin(level, (TransactionSettings)null);

?

hmottestad: Could redirect it in the sail wrapper, but it’ll introduce its own problems.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RDF4JProtocolSession.java
```diff
@@ -204,7 +207,7 @@ private synchronized String getTransactionURL() {
 	 *-----------------*/
 
 	public TupleQueryResult getRepositoryList()
-			throws IOException, RepositoryException, UnauthorizedException, QueryInterruptedException {
+		throws IOException, RepositoryException, UnauthorizedException, QueryInterruptedException {
```
jeenbroekstra, line 210: It looks like your formatter is still applying the single-indent rule, rather than sticking to 2 indents for wrapped lines. Did I make a mistake when I merged the formatting changes, or is there something out of sync on your end?

assembly/src/main/dist/docker/build.sh
```diff
@@ -30,4 +28,4 @@ cp $ZIP ./ignore/rdf4j.zip
 # build
 docker-compose build
 
-docker tag docker_rdf4j:latest eclipse/rdf4j-workbench:${MVN_VERSION}
\ No newline at end of file
+docker tag docker_rdf4j:latest eclipse/rdf4j-workbench:${MVN_VERSION}
```
jeenbroekstra, line 31: Why are there changes to the docker build script in this PR?

jeenbroekstra:
@hmottestad I have some availability today or over the weekend to help out with finalizing this design. Just a heads up that I may be pushing some changes on your branch.

assembly/src/main/dist/docker/build.sh
hmottestad: Because I needed to run up the server, and I use docker for that. Got annoyed with it taking so long, so made it faster while I was at it.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RDF4JProtocolSession.java
hmottestad: I’ll try to sync it up again.

hmottestad:
I invited you just now.

jeenbroekstra:
> I invited you just now.

Thanks! I pushed a formatting fix and it seems to work. 

Bit odd that you had to invite me. I thought forks automatically enable the original maintainers to push. Maybe they changed the behavior.

hmottestad:
Says you’ve got write permission. Could you try to edit a file in the web view? 

I had some trouble initially with the new token based authentication in GitHub when I started the HASMAC organization in GitHub. Had to create a new token with broader permissions.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RDF4JProtocolSession.java
```diff
@@ -606,7 +609,7 @@ public void getStatements(Resource subj, IRI pred, Value obj, boolean includeInf
 		pingTransaction();
 	}
 
-	public synchronized void beginTransaction(IsolationLevel isolationLevel)
+	public synchronized void beginTransaction(TransactionSetting... transactionSettings)
```
jeenbroekstra, line 612: This is backward incompatible in multiple ways.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RDF4JProtocolSession.java
```diff
@@ -620,11 +623,14 @@ public synchronized void beginTransaction(IsolationLevel isolationLevel)
...
+			for (TransactionSetting transactionSetting : transactionSettings) {
+				if (transactionSetting == null)
+					continue;
 				params.add(
 						new BasicNameValuePair(
-								Protocol.ISOLATION_LEVEL_PARAM_NAME,
-								isolationLevel.getURI().stringValue()
+								TRANSACTION_SETTINGS_PREFIX + transactionSetting.getName(),
+								transactionSetting.getValue()
 						)
```
jeenbroekstra, line 634: This is not backward compatible. The REST API _requires_ a transaction isolation level to be sent, but this opens the option of not supplying one. It also removes the actual Isolation level parameter that is part of the protocol, instead encoding everything as a generic transaction setting. 

I'll push some fixes for this.

core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/Protocol.java
```diff
@@ -181,9 +181,9 @@
 	public static final String ACCEPT_PARAM_NAME = "Accept";
 
 	/**
-	 * Parameter name for the isolation level used in transactions.
+	 * Prefix for transaction settings in the query param
 	 */
-	public static final String ISOLATION_LEVEL_PARAM_NAME = "isolation-level";
+	public static final String TRANSACTION_SETTINGS_PREFIX = "transaction-setting__";
```
jeenbroekstra, line 186: Removing a parameter from the public REST API and replacing it with something else is backward incompatible. I think the simplest solution is that at the level of the REST API protocol we keep IsolationLevel and other Transaction Settings distinct.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RDF4JProtocolSession.java
jeenbroekstra: Looks like I misremembered that it's required at the protocol level. We'll still need to keep the specific parameter around though.

core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/Protocol.java
jeenbroekstra: Actually what I ended up doing for this is that the client-side sends the isolation level twice (once using the legacy param, once with the name prefixed param name), and the server-side accepts both as well. This way, the change is backward compatible for both client and server: an older client communicating with a newer server, or vice versa.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/TransactionStartController.java
```diff
@@ -101,15 +115,9 @@ public String getValue() {
 
 					transactionSettings.add(transactionSetting);
 				}
-
 			}
-
 		});
 
-		if (isolationLevel[0] == null) {
-			throw new IllegalStateException("Isolation level must be specified");
```
jeenbroekstra: I took this check back out because it shouldn't be necessary: if no isolation level is supplied the connection should just use its default setting.

core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/Protocol.java
hmottestad: I was hoping our rest api wasn’t part of the semantic versioning contract. Ie. that it was considered an internal api between the server and client.

core/model/src/main/java/org/eclipse/rdf4j/TransactionSetting.java
```diff
@@ -0,0 +1,12 @@
...
+
+public interface TransactionSetting {
+
+	default String getName() {
```
hmottestad, line 5: @jeenbroekstra we really shouldn’t make these methods default in a brand new interface.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/RDF4JProtocolSession.java
hmottestad: I presume this was just me trying to get things working before I started considering backwards incompatability.

core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/Protocol.java
jeenbroekstra: It most definitely is I'm afraid. We have a full canonical spec published for it.

core/model/src/main/java/org/eclipse/rdf4j/TransactionSetting.java
jeenbroekstra: Why not? If this were an abstract class it would be natural. How significantly different from these defaults are you suspecting  implementing classes to be for these methods?

hmottestad: I've been thinking a bit more about this and I'm actually not so sure anymore. In general they are decent defaults, except in the case of IsolationLevel since the implementing class is IsolationLevels and not the interface IsolationLevel which blocks users from adding their own isolation levels.

assembly/src/main/dist/docker/build.sh
jeenbroekstra: @barthanssens can we get your expert docker eye on this? This change is not really related to the issue under development here but if you're happy that it's a decent improvement we can treat it as a "tidy as you go along" thing....

jeenbroekstra:
@hmottestad  given that this needs some more work, have you considered if maybe you want to postpone the 3.3 release?

hmottestad:
I’m not sure which at the moment. I still have a glimmer of hope that I can figure it out.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -926,4 +930,60 @@ public long getValidationResultsLimitTotal() {
...
+				&& Settings.ValidationApproach.Disabled.getValue().equals(value)) {
+			return Optional.of(Settings.ValidationApproach.Disabled);
+		}
+		if (Settings.ValidationApproach.Bulk.getName().equals(name)
+				&& Settings.ValidationApproach.Bulk.getValue().equals(value)) {
+			return Optional.of(Settings.ValidationApproach.Bulk);
+		}
+		return getBaseSail().internTransactionSetting(name, value);
+
+	}
```
hmottestad, line 951: I've introduced this concept here for handling transaction settings that come from http. Essentially we "intern" them just like you can intern a string. This makes it much easier to write the rest of the code since we can use things like `instanceOf` and `==` for enums. 

Also this opens up for supporting multiple isolation levels. I was thinking of adding our current isolation levels to the abstract sail class, so we retrieve them from there if possible. That way another sail can override that class and provide their own isolation levels. What do you think? @jeenbroekstra

hmottestad:
And there is still the issue of wrapping sails that override the transactional methods such as the two original `begin` methods, but don't override the new `begin` method with transaction settings and thus could skip out on handling some transactions correctly. This is easily fixed in our own sail wrappers (like the SchemaCachingRdfsInferencer), but there isn't a way to force other implementers/developers to do the same.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: I don't quite follow what you have in mind for multiple isolation levels. How about we park that for now? Let's get the basic transaction settings working first.

jeenbroekstra: As for the 'interning': it's a workable approach but I had an alternative with a registry for settings and a factory in mind, so that (de)serializing of TransactionSettings can happen in the server application code and is not a responsibility for the Sail itself. I'll try and sketch something.

jeenbroekstra:
> And there is still the issue of wrapping sails that override the transactional methods such as the two original `begin` methods, but don't override the new `begin` method with transaction settings and thus could skip out on handling some transactions correctly. This is easily fixed in our own sail wrappers (like the SchemaCachingRdfsInferencer), but there isn't a way to force other implementers/developers to do the same.

I think if we switch around the way you redirect methods a little we can get some improvements here. I'll try and push some fixes.

jeenbroekstra:
I'm working on a change that replaces the interning approach with a registry-based approach. It's almost complete, I'll push it as a single commit so you can easily see the diff and we can discuss (and if necessary roll back)

core/util/src/main/java/org/eclipse/rdf4j/common/transaction/TransactionSetting.java
```diff
@@ -0,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.common.transaction;
```
jeenbroekstra, line 8: Went ahead and moved the interface and related factory and registry classes to its own package.

core/util/src/main/java/org/eclipse/rdf4j/common/transaction/TransactionSetting.java
```diff
@@ -0,0 +1,29 @@
...
+ * the scope of that transaction only.
+ *
+ * @author Håvard Ottestad
+ * @author Jeen Broekstra
+ */
+public interface TransactionSetting {
+
+	// FIXME we should perhaps use getURI (with an IRI return type) to uniquely identify settings. Same as what we do
+	// for IsolationLevels.
+	default String getName() {
```
jeenbroekstra, line 21: Calling it out here as well: For IsolationLevel, we use getURI as the canonical identifier. If we do the same thing generally for TransactionSetting, it would make integration IsolationLevel with the generic handling of TransactionSetttings easier.

hmottestad:
Was looking into the SPI approach myself, but had never used it before so was a bit apprehensive of trying to figure it out. Thanks for knowing how to do this the right way!

It’s all starting to look fairly good now isn’t it? Tests all seem to be working. 

Only major thing left is figuring out how this is going to work for sails that don’t override the new begin method and extend the sail wrapper. I had another approach in mind actually. We could have the new begin method with transaction settings present in the repository code, but switch to using a setter for the sail code. 

For the SailRepository it would be something like this:
```
sailCon.setTransactionSettings(settings);
sailCon.begin(isolationlevel);
```

This would allow all the existing wrapper sails to work seamlessly.

core/util/src/main/java/org/eclipse/rdf4j/common/transaction/TransactionSetting.java
hmottestad: I think the use of the class/interface canonical name is the best option. Moving the class would not be backwards compatible anyway, and the class canonical name already needs to be unique in the java world. Since this is java code I prefer using a more javaish approach to uniquely identifying a class. It also doesn’t require us to manage namespaces or create IRIs.

jeenbroekstra:
> Only major thing left is figuring out how this is going to work for sails that don’t override the new begin method and extend the sail wrapper. I had another approach in mind actually. We could have the new begin method with transaction settings present in the repository code, but switch to using a setter for the sail code.
> 
> For the SailRepository it would be something like this:
> 
> ```
> sailCon.setTransactionSettings(settings);
> sailCon.begin(isolationlevel);
> ```
> 
> This would allow all the existing wrapper sails to work seamlessly.

Hm, interesting. Yes, that could work.

core/util/src/main/java/org/eclipse/rdf4j/common/transaction/TransactionSetting.java
jeenbroekstra: If even _we_ can't be bothered to manage a namespace.... :) But seriously: fair enough. We should consider making getURI in IsolationLevel deprecated then.

hmottestad:
@jeenbroekstra as long as the tests pass I believe I'm happy with this approach.

hmottestad:
@barthanssens I've made some modifications to the docker build in this branch. Essentially unified everything so it builds using the local code instead of downloading a release, I much prefer this since it means we can test new things locally. Also changed the maven command for building to one that skips a bunch of unnecessary stuff. I hope you are ok with this!

@jeenbroekstra I'm happy with this PR now and would like to merge it.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
hmottestad: I prefer an exception when calling code that doesn't do what the user expects.

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/RepositoryConnection.java
```diff
@@ -622,7 +623,18 @@ public void exportStatements(Resource subj, IRI pred, Value obj, boolean include
 	 * @see #rollback()
 	 * @see #setIsolationLevel(IsolationLevel)
 	 */
-	public void begin(IsolationLevel level) throws RepositoryException;
+	void begin(IsolationLevel level) throws RepositoryException;
+
+	default void begin(TransactionSetting... settings) {
```
ansell, line 628: Javadoc on this would be useful. Specifically the general contract and the default implementation behaviour of selecting the first `IsolationLevel` rather than using the last or another decision algorithm.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
ansell: The default method does backwards compatibility by calling `begin()` by default. This method should also call `begin()` for backwards compatibility unless there is a specific reason for not following the previous transactional implementation which hasn't been deleted or modified here.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/SailConnection.java
```diff
@@ -187,6 +188,19 @@ default boolean hasStatement(Resource subj, IRI pred, Value obj, boolean include
...
+	/**
+	 *
+	 * Override this method to receive the transaction settings. This method is always called (by the SailRepository)
+	 * before any .begin method is called, but may not be called if the sail is being used outside of the
+	 * SailRepository. Always remember to call super. receiveTransactionSettings(settings) if you override this method.
+	 *
+	 * @param settings the transaction settings on which this transaction operates. It may or may not contain the
+	 *                 isolation level.
+	 */
+	default void receiveTransactionSettings(TransactionSetting[] settings) {
```
ansell, line 200: The other APIs use the `...` form. This call seems to be compatible with that so there doesn't seem to be a reason why it should use the `[]` form instead of `...`.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -375,61 +385,70 @@ public final void prepare() throws SailException {
...
+				} finally {
+					updateLock.unlock();
 				}
 			} finally {
-				updateLock.unlock();
+				connectionLock.readLock().unlock();
 			}
 		} finally {
-			connectionLock.readLock().unlock();
+			transactionSettings.clear();
```
ansell, line 414: This seems like it will cause threadsafety issues as the implementation is a non-threadsafe Map and this is outside of the lock. Is there a reason why the transaction settings cannot be cleared from within the lock section?

core/sail/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencerConnection.java
```diff
@@ -767,17 +767,13 @@ public void rollback()
...
 			throws SailException {
-		this.begin(null);
+		this.begin(sail.getDefaultIsolationLevel());
 	}
 
 	@Override
 	public void begin(IsolationLevel level)
 			throws SailException {
 
-		if (level == null) {
```
ansell, line 777: If this check is being removed, then there should be a new null-guard on this parameter, given it is a public API method that previously supported null as a valid parameter value (even if `IsolationLevels.getCompatibleIsolationLevel` is currently null-safe)

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/RepositoryConnection.java
jeenbroekstra: Good point regarding javadoc. Fwiw at this point I would prefer _not_ to commit ourselves to a particular decision algorithm for this, and just say that "behavior is undefined when more than one isolation level is supplied".

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/SailConnection.java
jeenbroekstra: Piggybacking off this comment: couldn't we just call this 'setTransactionSettings' rather than 'receive...'?

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
```diff
@@ -155,7 +165,41 @@ public void begin() throws RepositoryException {
...
 		try {
-			sailConnection.begin(level);
+			// always call receiveTransactionSettings(...) before calling begin();
+			sailConnection.receiveTransactionSettings(new TransactionSetting[0]);
+
+			if (level != null) {
+				sailConnection.begin(level);
+			} else {
+				sailConnection.begin();
+			}
```
jeenbroekstra, line 175: This check is not necessary - you can just pass `null` as an argument to the `begin` method.

assembly/src/main/dist/docker/Dockerfile.amd64
```diff
@@ -1,30 +1,32 @@
...
 USER tomcat
 
-VOLUME /var/rdf4j
-VOLUME /usr/local/tomcat/logs
+WORKDIR /usr/local/tomcat/
 
-EXPOSE 8080
+# never got this syntax to work with docker-compose
+#VOLUME /var/rdf4j
+#VOLUME /usr/local/tomcat/logs
```
jeenbroekstra, line 30: I'm not a docker expert but there are a lot of changes in these files that I am unsure about or that look like they're temporary changes for your own test rig, rather than for our deployment. Can you roll these changes back please? If you think some of these are good general improvements let's create a separate issue and PR for it.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -375,61 +385,70 @@ public final void prepare() throws SailException {
...
```
jeenbroekstra, line 451: Shouldn't this be cleared _before_ releasing the update lock?

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/SailConnection.java
hmottestad: I'm using array notation because this method isn't supposed to be called from user code. I prefer "receive" over "set" for methods that the developer is supposed to override in order to receive something.

hmottestad: But not married to "receive", so can switch to "set" if you want.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
hmottestad: That code is actually from the initial implementation, will remove.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
hmottestad: I really prefer to move away from passing null to the begin method. We could consider deprecating and removing the `begin()` method and only keep the `begin(level)` method for the sail code.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
hmottestad: removed

assembly/src/main/dist/docker/Dockerfile.amd64
hmottestad: All of this just fixes up so that the ARM and normal docker images all build the same way (eg. using the code in your branch instead of downloading a release). We were already doing this for the default docker image, just copied it from there so everything was aligned. I've moved it out and made a separate PR.

core/sail/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencerConnection.java
hmottestad: fixed

assembly/src/main/dist/docker/Dockerfile.amd64
jeenbroekstra: Thanks :)

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
hmottestad: that's actually a good point

hmottestad: fixed

assembly/src/main/dist/docker/build.sh
hmottestad: We can move this discussion here: https://github.com/eclipse/rdf4j/pull/2375

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
hmottestad: Also, from seeing that there were no test failures when I removed the null check from the SchemaCachingRDFSInferencerConnection I'm starting to suspect that we don't have any tests for this null handling behaviour.

Here is from the docs:

```
	/**
	 * Begins a transaction with the specified {@link IsolationLevel} level, requiring {@link #commit()} or
	 * {@link #rollback()} to be called to close the transaction.
	 *
	 * @param level the transaction isolation level on which this transaction operates.
	 * @throws UnknownSailTransactionStateException If the IsolationLevel is not supported by this implementation
	 * @throws SailException                        If the connection could not start a transaction, if the supplied
	 *                                              transaction isolation level is not supported, or if a transaction is
	 *                                              already active on this connection.
	 */
	void begin(IsolationLevel level) throws UnknownSailTransactionStateException, SailException;

```

From looking at that documentation from the SailConnection interface there is no mention of how null should be handled and it would make just as much sense to throw a SailException if null were provided.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/SailConnection.java
jeenbroekstra: Let's go with 'set', and let's reword the javadoc slightly to go with that, something like this:

	/**
	 * Pass any transaction-specific settings to the Sail stack. This method needs to be called before the transaction is {@link #begin() started }.
	 * <p>
	 * Sail implementations can override this method to receive the transaction settings (to inspect and/or pass them 
	 * along to any wrapped sail objects). Remember to call <code>super.superTransactionSettings(settings)</code> 
	 * if you override this method.
	 *
	 * @param settings the transaction settings on which the next transaction operates. It may or may not contain the
	 *                 isolation level. 
	 * @since 3.3.0
	 */

I don't think we should say that only SailRepository is allowed to call it. That is no more or less true for this method than for anything else in the Sail API.

jeenbroekstra: As for the array vs vararg notation, I have a slight preference for a vararg for consistency, but can live with either. Your call.

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/RepositoryConnection.java
```diff
@@ -622,7 +623,39 @@ public void exportStatements(Resource subj, IRI pred, Value obj, boolean include
...
+	 *                             <li>the specified {@link IsolationLevel} is not supported by the store, and no
+	 *                             compatible level could be found.
+	 *                             </ul>
+	 * @see #begin()
+	 * @see #isActive()
+	 * @see #commit()
+	 * @see #rollback()
+	 * @see #setIsolationLevel(IsolationLevel)
+	 */
+	default void begin(TransactionSetting... settings) {
```
hmottestad, line 649: Added java docs now.

hmottestad: @jeenbroekstra what do you think?

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
hmottestad: Mind you, the RepositoryConnection code specifically states that it allows null as an isolation level.

jeenbroekstra: > Also, from seeing that there were no test failures when I removed the null check from the SchemaCachingRDFSInferencerConnection I'm starting to suspect that we don't have any tests for this null handling behaviour.
> 
> Here is from the docs:
> 
> ```
> 	/**
> 	 * Begins a transaction with the specified {@link IsolationLevel} level, requiring {@link #commit()} or
> 	 * {@link #rollback()} to be called to close the transaction.
> 	 *
> 	 * @param level the transaction isolation level on which this transaction operates.
> 	 * @throws UnknownSailTransactionStateException If the IsolationLevel is not supported by this implementation
> 	 * @throws SailException                        If the connection could not start a transaction, if the supplied
> 	 *                                              transaction isolation level is not supported, or if a transaction is
> 	 *                                              already active on this connection.
> 	 */
> 	void begin(IsolationLevel level) throws UnknownSailTransactionStateException, SailException;
> ```
> 
> From looking at that documentation from the SailConnection interface there is no mention of how null should be handled and it would make just as much sense to throw a SailException if null were provided.

Spec-wise perhaps, but it wouldn't in practice: most (base) SailConnection implementations inherit from `AbstractSailConnection`, which accepts `null` and falls back to the default isolation level. But you're right that it's not officially documented. That's an oversight I think. 

We can discuss if that needs to change some other time but I would suggest for the purpose of this PR to stick to how most Sails currently behave. That doesn't mean you need to take out this if-else construction by the way: I only commented because I considered it unnecessary, but it doesn't harm to have it there either.

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/RepositoryConnection.java
jeenbroekstra: Looks good to me! One last niggle: add a `@since 3.3.0` if you have time.

ansell: LGTM, and I agree that `@since` tags are very useful on new public API methods.

core/repository/sail/src/main/java/org/eclipse/rdf4j/repository/sail/SailRepositoryConnection.java
hmottestad: Btw. Created a new issue: https://github.com/eclipse/rdf4j/issues/2376

hmottestad:
@jeenbroekstra @ansell thanks for taking time to review this and give great feedback. I hope I have fixed everything now. Let me know if you are happy or if you have any other recommendations.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/BulkValidationSettingsTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+				connection.commit();
+
+			}
+
+		}
+
+	}
+
+}
```
hmottestad, line 199: Representative tests

Issue #2366, opened 2020-07-17
Allow custom ExecutorService to be injected in FedX for parallel query execution


jeenbroekstra:
FedX has built-in functionality for parallel query execution. To facilitate this, the FedX FederationManager creates an ExecutorService thread pool internally. However, the thread pool configuration can currently not be influenced from outside, which makes it hard to tune behavior for some use cases (for example, if security tokens that are tied to the current thread need to be passed on to allow each parallel query to execute succesfully).

We should allow injection of a custom executor service (similar to how we already allow this for the handling of background parsing in the http clients). 

Possible approach: add a `withExecutorService` method to `FedXConfig`, allowing the user to optionally pass an `ExecutorService` of their own creation/configuration. This can then be passed on to the FederationManager to use instead of its own internal thread pool object.

jeenbroekstra:
As an aside the `SharedHttpClientSessionManager` (used in SPARQLRepository and HTTPRepository to have a shared thread pool and HttpClient between various repositories) also creates its own ExecutorService and optionally allows a custom executor service to be injected. 

Instead of doing custom injection in FedX directly, should we consider reusing that same SessionManager mechanism in FedX? Although it's not necessarily used for querying over HTTP, it's still valid to say that a single shared thread pool for RDF4J as a whole makes sense.

jetztgradnet:
I guess it makes sense to allow both use cases, i.e. inject an external one and/or use the shared one (and falling back to current behavior of creating one for itself).
IIRC, this is also relevant wrt cleaning up (i.e. shutting down) the executor, so there needs to be some additional tracking there, similar to how it's done for the HttpClient.

jeenbroekstra:
> I guess it makes sense to allow both use cases, i.e. inject an external one and/or use the shared one (and falling back to current behavior of creating one for itself).
> IIRC, this is also relevant wrt cleaning up (i.e. shutting down) the executor, so there needs to be some additional tracking there, similar to how it's done for the HttpClient.

Just to be clear if we start relying on the shared one we _automatically_ also have a way to inject an external one, because the SharedHttpClientSessionManager allows this, and we can even, I think, have separate threadpools via this mechanism, that is not a pool that is shared between fedx and the httpclients.

I'll try and take a closer look at how this was supposed to work again (it's not something we have great documentation for, although there is some javadoc).

jeenbroekstra:
[...]

PR #2370, opened 2020-07-18
GH-2366 fedx custom executor service

jeenbroekstra:
GitHub issue resolved: #2366 

**UPDATED**  by @aschwarte10 

This change provides a configurable TaskWrapper infrastructure for
wrapping background tasks.

A TaskWrapper is a facility to wrap Runnable background tasks before
they are passed to the Executor. Use-cases include injection of
thread-local context variables, or more fine-granular error handling.

All sub-queries sent by the federation engine that make use of the
concurrency infrastructure (e.g. ControlledWorkerScheduler) are passing
this wrapper.

The concrete implementation can be configured using
FedXConfig.withTaskWrapper(TaskWrapper).

---- 


aschwarte10:
@jeenbroekstra as discussed I have modified the original proposed change on the branch.

When I had a closer look, it was not a viable option to use a factory for the executor service (as there are quite some specifics in the executor). Instead we now make sure to support wrapping all background tasks

Let me know if this is OK to solve the issue

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FederationManager.java
```diff
@@ -103,28 +104,43 @@ public void reset() {
...
 			leftJoinScheduler.abort();
 		}
 		leftJoinScheduler = new ControlledWorkerScheduler<>(federationContext.getConfig().getLeftJoinWorkerThreads(),
-				"Left Join Scheduler");
+				"Left Join Scheduler", taskWrapper);
 
 	}
 
+	/**
+	 * Returns the managed {@link Executor} which takes for properly handling any configured
```
jetztgradnet, line 129: `...which takes care of properly handling...`

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ControlledWorkerScheduler.java
```diff
@@ -46,23 +46,41 @@
...
+	}
+
+	/**
+	 * Construct a new instance with the specified number of workers, the given name and {@link TaskWrapper}.
+	 *
+	 * @param nWorkers
+	 * @param name
+	 * @param taskWrapper
+	 */
+	public ControlledWorkerScheduler(int nWorkers, String name, TaskWrapper taskWrapper) {
```
jetztgradnet, line 80: as a TaskWrapper might be a construct also used in other places and there are already quite some constructor parameters and constructor variants I suggest to rather create an interface TaskWrapperAware which allows setting  (and getting) a TaskWrapper and implement that here. This also allows usiing this in other places.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/TaskWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+ * @see ControlledWorkerScheduler
+ * @see FedXConfig#withTaskWrapper(TaskWrapper)
+ */
+public interface TaskWrapper {
+
+	/**
+	 * Default implementation of {@link TaskWrapper} which returns the unmodified original task
+	 */
+	static final TaskWrapper DEFAULT = new TaskWrapper() {
+	};
```
jeenbroekstra, line 39: Instead of adding a default no-op impl here (which automatically gets inherited by every implementation of TaskWrapper, imho a bit of a code smell) I suggest we instead allow TaskWrapper to be undefined (null or empty) and check for that in relevant places. Also removes the need for default implementations of methods in this interface (which makes discovering what methods to implement easier as well).

aschwarte10: Done, comment is addressed. Let me know if this works for you

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FederationManager.java
```diff
@@ -103,28 +105,46 @@ public void reset() {
 			log.debug("Scheduler for join and union are reset.");
 		}
 
+		Optional<TaskWrapper> taskWrapper = federationContext.getConfig().getTaskWrapper();
 		if (joinScheduler != null) {
 			joinScheduler.abort();
 		}
 		joinScheduler = new ControlledWorkerScheduler<>(federationContext.getConfig().getJoinWorkerThreads(),
 				"Join Scheduler");
+		joinScheduler.setTaskWrapper(taskWrapper);
```
jeenbroekstra, line 114: You're passing the Optional, rather than the TaskWrapper - will that work?

jeenbroekstra: Ah never mind I see what you did - added an overloaded setter. 

I personally prefer handling the Optional at the point of message passing rather than inside the setter:

    federationContext.getConfig().getTaskWrapper().ifPresent(joinScheduler::setTaskWrapper);

but YMMV.

aschwarte10: That looks better, will change and remove the default implementation for optional. Thanks

Issue #2374, opened 2020-07-21
Docker build should act like the maven build and use the local code instead of downloading a different version of RDF4J


hmottestad:
The docker build currently has three different dockerfiles that use two different approaches for building. One uses the local code and another uses a downloaded version of rdf4j. We should always use the local code. If we want to build an older version we should checkout git branch instead.

jeenbroekstra:
This issue is still in the inbox/triage column, but it has an associated pull request (closed, unmerged, unsure why). What's the status?

hmottestad:
Need to fix one thing based on feedback from Bart.

barthanssens:
See also #1439

jeenbroekstra:
@hmottestad has this now been addressed?

barthanssens:
If I understand the files correctly, docker will now indeed use the local version.

Perhaps useful to include the experimental docker option "--squash" to squash layers: if not, the COPY to copy the rdf4j.zip (which is only use to obtain the wars) from the local build directory into the docker image will make the docker image at lot larger, even if the file is removed from docker image afterwards

removing the file in docker and squashing should remove the temporary zip from the docker file system

PR #2375, opened 2020-07-21
GH-2374 fix docker build

hmottestad:
GitHub issue resolved: #2374 

---- 


assembly/src/main/dist/docker/Dockerfile.amd64
```diff
@@ -1,30 +1,32 @@
...
-	wget -q "https://www.eclipse.org/downloads/download.php?file=/rdf4j/eclipse-rdf4j-${VERSION}-sdk.zip&r=1" -O /tmp/rdf4j.zip && \
-	unzip -q /tmp/rdf4j.zip && \
-	rm -rf /usr/local/tomcat/webapps/* && \
-	cp /tmp/eclipse-rdf4j-${VERSION}/war/*.war /usr/local/tomcat/webapps && \
-	rm -rf /tmp/eclipse && \
-	rm /tmp/rdf4j.zip && \
-	mkdir -p /var/rdf4j && \
-	chown -R tomcat /var/rdf4j /usr/local/tomcat && \
-	chmod a+x /usr/local/tomcat /usr/local/tomcat/bin /usr/local/tomcat/bin/catalina.sh
+RUN	unzip -q /tmp/rdf4j.zip
```
barthanssens, line 13: IIRC, each RUN will add a layer, which will increase the size of the docker image (hence the '&&' hack, especially for not increasing the image size with unzipped files etc that will be eventually removed).

Perhaps the "new" docker multistage build feature https://docs.docker.com/develop/develop-images/multistage-build/ would be a better approach ?

hmottestad: Multistage isn't worth it unless we are doing the actual maven building inside the docker image. Single layer for this makes sense though.

assembly/src/main/dist/docker/build.sh
```diff
@@ -14,10 +14,10 @@ rm -rf ../../../../target/
 
 # go to root of project and do clean, format, install and assembly
 cd ../../../../..
-MVN_VERSION=$(mvn -q -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec)
+MVN_VERSION=$(xmllint --xpath "//*[local-name()='project']/*[local-name()='version']/text()" pom.xml)
```
jeenbroekstra, line 17: Like I also mentioned on the mailinglist: I may end up pulling this change out into a separate PR because it is blocking me from using this script.

assembly/src/main/dist/docker/build.sh
```diff
@@ -14,10 +14,10 @@ rm -rf ../../../../target/
...
+
 mvn clean
-mvn formatter:format
-mvn -Passembly install -DskipTests
+mvn -T 2 -Passembly install -DskipTests -Dmaven.javadoc.skip=true -Dmaven.test.skip=true -Dformatter.skip=true -Dimpsort.skip=true -Dxml-format.skip=true  -Djapicmp.skip -Denforcer.skip=true -Dbuildnumber.plugin.phase=none -Danimal.sniffer.skip=true
```
jeenbroekstra, line 20: I'm not entirely sure why but this command failed for me on the first run, reporting that it couldn't find the rdf4j-workbench.war (as part of the assembly process). On a second run (after removing the `-T 2`) it succeeded, but I'm not sure if that is because I removed the parallel execution flag or because I just ran it a second time.

assembly/src/main/dist/docker/build.sh
```diff
@@ -30,4 +30,4 @@ cp $ZIP ./ignore/rdf4j.zip
 # build
 docker-compose build
 
-docker tag docker_rdf4j:latest eclipse/rdf4j-workbench:${MVN_VERSION}
\ No newline at end of file
+docker tag docker_rdf4j:latest eclipse/rdf4j-workbench:${MVN_VERSION}
```
jeenbroekstra, line 33: I'm a little lost on what this script produces. Clearly, the end result of this will be a locally tagged docker image using just the version, so, e.g., `eclipse/rdf4j-workbench:3.4.4`. So where and how do we produce the different images for the ARM and AMD architectures, and how are they pushed to dockerhub?

barthanssens: FWIW, currently I push the dockers manually to dockerhub. 
There *might* (haven't checked) be some github action that could build/push (with some secret key file) for different architectures.

jeenbroekstra: @barthanssens can you talk me through how you currently produce the docker build, step by step? Perhaps on the dev list, rather than here.

jeenbroekstra:
We should also start looking into making this part of the maven build process, so that it can be produced alongside the other distribution assemblies (the SDK and onejar). I haven't looked into a maven docker build plugin yet, but if there is one we should consider using that, instead of tinkering about with shell scripts.

jeenbroekstra:
@hmottestad @barthanssens what's the status of this PR?

hmottestad:
Need to address Barts comment. And if your comment about that maven command still holds I guess that needs to be addressed too.

jeenbroekstra:
> Need to address Barts comment. 

Fair enough.

> And if your comment about that maven command still holds I guess that needs to be addressed too.

I'd keep that out of scope. I'd rather we merge this so we at least have an easy manual way to create docker image. We can address integration in the build process in a followup.

hmottestad:
You said the maven build command was failing for you on the first run through.

jeenbroekstra:
> You said the maven build command was failing for you on the first run through.

Ah right, forgot about that. If nobody can reproduce that problem though, I'm happy to call it a glitch on my end for now.

Issue #2377, opened 2020-07-22
LuceneSail#reindex() fails to add document for last resource


aschwarte10:
While working with the LuceneSail in our application, we found an issue in the re-indexing routine: with the current logic the last document is not persisted in the index.

Unit test for reproducing this (part of AbstractLuceneSailTest)

```
@Test
	public void testReindexing_SingleResource() throws Exception {

		// wipe the Lucene index to allow specific test data
		try (RepositoryConnection connection = repository.getConnection()) {
			connection.clear();
		}

		String query = "PREFIX search: <http://www.openrdf.org/contrib/lucenesail#> \n" +
				"SELECT ?subj ?text ?prop ?score WHERE { \n" +
				"  ?subj search:matches [ search:query \"one\" ; search:property ?prop; search:snippet ?text ; search:score ?score] }";

		List<BindingSet> res;

		// expected empty result => no data in the index
		res = Repositories.tupleQuery(repository, query, t -> Iterations.asList(t));
		Assert.assertEquals(Collections.emptyList(), res);

		try (RepositoryConnection connection = repository.getConnection()) {
			connection.add(SUBJECT_1, PREDICATE_1, vf.createLiteral("one"));
		}

		// expected single result
		res = Repositories.tupleQuery(repository, query, t -> Iterations.asList(t));
		Assert.assertEquals(1, res.size());

		// re-index
		this.sail.reindex();

		// expected single result
		res = Repositories.tupleQuery(repository, query, t -> Iterations.asList(t));
		Assert.assertEquals(1, res.size());
	}
```

PR #2378, opened 2020-07-22
GH-2377: fix LuceneSail#reindex() to create document for last resource

aschwarte10:
GitHub issue resolved: #2377  

With the current algorithm the document for the last resource was never
persisted. This is now change to explicitly persist the last document
after loop completion.

Behavior is covered with a unit test

---- 

 - [ x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [x ] I've added tests for the changes I made
 - [ x] I've applied [code formatting](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)
 - [x ] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change
 - [x ] every commit has been [signed off](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)

### DIFF
--- a/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
+++ b/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
@@ -551,6 +551,15 @@ public class LuceneSail extends NotifyingSailWrapper {
 						}
 						statements.add(vf.createStatement(r, p, o, c));
 					}
+
+					// make sure to index statements for last resource
+					if (current != null) {
+						if (logger.isDebugEnabled()) {
+							logger.debug("reindexing resource " + current);
+						}
+						// commit
+						luceneIndex.addDocuments(current, statements);
+					}
 				}
 			} finally {
 				repo.shutDown();
--- a/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
+++ b/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
@@ -553,7 +553,7 @@ public class LuceneSail extends NotifyingSailWrapper {
 					}
 
 					// make sure to index statements for last resource
-					if (current != null) {
+					if (current != null && !statements.isEmpty()) {
 						if (logger.isDebugEnabled()) {
 							logger.debug("reindexing resource " + current);
 						}
--- a/testsuites/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/AbstractLuceneSailTest.java
+++ b/testsuites/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/AbstractLuceneSailTest.java
@@ -21,7 +21,6 @@ import static org.junit.Assert.fail;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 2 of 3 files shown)


aschwarte10:
@hmottestad , @jeenbroekstra I know this is pretty much a last minute change (which we unfortunately discovered just now) - actually a pretty nasty bug. 

However, could you make sure to included this in the 3.3.0 release? Note: after talking to  @jeenbroekstra I explicitly added this change against the master branch, to potentially have it avaialble in 3.2.4 (if we decide to release another patch release).

core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
```diff
@@ -551,6 +551,15 @@ public void shutDown() {
...
 						statements.add(vf.createStatement(r, p, o, c));
 					}
+
+					// make sure to index statements for last resource
+					if (current != null) {
+						if (logger.isDebugEnabled()) {
+							logger.debug("reindexing resource " + current);
+						}
+						// commit
+						luceneIndex.addDocuments(current, statements);
```
hmottestad, line 561: Just a a general note, the whole code above should actually be using a peeking iterator. It's iterating over all the statements in the database and grouping by subject. It's doing it in a streaming fashion, which makes me assume that the iterator it has against the database is sorted by subject (which would take a massive amount of memory!). And it needs to check the next statement to see if it's got a different subject so it can add all the statements it's gathered for that subject to the lucene index as a document. Is that correct? 

I'm not proposing that we need to actually rewrite the code.

Could you check that !statements.empty() as well, so we don't add the last document twice?

testsuites/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/AbstractLuceneSailTest.java
```diff
@@ -121,7 +121,7 @@ public void setUp() throws Exception {
 
 		// create a Repository wrapping the LuceneSail
 		repository = new SailRepository(sail);
-		repository.initialize();
+		repository.init();
```
hmottestad, line 124: ps, explicit initialization is now optional

jeenbroekstra:
I'll take a closer look at this now, do any updates we need, and then get it merged in time for release.

core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
jeenbroekstra: > Just a a general note, the whole code above should actually be using a peeking iterator. It's iterating over all the statements in the database and grouping by subject. It's doing it in a streaming fashion, which makes me assume that the iterator it has against the database is sorted by subject (which would take a massive amount of memory!). And it needs to check the next statement to see if it's got a different subject so it can add all the statements it's gathered for that subject to the lucene index as a document. Is that correct?

Yeah I think that sums it up. This code is pretty old, and I agree that there are likely far more efficient ways to trigger a reindexing. 

> I'm not proposing that we need to actually rewrite the code.

Worth logging an issue for? The problem is that none of us have LuceneSail maintenance/improvements high on our priority list.

> Could you check that !statements.empty() as well, so we don't add the last document twice?

Good point. I'll add a check for that.

Issue #2187, opened 2020-05-06
Parent link in QueryModelNode is set wrong in some cases


hmottestad:
The parent link in QueryModelNode is set wrong in some cases when we reuse a node without cloning it first.

hmottestad:
See https://github.com/eclipse/rdf4j/pull/2180 for example of failures.

jeenbroekstra:
Can you point to somewhere where we reuse a node without cloning it first, or perhaps a particular example of a query where optimizing/evaluating results in the parent nodes being out of date? I've briefly gone over how some of our optimizers replace and move nodes, and from what I can tell they're fairly conscientious about the book keeping.

PR #2403, opened 2020-07-29
GH-2187 keep parent reference in querymodelnode consistent

jeenbroekstra:
GitHub issue resolved: #2187 

- ensure optimizers and other components that manipulate the algebra tree keep the parent node references consistent
- added "cleanup" at end of optimizer pipeline to catch any wrongdoers
- added regression tests for all optimizer implementations

---- 

### DIFF
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
@@ -8,6 +8,7 @@
 package org.eclipse.rdf4j.query.algebra.helpers;
 
 import java.util.ArrayDeque;
+import java.util.IdentityHashMap;
 
 import org.eclipse.rdf4j.common.annotation.Experimental;
 import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
@@ -26,6 +27,8 @@ public class QueryModelTreeToGenericPlanNode extends AbstractQueryModelVisitor<R
 	GenericPlanNode top = null;
 	QueryModelNode topTupleExpr;
 	ArrayDeque<GenericPlanNode> deque = new ArrayDeque<>();
+	ArrayDeque<QueryModelNode> deque2 = new ArrayDeque<>();
+	IdentityHashMap<QueryModelNode, QueryModelNode> visited = new IdentityHashMap<>();
 
 	public QueryModelTreeToGenericPlanNode(QueryModelNode topTupleExpr) {
 		this.topTupleExpr = topTupleExpr;
@@ -39,6 +42,9 @@ public class QueryModelTreeToGenericPlanNode extends AbstractQueryModelVisitor<R
 	// queue to maintain the effective parent stack.
 	@Override
 	protected void meetNode(QueryModelNode node) {
+		assert !visited.containsKey(node);
+		visited.put(node, node);
+
 		GenericPlanNode genericPlanNode = new GenericPlanNode(node.getSignature());
 		genericPlanNode.setCostEstimate(node.getCostEstimate());
 		genericPlanNode.setResultSizeEstimate(node.getResultSizeEstimate());
@@ -63,11 +69,14 @@ public class QueryModelTreeToGenericPlanNode extends AbstractQueryModelVisitor<R
 		if (!deque.isEmpty()) {
 			GenericPlanNode genericParentNode = deque.getLast();
 			genericParentNode.addPlans(genericPlanNode);
+			assert node.getParentNode() == deque2.getLast();
 		}
 
 		deque.addLast(genericPlanNode);
+		deque2.addLast(node);
 		super.meetNode(node);
 		deque.removeLast();
+		deque2.removeLast();
 
 	}
 
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
@@ -8,7 +8,6 @@
 package org.eclipse.rdf4j.query.algebra.helpers;
 
 import java.util.ArrayDeque;
-import java.util.IdentityHashMap;
 
 import org.eclipse.rdf4j.common.annotation.Experimental;
 import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
@@ -26,9 +25,7 @@ public class QueryModelTreeToGenericPlanNode extends AbstractQueryModelVisitor<R
 
 	GenericPlanNode top = null;
 	QueryModelNode topTupleExpr;
-	ArrayDeque<GenericPlanNode> deque = new ArrayDeque<>();
-	ArrayDeque<QueryModelNode> deque2 = new ArrayDeque<>();
-	IdentityHashMap<QueryModelNode, QueryModelNode> visited = new IdentityHashMap<>();
+	ArrayDeque<GenericPlanNode> planNodes = new ArrayDeque<>();
 
 	public QueryModelTreeToGenericPlanNode(QueryModelNode topTupleExpr) {
 		this.topTupleExpr = topTupleExpr;
@@ -38,13 +35,8 @@ public class QueryModelTreeToGenericPlanNode extends AbstractQueryModelVisitor<R
 		return top;
 	}
 
-	// node.getParentNode() is not reliable because nodes are reused and parent is not maintained! This is why we use a
-	// queue to maintain the effective parent stack.
 	@Override
 	protected void meetNode(QueryModelNode node) {
-		assert !visited.containsKey(node);
-		visited.put(node, node);
-
 		GenericPlanNode genericPlanNode = new GenericPlanNode(node.getSignature());
 		genericPlanNode.setCostEstimate(node.getCostEstimate());
 		genericPlanNode.setResultSizeEstimate(node.getResultSizeEstimate());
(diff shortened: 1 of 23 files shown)


jeenbroekstra:
From what I can tell most (all?) of the failing tests here are caused specifically by `Var` objects being reused in the path sequence processing. Looks like part of the code is assuming we're using java object identity to do variable unification. That's not right - I think this points to (another) bug in the path sequence processing.

jeenbroekstra:
I will revisit this shortly. We recently did some fairly intensive refactoring of the path expression processing, so there's a chance that we inadvertently addressed this problem as part of that (or if not, at least made it easier to deal with ;))

hmottestad:
I think I remember this being a fairly extensive issue, not limited to paths by any means. Is there a way we could just remove the parent field instead? My confidence with the parent field is really low because it seems much easier to mess it up than to actually get it right.

jeenbroekstra:
> I think I remember this being a fairly extensive issue, not limited to paths by any means. Is there a way we could just remove the parent field instead? My confidence with the parent field is really low because it seems much easier to mess it up than to actually get it right.

We need to be able to somehow refer to a node's parent in order to handle things like replacing nodes (done as part of query optimization/rewriting in several places). We could look into replacing the fixed references with a dynamic treewalker mechanism that starts at the root and then walks down until it finds the first node that has the current node as one of its children, but that would not solve anything I think: we would still run the risk of a node effectively being the child of more than one parent. 

I'd rather we look into fixing the mechanism and putting safeguards against regression in place.

jeenbroekstra:
@hmottestad I have found the immediate cause of the problem, which is a bit of a corner case in that during evaluation of the query a new algebra node is temporarily created (this happens in PathIteration), and that node used uncloned variables. This is the only instance of that kind of behavior that I could find.

I would appreciate it if you could, for the purpose of this PR, remove the bespoke deque you are using `QueryModelTreeToGenericPlanNode` to track these relations (instead of relying on `getParentNode()`). I am not fully following what's happening there, but I'd like to see if we can start relying on the parent node mechanism again for this.

jeenbroekstra:
Btw one idea to "harden" the parent node mechanism is to add a few more assertions (so that we pick up more cases in testing) or even think about putting some exceptions in place.

hmottestad:
@jeenbroekstra I added another failing scenario.

hmottestad:
I feel this issue needs good test coverage before I would trust the parent field.

jeenbroekstra:
Can you please stop adding "temp" tests and other throwaway code? Either put in a proper test or just describe it to me. This is not helping, it's just creating extra work.

jeenbroekstra:
An alternative we can consider is, as you suggested, getting rid of the parentNode field in the algebra model itself, and using a stack of ancestory nodes (very much like you're doing in this specific case) as a general mechanism in the AbstractQueryModelVisitor. That way, the ancestry is always determined dynamically at the point it's needed. Not sure it will solve all corner cases, but it will be easier to test I think.

jeenbroekstra:
From what I can tell, the main problem we're seeing (in all these different cases) is that up to now, using getParentNode  was something  exclusively used _before_ the optimization pipeline kicks in. Several of the visitors in the optimization pipeline mess it about, creating new model nodes without cloning the arguments. The reason this was never a problem before is that we don't really rely on it anywhere else _after_ optimization or evaluation has happened, that's something newly introduced by the explain feature.

I've had a bit of a play with the alternative suggested above, but there are several cases where using that approach is not straightforward at all. We are probably better off keeping `getParentNode()` in place and making sure we come up with some testing strategies that will help us detect "wrongdoers" in the optimization pipeline.

hmottestad:
> Can you please stop adding "temp" tests and other throwaway code? Either put in a proper test or just describe it to me. This is not helping, it's just creating extra work.

I don't know why it is failing. As I've said, when I first created this branch I went digging to try to find the culprit and got lost since there were so many places that seemed to be problematic. The point of the temp test is literally just to show you that the problem was considerably more wide spread.

hmottestad:
My recommendation is to make parent and child fields immutable (final) so that we force the code to always have to decide what the parent and child fields should contain.

jeenbroekstra:
> My recommendation is to make parent and child fields immutable (final) so that we force the code to always have to decide what the parent and child fields should contain.

That would prohibit any sort of optimization or query rewriting.

hmottestad:
> > My recommendation is to make parent and child fields immutable (final) so that we force the code to always have to decide what the parent and child fields should contain.
> 
> That would prohibit any sort of optimization or query rewriting.

No no, we would just make a mechanism that would replace the part of the tree that you want to optimise. Maybe making the child fields immutable isn't necessary, just the parent field.

jeenbroekstra:
Ah I see, if it's _only_ the parent that we make final.... It will still be non-trival though because it means we can't ever move nodes around in the tree. We'll _always_ have to create a copy. We'll also need to adjust the `clone()` method on all node implementations, or rather we'll need to migrate away from using `Object.clone()` an introduce a bespoke `copyOf` method that takes a new parent node as one of its arguments.

hmottestad:
Test wise I was thinking about creating a QueryModelVisitor that uses the stack approach with assertions like the QueryModelTreeToGenericPlanNode, except it would be in the test directory instead. Then we could see how we could integrate it so that it can utilize our current sparql tests. I know we have the W3C sparql tests, maybe we could extend these in the the MemoryStore tests to also test the parent field in the query plan at various levels. Or maybe add a more complex assert to the query execution in general that runs the new QueryModelVisitor to check the parent field against the stack approach.

hmottestad:
Yes, `cloneWithNewParent(....)` was what I had in mind.

hmottestad:
But I have another idea. If it wouldn't degrade performance too much. How about a "cleaner" that can be run after each optimizer to fix up the parent field? Wouldn't solve the issue of nodes being shared instead of cloned, but that problem should only really exist in the parser I hope. This approach would also eliminate any issue from users/developers who add their own optimizers (that could be buggy) and then run into trouble with another optimizer that uses the parent field.

jeenbroekstra:
> But I have another idea. If it wouldn't degrade performance too much. How about a "cleaner" that can be run after each optimizer to fix up the parent field?

Not sure if we should do this after_each_ optimizer, but it's perhaps something we can consider as a "final" optimizer.

hmottestad:
Do you still think this is worth fixing?

jeenbroekstra:
> If one optimizer messes up the parent field and another optimizer uses that field?

There's only so much I'm prepared to do to clean up after optimizers that don't behave properly. I think it's useful to put a failsafe at the end, but the real fix is obviously in the actual optimizers, making sure that they keep the tree consistent. 

I've pushed a draft fix along these lines. 

I suggest that for each individual optimizer we introduce a test case that do, essentially, the same check as I've now added to ParentReferenceCleanerTest. If we make that part of an abstract test suites for optimizer implementations we can cheaply use it as a unit test for each optimizer, making sure that we catch these problems early.

hmottestad:
I feel that one optimizer shouldn't be able to break something as fundamental as the parent field.

jeenbroekstra:
> I feel that one optimizer shouldn't be able to break something as fundamental as the parent field.

Having a testsuite in place that ensures an optimizer implementation behaves is a good step in that direction.

> A good solutions would maybe be to abstract away the code that modifies the query plan tree so that the parent field was always maintained. Though that's not necessarily going to be easy.

I don't think the effort is worth it, to be honest. I'm certainly not intending to spend _that_ much time on this. There's other things far more important to fix/improve.

jeenbroekstra:
I've added some tests for all standard optimizers. Obviously we can extend these test cases to involve different query constructions (filters, functions, aliases, aggregates, etc) but it's a decent start I think.

jeenbroekstra:
This is now ready for final review.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
```diff
@@ -873,4 +873,38 @@ public void testDotTimed() {
 
 	}
 
+	@Test
+	public void testArbitraryLengthPath() {
```
hmottestad, line 877: I would have been ok with you deleting this test, but it's nice of you to turn it into a functional test.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
```diff
@@ -35,10 +37,11 @@ public GenericPlanNode getGenericPlanNode() {
 		return top;
 	}
 
-	// node.getParentNode() is not reliable because nodes are reused and parent is not maintained! This is why we use a
-	// queue to maintain the effective parent stack.
 	@Override
 	protected void meetNode(QueryModelNode node) {
+		assert !visited.containsKey(node);
+		visited.put(node, node);
```
hmottestad, line 43: What was the reason you added this? Did you see any cycles in query plan?

jeenbroekstra: This is something _you_ added, I think. I assumed it was necessary, but was this part of the throwaway test code as well?

hmottestad: Yeah. It was.

hmottestad: I intended for you to just revert my commits when you had figured out how to test the issue properly and how to fix it.

hmottestad:
Looks good to me @jeenbroekstra 

Nice job!

I'm glad you have left my `QueryModelTreeToGenericPlanNode` class as is. Your fixes and tests go a long way, but I'm still nervous about relying on the parent field at the moment. The stack based approach works fine though, so I'm happy to keep it.

Thanks again for your patience with me :)

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/QueryModelTreeToGenericPlanNode.java
jeenbroekstra: I'll take it out. See, this is why I don't like throwaway code. It has a nast habit of sticking around.

hmottestad:
When you have removed the temporary cycle detection code, fell free to squash all the commits into one. I've not really contributed any code towards a solution here, so no need to keep my commits around.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
jeenbroekstra: It's still a useful test to have in place, making sure that the expected explanation is actually produced.

jeenbroekstra:
> When you have removed the temporary cycle detection code, fell free to squash all the commits into one. I've not really contributed any code towards a solution here, so no need to keep my commits around.

I'll stick to the 5 commits now in place :)

Issue #2324, opened 2020-06-21
Blank node inlining misses inlining and writes incorrect data


hmottestad:
Parsing the following data:

```
@prefix ex: <http://example.com/ns#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

ex:node a sh:NodeShape;
  sh:property _:node1ebbjap1tx2 .

_:node1ebbjap1tx2 a sh:PropertyShape;
  sh:datatype xsd:integer;
  sh:path ex:age .

_:node1ebbjanb2x103 a sh:ValidationReport;
  sh:conforms false;
  sh:result _:adfa1de2-335b-43d8-85db-9b262f5a9cd1 .

_:adfa1de2-335b-43d8-85db-9b262f5a9cd1 a sh:ValidationResult;
  sh:detail _:44c0da42-0a09-4c10-a8dd-3c6db2ad5571 .

_:44c0da42-0a09-4c10-a8dd-3c6db2ad5571 a sh:ValidationResult;
  sh:focusNode _:node1ebbjap1tx4;
  sh:resultPath ex:age;
  sh:resultSeverity sh:Violation;
  sh:sourceConstraintComponent sh:DatatypeConstraintComponent;
  sh:sourceShape _:node1ebbjap1tx2;
  sh:value "123" .

_:adfa1de2-335b-43d8-85db-9b262f5a9cd1 sh:focusNode ex:validPerson1;
  sh:resultPath ex:knows;
  sh:resultSeverity sh:Violation;
  sh:sourceConstraintComponent sh:NodeConstraintComponent;
  sh:sourceShape _:node1ebbjap1tx1 .

_:node1ebbjap1tx1 a sh:PropertyShape;
  sh:node ex:node;
  sh:path ex:knows .

_:adfa1de2-335b-43d8-85db-9b262f5a9cd1 sh:value _:node1ebbjap1tx4 .
```

Then writing out using:

```
		model.setNamespace("ex", "http://example.com/ns#");
		model.setNamespace(FOAF.PREFIX, FOAF.NAMESPACE);
		model.setNamespace(XSD.PREFIX, XSD.NAMESPACE);
		model.setNamespace(RDF.PREFIX, RDF.NAMESPACE);
[...]

PR #2404, opened 2020-07-29
GH-2324 inlining blank nodes

hmottestad:
GitHub issue resolved: #2324 

---- 

### DIFF
--- a/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriter.java
+++ b/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriter.java
@@ -16,7 +16,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.LinkedList;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


hmottestad:
@jeenbroekstra this fixes this bug, but it's a terrible solution. The issue is that there is no way to inform the delegate that we don't want to inline the object of a statement while at the same time telling it that we want to inline the subject. 

I feel that the overall handling on inlining is too split up, we have this code here which handles deciding if we should inline, and then we have some code somewhere else that does the actual inlining, and we keep a lot of stuff in memory but decide to build complex counters instead of just using a model an querying it while we are writing to decide what should be inlined.

I don't propose we do anything about it now though. I've already spent more time on this than I wanted, and at least now it outputs a correct result, even if it's not an optimal result.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriter.java
```diff
@@ -293,58 +299,74 @@ private boolean isStillReferenced(SubjectInContext key) {
...
-					BNode bNode = (BNode) obj;
-					if (!bNodeOccurence.containsKey(bNode)) {
-						bNodeOccurence.put(bNode, 0);
+					bNodeOccurrences.compute(
+							(BNode) obj,
+							(key, i) -> i == null ? 1 : i + 1
+					);
+
+					if (bNodeOccurrences.get(obj) > 1) {
+						noneInlinedNodes.add(st.getSubject());
```
jeenbroekstra, line 338: Shouldn't you check if the subject is a blank node before adding it?

jeenbroekstra: I've tweaked this locally and it seems to solve the inlining issue where you had to adjust the test case, without compromising on your own new test case. I'll push a change.

hmottestad: Not sure actually. But if it works better now than I'm fine with it.

Issue #528, opened 2016-09-07
Allow HTTP authentication to be configured in repository configuration templates


jeenbroekstra:
(Migrated from https://openrdf.atlassian.net/browse/SES-727)

Please allow repository configuration templates - http://www.openrdf.org/doc/sesame2/users/ch07.html#d0e606 - to set the username and password required for connecting to HTTP Repositories protected by HTTP Basic Authentication. I know that: - currently that is supported by the Repository API using the RemoteRepositoryManager - SES-400, SES-695 - support for the parameters http://www.openrdf.org/config/repository/http#username and #password is almost implemented. The relevant code has been completed and then commented in http://repo.aduna-software.org/svn/org.openrdf/sesame/trunk/core/repository/http/src/main/java/org/openrdf/repository/http/config/HTTPRepositoryConfig.java and HTTPRepositoryFactory.java Please reconsider supporting http:username and http:password in .ttl configuration templates, along the support through the remote repository manager. The .ttl has the following advantages: - by allowing connections to a remote password-protected HTTPRepository with a LocalRepositoryManager, one would be able to a local Sesame Workbench with all remote repositories. A commenter in SES-400 and SES-695 has requested that after the issues has been closed. - having .ttl configuration templates that contain all configuration of a sesame repository, local or remote, is extremely convinient as demonstrated by the Sesame Console. My company

PR #2407, opened 2020-07-31
GH-528 httprepo basic auth config

jeenbroekstra:
GitHub issue resolved: #528 

- allow basic auth credentials (username, password) to be included when exporting a HTTPRepositoryConfig to an RDF model
- obfuscate password when a HTTPRepositoryConfig is sent in an RDF4J Server response
- remove obsolete / undocumented alternate way to get at repository configs, via old System repository setup.

An advantage of this approach is that the client receives the properties related to credentials, and such gets an indication that they need to be set. A downside is that if a client receives this config, changes some other property but forgets to re-assert the correct password, the stored password will be overwritten. 

Possible fix for that is to explicitly scan for the password mask when receiving an updated config, and ignoring that property if it is set to the mask value. 

(all of this under the caveat that we do not pretend that RDF4J server is particularly secure when basic http auth is used) 

---- 

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/config/ConfigView.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/config/ConfigView.java
@@ -18,7 +18,9 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.eclipse.rdf4j.http.server.ServerHTTPException;
 import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.repository.RepositoryException;
+import org.eclipse.rdf4j.repository.http.config.HTTPRepositorySchema;
 import org.eclipse.rdf4j.rio.RDFFormat;
 import org.eclipse.rdf4j.rio.RDFHandlerException;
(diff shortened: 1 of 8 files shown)


barthanssens (approved):
lgtm

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/config/ConfigView.java
```diff
@@ -79,6 +81,13 @@ public void render(Map model, HttpServletRequest request, HttpServletResponse re
 
 				if (!headersOnly) {
 					Model configuration = (Model) model.get(CONFIG_DATA_KEY);
+					// obfuscate the password, if present
+					configuration.getStatements(null, HTTPRepositorySchema.PASSWORD, null).forEach(st -> {
+						configuration.add(st.getSubject(), HTTPRepositorySchema.PASSWORD,
+								SimpleValueFactory.getInstance().createLiteral("****"));
+						configuration.remove(st);
+					});
```
jeenbroekstra, line 89: A massive problem with this approach is that if a client is unaware credentials have been configured, and tries to modify some other configuration property by retrieving config from the server, changing one properrty, then sending back, will overwrite the existing password with one containing four asterisks.

jeenbroekstra:
I will be dropping this PR for now. It's got less priority for us than I though, and the current approach is flawed.

PR #2409, opened 2020-07-31
GH-2324 blank node inlining still doesn't work

hmottestad:
GitHub issue resolved: #2324 

---- 

### DIFF
--- a/core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriterTest.java
+++ b/core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriterTest.java
@@ -150,6 +150,63 @@ public class ArrangedWriterTest {
 
 	}
 
+	@Test
+	public void testBlankNodesNotInlinedCorrectly2() throws IOException {
+		Model expected = Rio.parse(
+				new StringReader(
+						String.join("\n", "",
+								"@prefix ex: <http://example.com/ns#> .",
+								"@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .",
+								"@prefix sh: <http://www.w3.org/ns/shacl#> .",
+								"@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .",
+								"",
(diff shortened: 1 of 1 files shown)


hmottestad:
@jeenbroekstra 

I now believe that the only way to properly fix this is to add control for if:
 1. only subject should be inlined
 1. only object should be inlined
 1. both should be inlined

jeenbroekstra:
I think we need to step back a bit. ArrangedWriter has become a rather convoluted mess and the logic around inlining is spread out, which is really not good - the role of ArrangedWriter should really only be to buffer and sort statements, I think. Or perhaps we should get rid of it completely and just roll the arranging/inlining logic completely into the TurtleWriter class itself.

jeenbroekstra:
Let's think about it: when can a blank node be inlined? Or actually, let's turn it around: when can a blank node _not_ be inlined?

As far as I'm aware the only real case in which a blank node can not be inlined is **if it is an object value for more than one statement.**, because in that case Turtle syntax needs to be able to refer to the same resource in multiple places. 

So that leaves us, I think, with a rather simple internal sorting algorithm: group all statements by subject then predicate. Sort subject groups such that subject blank nodes which also occur elsewhere as objects are ordered _directly after_ their object-occurrences. All blank nodes can no be handled as inlined with the exception of subjects that occur as objects in multiple places. Which is a simple check to make if we just buffer the entire model.

The only downside is that we will need to buffer the complete RDF document, but this is inevitable when we want blank node inlining.

jeenbroekstra:
What complicates matters a bit is that ArrangedWriter also handles sorting logic related to other pretty-printing objectives, and I suspect this is why it's there in the first place: to allow disabling blank node inlining but still pretty-printing turtle without the need for full model buffering.

jeenbroekstra:
I'm going to be tinkering with this a bit. I'm also toying with just rolling back the changes done for GH-1079 and looking at an alternative approach for that corner case.

jeenbroekstra:
I've pushed a new branch with a proposed somewhat radical fix for the inlining problems: removing ArrangedWriter from the equation completely. See #2414

PR #2414, opened 2020-08-01
GH-2324 obsolete ArrangedWriter; move node inlining and sorting logic into TurtleWriter

jeenbroekstra:
GitHub issue resolved: #2324 

- deprecate and disuse ArrangedWriter
- add internal buffer (a Model) to TurtleWriter, use basic grouping/sorting filtering to output statements in correct order
- determine if node can be inlined per node instead of per statement

---- 


jeenbroekstra:
I did some ill-advised commit reverting, and it seems to ahve confused the eca check. It's not really important as it reverts a change that I subsequently made obselete anyway.

jeenbroekstra:
Clearly still an issue in this - from the test failure it looks like it mucks up serialization of repository configs.

core/rio/api/src/test/resources/logback-test.xml
```diff
@@ -5,6 +5,9 @@
 			<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %msg%n</pattern>
 		</encoder>
 	</appender>
+	<logger name="org.eclipse.rdf4j.rio">
+		<level value="warn"/>
```
jeenbroekstra, line 9: locally change to `debug` to see results of writing on stdout in several tests. Don't forget to change back before committing to avoid blowing up the build log.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriter.java
```diff
@@ -205,29 +203,27 @@ protected synchronized void consumeStatement(Statement st) throws RDFHandlerExce
...
-
-		if (statementBySubject.isEmpty() && blanks.isEmpty()) {
+		if (stmtBySubject.isEmpty() && blanks.isEmpty()) {
 			assert queueSize == 0;
 			return null;
 		}
-		Set<Statement> statements = null;
-		while (statements == null) {
+		Set<Statement> stmts = null;
+		while (stmts == null) {
```
jeenbroekstra, line 211: All the changes in this file are just the result of me reverting some earlier attempts to fix the inlining issues. They're not really relevant as the ArrangedWriter class is no longer used and only left in place on the off-chance that some third party uses it (unlikely but you never know).

jeenbroekstra:
Happy with this now, appreciate some eyeballs.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriter.java
```diff
@@ -298,105 +294,25 @@ private boolean isStillReferenced(SubjectInContext key) {
...
-					);
-
-					if (bNodeOccurrences.get(obj) > 1) {
-						if (st.getSubject() instanceof BNode) {
-							nonInlinedNodes.add(st.getSubject());
-						}
-						nonInlinedNodes.add(obj);
-					}
-				}
+				delegate.handleStatement(st);
```
hmottestad, line 314: This looks like you've reverted quite a lot of history.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -70,12 +85,6 @@
...
 	protected IRI lastWrittenPredicate;
 
-	/**
-	 * Always null
-	 */
-	@Deprecated
-	protected Model prettyPrintModel;
-
 	private final Deque<Resource> stack = new LinkedList<>();
 	private final Deque<IRI> path = new LinkedList<>();
```
hmottestad, line 89: These should be ArrayDeque, unless we expect them to grow very very large (since expanding the ArrayDeque gets costlier the bigger it is, talking millions of entries).

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -108,6 +117,11 @@ public TurtleWriter(OutputStream out, ParsedIRI baseIRI) {
 		this.writer = new IndentingWriter(new OutputStreamWriter(out, StandardCharsets.UTF_8));
 	}
 
+	// TODO make injectable
+	private ModelFactory getModelFactory() {
+		return new LinkedHashModelFactory();
+	}
```
hmottestad, line 123: You could just make it protected and then others could override this method if they have a better Model implementation they want to use.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -145,6 +159,12 @@ public void startRDF() throws RDFHandlerException {
 			xsdStringToPlainLiteral = getWriterConfig().get(BasicWriterSettings.XSD_STRING_TO_PLAIN_LITERAL);
 			prettyPrint = getWriterConfig().get(BasicWriterSettings.PRETTY_PRINT);
 			inlineBNodes = getWriterConfig().get(BasicWriterSettings.INLINE_BLANK_NODES);
+
+			if (isBuffering()) {
+				this.bufferedStatements = getModelFactory().createEmptyModel();
+				this.bufferSize = inlineBNodes ? -1 : DEFAULT_BUFFER_SIZE;
```
hmottestad, line 165: Btw. If you use a `long` for bufferSize you can just set it to be `Long.MAX_VALUE` and you don't have to deal with -1 as a special case. Dunno if it would help, but it's a decent approach since you would run out of memory way before reaching that limit (minimum 4 611 686  terabytes).

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -173,6 +193,10 @@ public void startRDF() throws RDFHandlerException {
 	@Override
 	public void endRDF() throws RDFHandlerException {
 		checkWritingStarted();
+
+		synchronized (bufferLock) {
+			processBuffer();
+		}
```
hmottestad, line 199: Could processBuffer be synchronized instead?

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -256,57 +276,71 @@ protected void handleStatementInternal(Statement st, boolean endRDFCalled, boole
...
-					writer.decreaseIndentation();
-					writePredicate(pred);
-					writer.increaseIndentation();
-					wrapLine(true);
-					path.removeLast();
-					path.addLast(pred);
-					lastWrittenPredicate = pred;
-				}
+			if (inlineBNodes && (pred.equals(RDF.FIRST) || pred.equals(RDF.REST))) {
+				handleList(st);
```
hmottestad, line 282: List handling in the pretty printed turtle code has always been bad. I should make some tests for that to see if you've fixed it now.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
...
+			for (Resource context : bufferedStatements.contexts()) {
+				Model contextData = bufferedStatements.filter(null, null, null, context);
+				Set<Resource> processedSubjects = new HashSet<>();
+				Optional<Resource> nextSubject = nextSubject(contextData, processedSubjects);
+				while (nextSubject.isPresent()) {
+					processSubject(contextData, nextSubject.get(), processedSubjects);
+					nextSubject = nextSubject(contextData, processedSubjects);
+				}
+			}
+		} else {
```
hmottestad, line 758: Does this now support pretty printing TRIG? That would be awesome.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
 			writer.write(" ");
 		}
 	}
+
+	/*
+	 * not synchronized, assumes calling method has obtained a lock on bufferLock
+	 */
```
hmottestad, line 741: I wonder if this could be checked in some way. To assert that we're not calling this method from multiple threads.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
...
+	private void processBuffer() throws RDFHandlerException {
```
hmottestad, line 742: I wonder if this could be checked in some way. To assert that we're not calling this method from multiple threads.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
...
+			if (subject instanceof BNode && inlineBNodes) {
+				Set<Resource> otherSubjects = contextData.filter(null, null, subject).subjects();
+				if (otherSubjects.stream().anyMatch(s -> !processedSubjects.contains(s))) {
+					continue;
+				}
+			}
+			return Optional.of(subject);
+		}
+		return Optional.empty();
+	}
```
hmottestad, line 785: Nice use of optional!

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
...
+					handleStatementInternal(toWrite, false, inlineValue(contextData, toWrite.getSubject()),
+							canInlineObject);
+					if (canInlineObject && toWrite.getObject() instanceof BNode) {
+						processSubject(contextData, (BNode) toWrite.getObject(), processedSubjects);
+					}
+				}
+				processedPredicates.add(predicate);
+			}
+
+		}
```
hmottestad, line 824: Since these pieces of code are so similar I'm wondering if maybe you would prefer to use the java 8 streams grouping collector (https://www.baeldung.com/java-groupingby-collector) and then sort it so that `rdf:type` gets preference. 

A more performant alternative would be to sort `contextData.getStatements(subject, null, null)` by the predicate with preference to `rdf:type` and then grouping could be done by looking at the previous predicate.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
...
+
+	private boolean inlineValue(Model contextData, Value v) {
+		if (!inlineBNodes) {
+			return false;
+		}
+		if (v instanceof BNode) {
+			return (contextData.filter(null, null, v).size() <= 1);
+		}
+		return true;
+	}
```
hmottestad, line 837: Maybe this should be `canInlineValue` to indicate that it checks if inlining is allowed.

hmottestad: It's a nice and simple approach though. I like it a lot.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -701,4 +735,109 @@ private void wrapLine(boolean space) throws IOException {
...
+
+	private boolean isBuffering() {
+		return inlineBNodes || prettyPrint;
+	}
```
hmottestad, line 841: I'm wondering what happened to the buffer size value, shouldn't it be used in some form when deciding if we are buffering or not?

hmottestad:
This is looking great @jeenbroekstra! I'm going to merge this into my shacl ast branch and see how it works (since that was where I first discovered the inlining bug).

hmottestad:
> This is looking great @jeenbroekstra! I'm going to merge this into my shacl ast branch and see how it works (since that was where I first discovered the inlining bug).

Just finished checked it now and it works great! Thanks @jeenbroekstra !

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: I didn't check - I'm pretty sure TriGWriter already supported pretty printing to some extent though.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: None of what I've done really touches the list handling I think - the only thing that changes is how it is invoked.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/ArrangedWriter.java
jeenbroekstra: It's really only two pull requests - yours and Miro's. In retrospect it wasn't really necessary but I wanted to start with a fresh perspective instead of continuing to build on patches that didn't completely solve the problems.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: We've got a lot of pre-1.6 code in places :) Good point, will fix.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: Forgot about this. I prefer making it properly injectable instead of forcing people to subclass - I'll add an optional constructor argument and also make it protected.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: Good idea.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: I'll take a look at DRYing this up a bit.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: It's a private method, and expectations are documented. I'm not too worried about it tbh. I'll turn it into a javadoc comment though to be a bit more in-your-face about it.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: Buffer size value only influences the choice of when we flush the buffer. This private method is just meant to check if we are in fact using a buffer (which is what we do when we are either pretty-printing or inlining). It is in fact used to initialize the buffer as well, so it couldn't check the buffer size.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
jeenbroekstra: I had a quick look at a streaming/sorting approach, but to be honest I'm not convinced I can easily come up with an alternative that is both more performant and also still easy to read/maintain. Not really worth investing too much time in right now I think.

jeenbroekstra:
Thanks for the feedback @hmottestad! I have fixed most of the issues you raised, have another look when you can.

avataar:
ArrangedWriter had a handy trimNamespaces() method that reduced the number of exported prefixes. As a side effect of this change now TurtleWriter-based exports have all the namespaces that are in the repository regardless of whether they are used by any of the exported statements (I've seen some pretty extreme abuse of namespaces). I'm not sure if that was the desired behaviour when ArrangedWriter was deprecated.

hmottestad:
The purpose of this pr was to fix some inlining test cases I had found. The original approach was fundamentally flawed, so Jeen rewrote it and cleaned up the code at the same time. The intention was not to remove features. @jeenbroekstra will know what we ought to do!

jeenbroekstra:
@avataar  good catch - I hadn't noticed that. Raised a ticket https://github.com/eclipse/rdf4j/issues/2694.

Issue #2390, opened 2020-07-25
Documentation for Eclipse RDF4J SHACL extensions


jeenbroekstra:
Followup of #2371:

publish the namespace as

- Respec page
-  http://rdf4j.org/shacl-extensions.ttl file

See open PR at https://github.com/eclipse/rdf4j-doc/pull/177

(split into separate issue to allow closing the original feature ticket for the release milestone)

jeenbroekstra:
What's the status of this? There's an open and approved pull request, are we ready to merge and publish the first working draft?

hmottestad:
I need to double check this after a bug I found.

PR #2418, opened 2020-08-05
GH-2390 shacl rsx spec

hmottestad:
GitHub issue resolved: #2390 

 - Respec html document
 - turtle namespace file

---- 


site/static/shacl-extensions
```diff
@@ -0,0 +1,16 @@
...
+<body>
+<ul>
+<li>
+<a href="shacl-extensions.html">Specification</a>
+</li>
+<li>
+<a href="shacl-extensions.ttl">Namespace (turtle)</a>
+</li>
+</ul>
+</html>
```
hmottestad, line 16: I haven't tested if this work, or if it just ends up being downloaded.

site/static/shacl-extensions.html
```diff
@@ -0,0 +1,263 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Eclipse RDF4J SHACL Extensions</title>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer></script>
```
hmottestad, line 6: Do we need a CQ for this?

site/static/shacl-extensions.html
```diff
@@ -0,0 +1,263 @@
...
+    <script class="remove">
+        var respecConfig = {
+            specStatus: "unofficial",
+            editors: [
+                { name: "Håvard Ottestad" },
+            ],
```
hmottestad, line 12: We should add more editors. Probably Vladimir if he wants to be listed.

hmottestad:
@VladimirAlexiev

site/static/shacl-extensions.html
jeenbroekstra: Good point - I think we do. Actually worth verifying that it's acceptable before investing heavily. I doubt it's a problem, but W3C use a non-standard license for their software, so best be sure that the legal team are ok with it.

site/static/shacl-extensions
jeenbroekstra: I don't understand the setup. Why have an html file that links to another html file and a turtle file? Why not simply have a link to the turtle file _inside_ the specification? For that matter, why have a turtle file at all? 

Given that we have multiple files relating to the same thing, I'd prefer if we could organize this a little neater: a subdirectory, for one thing.

site/static/shacl-extensions.html
jeenbroekstra: Not sure if this is the place to do it, but I notice that the current rendered spec links to `https://eclipse.github.io/rdf4j/` at the top (under "latest editor's draft"). Can we adjust that to link to the correct page?

site/static/shacl-extensions
hmottestad: Goal was that if you go to the namespace url, you end up at a page where you can choose if you want to read the spec or download the turtle file. The turtle file contains the schema for the namespace. We could also add the SHACL shapes for validating our extension to that turtle file. That's what SHACL does with their schema file.

site/static/shacl-extensions.html
jeenbroekstra: Fixed this

site/static/shacl-extensions
jeenbroekstra: Fair enough. I've reorganized this slightly - the html file with the links has been moved to /content and changed to markdown (bonus is that it now uses our website theme). The spec and the schema file have been added to a subdir to keep things organized (also because in future we may want to publish more specs this way). Tested locally, the namespace resolves to the link page, so all works as intended.

site/static/shacl-extensions.html
jeenbroekstra: I have changed my mind on this actually. We are not distributing the respec code itself as part of our repo or as part of our project downloads, and from what I can tell of similar things in IPZilla (e.g. a CQ for a hugo theme that was withdrawn on advice of the IP team), they consider this kind of scenario low-risk enough that IP review is not necessary.

Issue #2431, opened 2020-08-08
Make JMH benchmarks executable from command line and IDE


jeenbroekstra:
Currently, most jmh benchmarks in the project are not (easily) executable without the use of a special plugin, which (apparently) is only available in the IntelliJ IDE. 

To make the benchmarks truly cross-platform and more easily executable on both the command line and in any IDE _without_ using a specialized plugin, we need to make sure that:

1. every benchmark class has a main method so it can be executed easily from within an IDE;
2. every maven module that has benchmarks produces a shaded jar so benchmarks can be run directly from the command line (see https://mkyong.com/java/java-jmh-benchmark-tutorial/ for some examples on how to set this up)

jeenbroekstra:
The shade plugin is not an option for us unfortunately, because it cannot easily be used to generated a jar for the test scope in a format that jmh can work with. I'm investigating the use of the assembly plugin to produce a benchmark jar.

jeenbroekstra:
The assembly plugin is not an option either because of its inability to merge SPI registry files.

jeenbroekstra:
To make this work for the project as a whole, we'll move benchmark tests to their own modules. Each module will mirror the core module for which it provides benchmarks, for example for the memory store we'll have a separate module `rdfj4-benchmark-sail-memory`. The entire tree will live in a new parent module `benchmarks`, alongside core, compliance, and testsuites.

This way, we can easily:

1. run benchmarks from the IDE directly;
2. use the shade plugin to produce benchmark jar artifacts for execution from the command line;
3. ensure that benchmark code is _not_ deployed as part of releases

hmottestad:
I don't see a need to run the benchmarks from command line. I've sometimes wondered why no one has created a benchmark saas solution like GitHub CI that would check the performance of a PR automatically. 

Main method for running the benchmarks is a good idea though. Pity no one has made a plugin for eclipse for running jmh benchmarks.

jeenbroekstra:
> I don't see a need to run the benchmarks from command line.

There's a couple of reasons, but an important is this: running benchmarks from within an IDE can distort benchmark results. Almost every JMH article I've come across talks about this. That may be less of a problem if you control your environment carefully, but it doesn't hurt to take the IDE out of the equation. 

[...]

PR #2433, opened 2020-08-09
GH-2431 all benchmarks now have a main, and a fat jar

jeenbroekstra:
GitHub issue resolved: #2431 

- main method makes direct execution from IDE possible
- fat jar (target/jmh-benchmark.jar) makes command line execution
possible

to run benchmarks from command line (example: memory store benchmarks):

1. `cd core/sail/memory`
2. `mvn package -Pquick` 
3. `java -jar target/jmh-benchmarks.jar MemoryBenchmark.load` (to run the specific load benchmark test). 

To run from Eclipse, you can right-click on the benchmark and select "Run as Java application". Occassionally you may need to run a `mvn clean install` beforehand to make this work. Adapt the OptionBuilder in the benchmark's main class to control which exact benchmark tests get run.

---- 

### DIFF
--- a/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/MemoryBenchmark.java
+++ b/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/MemoryBenchmark.java
@@ -30,6 +30,10 @@ import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
 
 @State(Scope.Benchmark)
 @Warmup(iterations = 20)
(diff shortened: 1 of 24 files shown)


core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/MemoryBenchmark.java
```diff
@@ -45,6 +49,16 @@
 
 	private List<Statement> statementList = getStatements();
 
+	public static void main(String[] args) throws RunnerException {
+		Options opt = new OptionsBuilder()
+				.include("MemoryBenchmark.load") // adapt to run other benchmark tests
+				// .addProfiler("stack", "lines=20;period=1;top=20")
```
hmottestad, line 55: Btw. The stack profiler is terrible. It's more often than not gives the wrong results because it only benchmarks safe points. 20 stack lines is also a bit little. I think you need a hundred or maybe more. 

I just Java Flight Recorder for profiling.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/QueryBenchmark.java
```diff
@@ -68,6 +72,16 @@
 
 	List<Statement> statementList;
 
+	public static void main(String[] args) throws RunnerException {
```
hmottestad, line 75: I also noticed that you added main methods to each file. I'll have to try with IntelliJ to see if it doesn't mess up the plugins detection of what to run. I'm away from my computer for a week now, so just merge it and I'll give it a go sometime later. ￼

jeenbroekstra: No rush, I'm supposed to be on holiday myself (at home, we're still in lockdown here).

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/MemoryBenchmark.java
jeenbroekstra: I just copy pasted that from somewhere but good to know.

Issue #2415, opened 2020-08-02
Merge SPARQL* JSON result parser into regular SPARQL result JSON parser


jeenbroekstra:
We currently support working with serialized select query results for SPARQL* / RDF* in JSON through an extension of the regular SPARQL Results Json parser/writer. It uses a separate MIME-type (`application/x-sparqlstar-results+json`) to distinguish it from the standard format.

However, this approach has some downsides:

- when someone uses a query against the RDF4J Server where the result involves an RDF* triple, if they've not selected the specific RDF*-support result formats, they will receive the RDF* triple "encoded" as a base 64 URN (which is parsable, but likely opaque to the client);
- when someone uses RDF4J against a third party triplestore that supports RDF*, that third party store is unlikely to send the result back in a format that RDF4J can process.

An example of the latter is Stardog, which implements a SPARQL Query Results JSON extension that is almost identical to our own extension (see https://community.stardog.com/t/documentation-for-json-binary-query-result-format-extensions-for-edge-properties/2622) , but which gets sent back using the 'standard' `application/sparql-results+json` format, leading to our SPARQLRepository using the wrong result parser and crashing with an error.

We should probably also make some sort of proposal to the wider community to settle on a de-facto standard extension, to facilitate greater interoperability between the various tools that support RDF*.

jeenbroekstra:
I've started a discussion on the RDF* community mailinglist around this issue to see if we can come up with some sort of convergence and perhaps even a draft spec. See https://lists.w3.org/Archives/Public/public-rdf-star/2020Aug/0000.html. Interesting point is that Apache Jena also extended the JSON result syntax, with yet another _slightly_ different variant.

jeenbroekstra:
Possibility: merge  parsers, but not writers. keep both mime types. This will allow rdf4j clients to communicate  with third party stores like stardog, while rdf4j can still be  explicit.  We can modify the parser to gracefully handle the variants in syntax keywords.

PR #2434, opened 2020-08-12
GH-2415 allow parsing RDF* using standard sparql json result parser

jeenbroekstra:
GitHub issue resolved: #2415 

- allow parsing of RDF* extended syntax when receiving "normal" sparql json results mime type
- support Stardog syntax variant
- support Jena syntax variant

Note that RDF4J still uses the base-64 encoding approach when writing / sending RDF* triples in `application/sparql-results+json` format, a client would have to explicitly ask for `application/x-sparqlstar-results+json` to receive the extended format from an RDF4J-based endpoint. Depending on which way community opinions go with respect to this, we may change that behavior in a followup.

---- 

### DIFF
--- a/core/queryresultio/sparqljson/src/test/resources/sparqljson/rdfstar-extendedformat-rdf4j.srj
+++ b/core/queryresultio/sparqljson/src/test/resources/sparqljson/rdfstar-extendedformat-rdf4j.srj
@@ -1,0 +1,41 @@
+{
+  "head" : {
+    "vars" : [
+      "a",
+      "b",
+      "c"
+    ]
+  },
+  "results" : {
+    "bindings": [
+      { "a" : {
+          "type" : "triple",
+          "value" : {
+            "s" : {
+              "type" : "uri",
+              "value" : "http://example.org/bob"
+            },
+            "p" : {
+              "type" : "uri",
+              "value" : "http://xmlns.com/foaf/0.1/age"
+            },
+            "o" : {
+              "datatype" : "http://www.w3.org/2001/XMLSchema#integer",
+              "type" : "literal",
+              "value" : "23"
+            }
+          }
+        },
+        "b": { 
+          "type": "uri",
+          "value": "http://example.org/certainty"
+        },
+        "c" : {
+          "datatype" : "http://www.w3.org/2001/XMLSchema#decimal",
+          "type" : "literal",
+          "value" : "0.9"
+        }
+      }
+    ]
+  }
+}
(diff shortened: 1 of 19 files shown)


core/queryresultio/sparqljson/src/test/resources/sparqljson/rdfstar-extendedformat-rdf4j.srj
```diff
@@ -0,0 +1,41 @@
+{
```
jetztgradnet, line 1: Does it make sense to also add a test for the statement-encoded-as-base64 variant of the backwards-compatible output? And speaking of which, maybe even any generic plain JSON-LD output, as the SPARQL* parser should also be able to read that, right?

jetztgradnet:
Would it make sense to also add a test for the statement-encoded-as-base64 variant of the backwards-compatible output produced by RDF4J? 
And speaking of which, maybe even any generic plain JSON-LD output, as the SPARQL* JSON-LD parser should also be able to read that, right?

core/queryresultio/sparqljson/src/test/resources/sparqljson/rdfstar-extendedformat-rdf4j.srj
jeenbroekstra: > Does it make sense to also add a test for the statement-encoded-as-base64 variant of the backwards-compatible output?

The base-64 encoding is something the _writer_ uses to transform a triple value to a URI, so that it can be added to a result format serialization without the need for special extensions. It doesn't really make sense to add tests for it at the parser end because the whole point of it is that the parser treats it like any other URI. 

Arguably we could have some tests in place for the writer to check that the encoding is properly applied when the selected format doesn't support RDF* extensions, but that's a little out of scope for what we're doing here I think. I may look at that in a followup though as I may want to change the logic there a little bit.

> And speaking of which, maybe even any generic plain JSON-LD output, as the SPARQL* parser should also be able to read that, right?

Note that what we're fixing here is specifically the parser/writer for the SPARQL _table results_ format. JSON-LD is a different format, a serialization for RDF graphs (rather than table results), and it has a separate parser/writer. Currently, RDF4J's JSON-LD parser/writer does not support any RDF* extensions - the only RDF serialization formats that have been extended are Turtle (Turtle*), TriG (TriG*), and the RDF4J Binary RDF format. For all other formats, you can configure the writer to either convert RDF* triples to "classic" reification, or to use the same base64 encoding trick.

jeenbroekstra: I've created a followup issue for more configurable writing behavior: https://github.com/eclipse/rdf4j/issues/2437 . Separate PR coming up soon

Issue #2458, opened 2020-08-22
SHACL - Scoping of logical operators with sh:hasValue and dash:hasValueIn


hmottestad:
There is a scoping issue for sh:or when used node shapes vs property shapes.

```
@prefix ex: <http://example.com/data/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .

ex:shape1
	a sh:NodeShape ;
	sh:targetClass ex:Test ;
	sh:property [
		sh:path ex:info ; 
		sh:or ([ sh:hasValue "blue" ; ] [ sh:hasValue "green" ; ] [ sh:hasValue "red" ; ]) ;
	]  .

```

```
@prefix ex: <http://example.com/data/> .

ex:test1 a ex:Test  .
```

The above example fails validation in the ShaclSail, but should pass since `sh:hasValue` is inside a `sh:nodeShape` and thus each value is validated independently.

jeenbroekstra:
I'm not sure I understand the example. Why should this succeed? The contraint appears to expresss that every instance of test must have a value for `ex:info` that is either blue, green or, red. Why should a shape that doesn't have _any_ value for that property be valid?

hmottestad:
The definition for sh:and is "for every value" so if there are no values then it should succeed. Further more, each shape inside the sh:and is a node shape. So we are actually saying that each value for sh:info should be "red" and "blue" and ...  

It took me a while to understand it. It's fairly counter intuitive.

jeenbroekstra:
...but the example is an `sh:or`, not an `sh:and`.... I'm confused.

hmottestad:
I'll update the title. The definition sh:and and sh:or are the same wrt. how they operate if there are not elements and how they operate on value nodes vs. targets.

rdstn:
If I'm getting this correctly, the way to retain the old behaviour for target shapes is to use paths?

So, for example, if I want this Shape to target people named Aerith or Bob, and not to match on null names, I'd do:

```
rsx:targetShape [
    sh:or (
        [ sh:path rdfs:name ; sh:hasValue "Aerith" ]
        [ sh:path rdfs:name ; sh:hasValue "Bob" ]
    ) ; ] ;
```

Whereas if I want to match null names, I'd go for:
```
rsx:targetShape [
    dash:hasValueIn ( "Aerith", "Bob" )
]
```

[...]

PR #2459, opened 2020-08-22
GH-2458 sh:hasValue and dash:hasValueIn support as node shapes

hmottestad:
GitHub issue resolved: #2458  

Make sh:hasValue and dash:hasValueIn work when part of a sh:NodeShape. This fixes when they are used as part of a logical operator like sh:or. 

---- 


hmottestad:
@jeenbroekstra I would appreciate it if you could take a look at this. I would like to merge it and release a bug fix release once ontotext gives it a quick test.

jeenbroekstra:
I'll take a look. While I do that, can you please link to the issue and update the PR description?

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/TempTest.java
```diff
@@ -500,4 +500,50 @@ private void add(SailRepositoryConnection connection, String data) throws IOExce
 		connection.commit();
 	}
 
+	@Test()
+	public void temp() throws Throwable {
+
```
jeenbroekstra, line 505: Please remove this class.

core/sail/shacl/src/test/resources/test-cases/hasValueIn/or/shacl.ttl
```diff
@@ -0,0 +1,16 @@
...
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix dash: <http://datashapes.org/dash#>.
+
+ex:shape1
+	a sh:NodeShape ;
+	sh:targetClass ex:Person ;
+	sh:property [
+		sh:path ex:info ;
+		sh:or ([ dash:hasValueIn ("blue") ; ] [ dash:hasValueIn ("green") ; ] [ dash:hasValueIn ("red"); ]) ;
```
jeenbroekstra, line 15: I don't object to having a test for a corner case, but isn't this an odd use of `dash:hasValueIn`? I thought the point was to allow use of a list so you don't have to use an or constraint at all. In other words, isn't the above equivalent to just saying:

     sh:path ex:info;
     dash:hasValueIn ("blue" "green" "red");

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
```diff
@@ -227,15 +227,45 @@ public PlanNode getAllTargetsPlan(ConnectionsGroup connectionsGroup, boolean neg
...
+					.map(Optional::get)
+					.collect(Collectors.joining(" ) && ( ", "( ",
+							" )"));
+
+			String query = pathQuery1 + "\n FILTER (! EXISTS {\n" + pathQuery1.replaceAll("(?m)^", "\t")
+					+ "\n\tFILTER(!(" + collect + "))\n})";
+
+			String pathQuery2 = getPath().getQuery(targetVar, randomVariable(), null);
+
+			query = "{\n#VALUES_INJECTION_POINT#\n " + query.replaceAll("(?m)^", "\t")
```
jeenbroekstra, line 250: what is `VALUES_INJECTION_POINT`?

jeenbroekstra (changes requested):
Like I mentioned on the actual issue ticket: I don't follow what exact problem you're solving. Can you clarify  on the issue and by updating the PR description?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
hmottestad: It's where we inject SPARQL values. Since we have unions we end up introducing a new scope, which leads to poor performance. So we add points where we can inject sparql values, for the cases where we need to run the query against a subset of nodes.

jeenbroekstra: Ah I see. ObNitpick: it might be good to declare a constant for this substitution pattern.

core/sail/shacl/src/test/resources/test-cases/hasValueIn/or/shacl.ttl
hmottestad: That would be equivalent to:
```
sh:or ([sh:path ex:info ; dash:hasValueIn ("blue") ; ] [sh:path ex:info ; dash:hasValueIn ("green") ; ] [sh:path ex:info ; dash:hasValueIn ("red"); ]) ;
```

jeenbroekstra: Ah I see! I think I'm beginning to understand what the actual issue here is. At least, I think so...

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
```diff
@@ -227,15 +227,45 @@ public PlanNode getAllTargetsPlan(ConnectionsGroup connectionsGroup, boolean neg
...
-						.map(propertyShape -> propertyShape.buildSparqlValidNodes(targetVar))
-						.reduce((a, b) -> a + "\n" + b))
-				.filter(Optional::isPresent)
-				.map(Optional::get)
-				.reduce((a, b) -> a + "\n" + b)
-				.orElse("");
+
+		if (hasOwnPath()) {
+			// within property shape
+			String objectVariable = "?b";
```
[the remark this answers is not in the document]
hmottestad: I'll take a look and check.

hmottestad:
I believe this should be done now. I found another edge case to handle, but that's fixed now.

jeenbroekstra (approved):
LGTM, thanks!

jeenbroekstra:
@hmottestad on the issue ticket @rdstn pointed out that the whole thing is counter-intuitive and we perhaps should add some documentation to explain how to use this. Is that something you want to tackle as part of this issue, or will you log a followup for that (or do you think it's not our problem to document this)?

hmottestad:
The issue is with SHACL being counter intuitive and not so much this implementation. So I'm not sure where to document it.

Issue #2442, opened 2020-08-16
Sorting values with datatype - performance


hmottestad:
Values with datatype are slow to sort because there is effectively a lot of string equals to determine which datatype is being used. 

We should look at what options we have to reduce the amount of string equals operations. Maybe some sort of interning, or a new class with helper methods, or enums?

hmottestad:
Step one would be to profile sorting of various values to show that comparing data types is the issue.

jeenbroekstra:
Can you give some context? Are you talking about SPARQL ordering, or one of our comparator utilities, or both? A particular triplestore, or more generally?

hmottestad:
<img width="1680" alt="Screenshot 2020-08-17 at 12 08 41" src="https://user-images.githubusercontent.com/797185/90384784-67cbc800-e082-11ea-8b5f-4d2e3db3dfda.png">

hmottestad:
I created a simple test to show that SimpleIRI.equals is the method that takes the most time during sorting. Mostly from determining the datatype. This is with the ValueComparator.

hmottestad:
<img width="1529" alt="Screenshot 2020-08-17 at 12 50 17" src="https://user-images.githubusercontent.com/797185/90388539-3f46cc80-e088-11ea-88ad-ca97e572411d.png">

Here is part of the flame graph showing just the comparisons. 

Take away from this is that `compareLiterals(...)` uses most of its time trying to determine the datatype. Which it needs to do many times for the same value when sorting. There is also the ´nullEquals(...)` method which I think is unneeded. I'll look into that.

hmottestad:
Made a little PR with my code: https://github.com/eclipse/rdf4j/pull/2447

hmottestad:
The merged PR only solves part of this issue. We still have the issue of the large amount of time being used to determine the datatype of literals.

jeenbroekstra:
> The merged PR only solves part of this issue. We still have the issue of the large amount of time being used to determine the datatype of literals.

Just from looking at the code in XMLDatatypeUtil: we're doing a lot of `uri.equals(XSD.DT1) || uri.equals(XSD.D2) || ...`. to check if a datatype is numerical, or a calendar type, etc. I wonder if a simple optimization would be to stick those types in a fixed collection and do a `collection.contains(uri)`. If we use something like a LinkedHashSet this would gives us a better average lookup performance.

jeenbroekstra:
[...]

PR #2477, opened 2020-08-29
GH-2442 improve performance for sorting literals with datatype by introducing an enum for known types

hmottestad:
GitHub issue resolved: #2442  

 - introduce enum for known datatypes with helper methods that return precomputed data for things like `isNumber()`
 - reduce use of `instanceof` by introducing helper methods on `Value` like `isLiteral()` and `isBnode()`

---- 


hmottestad:
@jeenbroekstra this was my general idea. It's not helping as much as I had though. I was expecting to increase the performance by maybe 2x, but it's only around 25%.

I'm wondering it the `instanceOf` usage in the value comparator is what is actually slowing things down. Might look at bit more at that.

You can take a look at this PR though. It gives a good idea of the solution that I was thinking about. We don't need to consider merging it yet though. Since the performance isn't what I was hoping for I think we would need to find more performance in other aspects of the code (like maybe the memory store or maybe the sparql engine) in order to justify the new code.

hmottestad:
Btw. I had initially written this code to use optionals, but the profiling showed that it was a big bottleneck. Removing the use of optional didn't actually make a difference though, so we could easily go back to using optionals instead of returning null when the xml datatype is unknown.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -130,11 +145,26 @@ protected void setDatatype(IRI datatype) {
...
+	}
+
 	@Override
 	public IRI getDatatype() {
 		return datatype;
 	}
 
+	@Override
+	public XmlDatatypeEnum getXmlDatatypeEnum() {
+		if (!xmlDatatypeCached) {
```
jeenbroekstra, line 161: Is this separate boolean flag necessary? You could just do this with a simple null check, right?

core/model/src/main/java/org/eclipse/rdf4j/model/Literal.java
```diff
@@ -47,6 +48,10 @@
 	 */
 	public IRI getDatatype();
 
+	default public XmlDatatypeEnum getXmlDatatypeEnum() {
```
jeenbroekstra, line 51: I agree with you that use of Optional is better here (especially if it makes no difference in performance) - it will also make some of your usage code a bit more fluent. Btw the `public` modifier is redundant here (and on everything else in this interface as well I guess).

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/Temp.java
```diff
@@ -0,0 +1,23 @@
...
+					XMLDatatypeUtil.isDecimalDatatype(value.iri) + " ," +
+					XMLDatatypeUtil.isFloatingPointDatatype(value.iri) + " ," +
+					XMLDatatypeUtil.isCalendarDatatype(value.iri) + " " +
+					") ,"
+			);
+
+		}
+		;
+
+	}
```
jeenbroekstra, line 21: You will need to remove this before merging.

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XmlDatatypeEnum.java
```diff
@@ -0,0 +1,213 @@
...
+
+import javax.xml.datatype.Duration;
+import javax.xml.datatype.XMLGregorianCalendar;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.vocabulary.XSD;
+
+import com.google.common.collect.Sets;
+
+public enum XmlDatatypeEnum {
```
jeenbroekstra, line 15: I really dislike having the `Enum` suffix on the class name. How about we move this enum to an inner type of `XSD` and call it `Datatype` instead? You'd refer to it as `XSD.Datatype.DURATION` etc.

core/model/src/main/java/org/eclipse/rdf4j/model/Literal.java
jeenbroekstra: Naming: as with the class name, I dislike having the type in the method name. Can we call the method `getXmlDatatype` instead?

jeenbroekstra (changes requested):
Thanks, looks good, I think this is more or less what I thought you had in mind as well. I like the general approach and if we can demonstrate a significant performance advantage then it's worth the addition. The way it's done introduce little or no implementation burden for third party developers so that's good.  

You'll need to rebase this against the develop branch though. Also a couple of inline remarks to take into account at some point between now and finalizing this.

hmottestad:
Performance wise I'm seeing in the order of 20-30% performance improvement, which is far away from the 2x I was hoping for. Should we still go for this @jeenbroekstra ?

jeenbroekstra:
That's still significant, I think. Though I don't quite know what the absolute numbers are and how much of a performance issue this was in the first place.

hmottestad:
It's the hottest code when doing sorting. So it will make all sorting of literals faster.

hmottestad:
### Original
Benchmark                                       Mode  Cnt    Score    Error  Units
ValueComparatorBenchmark.sortManyDeepEquals     avgt    5  282.831 ±  5.465  ms/op
ValueComparatorBenchmark.sortManyPointerEquals  avgt    5  128.369 ±  3.106  ms/op
ValueComparatorBenchmark.sortObjects            avgt    5  147.940 ± 14.908  ms/op
ValueComparatorBenchmark.sortPredicates         avgt    5    0.111 ±  0.028  ms/op
ValueComparatorBenchmark.sortSubjects           avgt    5   42.832 ± 12.686  ms/op

### Current

```
Benchmark                                       Mode  Cnt    Score   Error  Units
ValueComparatorBenchmark.sortLiterals           avgt    5   22.420 ± 8.165  ms/op
ValueComparatorBenchmark.sortManyDeepEquals     avgt    5  198.964 ± 2.486  ms/op
ValueComparatorBenchmark.sortManyPointerEquals  avgt    5   16.796 ± 0.338  ms/op
ValueComparatorBenchmark.sortPredicates         avgt    5    0.052 ± 0.005  ms/op
ValueComparatorBenchmark.sortSubjects           avgt    5   20.751 ± 9.934  ms/op
```

hmottestad:
@jeenbroekstra I figured out the biggest issue was the use of `instanceof` in the comparator. I've introduced helper methods on `Value` like `.isBnode()` with default implementations which makes everything considerably faster!

The subjects benchmark (IRIs and BNodes) is **2x** faster, only because of this change.

Sorting objects (mostly literals) is now **6x** faster!

hmottestad:
There are two things to still consider:
 - reintroduce `Optional` instead of returning `null` when getting the enum based datatype from a literal
 - remove the datatype IRI comparison pre calculation for known datatypes

Now that everything is as fast as I was expecting originally we can consider if the above two things will affect performance noticeably. If not then we should prefer readable code over highly optimised code with low impact.

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XmlDatatype.java
```diff
@@ -0,0 +1,2470 @@
...
+			valueComparatorLookup.put(GYEAR, -3);
+			valueComparatorLookup.put(GMONTHDAY, -3);
+			valueComparatorLookup.put(GDAY, -3);
+			valueComparatorLookup.put(GMONTH, -3);
+			valueComparatorLookup.put(STRING, -15);
+			valueComparatorLookup.put(BOOLEAN, 2);
+			valueComparatorLookup.put(BASE64BINARY, 2);
+			valueComparatorLookup.put(HEXBINARY, -4);
+			valueComparatorLookup.put(FLOAT, -2);
+			valueComparatorLookup.put(DECIMAL, 16);
```
hmottestad, line 227: Pre-calculation of `new ValueComparator().compare(d1, d2)` for all datatype combinations.

hmottestad:
Followup from this would be to reduce use of `instanceof` in the rest of the codebase and use the new helper methods instead (eg. `value.isLLiteral()` instead of `value instanceof Literal´).

hmottestad:
Would really appreciate if you could take a look at this @jeenbroekstra . I'll test out optional today and then I would like to merge this PR.

hmottestad:
Use of optionals did not change the performance significantly and it makes the code more intuitiv.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
hmottestad: removed now that we've moved to optionals

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XmlDatatypeEnum.java
hmottestad: I followed your recommendation here and it looks pretty decent. Thanks!

core/model/src/main/java/org/eclipse/rdf4j/model/Literal.java
hmottestad: I've gone with `getXsdDatatype()` which aligns with having `Datatype` be an enum within the `XSD` class.

hmottestad: I've also moved to using `Optional`.

core/model/src/main/java/org/eclipse/rdf4j/model/BNode.java
```diff
@@ -40,4 +40,24 @@
 	 */
 	@Override
 	public int hashCode();
+
+	@Override
+	default boolean isBnode() {
```
jeenbroekstra, line 45: Spelling: it should be `isBNode` (capital N)

core/model/src/main/java/org/eclipse/rdf4j/model/IRI.java
```diff
@@ -72,4 +72,24 @@
...
+
+	@Override
+	default boolean isLiteral() {
+		return false;
+	}
+
+	@Override
+	default boolean isTriple() {
+		return false;
+	}
```
jeenbroekstra, line 94: I suggest we move these defaults to `Value` instead, and then have each sub-interface override only the one for which the result should be true.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -159,11 +192,8 @@ public boolean equals(Object o) {
 				return getLanguage().get().equalsIgnoreCase(other.getLanguage().get());
 			}
 			// If only one has a language, then return false
-			else if (getLanguage().isPresent() || other.getLanguage().isPresent()) {
-				return false;
-			}
-
-			return true;
+			else
+				return !getLanguage().isPresent() && !other.getLanguage().isPresent();
```
jeenbroekstra, line 196: Missing curly braces.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -54,6 +54,12 @@
 	 */
 	private IRI datatype;
 
+	// The XSD.Datatype enum that matches the datatype IRI for this literal. This value is calculated on the fly and
+	// cached in this variable. `null` means we have not calculated and cached this value yet. We are not worried about
+	// race conditions, since calculating this value multiple times must lead to the same effective result. Transient is
+	// only used to stop this field from be serialised.
+	transient private Optional<XSD.Datatype> xsdDatatype;
```
jeenbroekstra, line 61: It's a little odd to be using an Optional value that can also be `null`. I understand why it's done (and the comment also explains it well), but to make it glaringly obvbious perhaps initialize the value as `null` here as well, explicitly.

jeenbroekstra (commented):
Overall looks good to me @hmottestad, though I am frankly astonished that an `instanceof` check would be such a performance bottleneck - this is something where I would expect the JRE to optimize ferociously.

I only have couple of minor editorial remarks.

One other point: this PR is likely to cause conflicts with #2546 when one or the other is merged. Regardless of which order we merge them and where we resolve conflicts, I'll be available to help clean it up though I could use some help.

core/model/src/main/java/org/eclipse/rdf4j/model/IRI.java
hmottestad: I prefer to have Value have functional defaults (eg. instanceof), but it wouldn't actually make a difference I think.

Issue #2485, opened 2020-09-05
Fix failing test problem


jeenbroekstra:
We currently have a consistently failing test on both master and develop. It started after the merge of https://github.com/eclipse/rdf4j/pull/2474 - the guava version bump. Oddly neither the feature branch nor master initially failed. The first failing build was on develop, after https://github.com/eclipse/rdf4j/pull/2480 (the develop sync merge of the guava version bump).

PR #2486, opened 2020-09-05
GH-2485 fix failing test in RDFCollections

jeenbroekstra:
GitHub issue resolved: #2485 

- make sure test for cycle detection uses a collection that is well-formed (apart from the cycle obviously)
---- 

### DIFF
--- a/core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
+++ b/core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
@@ -82,7 +82,7 @@ public class RDFCollectionsTest {
 
 	@Test(expected = ModelException.class)
 	public void testNonWellformedCollection_Cycle() {
-		Resource head = vf.createBNode();
+		Resource head = vf.createBNode("z");
 		Model m = RDFCollections.asRDF(values, head, new TreeModel());
 		m.add(head, RDF.REST, head);
 		RDFCollections.asValues(m, head, new ArrayList<>());
--- a/core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
+++ b/core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
@@ -82,13 +82,11 @@ public class RDFCollectionsTest {
 
 	@Test(expected = ModelException.class)
 	public void testNonWellformedCollection_Cycle() {
-		for (int i = 0; i < 1000; i++) {
-			Resource head = vf.createBNode();
-			Model m = RDFCollections.asRDF(values, head, new TreeModel());
-			m.add(head, RDF.REST, head);
-			RDFCollections.asValues(m, head, new ArrayList<>());
-			fail("collection with cycle should result in error");
-		}
+		Resource head = vf.createBNode();
+		Model m = RDFCollections.asRDF(values, head, new TreeModel());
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Ok. So it looks as if the guava version bump was not directly responsible for this problem: even when going back to the old version, we still have a failing test.

jeenbroekstra:
I'm going to take a closer look at the test itself.

jeenbroekstra:
🤨

hmottestad:
So you think it's the test that's actually broken, and not the underlying code?

jeenbroekstra:
I don't know what to think. I reorganised the tests not because I thought that would fix it, but to isolate the failing function (cycle detection).

Then again I can't spot anything in the code that would make this test fail in some circumstances but not others. There's no IO involved, no concurrency, so no possible race conditions that I can see.

hmottestad:
I found out the issue only occurs when the head node is lexicographically after the other nodes in the rdf list. So I pushed a commit where the head node is "z", which on my computer is always after since the other blank nodes all start with node... .

jeenbroekstra:
> I found out the issue only occurs when the head node is lexicographically after the other nodes in the rdf list. So I pushed a commit where the head node is "z", which on my computer is always after since the other blank nodes all start with node... .

That is an interesting find, thanks! Gives us a good lead (though why lexicographic order of bnods should play a role in this is currently a mystery to me - perhaps a hash issue in the specific Model implementation we use?)

jeenbroekstra:
Part of the problem is that the test introduces a cycle by adding a _second_ `rdf:rest` statement for the head node. The original `rdf:rest` statement is not replaced. If the head node is lexicographically _behind_ the object node of that original `rdf:rest` statement, the algorithm just ignores the second statement.

jeenbroekstra:
Since the utility function is documented as expecting a well-formed collection, I've fixed this in the short term as a problem in the test, rather than in the code itself. The test now replaces the `rdf:rest` relation, rather than adding a second one, so that (apart from the introduced cycle) the collection is well-formed.

jeenbroekstra:
<insert noughty word>! I _really_ hate it when the build fails on failing to download a dependency... multiple times.

jeenbroekstra:
Alright, build is green, I'd appreciate a final review (and if possible, a green tick :)

core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
```diff
@@ -71,35 +70,33 @@ public void testConversionRoundtrip() {
...
-		}
+		RDFCollections.asValues(m, head, new ArrayList<>());
+	}
+
+	@Test(expected = ModelException.class)
+	public void testNonWellformedCollection_IncorrectHeadNode() {
+		Resource head = vf.createBNode();
+		Model m = RDFCollections.asRDF(values, head, new TreeModel());
+
+		RDFCollections.asValues(m, vf.createBNode(), new ArrayList<>());
```
hmottestad, line 99: Could you add the comment: `// supply incorrect head node` 

Makes the test easier to read.

hmottestad (approved):
Looks good to me. Just added one comment. 👍

core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
jeenbroekstra: Fixed.

Issue #2497, opened 2020-09-14
Turtle Blank node inlining loses data when processing malformed rdf collection structures


hmottestad:
Found another blank node inlining bug where the inlined turtle file is not isomorphic with the non-inlined version.

The problem in this case is that the input data uses a malformed list structure: there are several places where nodes have _multiple_ rdf:first or rdf:rest properties, which is not valid as an RDF collection structure. In such cases, the writer should not attempt to use list shorthand syntax (e.g. `(1 2 3)`) but fall back to just serializing as regular statements (`[ rdf:first 1; rdf:rest ...` etc). 

```
@prefix ex: <http://example.com/ns#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .

_:node1ei5i7ns4x11752 a sh:ValidationReport;
  sh:conforms false;
  <http://rdf4j.org/schema/rdf4j#truncated> false;
  sh:result _:5f9908e9-caee-4059-aa52-69227a27ac20, _:3a37be3d-7af5-4b74-b7cf-b6358196bf0e,
    _:ba3bb72d-eb40-4aec-987a-bf2a3d925f06, _:6c86bbd3-a6ea-4b4f-8881-d479cff3aff5, _:ff167b12-c1b4-4341-bf63-187bed54e1ed,
    _:c3c6878c-7a69-41c8-b65b-50ab02da9519, _:f07f8ae4-1137-454c-ac6a-f385631b592e, _:a7b959ba-a7b4-4922-9974-9f363fdaccdf .

_:5f9908e9-caee-4059-aa52-69227a27ac20 a sh:ValidationResult;
  sh:focusNode ex:validPerson1;
  sh:value ex:steve;
  sh:resultPath ex:knows;
  sh:sourceConstraintComponent sh:NotConstraintComponent;
  sh:resultSeverity sh:Violation;
  sh:sourceShape _:node1ei5i7sdex1 .

_:node1ei5i7sdex1 a sh:PropertyShape;
  sh:path ex:knows;
  sh:not _:node1ei5i7sdex2 .

_:node1ei5i7sdex2 a sh:NodeShape;
[...]

PR #2498, opened 2020-09-14
GH-2497 fix RDF collection processing for malformed collection structures

hmottestad:
GitHub issue resolved: #2497  

- Fix handling of malformed collection structures by checking beforehand, and falling back to serializing as regular RDF statements if the collection is not well-formed
- Added several regression test cases for various kinds of malformedness.

---- 


hmottestad:
@jeenbroekstra I found another blank node inlining bug. Maybe you want to take a look since you fixed the last one.

jeenbroekstra:
Bit hard to tell where the discrepancy is from this. Can you narrow down where in the example input things get mapped wrong?

hmottestad:
Narrowed it down and also used Jena to cleanup the input data somewhat.

core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleWriterTest.java
```diff
@@ -194,4 +194,170 @@ public void testNoBuffering() throws Exception {
...
+		String data = "" +
+				"@prefix ex:    <http://example.com/ns#> .\n" +
+				"@prefix sh:    <http://www.w3.org/ns/shacl#> .\n" +
+				"@prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .\n" +
+				"@prefix xsd:   <http://www.w3.org/2001/XMLSchema#> .\n" +
+				"@prefix rdfs:  <http://www.w3.org/2000/01/rdf-schema#> .\n" +
+				"@prefix foaf:  <http://xmlns.com/foaf/0.1/> .\n" +
+				"\n" +
+				"ex:a  ex:list   (_:b0 _:b0) .";
+
```
hmottestad, line 344: This seems to be a representative example.

hmottestad: This is what RDF4J outputs:

```
@prefix ex: <http://example.com/ns#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

ex:a ex:list ([] []) .
```

jeenbroekstra: ...that is correct, as far as I can tell. In this case, the bug might actually be in the isomorphism detection, because as far as I can tell those two graphs _should_ be isomorphic.

jeenbroekstra: Then again, what we're actually after here is no loss of information of course. I'll take a closer look at the writer, see why this node gets inlined when it's referred to more than once.

jeenbroekstra:
I've pushed a partial fix that addresses the simplified use case - unfortunately the more complex one still fails. The crux is probably more generally the list handling (which is a separate code path from how other statements with blank nodes are inlined, hence this being a corner case).

hmottestad:
That's great @jeenbroekstra . I'm glad I kept the more complex one, but it's really hard to debug so only really useful as a validation and not for figuring out what's going on.

jeenbroekstra:
I think we can isolate further by creating a few more small list-specific corner cases. We'll keep the original test until we're happy the corner cases cover all the problems.

core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleWriterTest.java
```diff
@@ -194,4 +194,196 @@ public void testNoBuffering() throws Exception {
...
+	@Test
+	public void anotherBnodeTestReduced2() throws Exception {
+		String data = "@prefix ex:    <http://example.com/ns#> .\n" +
+				"@prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .\n" +
+				"\n" +
+				"ex:ex1    ex:list  [ rdf:first  [  ] ;\n" +
+				"                             rdf:rest   rdf:nil ;\n" +
+				"                             ex:abc   ex:def\n" +
+				"                           ]\n" +
```
jeenbroekstra, line 367: I am not sure this one really counts as a bug, to be honest - and even if it does it is a _very_ exotic corner case. 

This case is an invalid collection structure in the sense that it links a non-list related poperty (`ex:abc`) to a bnode that that is directly part of the list structure itself. We are not so much inlining that blank node as just removing it, because the list gets converted to a syntactical shortform where the blank nodes and (rdf:first/rest properties) are just erased.

The actual output for this is:

```
@prefix ex: <http://example.com/ns#> .

ex:ex1 ex:list ([]) .
[] ex:abc ex:def .
```

In other words, one list with a single item (a blank node), and a statement with a blank node subject. The fact that in the original file that blank node was the _same_ blank node as the one used in the internal collection structure is lost, but that does not make the output incorrect. There's also a strong whiff of "what the heck are you even _doing_ using a collection blank node like this?!"

We could potentially handle this with a separate test to check if the collection is well-formed before serializing it, but it's going to be costly.

hmottestad: I know it's pretty funky. I have no idea how it got produced to be honest. I still think we should consider it a bug though.

hmottestad: I think the output is incorrect. The pretty printer and bnode inliner should return the same data if you first output n-triples, then pretty print and inline it, then convert it to n-triples again.

hmottestad:
I would like to merge this, since we at least have a fix for one of the edge cases. The others I have `@ignored`.

jeenbroekstra:
@hmottestad I may spend another hour or so today on a fix. I'll merge when I'm done either way.

core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleWriterTest.java
jeenbroekstra: > I think the output is incorrect. The pretty printer and bnode inliner should return the same data if you first output n-triples, then pretty print and inline it, then convert it to n-triples again

My point is that effectively, it does.

jeenbroekstra:
I figured out that the original big failing test is also caused by what is effectively a malformed list structure: there are places where list nodes have _multiple_ rdf:rest properties, which is not valid. 

I've pushed a fix that does an extensive check before attempting to use list shorthand syntax, to make sure the list is not malformed. If it is malformed, it falls back to "normal" bnode inlining, not using list shorthand. This seems to address all corner cases I could come up with in additional tests, as well as the original large test case.

hmottestad:
That's awesome Jeen. I'm going to try to figure out where I was producing the malformed lists from.

Issue #2510, opened 2020-09-17
Cache dependencies for our github actions


jeenbroekstra:
See https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

PR #2512, opened 2020-09-17
GH-2510 bump version of cache action and use in status CI jobs as well

jeenbroekstra:
GitHub issue resolved: #2510 

- bump to new version of cache action
- add config to all build jobs, not just pr verify

This may need some tweaking / testing. 

---- 

### DIFF
--- a/.github/workflows/develop-status.yml
+++ b/.github/workflows/develop-status.yml
@@ -19,6 +19,13 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: ${{ matrix.jdk }}
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
     - name: Build
       run: mvn -B clean install -Pquick,\!formatting
     - name: Compliance tests
--- a/.github/workflows/develop-status.yml
+++ b/.github/workflows/develop-status.yml
@@ -23,9 +23,9 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ~/.m2/repository
-        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        key: ${{ runner.os }}-jdk${{ matrix.jdk }}-maven-${{ hashFiles('**/pom.xml') }}
         restore-keys: |
-          ${{ runner.os }}-maven-
+          ${{ runner.os }}-jdk${{ matrix.jdk }}-maven-
     - name: Build
       run: mvn -B clean install -Pquick,\!formatting
     - name: Compliance tests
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Looks like the setup works. But then I don't quite see why the older setup didn't also work. Unless it was to do with the fact that it was only use on the pr verify jobs.

.github/workflows/develop-status.yml
```diff
@@ -19,6 +19,13 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: ${{ matrix.jdk }}
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
```
hmottestad, line 28: What is the difference between key and restore key? Are they supposed to be different?

hmottestad:
Screenshot 2020-09-17 at 07 34 49

Does this mean that it never overwrites the cache? Or does it overwrite the cache?

Do you know if the caching step is skipped if the build fails?

jeenbroekstra:
> Does this mean that it never overwrites the cache? Or does it overwrite the cache?

I think that error happens because we have two parallel builds on the same PR, who are therefore using the same cache entry key. This error happens on the java 14 build because the other build (1.8) already pushed a key. Perhaps we can adjust the key to make it specific to each build. 

> Do you know if the caching step is skipped if the build fails?

No, I don't know.

.github/workflows/develop-status.yml
jeenbroekstra: See https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows . I just copy-pasted this setup. As I inderstand it the restore keys are optional, and can be used as alternative lookup keys if the main key gives no hit.

jeenbroekstra:
@hmottestad feel free to play with this a little bit as well btw, I am not able to spend more time on this today or tomorrow I'm afraid.

hmottestad:
Will do. Thanks for putting this up here :)

jeenbroekstra:
@hmottestad I think I have an idea for making that error you saw above go away. Going to push a small fix.

jeenbroekstra:
Ok so that answers that question: if the build fails, the post cache update step is skipped.

jeenbroekstra (commented):
I'm happy with this now - looks like the adapted cache key make the system more robust.

hmottestad:
I triggered a re-run of the ci and I can see that it's not downloading the cached files. So as long as we do not update the pom.xml files it should use the cache.

hmottestad (approved):

jeenbroekstra:
You sure? Looks to me as if it does download the cache.

hmottestad:
Hahahaha. I meant that it does download the cache. Not sure where the "not" came from :P Sorry for the confusion.

PR #2513, opened 2020-09-17
GH-2489 Indexing and searching over geospatial data using lucene

AlyHdr:
GitHub issue resolved: #2489 

With the current implementation in rdf4j, the lucene index have had some troubles in indexing and searching over geospatial data. After trying some modifications as presented in the issue #2489, no good results came out. So in this pull request the new way to index data introduced by Lucene is integrated and has been tested on very huge geo-shapes, the indexing works much much faster and it doesn't fail at all, adding to that the very fast search capability over such data is attained.

---- 

### DIFF
--- a/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
+++ b/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
@@ -8,15 +8,29 @@
 package org.eclipse.rdf4j.sail.lucene;
 
 import java.text.ParseException;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 import org.apache.lucene.document.Document;
+import org.apache.lucene.document.Field;
+import org.apache.lucene.document.LatLonPoint;
+import org.apache.lucene.document.LatLonShape;
+import org.apache.lucene.geo.Line;
+import org.apache.lucene.geo.Polygon;
+import org.apache.lucene.geo.SimpleWKTShapeParser;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.spatial.SpatialStrategy;
+import org.apache.lucene.spatial.spatial4j.Geo3dSpatialContextFactory;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.spatial4j.context.SpatialContext;
+import org.locationtech.spatial4j.context.SpatialContextFactory;
+import org.locationtech.spatial4j.context.jts.JtsSpatialContext;
+import org.locationtech.spatial4j.context.jts.JtsSpatialContextFactory;
+import org.locationtech.spatial4j.io.GeoJSONWriter;
+import org.locationtech.spatial4j.io.ShapeReader;
+import org.locationtech.spatial4j.io.WKTReader;
+import org.locationtech.spatial4j.shape.Point;
 import org.locationtech.spatial4j.shape.Shape;
+import org.locationtech.spatial4j.shape.jts.JtsGeometry;
 
 import com.google.common.base.Function;
 
@@ -140,13 +154,33 @@ public class LuceneDocument implements SearchDocument {
 	public void addGeoProperty(String field, String value) {
 		LuceneIndex.addStoredOnlyPredicateField(field, value, doc);
 		try {
-			SpatialStrategy geoStrategy = geoStrategyMapper.apply(field);
-			Shape shape = geoStrategy.getSpatialContext().readShapeFromWkt(value);
-			for (IndexableField f : geoStrategy.createIndexableFields(shape)) {
-				doc.add(f);
+			String wkt = value;
+			// wkt = wkt.replace("\"", "").replace("^^<http://www.opengis.net/ont/geosparql#wktLiteral>", "");
+			// System.out.println(wkt);
+			Object shape = SimpleWKTShapeParser.parse(wkt);
+			if (shape instanceof Polygon[]) {
+				for (Polygon p : (Polygon[]) shape) {
+					for (Field f : LatLonShape.createIndexableFields(field, p)) {
+						doc.add(f);
+					}
+				}
+			} else if (shape instanceof Polygon) {
+				for (Field f : LatLonShape.createIndexableFields(field, (Polygon) shape)) {
+					doc.add(f);
+				}
+			} else if (shape instanceof Line) {
+				for (Field f : LatLonShape.createIndexableFields(field, (Line) shape)) {
+					doc.add(f);
+				}
+			} else if (shape instanceof double[]) {
+				double point[] = (double[]) shape;
+				// System.out.println(point[0]+" "+point[1]);
+				doc.add(new LatLonPoint(field, point[1], point[0]));
+			} else {
+				throw new IllegalArgumentException("Geometry is not supported");
 			}
-		} catch (ParseException e) {
-			// ignore
+		} catch (Exception e) {
+			e.printStackTrace();
 		}
 	}
 }
--- a/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
+++ b/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
@@ -7,36 +7,32 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.lucene;
 
+import java.io.IOException;
 import java.text.ParseException;
-import java.util.*;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
-import org.apache.lucene.document.Document;
-import org.apache.lucene.document.Field;
-import org.apache.lucene.document.LatLonPoint;
-import org.apache.lucene.document.LatLonShape;
+import org.apache.lucene.document.*;
 import org.apache.lucene.geo.Line;
 import org.apache.lucene.geo.Polygon;
+import org.apache.lucene.geo.Rectangle;
 import org.apache.lucene.geo.SimpleWKTShapeParser;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.spatial.SpatialStrategy;
-import org.apache.lucene.spatial.spatial4j.Geo3dSpatialContextFactory;
-import org.locationtech.jts.geom.Coordinate;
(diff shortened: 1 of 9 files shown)


jeenbroekstra:
Thanks for providing a fix @AlyHdr! I'll do a review as soon as possible, but could you also take care that you've signed the ECA and have signed off _all_ your commits? We will not able to accept your contribution, otherwise. Have a look at the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md) for more details.

Tip: the simplest way to make sure your existing commits are signed off is to [squash your commits using the `--signoff` flag](https://rdf4j.org/documentation/developer/squashing/) and then do a force-push. For new commits, just use the `-s` flag.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
```diff
@@ -140,13 +154,33 @@ public boolean hasProperty(String fieldName, String value) {
...
+			} else if (shape instanceof double[]) {
+				double point[] = (double[]) shape;
+				// System.out.println(point[0]+" "+point[1]);
+				doc.add(new LatLonPoint(field, point[1], point[0]));
+			} else {
+				throw new IllegalArgumentException("Geometry is not supported");
 			}
-		} catch (ParseException e) {
-			// ignore
+		} catch (Exception e) {
```
jeenbroekstra, line 182: Let's be specific about the exceptions we're catching here:

     catch (ParseException | IllegalArgumentException e) {

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
```diff
@@ -8,15 +8,29 @@
 package org.eclipse.rdf4j.sail.lucene;
 
 import java.text.ParseException;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
jeenbroekstra, line 11: A minor thing, but: our [code conventions](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#code-formatting) don't use wildcard imports, could you roll this back here (and in other places where this has been changed)?

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
```diff
@@ -140,13 +154,33 @@ public boolean hasProperty(String fieldName, String value) {
...
+			} else if (shape instanceof Line) {
+				for (Field f : LatLonShape.createIndexableFields(field, (Line) shape)) {
+					doc.add(f);
+				}
+			} else if (shape instanceof double[]) {
+				double point[] = (double[]) shape;
+				// System.out.println(point[0]+" "+point[1]);
+				doc.add(new LatLonPoint(field, point[1], point[0]));
+			} else {
+				throw new IllegalArgumentException("Geometry is not supported");
```
jeenbroekstra, line 180: Could we add the shape object to the exception message? Not sure it has a readable string representation but on the off chance, it could help pinpoint the error.

jeenbroekstra: More generally: I don't know the geospatial support well enough to really follow, but I'm a little uncomfortable to have this if..else construction in our code base. Here, we make a choice to just support Polygon, Line, and coordinate tuples as possible Shapes, and everything else we just throw an exception. Are there no other possible shapes that could come from processing the WKT literal that could be indexed by Lucene? For example what about multiline shapes, or rectangles/envelopes, or more generally Geometry collections?

core/sail/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/examples/LuceneSailExample.java
```diff
@@ -69,45 +75,35 @@ public static void createSimple() throws Exception {
...
+			/*
+			 * String queryString = "PREFIX geo: <http://www.opengis.net/ont/geosparql#>\n" +
+			 * "PREFIX geof: <http://www.opengis.net/def/function/geosparql/>\n" +
+			 * "PREFIX uom: <http://www.opengis.net/def/uom/OGC/1.0/>\n" + "PREFIX ex: <http://example.org/>\n" +
+			 * "SELECT *\n" + "WHERE {\n" + "  ?lmA a ex:Landmark ;\n" +
+			 * "       geo:hasGeometry [ geo:asWKT ?coord1 ].\n" + "  ?lmB a ex:Landmark ;\n" +
+			 * "       geo:hasGeometry [ geo:asWKT ?coord2 ].\n" +
+			 * "  BIND((geof:distance(?coord1, ?coord2, uom:metre)/1000) as ?dist) .\n" +
+			 * "  FILTER (str(?lmA) < str(?lmB))\n" + "}";
+			 */
```
jeenbroekstra, line 105: This looks like you've used the example for your own tests - which is fine of course, but I think it needs some cleaning up.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneIndex.java
```diff
@@ -753,11 +751,15 @@ public DocumentScore apply(ScoreDoc doc) {
 		SpatialStrategy strategy = getSpatialStrategyMapper().apply(geoField);
 		final Shape boundingCircle = strategy.getSpatialContext().getShapeFactory().circle(p, degs);
 		Query q = strategy.makeQuery(new SpatialArgs(SpatialOperation.Intersects, boundingCircle));
+
+		// use LatLonPoint for distance query after indexing it with the same data structure
+
+		q = LatLonPoint.newDistanceQuery(geoField, p.getY(), p.getX(), distance);
```
jeenbroekstra, line 757: This brings up a point: are these changes compatible with existing indexes? What would happen if you ran this code on a Lucene index that was generated with an older version of RDF4J?

jeenbroekstra (commented):
Thanks for putting up this PR! I like the general idea of using Lucene's own WKTShapeParser and specific geospatial classes to make this more streamlined (the spatial4j wkt parser was a bit of a workaround in the first place, if I remember correctly).

Couple of inline remarks and questions.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneDocument.java
AlyHdr: Yes you are totally right, Lucene could parse all those types:("MULTIPOLYGON","POLYGON","POINT","MULTIPOINT","LINESTRING","MULTILINESTRING","ENVELOPE","GEOMETRYCOLLECTION"). 
I can support those extra shapes at index time but what would you suggest to remove this if else construction ? Is there a specific style you prefer so that I can follow ?

jeenbroekstra: I had a closer look at how createIndexableFields works and I guess we're stuck with the approach. It's somewhat inevitable since the output of SimpleWKTShapeParser is an Object - so the core of the problem is not in our code but in how that parser's api is designed. 

That's a long-winded way of saying that I'm fine with sticking with the if-else construction :)

AlyHdr: At some point this logic has to be done somewhere, in the current implementation in spatial4j, it was done by inherting the Shape class to any other geometry and at index time the indexer wouldn't care about the type. But with the Lucene parser and the new indexing scenario there should some logic because (shapes are not indexed with points.. etc). I have changed the code again based on your demand to cover all the available WKT geomerties, and even at search time I did some modifications for the new parser,indexer and searcher.. I will commit the modifications soon.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneIndex.java
AlyHdr: @jeenbroekstra  In this scenario it would not work and basically it is better to re-index things in Lucene because in the current implementation it was failing to index all other geometries - check the issue #2489 . Adding to this, that I tried to index some polygons using the current indexing algorithm after doing some modifications and it took a very long time while using this new algorithm it is much faster. That is to say that this modification is worth it and if you would like to have a turnaround to still support old index at search time, one have to figure out the shape of the index and then choose to search using the old or the new algorithm which is something not so practical.

core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
```diff
@@ -71,35 +70,34 @@ public void testConversionRoundtrip() {
 
 	}
 
-	@Test
-	public void testNonWellformedCollection() {
+	@Test(expected = ModelException.class)
+	public void testNonWellformedCollection_MissingTerminator() {
```
jeenbroekstra, line 74: Not sure why this shows up in your PR, as these changes are already present on the master branch. Possibly caused by your fork being out of sync with the main repo. Puzzling, but not a major problem, it will solve itself when we merge.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneIndex.java
jeenbroekstra: Oh I'm not disputing the usefulness of this fix, the reason I asked is that I wanted to know if we will need to put anything in the release notes about updating existing indexes. Which, apparently, we do, so I'll make a note of that.

core/model/src/test/java/org/eclipse/rdf4j/model/util/RDFCollectionsTest.java
jeenbroekstra: Apart from this, by the way, I think the fix now looks really good, thanks!

AlyHdr: Ok sounds good ! Thank you for your collaboration and I fixed this it was my fault ..

jeenbroekstra (approved):
LGTM. @AlyHdr can I have the name by which you wish to be added to the acknowledgements in the release notes?

AlyHdr:
@jeenbroekstra the name is: Ali Haidar. Thank you !

Issue #2083, opened 2020-04-10
SHACL - create new AST that supports the full spectrum of SHACL


hmottestad:
Todays AST does not extend to support sh:node among others. To support this I think it's best to create a new AST instead of trying to extend the current implementation. At the same time we should leverage the SHACL RDFS (schema) to reasoning.

hmottestad:
Example of something to support:

```
ex:Shape1
      a sh:NodeShape  ;
      sh:targetNode ex:CustomType;

      sh:property [
             sh:path [sh:inversePath dct:type ];
             sh:property [
               sh:path rdfs:label;
               sh:minCount 1 ;
               sh:maxCount 1 ;
             ]

      ] .
```

This should end up as.

Target: ex:CustomType [sh:inversePath dct:type ]
Path: rdfs:label
Rule: sh:minCount = 1

Target: ex:CustomType [sh:inversePath dct:type ]
Path: rdfs:label
Rule: sh:maxCount = 1

hmottestad:
This could become a bit trickier:

```
@prefix dash: <http://datashapes.org/dash#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix schema: <http://schema.org/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix ex: <http://example.org/ns#> .

ex:PersonShape
      a sh:NodeShape  ;
      sh:targetNode ex:Norway;

      sh:property [
             sh:path [sh:inversePath ex:nationality ];
             sh:or (
             [
              sh:property [
               sh:path ex:norwegianID ;
               sh:minCount 1 ;
               sh:maxCount 1 ;
             ]
            ]

            [
              sh:property [
               sh:path ex:nationalID ;
               sh:minCount 1 ;
               sh:maxCount 1 ;
             ]
            ]

            )

      ] .
```


Or something like:

```
@prefix dash: <http://datashapes.org/dash#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix schema: <http://schema.org/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
[...]

PR #2539, opened 2020-09-27
GH-2083 New SHACL Abstract Syntax Tree

hmottestad:
GitHub issue resolved: #GH-2083 

 - New AST that supports the entire spectrum of SHACL
     - export SHACL shapes back to RDF
 - Support nested `sh:property`
 - Support `sh:node`

---- 


hmottestad:
This PR tracks my work on the new AST for the ShaclSail. There are still a few features missing, but overall we are in a good state with most things functioning correctly.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
```diff
@@ -42,6 +45,11 @@ public Var(String name, Value value) {
 		setValue(value);
 	}
 
+	public Var(Value value) {
+		this(UUID.randomUUID() + "_anon");
```
jeenbroekstra, line 49: Nice idea to create an anonymous variable this way, but this will cause syntactically illegal variable names - in SPARQL, variable names cannot contain '-'. See createAnonVar in the SPARQL TupleExpBuilder, which use this to create the name: 
    
     new Var("_anon_" + UUID.randomUUID().toString().replaceAll("-", "_")); 

You probably also need to set the `isAnonymous` flag.

Might be a good idea to handle this in one place (here in the constructor) and refactor the TupleExprBuilder and other places where we create anonymous vars. We've had problems in the past with the naming pattern not being consistent, so DRYing this up is a good idea.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ConnectionsGroup.java
```diff
@@ -14,11 +14,10 @@
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
-import org.eclipse.rdf4j.common.transaction.TransactionSetting;
 import org.eclipse.rdf4j.sail.Sail;
 import org.eclipse.rdf4j.sail.SailConnection;
-import org.eclipse.rdf4j.sail.shacl.planNodes.BufferedSplitter;
-import org.eclipse.rdf4j.sail.shacl.planNodes.PlanNode;
+import org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree.planNodes.BufferedSplitter;
```
jeenbroekstra, line 19: Two things:

1. let's call the package 'ast' rather than 'abstractsyntaxtree', to be consistent with how we do this for SPARQL and SeRQL as well.
2. I thought part of the idea was to have an AST that is not "just" usable by the SAIL, but also potentially for other SHACL implementations. If that is still the case it should be put in a in a more generic package: `org.eclipse.rdf4j.shacl.ast`, and ideally a separate module as well.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/ValidationApproach.java
```diff
@@ -0,0 +1,16 @@
+package org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree;
+
+public enum ValidationApproach {
+
+	Transactional,
```
jeenbroekstra, line 5: ObNitpick, this should be all caps.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/constraintcomponents/AndConstraintComponent.java
```diff
@@ -0,0 +1,128 @@
...
+import org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree.planNodes.PlanNodeProvider;
+import org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree.planNodes.UnionNode;
+import org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree.planNodes.Unique;
+import org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree.targets.TargetChain;
+
+public class AndConstraintComponent extends AbstractConstraintComponent {
+	List<Shape> and;
+
+	public AndConstraintComponent(Resource id, RepositoryConnection connection,
+			Cache cache, ShaclSail shaclSail) {
```
jeenbroekstra, line 33: Ah, this suggests that this AST is still tightly coupled to the ShaclSail, in which case separating it out to a generic package probably makes little sense.

jeenbroekstra (commented):
Good to see so much progress on this! Nowhere near a complete review on my part but some initial remarks inline.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ConnectionsGroup.java
hmottestad: It's called abstract syntax tree so that it can exist together with the old AST. We can rename it before merging.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/constraintcomponents/AndConstraintComponent.java
hmottestad: It's not that tightly coupled. The SHACL sail is only used to get hold of some user settings. Could make an interface for that instead.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
hmottestad: I'm not sure if I use this code anymore, will have to take another look. Might be something left over from testing.

hmottestad:
Thanks Jeen. There is still a lot of cleanup needed before we can merge this. I've posted the PR since I'm  getting close to being feature complete and will start to encourage user testing very soon. 

Was also wondering if we could do an alpha release of this feature? Not sure how we would accomplish it, but would be nice to get some feedback from users without them having to checkout and build the code.

hmottestad:
TODO-list
 - [ ] sh:deactivated
 - [ ] target shapes
 - [ ] remove old AST and old plan nodes
 - [ ] decouple new AST from ShaclSail

jeenbroekstra:
> Thanks Jeen. There is still a lot of cleanup needed before we can merge this. I've posted the PR since I'm getting close to being feature complete and will start to encourage user testing very soon.

No worries, sounds good.

> Was also wondering if we could do an alpha release of this feature? Not sure how we would accomplish it, but would be nice to get some feedback from users without them having to checkout and build the code.

Well, there's always the snapshots of course, but if you want a tagged build on maven central we could do a milestone build at some point. It's not something we want to use too lightly, as I'd rather not have 10 different "alpha releases" floating about on central, but if we are a bit careful I don't see a problem with it.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/planNodes/Formatter.java
```diff
@@ -0,0 +1,50 @@
...
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree.planNodes;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.model.vocabulary.SHACL;
+
+public class Formatter {
```
rdstn, line 16: This could be moved to the util class?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/planNodes/LanguageInFilter.java
```diff
@@ -0,0 +1,70 @@
...
+		boolean languageMatches = language.map(String::toLowerCase).filter(lowerCaseLanguageIn::contains).isPresent();
+		if (languageMatches) {
+			return true;
+		}
+
+		// test according to BCP47
+		String langTag = language.get();
+
+		for (String languageRange : languageRanges) {
+			if (Literals.langMatches(langTag, languageRange)) {
```
rdstn, line 54: Looks like we have some subtagging, which is good.

rdstn:
Looking great, but it's a lot to read through! I'll try to add some meaningful comments over the next few days.

A tagged build when the feature is considered "stable" would be good. However, it is easy enough to build from source.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/Cache.java
```diff
@@ -0,0 +1,24 @@
+package org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+import org.eclipse.rdf4j.model.Resource;
+
+public class Cache {
```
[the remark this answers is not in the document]
hmottestad: I think it's just the cache used to allow parsing of SHACL with recursion. We don't allow for executing SHACL with recursion though, but parsing is a first step. The actual cache, which is also available in the current SHACL implementation is broken at the moment.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/ShaclUnsupportedException.java
```diff
@@ -0,0 +1,23 @@
+package org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree;
+
+public class ShaclUnsupportedException extends UnsupportedOperationException {
```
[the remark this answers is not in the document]
hmottestad: I'm not very consisten with this. I'm unsure if I want to throw and exception in general for unsupported things or just simply log a message.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/ValidationApproach.java
```diff
@@ -0,0 +1,18 @@
+package org.eclipse.rdf4j.sail.shacl.abstractsyntaxtree;
+
+public enum ValidationApproach {
```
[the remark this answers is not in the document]
hmottestad: When generating the validation plan we ask what the preferred validation approach is. Things aren't working that well at the moment so it's mostly disabled. But the code is here so we have somewhere to start.

rdstn:
This is supposed to have validations for `sh:node` already implemented, correct? Because when testing, I'm not seeing it get picked up. The following SHACL shapes:

```
vocsh:Droid
	a sh:NodeShape ;
	sh:targetClass voc:Droid ;
	sh:property [
		sh:path voc:owner ; 
		sh:maxCount 1 ; 
		sh:node vocsh:HumanRef ; 
	] .

vocsh:HumanRef
	a sh:NodeShape ;
	sh:targetClass voc:Human.
```
Do not produce any validation errors for:
```
    <https://starwars.org/resource/droid/word/> a voc:Droid ;
                                                                       voc:owner voc:Human .
    voc:Human a voc:Wrong .
```

hmottestad:
Maybe it should be 

```
vocsh:HumanRef
	a sh:NodeShape ;
	sh:class voc:Human.
```

hmottestad:
There is one thing I need to look at. It's when we have nested `sh:property` and the data doesn't nest is a tree like fashion.

Eg.

```
ex:a ex:prop1 ex:b.
ex:b ex:prop2 ex:d.
ex:b ex:prop2 ex:e.
ex:e ex:prop3 ex:f.
```

So we can see that the chain `prop1 -> prop2 ->prop3` has two statements for `ex:prop2` even though there is only one value at the end of the chain `ex:f`.

I'm curious if this somehow break our join algorithms that assume that the left argument is alway unique.

rdstn:
> 
> 
> Maybe it should be
> 
> ```
> vocsh:HumanRef
> 	a sh:NodeShape ;
> 	sh:class voc:Human.
> ```

Yes, that works, provided that we use `rdf:type` (or one of its subclasses) as the predicate to identify `HumanRef` by. However, once we have ported the advanced targeting statements over to the new AST, this would break. Do we plan to do that?

hmottestad:
I'm planning on porting the advanced targeting over yes! 

If you want to limit on something other than `rdf:type` you could try something like this:

```
vocsh:Droid
	a sh:NodeShape ;
	sh:targetClass voc:Droid ;
	sh:property [
		sh:path voc:owner ; 
		sh:maxCount 1 ; 
		sh:property vocsh:HumanRef ; 
	] .

vocsh:HumanRef
	a sh:PropertyShape ;
        sh:path voc:type ;
	sh:in (voc:Human).
```

You would loose the basic RDFS support that `sh:class` provides.

jeenbroekstra:
I realize this is still a work in progress, so no urgency on this, but I think we _definitely_ need a package-level description on the SHACL sail in general, and the AST in particular, that explains the basic design ideas behind how this all works. There's an awful lot of code and classes that all talk to each other, and having some high-level understanding of the design principles will go a long way towards being able to understand (let alone help maintain) the code.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
```diff
@@ -42,6 +45,12 @@ public Var(String name, Value value) {
 		setValue(value);
 	}
 
+	public Var(Value value) {
+		this(UUID.randomUUID().toString().replace("-", "") + "_anon");
```
jeenbroekstra, line 49: Let's make sure this uses the same algorithm as `oer.query.parser.sparql.TupleExprBuilder#createAnonVar`:

```
"_anon_" + UUID.randomUUID().toString().replaceAll("-", "_")
```

In fact if we add this as a convenience thing here in the `Var` class, let's also make sure `TupleExprBuilder` reuses it.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/ShaclProperties.java
```diff
@@ -48,39 +61,70 @@
...
-
-		try (Stream<Statement> stream = connection.getStatements(propertyShapeId, null, null).stream()) {
+	public ShaclProperties(Resource id, RepositoryConnection connection) {
+		this.id = id;
+		try (Stream<Statement> stream = connection.getStatements(id, null, null, true).stream()) {
 			stream.forEach(statement -> {
 				String predicate = statement.getPredicate().toString();
 				Value object = statement.getObject();
 				switch (predicate) {
+				case "http://www.w3.org/1999/02/22-rdf-syntax-ns#type":
```
jeenbroekstra, line 100: Can we replace this massive switch statement with a simple strategy pattern? Bit more coding but in the end probably a lot more readable and maintainable.

Also: can we replace the hardcoded strings with constant references?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/ShaclProperties.java
```diff
@@ -319,10 +404,58 @@ public boolean isUniqueLang() {
...
-	public Value getHasValue() {
+	public Resource getId() {
+		return id;
+	}
+
+	public IRI getType() {
+		return type;
+	}
+
+	public List<Literal> getMessage() {
```
jeenbroekstra, line 415: Given that it returns a list, rename to `getMessages()`?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/ShaclProperties.java
```diff
@@ -319,10 +404,58 @@ public boolean isUniqueLang() {
...
+		return message;
+	}
+
+	public List<Resource> getProperty() {
```
jeenbroekstra, line 419: Same here: `getProperties()`?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/HelperTool.java
```diff
@@ -0,0 +1,85 @@
...
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.repository.RepositoryConnection;
+
+public class HelperTool {
```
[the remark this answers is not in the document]
jeenbroekstra: We already have generic utility functions for converting between java collections and rdf collections. @hmottestad any reason you're not just reusing `org.eclipse.rdf4j.model.util.RDFCollections` (or perhaps `org.eclipse.rdf4j.repository.util.Connections`) here?

jeenbroekstra (commented):
I did another cursory review (got about a third of the way down this time I think). See inline comments. Also a general question: I notice this PR contains both changes to the existing AST package, and also introduces a new abstractsyntaxtree package. I thought the idea was that the new package would replace the old code? If so why still make changes to the old code?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/HelperTool.java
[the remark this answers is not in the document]
hmottestad: Doesn't supporting generics.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/ShaclProperties.java
hmottestad: This code was written for performance. It was originally a bunch of getStatement calls, but those ended up being considerably slower. This class essentially just converts the raw graph in a java object to make things easier to work with. Hence the naming of the variables and getters. They all reflect the rdf property.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
hmottestad: I'm actually going to redo some of the use of Var and StatementPattern because things aren't working as I want with the last bit of the support for rsx:targetShape. I was actually considering making some more constructors in Var and also a static constant anonymous instance. Also quite annoying that there is no documentation in the Var class!

hmottestad:
> I did another cursory review (got about a third of the way down this time I think). See inline comments. Also a general question: I notice this PR contains both changes to the existing AST package, and also introduces a new abstractsyntaxtree package. I thought the idea was that the new package would replace the old code? If so why still make changes to the old code?

Those changes were just some minor cleanup before I ported over the code into the new abstract syntax tree.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/HelperTool.java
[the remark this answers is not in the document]
jeenbroekstra: > Doesn't supporting generics.

Ok, if you need generics, how about we add this to the RDFCollections util instead?

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
jeenbroekstra: (I think github might be having some issues, because some of my comments are showing up when I look at the diff, but not here - forgive me if I appear to say things twice)

A few more constructors to make (anonymous) Var creation sounds fine. Not sure about a static constant, at least not for general inside the Var class itself - that sounds problematic in the sense that even if it's anonymous, it still has identity. You can't just reuse the same anonymous var everywhere.

As for documentation: I agree that it could be better, but `Var` is a relatively simple container class, anything specific you're missing specs on?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/abstractsyntaxtree/HelperTool.java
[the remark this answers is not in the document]
hmottestad: That's essentially what Rado's comment boils down to.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
hmottestad: Would be nice with some examples of combinations that are legal and what they match to in the query. And simple definitions for all the fields.

hmottestad: I presume that anonymous means that it doesn't have a variable name (which makes me wonder why null doesn't suffice). However it can still have a value. Var with variable name and constant value is probably equivalent to a variable bound at query time.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/ShaclProperties.java
jeenbroekstra: I'm not convinced about the performance argument to be honest - I don't doubt that this code performs significantly better than what was previously there, but I do doubt that using hardcoded strings instead of reusing `RDF.TYPE.stringValue()`  is a significant factor in that (or indeed that a case statement is so much quicker than a simple dictionary lookup / strategy pattern impl).  However, I also appreciate that the focus should be on getting this in a mergable state, so maybe this isn't the highest priority.

I would suggest to liberally comment and javadoc the code though, to make it clear what is going on (in particular javadoc on the getters to say which RDF property they correspond to might be useful).

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
jeenbroekstra: To explain: anonymous vars are vars that have no specific _user-defined_ name in the query. In SPARQL this happens, for example, when a query involves a blank node:

```
 SELECT * WHERE {
      [] rdf:type :Foo ;
         rdfs:label ?label .
 }
```

The `[]` in the above query is represented as an anonymous variable in the algebra. However it still needs a separate identity because (as you can see) it needs to appear in two statement patterns - we are executing a join on the anonymous var. And there may be more than one anonymous var in the same query (a second blank node), which means we can't just reuse the same Var instance everywhere.  

Another point where they occur in SPARQL queries is in path expressions. For example this SPARQL query:

```
?start :p1/p2 ?end.
```

internally gets represented by two statement patterns, where the object-var of the first is the subject-var of the second. Again, the variable is anonymous, but its identity is still important as the 'glue-component' in the path expression.

I agree though that there is some ambiguity and lack of documentation in how the Var class handles this. Having to set the the `isAnonymous` flag _in addition_ to assigning it a generated name sounds superfluous.

jeenbroekstra: As for why null doesn't suffice: we need to be able to serialize the algebra. And the internal mechanics of the evaluation engine rely on being able to set values for vars by means of using the variable name as a key in a Binding(Set) object.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/ShaclProperties.java
hmottestad: It's just a java limitation. Can't use SHACL.PROPERTY.toString() because it's not bound at compile time.

hmottestad:
Currently answering comments is about the amount of time I have. I know that there is a big need for documenting the SHACL sail. The most important things that need to be documented are:

 - the overall flow and architecture
 - how the validation plan system works
 - each individual validation plan

jeenbroekstra:
> Currently answering comments is about the amount of time I have. I know that there is a big need for documenting the SHACL sail. The most important things that need to be documented are:
> 
> * the overall flow and architecture
> * how the validation plan system works
> * each individual validation plan

No rush, I'm not providing feedback expecting you to immediately fix it.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
hmottestad: I think I will migrate away from Var and StatementPattern then since my usecase is similar but sufficiently different. I'm using it as a two step process for retrieving the minimal amount of data from the backing store. Essentially I want to express things like:

For a query:
?a rdf:type ex:Company.
?b ex:subsidiary ?c. 

Bound to whatever matches the patterns:
- ?a rdf:type ex:company
- ?b ex:subsidiary ?c

But sometimes it gets more complicated, for some of the SHACL components and I need a pattern like:
- ?b null ?c
Or
- ?a rdf:type null

For the same query. 

These patterns I run against the changeset for the transaction and then use the results to bind the variables in the query to retrieve less data when validating. 

As an example. If I say that every person that has an address they must have max 1 street name. Then we only need to retrieve people and their street names if either a street name is added or a person is added. We can ignore any removals. And we don't need to revalidate all people and their street names, only ones that could actually now violate the rule based on the changes in the transaction.

hmottestad:
I'm hoping that I'm passing all the tests now.

hmottestad:
```
                            log on this level

                           +---------+
                           |         |                        or potentially log on this level
                           |         |
     +--------------------->         +---------------------------------+
     |                     |         |                                 |
     |                     |         |                                 |
     |                     +---------+                                 |
     |                                                                 |
     |                                                            +----v----+
     |                                                            |         |
     |                                                            |         |
     |                                                           >+         +----------------------+
     |                      +----------+                         ||         |                      |
     |                      |          |                         ||         |                      |
+----+-----+                |          |                         +----------+                      v
|          |                |          +-------------------------+                              +--+-------+
|          |                |          |                                                        |          |
|  target  +---------------->          |                                                        | failing  |
|          |                |          |                                                        | node     |
|          |                +----------+                                                        |          |
+----+-----+                                                                                    |          |
     |                                                                                          +-+--------+
     |                                                            +----------+                    ^
     |                                                            |          |                    |
     |                                   +----------------------->+          |                    |
     |                     +-----------+ |                        |          +--------------------+
     |                     |           | +                        |          |
     |                     |           |                          +----------+
     |                     |           |
     +--------------------->           |
                           |           |
                           |           |
                           +-----------+

```

hmottestad:
This is currently the last scenario that I want to reassure myself that works. Essentially where you start on one node, then traverse out to several nodes that all point to the same final node that gets validated. The question then becomes what actually gets reported in the validation report.

core/sail/shacl/src/test/resources/test-cases/hasValueIn/targetShapeOr/invalid/case2/report.ttl
```diff
@@ -4,3 +4,17 @@
...
+  <http://rdf4j.org/schema/rdf4j#truncated> false;
+  sh:result [ a sh:ValidationResult;
+      sh:focusNode ex:validPerson1;
+      sh:resultPath ex:info2;
+      sh:sourceConstraintComponent sh:MinCountConstraintComponent;
+      sh:resultSeverity sh:Violation;
+      sh:sourceShape [ a sh:PropertyShape;
+          sh:path ex:info2;
+          sh:minCount 1
+        ]
```
jeenbroekstra, line 19: I haven't seen the actual shape itself being included in a validation report like this before - it's usually just a reference (which is really annoying when the shape has a bnode identifier). Is this new for this reimplementation? And is this what the SHACL spec actually says we _should_ do? If so, I like it a lot :)

jeenbroekstra:
@hmottestad as I mentioned on the dev list, I am tentatively aiming for a 3.5 release next week - but I planned that under the impression that you would need a lot more time to get this PR in a stable shape. You seem to be making great progress again though. If I were to postpone release to, say, December 23, do you think it might be realistic to get this done in time for release? 

If so, I'll do a second milestone next week instead of a release, and we can do a proper release with this improvement in just in time before Christmas :)

core/sail/shacl/src/test/resources/test-cases/hasValueIn/targetShapeOr/invalid/case2/report.ttl
hmottestad: It's not part of the spec, but due to the way the new AST is implemented we are able to extract a much more minimal shape and add it to the report. Essentially if the shape has both minCount and maxCount but only minCount fails the validation, then maxCount is left out of the shape in the report.

hmottestad:
There is at least one backwards incompatible change, and also some of the logging has not been reimplemented. I had an idea to release 4.0.0-M1 directly after you have release 3.5.0. Do you think that that's a decent plan?

jeenbroekstra:
> There is at least one backwards incompatible change, and also some of the logging has not been reimplemented. I had an idea to release 4.0.0-M1 directly after you have release 3.5.0. Do you think that that's a decent plan?

What's the backwards incompatible change?

I'm not against a 4.0 release, but major releases normally require some planning and discussion, and dedicated development time. There is a long list of planned backward incompatible changes in the tracker that have been put on hold for the next major release, and we'll need time to prioritize them, decide what we minimally want in, and then actually _do_ them too. Major releases don't happen too often, so when we do them it's important that we include as many changes as we can in one go. 

What's your timeline, release-wise? I don't mean milestone, proper release.

A further point to consider: it's coming up to the time again where we'll need to do another release review (we're supposed to do one once a year). A major release would be a natural event for that I think, if we can plan it that way. If we start thinking about a major release now, give ourselves permission to do a few milestones, and then a final release in February, would that work?

jeenbroekstra:
Btw as for my original question: I'll take it to mean that I should just go ahead with a 3.5 release next week, _without_ the AST changes. Whether the AST changes require a 4.0 milestone/release or we can get away with a 3.6 minor is a separate discussion.

hmottestad:
isUndefinedTargetValidatesAllSubjects

That was always a non standard feature that was quickly deprecated. But it's no longer possible to support. 

I was hoping to have a release release this year.

hmottestad:
If you are happy with one or maybe two more milestone builds, and that change can be argued to not be big enough to have to trigger a major release then I would love to include the new ast in 3.5.

jeenbroekstra:
> If you are happy with one or maybe two more milestone builds, and that change can be argued to not be big enough to have to trigger a major release then I would love to include the new ast in 3.5.

Ok cool. I had a look at the feature flag you mentioned, and I guess I can live with that as a "semi-breaking" change in a minor release (as long as you keep the code so compilation doesn't fail, but it just becomes a no-op or throws some error if you try to use it). 

I do have a hard requirement for a 3.5 release this year. So I suggest we don't merge any of this for the next milestone just yet. that way any unforeseen things you run into won't block the release. 

I'll send an e-mail to the rdf4j-dev list with a proposed schedule for a 3.5 release, and a tentative plan for a 4.0 release as well (which I think we should start focusing on anyway).

hmottestad:
Since you're ok with the semi-breaking change I propose you release 3.5 next week and I follow with a 3.6-M1 directly after that.

jeenbroekstra:
> Since you're ok with the semi-breaking change I propose you release 3.5 next week and I follow with a 3.6-M1 directly after that.

Ok, I just sent a schedule to the rdfj4-dev list which proposes a milestone build next week instead of a release. But if you prefer this then yes, we can go that way as well. I'll send a followup to the list to rectify things.

jeenbroekstra:
If you have any further remarks can you ping me on gitter?

hmottestad:
@jeenbroekstra I hope it's ok I merge this into develop now and do a milestone release.

jeenbroekstra:
@hmottestad if you're confident, go for it! I think it's pretty well agreed that the next minor is all about the new SHACL AST, so you're not blocking anything.

jeenbroekstra:
Just as an aside though: I know there were several comments on the draft PR, and some outstanding "clean up" work (including package names etc). I haven't checked, maybe you've taken care of all that, but If you haven't addressed those here yet could you make sure to create followup tickets so we don't lose track?

hmottestad:
Yes, I'm confident. I know of one bug, but that only affects the validation report. Essentially it's the case outlined above, and only the first node gets reported. 

I've addressed the show stoppers. Including the use of Var and the package name. Also deleted the old AST. The new AST is also marked correctly as internal/deprecated so people don't rely on it.

jeenbroekstra:
don't forget to use 'squash and merge' :)

Issue #2544, opened 2020-09-28
Support Java 15


hmottestad:
Make sure that we both build and can run all tests on Java 15.

hmottestad:
For CQ see Jeen's comments about the build plugin CQ: https://github.com/eclipse/rdf4j/pull/2545

jeenbroekstra:
Oops, bit of a mixup here: we should never merge a PR if there are still outstanding CQs - it blocks us from doing a release.

Fortunately, it turns out that we don't need a new CQ for the bundle plugin. I asked the EMO for clarification and they say that it's a bit of a gray area for some things, but generally speaking things like the java compiler itself, maven and its core plugins etc are considered "part of the furniture", and we don't need to log CQs for them (it's not wrong to do CQs, just not required). So we're in the clear :)

PR #2545, opened 2020-09-28
GH-2544 Java 15

hmottestad:
Signed-off-by: hmottestad <hmottestad@gmail.com>

GitHub issue resolved: #2544  

Support Java 15.

---- 

### DIFF
--- a/.github/workflows/develop-status.yml
+++ b/.github/workflows/develop-status.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest    
     strategy:
       matrix:
-        jdk: [1.8, 14]
+        jdk: [1.8, 15]
 
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/master-status.yml
+++ b/.github/workflows/master-status.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest    
     strategy:
       matrix:
-        jdk: [1.8, 14]
+        jdk: [1.8, 15]
 
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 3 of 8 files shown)


hmottestad:
This is currently failing since spin uses the js script engine (nashorn?) and it was removed in java 15.

hmottestad:
@jeenbroekstra we need to bump the maven build plugin, I think I remember that we don't need to do CQs for that anymore.

hmottestad:
Looks like the merge block is stopping us from merging this because it is expecting a build for java 14. So we are going to have to disable the merge block if we don't want to build on both 8, 14 and 15!

jeenbroekstra:
> @jeenbroekstra we need to bump the maven build plugin, I think I remember that we don't need to do CQs for that anymore.

That depends - we have a single "list" CQ that covers these kinds of dependencies, usually with a version-range for each one. See https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20318. I see it's the bundle plugin you're updating, which is covered up to version 4.2.0, but you're bumping to 5.1.1, so it's not covered. 

I will get in touch with the IP team on how to log these kinds of things, as I am not sure how we are supposed to update that list CQ.

jeenbroekstra:
> Looks like the merge block is stopping us from merging this because it is expecting a build for java 14. So we are going to have to disable the merge block if we don't want to build on both 8, 14 and 15!

Ah, that is a pain. I'll ask them to update the settings so that only the 1.8 build is actually required, gives us a bit more flexibility in the future.

jeenbroekstra:
Branch protection has been adapted.

jeenbroekstra:
> This is currently failing since spin uses the js script engine (nashorn?) and it was removed in java 15.

Out of curiosity: what does the spin implementation use a javascript engine for?

hmottestad:
I think users can implement transformations of data using JS in SPIN functions. Haven't looked more into it though, so not sure. 

I know TopQuadrant are doing something similar in their SHACL implementation by using GraalVM. I assume that was due to people using the JS support in SPIN.

jeenbroekstra (approved):

Issue #2538, opened 2020-09-26
Decouple RDF model interfaces to a dedicated maven module


knoan:
The interfaces defining the RDF abstract data model (`org.eclipse.rdf4j.model`) are currently bundled in the `rdf4j-model` maven module with their simple implementations and a number of supporting utilities, totalling some 425 kb when packaged in a jar file.

Decoupling them to a dedicated module without dependencies would make it easier to port third-party interoperable libraries to limited devices supporting for instance Web of Thing applications.

The operation can be carried out without breaking changes; if there is consensus, I can submit a PR: just let's agree the artifactId/name/description for the new abstract module and how to update current name/description for `rdf4j-model`.

jeenbroekstra:
`rdf4j-model-api` is a good name for the interface module, that follows the same pattern we also use for the repository and sail apis.

I would ideally like things split up even further, for example a separate 'rdf4j-vocabularies' module would be useful, but perhaps we should leave that out of scope for now. 

I would suggest _not_ including the classes and interfaces in package `org.eclipse.rdf4j` (IsolationLevel, RDF4JException etc) in the new api package. They should probably be moved to an rdf4j-common module at some point (again though, out of scope for this).

knoan:
Initial PR submited for review. I had to manage two unforeseen dependencies:

- `Triple` is annotated as `@Experimental`
  - migrated `org.eclipse.rdf4j.common.annotation` package to `rdf4j-model-api` module
  - seems to make sense (as long as other annotations are not added), as `@Experimental/@InternalUseOnly` relate mainly to APIs
- the default implementation of `Model.setNamespace(String prefix, String name)` created `SimpleNamespace` objects
  - replaced `SimpleNamespace` with lightweight  read-only package-local `BasicNamespace` implementation

Concerning the latter point I wonder if we should:

1. migrate `org.eclipse.rdf4j.model.impl` to the `rdf4j-model-api` module
   1. seems to negate the original goal of the proposal, especially when taking into account all of the specific implementation classes related to `Model`
2. leave `Model` in the `rdf4j-model` module
   1. splitting the `org.eclipse.rdf4j.model` package across multiple modules feels quite hacky
3. provide additional lightweight implementations of  `org.eclipse.rdf4j.model` interfaces, wherever feasible directly as default methods 
   1. generates some overlap with `org.eclipse.rdf4j.model.impl.Simple*` implementations
   2. on the plus side, would provide a self-contained model definition working out of the box

[...]

PR #2546, opened 2020-09-28
GH-2538 Standalone Model API

knoan:
GitHub issue resolved: #2538

- created `rdf4j-model-api` maven module with no external dependencies
- added `rdf4j-model-api` module as a dependency to `rdf4j-model` module 
  - tweaked  `rdf4j-model` description
- removed references to `rdf4j-model` classes from javadocs in `org.eclipse.rdf4.model` interfaces
- migrated `org.eclipse.rdf4.model` test classes to `org.eclipse.rdf4.model.impl`
  - renamed `ValueFactoryTest` to `SimpleValueFactoryTest`
- added default implementation of `compareTo()` to `Namespace`
- replaced `SimpleNamespace` dependency in `Model` with lightweight  read-only package-local `BasicNamespace` implementation
- migrated `org.eclipse.rdf4j.common.annotation` package to `rdf4j-model-api` module to allow `@Experimental` annotation of `Triple` interface
  - seems to make sense as long as other annotations are not added, as `@Experimental/@InternalUseOnly` relate mainly to APIs
- migrated `org.eclipse.rdf4j.model` package to `rdf4j-model-api` module

----

 - [ x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [ x] I've added tests for the changes I made
 - [ x] I've applied [code formatting](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)
 - [x ] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change
 - [ x] every commit has been [signed off](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)


jeenbroekstra:
@knoan I've rebased this against the develop branch, since this is something we won't be releasing as a simple patch, but as part of the next minor release.

knoan:
- created `oer.model.base` package
- created abstract `Value` test suites in `oer.model` package
- created abstract `Value` classes in `oer.model.base` package
- migrated `Model.setNamespace()` default implementation to `AbstractNamespace`
- derived `Simple*` value classes from `Abstract*` companions
- Add `Simple*Test` test classes  based on abstract value test suites

barthanssens:
thanks for the PR, but it seems to fail on the code formatting of the AbstractTriple class... so could you perhaps run `mvn process-resources`  again ?

jeenbroekstra:
I went ahead and pushed code formatting fixes

core/model-api/src/main/java/org/eclipse/rdf4j/model/ValueFactory.java
```diff
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * Copyright (c) ${year} Eclipse RDF4J contributors.
```
jeenbroekstra, line 2: Your IDE forgot to substitute the number here. Also: I don't think we should replace the license header on this just because we moved the file.

core/model-api/src/main/java/org/eclipse/rdf4j/model/ValueFactory.java
```diff
@@ -30,7 +27,7 @@
 	 *
 	 * @param iri A string-representation of a IRI.
 	 * @return An object representing the IRI.
-	 * @throws IlllegalArgumentException If the supplied string does not resolve to a legal (absolute) IRI.
+	 * @throws IllegalArgumentException If the supplied string does not resolve to a legal (absolute) IRI.
```
jeenbroekstra, line 30: good catch. I had to look several times before I even realized what you had changed.

core/model/src/test/java/org/eclipse/rdf4j/model/impl/SimpleBNodeTest.java
```diff
@@ -0,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * All rights reserved. This program and the accompanying materials
```
jeenbroekstra, line 3: This is a new file I think? If so this is the wrong header.

core/model/src/test/java/org/eclipse/rdf4j/model/impl/SimpleTripleTest.java
```diff
@@ -0,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.model.impl;
+
+import org.eclipse.rdf4j.model.*;
```
jeenbroekstra, line 10: Wildcard import needs to be expanded.

core/model/pom.xml
```diff
@@ -8,13 +8,25 @@
...
 		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>rdf4j-util</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-model-api</artifactId>
+			<version>${project.version}</version>
+			<type>test-jar</type>
```
jeenbroekstra, line 27: This won't work unless the model-api module actually produces and publishes a test jar. I'm not sure yet if that is desirable - we use this approach in one other place (the rio api) but for everything else we have set up separate test suite projects instead. I'll take a look at this and if necessary refactor.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
```diff
@@ -0,0 +1,40 @@
...
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.model.base;
+
+import java.util.Objects;
+
+import org.eclipse.rdf4j.model.BNode;
+
+public abstract class AbstractBNode implements BNode {
```
jeenbroekstra, line 15: This class, and the other new abstract classes, could use a minimal bit of Javadoc. All it really needs:

> Base class for {@link BNode}, offering common functionality.

and preferable an author tag, and a `@since 3.5.0` tag as well.

jeenbroekstra (commented):
Looking good mostly. There may be some work in getting the test suite set up properly (see inline comments), and I've noticed a few minor things. But it's coming together nicely!

knoan:
> thanks for the PR, but it seems to fail on the code formatting of the AbstractTriple class... so could you perhaps run mvn process-resources again ?

My bad… still honing the workflow :-(

knoan:
> This won't work unless the model-api module actually produces and publishes a test jar. I'm not sure yet if that is desirable - we use this approach in one other place (the rio api) but for everything else we have set up separate test suite projects instead. I'll take a look at this and if necessary refactor.

The goal here is to factor all tests that could be possibly applied to values; I incorrectly assumed that all modules were configured to attach a test jar by default. Let me know if you prefer to revert the dependency in `rdf4j-model`.

knoan:
> This class, and the other new abstract classes, could use a minimal bit of Javadoc. 

Fixed, as well as other import/copyright issues.

jeenbroekstra:
I'll do a bit of work on resolving these conflicts (introduced by the merge of @hmottestad 's PR regarding sorting optimization) and if time permits will also take a look at the testsuite setup.

jeenbroekstra:
Unfortunately resolving these conflicts is not quite straightforward. For one thing the recent changes introduce a direct dependency between the Literal interface and the XSD vocabulary class (which is no problem if they're both in the same module but now they're separate). 

I'm tempted to move the vocabulary classes into the api module, or at the very least the 'core' ones, like XSD, RDF, RDFS.

knoan:
I resorted to minimal private or package-visible constants in `LiteralTest` and `AbstractLiteral`, reworking code as to require strings rather than IRIs.

jeenbroekstra:
I've merged develop into your branch and resolved the conflicts.  We'll need to review if this has actually been done correctly, but hopefully it at least compiles.

knoan:
> I've merged develop into your branch and resolved the conflicts.

Merging broke some tests in `QueryEvaluationUtilTest` and `ValueComparatorTest`

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
```diff
@@ -57,6 +58,22 @@ public static String getLabel(Optional v, String fallback) {
...
+	 *         <i>not</i> indicate that the literal has no datatype, merely that it has no cached enum representation of
+	 *         that datatype.
+	 * @since 3.5.0
+	 */
+	public static Optional<XSD.Datatype> getXsdDatatype(Literal l) {
+		if (l instanceof SimpleLiteral) {
+			return ((SimpleLiteral) l).getXsdDatatype();
+		}
+		return Optional.empty();
+	}
```
hmottestad, line 75: Can we move this back into the Literal interface. There is no need to introduce another instanceof call for every Literal again!

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
```diff
@@ -152,8 +152,8 @@ public static boolean compareLiterals(Literal leftLit, Literal rightLit, Compare
 		IRI leftDatatype = leftLit.getDatatype();
 		IRI rightDatatype = rightLit.getDatatype();
 
-		XSD.Datatype leftXsdDatatype = leftLit.getXsdDatatype().orElse(null);
-		XSD.Datatype rightXsdDatatype = rightLit.getXsdDatatype().orElse(null);
+		XSD.Datatype leftXsdDatatype = Literals.getXsdDatatype(leftLit).orElse(null);
+		XSD.Datatype rightXsdDatatype = Literals.getXsdDatatype(leftLit).orElse(null);
```
hmottestad, line 156: This is probably what is causing the test failures. Notice both methods call on leftLit

knoan:
`MemoryCascadeValueExceptionTest` broken as well.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
hmottestad: I'll run some benchmarks now to see the effects, but I understand that's it's hard to move back since it makes the Literal class dependent on the XSD class.

hmottestad: This branch: 

```
Benchmark                                       Mode  Cnt    Score   Error  Units
ValueComparatorBenchmark.sortManyDeepEquals     avgt   10  330.260 ± 2.762  ms/op
ValueComparatorBenchmark.sortManyPointerEquals  avgt   10   19.662 ± 0.335  ms/op
ValueComparatorBenchmark.sortObjects            avgt   10   26.762 ± 0.800  ms/op
ValueComparatorBenchmark.sortPredicates         avgt   10    0.049 ± 0.016  ms/op
ValueComparatorBenchmark.sortSubjects           avgt   10   24.441 ± 0.448  ms/op
```

Develop: 
```
Currently running benchmarks
```

knoan:
And `EvaluationStrategyTest.testDatetimeSubtypesStrict()`

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
jeenbroekstra: No, we can't, because it uses a class from the model module, which introduces a cyclic dependency.

jeenbroekstra: Unless we can interface/abstract that out as well, somehow.

jeenbroekstra: Actually might not be a bad idea. It's late and I'm in the middle of a whisky tasting but I'll have a look tomorrow unless you best me to it.

hmottestad: Enjoy your whisky tasting! No need to worry about this. I checked the performance, and using instanceof against a class seems to have no impact compared to how slow we were seeing instanceof be against an interface.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Namespace.java
```diff
@@ -11,9 +11,20 @@
...
+ * 
  */
 public interface Namespace extends Serializable, Comparable<Namespace> {
 
+	/**
+	 * Gets the prefix of the current namespace. The default namespace is represented by an empty prefix string.
+	 *
+	 * @return prefix of namespace, or an empty string in case of the default namespace.
+	 */
+	public String getPrefix();
```
jeenbroekstra, line 26: Adding new methods to an existing public interface (without a default implementation) is not backwards compatible.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Namespace.java
```diff
@@ -22,9 +33,23 @@
...
+	boolean equals(Object o);
+
+	/**
+	 * Computes the hash code of this namespace.
+	 *
+	 * @return a hash code for this namespace computed as {@link java.util.Objects#hash Objects.hash}(
+	 *         {@link #getPrefix()}, {@link #getName()})
+	 */
+	@Override
+	int hashCode();
```
jeenbroekstra, line 53: This, also, is not backwards compatible. I don't think we need to declare these at the interface level right now to get this to work.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Namespace.java
jeenbroekstra: Er, forget I said that, this is obviously not a new method. It was early, before coffee.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Namespace.java
jeenbroekstra: I still blame lack of coffee. This is not a problem in the sense that any subclass that does not explicitly implement these methods will just inherit them from Object. Which is not correct according the documented exepcted behavior we give here, but at least it won't cause compilation errors.

jeenbroekstra: I had a closer look at this and I think that in fact you are correct to specify this, and we should treat this not as a backward-incompatible change, but as a bug fix. Namespace extends Comparable, which _requires_ that we have properly defined contract for equals / hashCode. 

I'll add the necessary exceptions to the api comparison check.

core/model-api/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+	<description>RDF model interfaces.</description>
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<executions>
+					<execution>
+						<goals>
+							<goal>test-jar</goal>
```
jeenbroekstra, line 20: I've added this to make sure the test-jar gets produced. I originally wanted to move your abstracts tests to the rdf4j-model-testsuite, but that then results in a circular dependency between rdf4j-model-api and rdf4j-model-testsuite. So let's go with the test-jar approach instead.

jeenbroekstra (approved):
This now looks fine to me. Thanks for your efforts @knoan !

knoan:
> This now looks fine to me. Thanks for your efforts @knoan !

Happy to contribute to the good work.

PR #2562, opened 2020-09-30
Gh 2558 memory store sorting

hmottestad:
GitHub issue resolved: # 

---- 


core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -47,7 +47,7 @@
 	/**
 	 * The literal's language tag.
 	 */
-	private String language;
+	private Optional<String> language = Optional.empty();
```
hmottestad, line 50: This reduces the number og optional we have to instantiate.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/OrderIterator.java
```diff
@@ -270,38 +270,37 @@ public OrderIterator(CloseableIteration<BindingSet, QueryEvaluationException> it
...
+//					if (sorted.size() >= limit) {
+//						threshold = sorted.get(sorted.size() - 1);
+//					}
+//				}
+//				BindingSet next = iter.next();
+//				if (threshold == null || comparator.compare(next, threshold) < 0) {
+//					list.add(next);
+//					increment();
+//				}
+				list.add(iter.next());
```
hmottestad, line 302: this is just for testing performance at the moment

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemIRI.java
```diff
@@ -130,6 +130,19 @@ public int hashCode() {
...
+			int compareTo = namespace.compareTo(((MemIRI) o).namespace);
+			if (compareTo == 0) {
+				return localName.compareTo(((MemIRI) o).localName);
+			} else
+				return compareTo;
+		}
+
+		return toString().compareTo(o.toString());
+	}
+
```
hmottestad, line 145: This means we don't have to instantiate new string objects (concat) when comparing two MemIRIs. Biggest difference to be honest.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/EvaluationStrategy.java
```diff
@@ -95,6 +96,16 @@ TupleExpr optimize(TupleExpr expr, EvaluationStatistics evaluationStatistics,
...
+	/**
+	 * Gets the value of this expression.
+	 *
+	 * @param expr
+	 * @param bindings The variables bindings to use for evaluating the expression, if applicable.
+	 * @return The Value that this expression evaluates to, or <tt>null</tt> if the expression could not be evaluated.
+	 */
+	Value evaluate(Var expr, BindingSet bindings)
+			throws ValueExprEvaluationException, QueryEvaluationException;
+
```
hmottestad, line 108: Thought adding this would cut down on the number of instanceof calls we were making, but doesn't actually seem to make a difference.

hmottestad:
Also found out that one of the biggest performance bottlenecks in query evaluation is creating and using the HashMaps within the binding sets.

core/model/src/main/java/org/eclipse/rdf4j/model/util/LexicalValueComparator.java
```diff
@@ -94,7 +94,7 @@ private int compareBNodes(BNode leftBNode, BNode rightBNode) {
 	}
 
 	private int compareURIs(IRI leftURI, IRI rightURI) {
-		return leftURI.toString().compareTo(rightURI.toString());
+		return leftURI.compareTo(rightURI);
```
jeenbroekstra, line 97: The contract of the LexicalValueComparator is that it compares on the lexical value of the RDF term. This change is not valid because you don't know if the IRI implementations actually use lexical comparison internally. 

Except of course if, as mentioned above, we explicitly document that IRI should always implement Comparable using lexical value comparison.

core/model/src/main/java/org/eclipse/rdf4j/model/IRI.java
```diff
@@ -27,7 +27,7 @@
  * @author Jeen Broekstra
  */
 @SuppressWarnings("deprecation")
-public interface IRI extends URI, Resource {
+public interface IRI extends URI, Resource, Comparable<IRI> {
```
jeenbroekstra, line 30: We use Comparators for our RDF term types instead of declaring them Comparable, precisely because there is not a single mandated natural ordering for them.

jeenbroekstra: To give but one example: the native store keeps track of revisions for IRIs. Either you take that revision into account for total ordering (in which case the comparison is no longer purely lexical) or you don't, in which case comparing different revisions of the same NativeIRI will appear to be equal.

jeenbroekstra: We can get away with introducing this, but only if we _very_ explicitly document that we expect all IRI implementations to use lexical ordering for implementing Comparable. This also should make it clear that for some (implementation-specific) types of comparison (e.g. the nativestore one I mentioned above) you should continue to use a bespoke Comparator rather than rely on the Comparable implementation.

jeenbroekstra: Regardless, this makes the change backwards incompatible because it is quite conceivable that third party implementors have made their own IRI implementations implement Comparable already.

hmottestad: I don't this this change makes it backwards incompatible since there is a default method. And anyone who has made their class also comparable shouldn't really have any issue since it would be a subclass of IRI.

core/model/src/main/java/org/eclipse/rdf4j/model/util/LexicalValueComparator.java
hmottestad: Both out comparators use the same way of comparing IRIs.

hmottestad:
Screenshot 2020-10-06 at 16 22 25

This is what happens when we use `toString()` in the ValueComparator instead of making IRI comparable. Essentially what you are seeing is GC taking more and more time as it fills up with new Strings every time two IRIs are compared. The concatenation in the `toString()` method in IRI is taking massive amounts of RAM.

hmottestad:
Screenshot 2020-10-06 at 16 30 28

This is the most interesting thing though. Here is the exact same benchmark and code running on Java 15 instead of 8. No GC problems. I think this means that the JVM might be able to do escape analysis so the new string is allocated on the stack instead of the heap, thus no need for GC!

hmottestad:
Screenshot 2020-10-06 at 16 38 55

So here is Java 11! Same behaviour as Java 15.

Issue #2574, opened 2020-10-02
Improve SPARQL parsing performance


hmottestad:
Big methods in the SPARQL parser are not optimized properly by the JIT.

PR #2575, opened 2020-10-02
GH-2574 make SPARQL parsing faster

hmottestad:
GitHub issue resolved: #2574 

Split a big method.

---- 

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderTokenManager.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/SyntaxTreeBuilderTokenManager.java
@@ -1870,1282 +1870,11 @@ public class SyntaxTreeBuilderTokenManager implements SyntaxTreeBuilderConstants
 				ReInitRounds();
 			}
 			if (curChar < 64) {
-				long l = 1L << curChar;
-				do {
-					switch (jjstateSet[--i]) {
-					case 0:
-						if ((0x3ff000000000000L & l) != 0L) {
-							if (kind > 152) {
-								kind = 152;
-							}
-							jjCheckNAddStates(0, 6);
-						} else if ((0x100003600L & l) != 0L) {
-							if (kind > 2) {
-								kind = 2;
-							}
-						} else if (curChar == 46) {
-							jjCheckNAddTwoStates(99, 153);
-						} else if (curChar == 45) {
-							jjCheckNAddStates(7, 11);
-						} else if (curChar == 43) {
-							jjCheckNAddStates(12, 16);
-						} else if (curChar == 58) {
-							if (kind > 146) {
-								kind = 146;
-							}
-							jjCheckNAddStates(17, 19);
-						} else if (curChar == 34) {
-							jjstateSet[jjnewStateCnt++] = 60;
-						} else if (curChar == 39) {
-							jjstateSet[jjnewStateCnt++] = 49;
-						} else if (curChar == 36) {
-							jjstateSet[jjnewStateCnt++] = 24;
-						} else if (curChar == 60) {
-							jjCheckNAddTwoStates(13, 14);
-						} else if (curChar == 40) {
-							jjCheckNAddTwoStates(4, 5);
-						} else if (curChar == 35) {
-							if (kind > 3) {
-								kind = 3;
-							}
-							jjCheckNAdd(2);
-						} else if (curChar == 63) {
-							jjstateSet[jjnewStateCnt++] = 21;
-						}
-						if (curChar == 34) {
-							jjCheckNAddStates(20, 22);
-						} else if (curChar == 39) {
-							jjCheckNAddStates(23, 25);
-						}
-						break;
-					case 1:
-						if (curChar != 35) {
-							break;
-						}
-						if (kind > 3) {
-							kind = 3;
-						}
-						jjCheckNAdd(2);
-						break;
-					case 2:
-						if ((0xffffffffffffdbffL & l) == 0L) {
-							break;
-						}
-						if (kind > 3) {
-							kind = 3;
-						}
-						jjCheckNAdd(2);
-						break;
-					case 3:
-						if (curChar == 40) {
-							jjCheckNAddTwoStates(4, 5);
-						}
-						break;
-					case 4:
-						if ((0x100003600L & l) != 0L) {
-							jjCheckNAddTwoStates(4, 5);
-						}
-						break;
-					case 5:
-						if (curChar == 41 && kind > 30) {
-							kind = 30;
-						}
-						break;
-					case 7:
-						if ((0x100003600L & l) != 0L) {
-							jjAddStates(26, 27);
-						}
-						break;
-					case 12:
-						if (curChar == 60) {
-							jjCheckNAddTwoStates(13, 14);
-						}
-						break;
-					case 13:
-						if ((0xaffffffa00000000L & l) != 0L) {
-							jjCheckNAddTwoStates(13, 14);
-						}
-						break;
-					case 14:
-						if (curChar == 62 && kind > 145) {
-							kind = 145;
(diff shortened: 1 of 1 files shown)


hmottestad:
@jeenbroekstra I've split big SPARQL parsing methods before, and I have a sinking feeling it might have been this one! I noticed it taking a lot of time in the new SHACL AST (which is currently running far too many SPARQL queries, but that's another matter). Splitting this method made things 30-40% faster.

Last time I think James did something that would allow it to persist even when regenerating the SPARQL parsing code.

jeenbroekstra:
Can you find back the place you last did that, either in the issue tracker or in the commit log? It would help me see what you did, what James did to preserve it, and why it got reverted again.

jeenbroekstra:
Also, for what's it's worth, it's generally recommend _not_ to manually edit the TokenManager for a JavaCC-generated parser. Ideally we should figure out a way change the declaration such that the bottleneck is taken out (typically this kind of thing is caused by too many branching factors in the grammar and/or too much use of lookahead).

hmottestad:
Found it: https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/sparql.patch

It's just a simple git patch.

hmottestad:
GIT patch doesn't work anymore. I thought I would give JavaCC a try to figure out if there was anything to adjust, but getting build error:

```
MacBook-Pro:scripts havardottestad$ ./javacc /Users/havardottestad/Documents/Programming/RDF4J/rdf4j/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/sparql.jjt
Java Compiler Compiler Version 7.0.4 (Parser Generator)
(type "javacc" with no arguments for help)
Reading from file /Users/havardottestad/Documents/Programming/RDF4J/rdf4j/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/sparql.jjt . . .
Warning: Line 17, Column 9: Bad option name "MULTI".  Option setting will be ignored.
Warning: Line 18, Column 9: Bad option name "VISITOR".  Option setting will be ignored.
Warning: Line 19, Column 9: Bad option name "VISITOR_EXCEPTION".  Option setting will be ignored.
org.javacc.parser.ParseException: Encountered " "#" "# "" at line 379, column 15.
Was expecting one of:
    "throws" ...
    ":" ...
    
Detected 1 errors and 3 warnings.
```

I can't find much documentation on what I'm supposed to do with the `sparql.jit` file. So this is about as far as I've managed to come.

jeenbroekstra:
The .jjt file is input for the `jjtree` command, which is a pre-processor for javacc (see https://www.cs.purdue.edu/homes/hosking/javacc/doc/JJTree.html). It produces a `.jj` file, which then in turn is input for the actual javacc command. 

All of this is rather arcane - the javacc toolkit is somewhat sparsely documented itself. It's also, if I may say so, a pain to install and use properly, the way they package and distribute their releases is a bit of a nightmare (it's not just hard to install, it also _changes_ every other release). 

I should also point out that the last time I remember generating a new parser, JavaCC was still at version 5.0, so no clue how two major version bumps will influence things. I did notice, by the way, that apparently from version 7.0.5 their lookahead processing is broken. Since we use lookahead in our grammar, we should use version 7.0.4 (which I see you already do). 

I'm taking a closer look at the grammar itself this week, see if there are some parts that can be rewritten to reduce use of >1 lookaheads, as those are likely to be a major cause of performance issues.

jeenbroekstra:
Is the current SPARQLParseBenchmark test how you are evaluating the performance of this?

jeenbroekstra:
Note: we need proper benchmarking of the SPARQL parser itself, outside the context of SHACL.

hmottestad:
Yeah. That's the benchmark I created! And I found the original issue too: https://github.com/eclipse/rdf4j/issues/1002

jeenbroekstra:
So is the performance problem you're seeing specific to queries with VALUES clauses?

I'd like to get a better understanding of where in the parser the problem is, in terms of specific kinds of sparql operations or particular kinds of queries. That can hopefully help me pinpoint the point in the grammar production rules that corresponds to that big tokenizer method.

hmottestad:
I don't know. I never looked further into it. I can try some other queries. This might also be something that we would consider a bug in JavaCC.

jeenbroekstra:
FWIW I am not seeing any performance improvement with this method split in place.

3.4.3-SNAPSHOT, unmodified:

```
Benchmark                               Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.selectValuesQuery  avgt   40  14.751 ± 0.090  ms/op
```

This branch (with split method):

```
Benchmark                               Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.selectValuesQuery  avgt   40  15.468 ± 0.398  ms/op
```

The difference appears non-significant, and probably explained by small variations in CPU load on my machine.

I'm no optimization expert of course, but I have never heard of large method bodies being a performance burden for the JVM - it seems to me that simply splitting the method does nothing to reduce branching complexity, it's just organised slightly differently at the source level.

hmottestad:
That's very strange. How did you run the 3.4.3 benchmark?

hmottestad:
Btw. I've sometimes struggled with the benchmarks, when I switch branch it will sometimes not recompile properly and actually run the old code. Could that be the case for you?

hmottestad:
I've made two videos to show what's happening on my laptop and explain what I think the jdk is doing behind the curtains.

Video showing my benchmark run (no audio): https://www.youtube.com/watch?v=unkSJD70pBY&feature=youtu.be

Video explaining what I think is happening: https://www.youtube.com/watch?v=_5Q6VXIrdSY&feature=youtu.be

The last one is still being processed, so if it's all blurry then give it a half hour and see if it's any better.

jeenbroekstra:
> Btw. I've sometimes struggled with the benchmarks, when I switch branch it will sometimes not recompile properly and actually run the old code. Could that be the case for you?

I'll do another run to make sure but I was careful to run a complete mvn clean install before re-running, so it's unlikely it used old code.

jeenbroekstra:
> > Btw. I've sometimes struggled with the benchmarks, when I switch branch it will sometimes not recompile properly and actually run the old code. Could that be the case for you?
> 
> I'll do another run to make sure but I was careful to run a complete mvn clean install before re-running, so it's unlikely it used old code.

Ok, so, false alarm. I had been very careful to do a clean install on the second run, but apparently I forgot to do so for the first run, so it used your adapted code for both. 

I've ran it again, cleaning both times, and this time I get:

master branch
```

Benchmark                               Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.selectValuesQuery  avgt   40  29.734 ± 0.142  ms/op
```

split method branch

```
Benchmark                               Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.selectValuesQuery  avgt   40  14.395 ± 0.106  ms/op
```

jeenbroekstra:
Thanks for the videos by the way - an interesting insight in the limitations of the JIT compiler I guess (it's not a level at which I typically look at the code).

I'd like to add one or two more test cases to the benchmark, maybe one very large / complex query, and perhaps also one update operation. 

Apart from that, I think as a quick fix we should for now go with your manual split. We'll need a followup on this to investigate in more depth why javacc produces such a large method, if that is something we can control by modifying our grammar, and if not if we should log a bug or a feature request against javacc. We also generally need to start looking into updating the parser to be able to generate it with newer versions of javacc (currently, producing a new parser using javacc 7.0 run into compilation errors, because we use a custom Charstream implementation that is no longer compatible with their interface).

hmottestad:
I'll merge this then and create a new issue for the followup.

jeenbroekstra:
> I'll merge this then and create a new issue for the followup.

Go ahead with the merge. I already created a followup issue: #2580.

jeenbroekstra (approved):

Issue #2589, opened 2020-10-10
Decouple vocabularies to a dedicated maven module


knoan:
As suggested in the initial discussion for #2538.

> I would ideally like things split up even further, for example a separate 'rdf4j-vocabularies' module would be useful

Standalone vocabularies would improve modularisation and portability to limited devices.

PR in progress (branched from `develop` to leverage the abstract value base classes introduced with #2546).

barthanssens:
Would also make it easier to add additional vocabulary modules from other sources.

IIRC most (all ?) of the vocabularies are either RDF4j-specific vocabularies or W3C vocabularies, but in the future there could also be a separate e.g. rdf4j-vocabularies-schemaorg and rdf4j-vocabularies-semiceu.

knoan:
> most (all ?) of the vocabularies are either RDF4j-specific vocabularies or W3C vocabularies

Most, but not all: we have things like DC, GEO, FOAF/DOAP, …

knoan:
PR submitted; I opted for naming the new module `rdf4j-model-vocabulary`, after the migrated `o.e.r.model.vocabulary` package and to keep things consistent with #2546.

PR #2590, opened 2020-10-10
GH-2589 standalone vocabularies

knoan:
GitHub issue resolved: #2589

- migrate `o.e.r.model.vocabulary` package to the dedicated `rdf4j-model-vocabulary` maven module

---- 


core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/Vocabularies.java
```diff
@@ -0,0 +1,61 @@
...
+			@Override
+			public String getName() {
+				return namespace;
+			}
+
+		};
+	}
+
+	static IRI createIRI(final String namespace, final String localName) {
+		return new AbstractIRI() {
```
jeenbroekstra, line 43: This shift away from using the SimpleValueFactory might impact performance. We'll need to take a closer look at anything involving IRI object comparison.

jeenbroekstra (commented):
Mostly looks fine, though I want to see if (and if so, how much) this affects performance. The build currently fails because this is flagged as a backward incompatible change. I'll take a look at that and if it turns out it is similar to what we did for the new api module, I'll adapt the check to make an exception.

core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/Vocabularies.java
knoan: I considered this issue in the design of the current approach and I agree that a closer look is deserved; however, any performance degradation is likely to be traced back to #2546, which factored object equality from `SimpleIRI` to the shared `AbstractIRI` base class.

As far as I can see, the only relevant difference is in the timing of `IRI.toString()`:
failing string object identity, both the original `SimpleIRI.equals()` and the updated `AbstractIRI.equals()` methods compare the return value of `IRI.toString()`, but:

- in the case of `SimpleIRI` the value is precomputed in `AbstractValueFactory` as

```java
public IRI createIRI(String namespace, String localName) {
	return createIRI(namespace + localName);
}
```

- in the case of `AbstractIRI` the value is computed on demand on the basis of the public API as

```java
public String stringValue() {
	return getNamespace() + getLocalName();
}
```

Back to you with some hard figures…

knoan:
Some quick and dirty figures gathered with the attached harness and the following setup:

- macOs 10.15.7
- Oracle JDK 1.8.0_241
- java -Xmx4G
- 1M equality checks between randomly generated IRIs
- median time in ms for over 5 rounds after 3 warm-up rounds

|branch|factory|elapsed|ratio|notes|
 |------|-------|-------|-----|-----|
 |master     | SimpleIRI                | 7426 | 1.00 | baseline |
 |develop    | SimpleIRI                | 7213 | 0.97 | measurement error? |
 |GH-2589    | SimpleIRI                | 7295 | 0.98 | measurement error? |
 |GH-2589    | Vocabularies.createIRI   | 123081  | 16.57 | current implementation |
 |GH-2589    | Vocabularies.createIRI   | 16869  | 2.27 | precomputed stringValue() |

`SimpleIRI.equals()` doesn't appear to have suffered performance-wise from #2546: the observed discrepancies are likely to be measurement errors, as I wasn't able to set up a dedicated testing environment.

On the other hand, the current implementation of `Vocabularies.createIRI()` is not optimal: quite puzzlingly the obvious optimisation (already implemented) based on a pre-computed `stringValue` doesn't bring it on a par with `SimpleIRI`.

Having a closer look at it…

[IRIPerformance.java.zip](https://github.com/eclipse/rdf4j/files/5371794/IRIPerformance.java.zip)

knoan:
After reworking the harness to exclude the effects of array allocations as per the attached code, we have the following updated figures (not immediately comparable with the previous ones).

 |branch|factory|elapsed|ratio|notes|
 |------|-------|-------|-----|-----|
 |GH-2589    | SimpleIRI   | 1829  | 1 | baseline |
 |GH-2589    | VocabularyIRI   | 3978  | 2.17 | optimized implementation |
 |GH-2589    | VocabularyIRI   | 1819  | 0.99 | with .toString() |
 
For reasons far beyond my wisdom touching any `String` method in the IRI constructor like

```java
private final String stringValue = (namespace + localName).toString();
```

brings back performances on a par with `SimpleIRI`, which happens to touch a `String` method in the first line of 

```java
if (iriString.indexOf(':') < 0) { 
	throw new IllegalArgumentException("Not a valid (absolute) IRI: " + iriString);
}
```

Removing this check degrades `SimpleIRI` performance with the same factor…

May only wonder if this behaviour is connected with some string interning magic, but I wasn't able to chase it down in JDK code.
 
[IRIPerformance.java.zip](https://github.com/eclipse/rdf4j/files/5372888/IRIPerformance.java.zip)

jeenbroekstra:
Great work @knoan , I wasn't even suggesting that you should be doing this analysis, but very happy to you've picked it up! 

Fully satisfied that the minor peformance differences you for SimpleIRI are measurement errors. I'll try and take a closer look at what black magic is involved in the performance being affected so radically by using `toString()`, though to be fair at this point I'm mostly fine with just knowing that it does...

knoan:
The issue seems to be indeed related to the internals of JDK string handling, with OpenJDK _adversely_ affected by `.toString()`…

|branch|jdk|factory|w/ .toString()|w/o .tostring()|ratio|
|---|---|---|---|---|---|
|GH-2589|jdk1.8.0_241|Vocalbularies.createIRI|1822|4048|2.22|
|GH-2589|jdk-11.0.8|Vocalbularies.createIRI|2057|4595|2.23|
|GH-2589|adoptopenjdk-11-openj9|Vocalbularies.createIRI|1753|1583|0.90|

jeenbroekstra (commented):
I've just noticed that we should probably refactor AbstractValueFactory, and move it to the model-api module (the base package). This will require that it creates AbstractValue objects, rather than (as it does now) SimpleValue objects. 

This is not directly related to this change, however, so I am fine with calling this one done and doing a bit of a cleanup in a separate PR (either of us can pick that up). 

One last minor request: could you please [squash  your commits](https://rdf4j.org/documentation/developer/squashing), at least partially? 

In case you missed it: we've had to change our PR merge strategy. We previously used github squash-and-merge so that the entire PR is one nice tidy commit, but this turns out to have unexpected side effects for authorship tracking, so we're back to using regular merges. 

You dn't have to squash everything down to a single commit, but perhaps we can get at least some of the intermediate commits squashed together.

knoan:
> I've just noticed that we should probably refactor AbstractValueFactory, and move it to the model-api module (the base package). This will require that it creates AbstractValue objects, rather than (as it does now) SimpleValue objects.

I've already considered this option and unfortunately is not straightforward to implement…

`o.e.r.model.impl` exposes 7 specialised subclasses of `SimpleLiteral` handling native Java objects: in order to preserve binary compatibility **and** to prevent extensive duplications in `o.e.r.model.base`, their dependencies, the implemented optimisations and possible performance issues must be carefully evaluated; time permitting, I'll reassess the whole thing in more detail.

knoan:
> One last minor request: could you please squash your commits, at least partially?

Done.

> You dn't have to squash everything down to a single commit, but perhaps we can get at least some of the intermediate commits squashed together.

In order to trace module-related edits in poms, I felt useful to keep the following distinct:

- GH-2589 Create rdf4j-model-vocabulary module
- GH-2589 Migrate o.e.r.model.vocabulary package to rdf4j-model-vocabulary module

jeenbroekstra (approved):
Happy with this, unless I hear objections I will merge this tomorrow.

Issue #2603, opened 2020-10-18
Move AbstractValueFactory to the model-api module


knoan:
As discussed in https://github.com/eclipse/rdf4j/pull/2590#pullrequestreview-508869927

> I've just noticed that we should probably refactor AbstractValueFactory, and move it to the model-api module (the base package). This will require that it creates AbstractValue objects, rather than (as it does now) SimpleValue objects.

Starting work on a new PR with an incremental approach.

Issue #2563, opened 2020-10-01
Upgrade to java.time API for date and time handling


jeenbroekstra:
We use the old java.util.Date in several places in the core API. We should upgrade to the new java.time API since it is far superior and the recommended way of handling dates and times since Java 8.

See https://www.baeldung.com/java-8-date-time-intro

knoan:
Likely to be an epic touching SPARQL evaluation, value comparison, …

Opened a dedicated issue to start groundwork on `Value/ValueFactory`.

knoan:
A tentative list of other intervention areas:

- Literals.createLiteral()
- Literals.getTemporalAccessor/AmountValue() (see https://github.com/eclipse/rdf4j/pull/2659#issuecomment-728227741)
- org.eclipse.rdf4j.query.QueryResults.bindingSetsMatch()
- org.eclipse.rdf4j.query.algebra.evaluation.util.ValueComparator

knoan:
+ lenient temporal literal parsing in RIO

PR #2607, opened 2020-10-20
GH-2603 o.e.r.model.base.AbstractValueFactory

knoan:
GitHub issue resolved: #2603

Provide an abstract base implementation of the `ValueFactory` interface in package `o.e.r.model.base`, generating generic values based on the existing abstract base implementations of `Value` interfaces .

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
@@ -20,8 +20,6 @@ public abstract class AbstractBNode implements BNode {
 
 	private static final long serialVersionUID = -437354568418943981L;
 
-	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-
 	@Override
 	public String stringValue() {
 		return getID();
@@ -42,8 +40,6 @@ public abstract class AbstractBNode implements BNode {
 	public String toString() {
 		return "_:" + getID();
 	}
-
-	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 	static class GenericBNode extends AbstractBNode {
 
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
@@ -8,8 +8,6 @@
 
 package org.eclipse.rdf4j.model.base;
 
-import java.util.Objects;
-
 import org.eclipse.rdf4j.model.BNode;
 
 /**
@@ -22,6 +20,8 @@ public abstract class AbstractBNode implements BNode {
 
 	private static final long serialVersionUID = -437354568418943981L;
 
+	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 	@Override
 	public String stringValue() {
 		return getID();
@@ -30,17 +30,36 @@ public abstract class AbstractBNode implements BNode {
 	@Override
 	public boolean equals(Object o) {
 		return this == o || o instanceof BNode
-				&& Objects.equals(getID(), ((BNode) o).getID());
+				&& getID().equals(((BNode) o).getID());
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hashCode(getID());
+		return getID().hashCode();
 	}
 
 	@Override
 	public String toString() {
 		return "_:" + getID();
+	}
+
+	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+	static class GenericBNode extends AbstractBNode {
+
+		private static final long serialVersionUID = -617790782100827067L;
+
+		private String id;
+
+		GenericBNode(String id) {
+			this.id = id;
+		}
+
+		@Override
+		public String getID() {
+			return id;
+		}
+
 	}
 
 }
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
@@ -23,8 +23,6 @@ public abstract class AbstractIRI implements IRI {
 
 	private static final long serialVersionUID = 7799969821538513046L;
 
-	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-
 	@Override
 	public String stringValue() {
 		return getNamespace() + getLocalName();
@@ -33,20 +31,18 @@ public abstract class AbstractIRI implements IRI {
 	@Override
 	public boolean equals(Object o) {
 		return this == o || o instanceof IRI
-				&& toString().equals(o.toString()); // !!! use stringValue()
+				&& toString().equals(o.toString()); // TODO stringValue() (https://github.com/eclipse/rdf4j/issues/2565)
 	}
 
 	@Override
 	public int hashCode() {
-		return toString().hashCode(); // !!! use stringValue()
+		return toString().hashCode(); // TODO stringValue() (https://github.com/eclipse/rdf4j/issues/2565)
 	}
 
 	@Override
 	public String toString() {
 		return stringValue();
 	}
-
-	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 	static class GenericIRI extends AbstractIRI {
 
@@ -62,7 +58,7 @@ public abstract class AbstractIRI implements IRI {
 		}
 
 		GenericIRI(String namespace, String localName) {
-			this.iri = namespace + localName; // !!! concatenation performance impact?
+			this.iri = namespace + localName;
 			this.split = namespace.length();
 		}
 
@@ -82,11 +78,27 @@ public abstract class AbstractIRI implements IRI {
 		}
 
 		private int split() {
-			return (split > 0) ? split
-					: (split = iri.indexOf('#') + 1) > 0 ? split
-							: (split = iri.lastIndexOf('/') + 1) > 0 ? split
-									: (split = iri.lastIndexOf(':') + 1) > 0 ? split
-											: 0; // unexpected: colon presence already tested in factory methods
+			if (split > 0) {
+
+				return split;
+
+			} else if ((split = iri.indexOf('#') + 1) > 0) {
+
+				return split;
+
+			} else if ((split = iri.lastIndexOf('/') + 1) > 0) {
+
+				return split;
+
+			} else if ((split = iri.lastIndexOf(':') + 1) > 0) {
+
+				return split;
+
+			} else {
+
+				return 0; // unexpected: colon presence already tested in factory methods
+
+			}
 		}
 
 	}
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
@@ -8,10 +8,7 @@
 
 package org.eclipse.rdf4j.model.base;
 
-import java.util.Objects;
-
 import org.eclipse.rdf4j.model.IRI;
-import org.eclipse.rdf4j.model.Value;
 
 /**
  * Base class for {@link IRI}, offering common functionality.
@@ -26,6 +23,8 @@ public abstract class AbstractIRI implements IRI {
 
 	private static final long serialVersionUID = 7799969821538513046L;
 
+	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 	@Override
 	public String stringValue() {
 		return getNamespace() + getLocalName();
@@ -34,17 +33,62 @@ public abstract class AbstractIRI implements IRI {
 	@Override
 	public boolean equals(Object o) {
 		return this == o || o instanceof IRI
-				&& Objects.equals(toString(), ((Value) o).toString()); // !!! use stringValue()
+				&& toString().equals(o.toString()); // !!! use stringValue()
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hashCode(toString()); // !!! use stringValue()
+		return toString().hashCode(); // !!! use stringValue()
 	}
 
 	@Override
 	public String toString() {
 		return stringValue();
+	}
+
+	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+	static class GenericIRI extends AbstractIRI {
+
+		private static final long serialVersionUID = 2209156550690548467L;
+
+		private String iri;
+
+		private int split;
+
+		GenericIRI(String iri) {
+			this.iri = iri;
+			this.split = 0;
+		}
+
+		GenericIRI(String namespace, String localName) {
+			this.iri = namespace + localName; // !!! concatenation performance impact?
+			this.split = namespace.length();
+		}
+
+		@Override
+		public String stringValue() {
+			return iri;
+		}
+
+		@Override
+		public String getNamespace() {
+			return iri.substring(0, split());
+		}
+
+		@Override
+		public String getLocalName() {
+			return iri.substring(split());
+		}
+
+		private int split() {
+			return (split > 0) ? split
+					: (split = iri.indexOf('#') + 1) > 0 ? split
+							: (split = iri.lastIndexOf('/') + 1) > 0 ? split
+									: (split = iri.lastIndexOf(':') + 1) > 0 ? split
+											: 0; // unexpected: colon presence already tested in factory methods
+		}
+
 	}
 
 }
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
@@ -8,24 +8,25 @@
 
 package org.eclipse.rdf4j.model.base;
 
-import static java.lang.Boolean.FALSE;
-import static java.lang.Boolean.TRUE;
 import static java.util.Objects.requireNonNull;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.Locale;
-import java.util.Objects;
+import java.util.GregorianCalendar;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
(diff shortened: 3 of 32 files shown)


knoan:
An initial, functionally complete attempt: some performance issues still to be ironed out.

knoan:
Performance improved after some obvious functional optimizations and generally more or less on a par with `master`; still wrestling with some weird details (final fields apparently slower than writable ones, string concatenation apparently less efficient on equals than `new String()`, …).

Back to you as soon as figures are sensible.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
```diff
@@ -22,6 +23,37 @@
...
+	/**
+	 * Creates a new blank node value.
+	 *
+	 * @param nodeID the identifier of the blank node
+	 *
+	 * @return a new generic blank node value
+	 *
+	 * @throws NullPointerException if {@code nodeID} is {@code null}
+	 */
+	public static BNode createBNode(String nodeID) {
```
jeenbroekstra, line 46: Why are we adding factory methods to the class itself? This will be problematic to support. For example concrete subclasses will have their own factories, but if you do `MyConcreteBNode.createBNode` you suddenly get a `GenericBNode` instance again. It also somewhat defeats the purpose of having a separate ValueFactory.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
```diff
@@ -26,6 +23,61 @@
...
+	 * Creates a new IRI value.
+	 *
+	 * @param iri the string representation of the IRI
+	 * 
+	 * @return a new generic IRI value
+	 * 
+	 * @throws NullPointerException     if {@code iri} is {@code null}
+	 * @throws IllegalArgumentException if {@code iri} is not an absolute IRI
+	 */
+	public static IRI createIRI(String iri) {
```
jeenbroekstra, line 36: Same here. Let's not put factory methods in the abstract classes themselves.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
```diff
@@ -34,17 +86,92 @@ public String stringValue() {
...
 	}
 
 	@Override
 	public String toString() {
 		return stringValue();
 	}
 
+	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+	private static class UnaryIRI extends AbstractIRI {
```
jeenbroekstra, line 104: I think this may be over-optimising things somewhat. Keep in mind that most database implementations will have their own ValueFactory (and therefore own IRI) implementations.

jeenbroekstra (commented):
This looks a lot more extensive than I had anticipated, to be honest. I haven't fully reviewed yet, but I noticed a few things that I think we need to discuss before you carry on with this.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
knoan: More of a matter of convenience, as the merging/splitting logic is reversed in the two cases, but I'll review the matter.

knoan:
Reworked along the suggested lines: now the public API surface is minimal.

Collecting some final performance figures…

knoan:
Attached the results of the latest performance benchmarks for factory methods and `Value.equals/hashCode()` methods for `SimpleValueFactory` on the `master` branch and `base.AbstractValueFactory` on `GH-2603`: figures are in the same ballpark, with no major deviations.

[20201022 Factory Benchmarks.zip](https://github.com/eclipse/rdf4j/files/5425043/20201022.Factory.Benchmarks.zip)

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
```diff
@@ -22,6 +20,8 @@
 
 	private static final long serialVersionUID = -437354568418943981L;
 
+	////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
jeenbroekstra, line 23: This is somewhat minor (so don't feel the need to immediately take these out again), but: open question (to @hmottestad and @barthanssens as well) - what's your opinion on the usefulness of this kind of "visual divider" commenting? I've been taking most of the legacy marker comment blocks ("constructors" / "public methods") etc out because they tend to get malformed and also just wrong over time (people putting private methods in the public methods bit etc). 

I'm personally not in favor of adding this kind of thing at all. But if we feel there's value in organising code this way then I'd like a consistent style adopted, at least.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractValueFactory.java
```diff
@@ -0,0 +1,280 @@
...
+import org.eclipse.rdf4j.model.base.AbstractTriple.GenericTriple;
+
+/**
+ * Base class for {@link ValueFactory}, offering common functionality.
+ *
+ * @author Alessandro Bollini
+ * @since 3.5.0
+ */
+@SuppressWarnings("UseOfObsoleteDateTimeApi")
+public abstract class AbstractValueFactory implements ValueFactory {
```
jeenbroekstra, line 49: I've just realized that this new class was introduced, but there's no relation to the existing AbstractValueFactory. I'll take a look at that, we may want to have the old class inherit from the new and/or deprecate it.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractValueFactory.java
```diff
@@ -0,0 +1,280 @@
...
+	public BNode createBNode() {
+		return new GenericBNode(Long.toHexString(Math.abs(nodeID.getAndIncrement())));
+	}
+
+	@Override
+	public BNode createBNode(String nodeID) {
+
+		if (nodeID == null) {
+			throw new NullPointerException("null nodeID");
+		}
```
jeenbroekstra, line 68: `Objects.requireNonNull(nodeID);`?

jeenbroekstra (commented):
Thanks @knoan it's starting to look really good. There's a couple of minor things, most important thing left open is clarifying the relation between the old AbstractValueFactory and the new one I think.

jeenbroekstra:
add to whitelist

jeenbroekstra:
retest this please

jeenbroekstra:
(note to others - please disregard the above two comments, I am running a few tests with our Jenkins CI setup as an experiment, it's nothing to do with this particular PR and is not blocking anything)

jeenbroekstra:
retest this please

jeenbroekstra:
@rdf4j-bot  retest this please

jeenbroekstra:
@rdf4j-bot (re)run tests. Sigh, forgot that we customized this. But this should work.

jeenbroekstra:
@rdf4j-bot rerun tests

jeenbroekstra:
@rdf4j-bot  rerun tests. Last one I hope. Sorry about the noise, just weeding out minor config issues in the old Jenkins job.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
```diff
@@ -8,56 +8,88 @@
...
-import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.base.AbstractIRI.GenericIRI;
 
 /**
  * Base class for {@link Literal}, offering common functionality.
  *
  * @author Alessandro Bollini
  * @since 3.5.0
  */
+@SuppressWarnings("UseOfObsoleteDateTimeApi")
```
jeenbroekstra, line 37: Where does this come from? I've never seen this particular warning / annotation before.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
```diff
@@ -34,17 +33,62 @@ public String stringValue() {
...
+		public String getLocalName() {
+			return iri.substring(split());
+		}
+
+		private int split() {
+			return (split > 0) ? split
+					: (split = iri.indexOf('#') + 1) > 0 ? split
+							: (split = iri.lastIndexOf('/') + 1) > 0 ? split
+									: (split = iri.lastIndexOf(':') + 1) > 0 ? split
+											: 0; // unexpected: colon presence already tested in factory methods
```
jeenbroekstra, line 89: Could we rewrite this using if...else instead of a collection of ternary operators? It will be easier to read I think.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
```diff
@@ -148,41 +179,361 @@ public BigDecimal decimalValue() {
...
+		}
+
+		@Override
+		public IRI getDatatype() {
+			return RDF_LANGSTRING;
+		}
+
+	}
+
+	static class BooleanLiteral extends AbstractLiteral {
```
jeenbroekstra, line 290: We already have a `BooleanLiteral` class. Do we need this?

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractStatement.java
```diff
@@ -30,15 +35,15 @@ public boolean equals(Object o) {
...
+				&& getObject().equals(((Statement) o).getObject())
+				&& getSubject().equals(((Statement) o).getSubject())
+				&& getPredicate().equals(((Statement) o).getPredicate())
 				&& Objects.equals(getContext(), ((Statement) o).getContext());
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hash(getSubject(), getPredicate(), getObject(), getContext());
+		return Objects.hash(getSubject(), getPredicate(), getObject(), getContext()); // !!! avoid array creation
```
jeenbroekstra, line 46: I genuinely don't understand what this comment means (also not sure I understand what the `!!!` signifies).

jeenbroekstra (commented):
Ok, taking a closer look, and it's all beautifully done I think. However, at the risk of sounding ungrateful (and fully conscious of the fact that I originally suggested migrating AbstractValueFactory), I can't help but wonder: with this change, we have migrated almost all interface/abstract code that was originally in rdf4j-model to rdf4j-model-api, but what, exactly, have we won with this change? 

I fully agree that the principle of separating the public api from implementations is a good one, and I also think that taking out the vocabulary stuff to make the model-api module leaner was a good idea, but we now have two sets of base implementations of the model objects: there's all these private static classes in here, and then there's the Simple... implementations and various subtypes (including BooleanLiteral etc) that basically redo the entire thing and don't actually reuse any of the stuff you've added here.

I also note we're (semi-)duplicating a lot of internal code, and even things like vocabulary constants are now declared in two places (e.g. AbstractLiteral has its own constant for the RDF namespace, and so on). Are we aiming to the make the entire `Simple...` stuff deprecated? Or are we aiming to have all `Simple` concrete classes derive from these new abstract classes? If so, what exactly can they inherit? 

So stepping back a bit: what precise problem have we actually solved by creating this AbstractValueFactory (which, by the way, is not  _actually_ an abstract factory, as all methods are fully implemented) in rdf4j-model-api?

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
knoan: When working with a folding code editor on a complex class, I find it useful to be able to visually group related methods.

This particular graphic solution is probably too imposing: my personal suggestion would be to relax formatting rules to allow for two consecutive blank lines as in the attached example.

Screenshot 2020-10-26 at 08 23 41

knoan: Provisionally removed from this PR.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractValueFactory.java
knoan: No problem migrating to the compact form, if that is the project standard; however, in the context of argument checking, I personally find the extended form more explicit and more consistent when accompanied by other checks as in:

```java
if (iri == null) {
	throw new NullPointerException("null iri");
}

if (iri.indexOf(':') < 0) {
	throw new IllegalArgumentException("missing colon in absolute IRI");
}
```

knoan: Migrated to `Objects.requireNonNull()`.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
knoan: Technically speaking that annotation prevents IntelliJ automated QA from complaining about the use of `Date` and `Calendar`. Probably also useful as a human-readable reminder of the need to upgrade (e.g. https://github.com/eclipse/rdf4j/issues/2563).

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractIRI.java
knoan: Done

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
knoan: See #???

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractValueFactory.java
knoan: See #???

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractStatement.java
knoan: `!!!` is just a leftover personal annotation, now removed

knoan: `Objects.hash()` is a variadic method: invoking it requires the instantiation of an array containing arguments, which in the context of `.hasCode()` has a large performance impact (more than 25% according to my rough figures).

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractValueFactory.java
jeenbroekstra: I don't think we have a consistent standard for this throughout the project - I tend to prefer the more compact form myself but there's some room for personal preference here :)

knoan:
> Are we aiming to the make the entire Simple... stuff deprecated? Or are we aiming to have all Simple concrete classes derive from these new abstract classes? If so, what exactly can they inherit?

That was my original concern: the current `Simple*` implementation exposes lots of internals in the form of specialised literal classes, making it very difficult if ever possible to factor it to the `model.base` package without reimplementing the whole hierarchy, which would negate the initial goal.

As far as i can see, the new `AbstractValueFactory` and the hidden concrete implementations it serves could provide a useful starting point with a refactored code base, if the plan is to eventually deprecate the current implementations in `model.impl`, but that is imho definitely beyond the scope of this PR.

The current implementation could stil be useful in easing interoperability, as it would make it possible for third parties to depend only on `model.base` with no need to reimplement `ValueFactory` times and again.

That said the partial overlapping with `model.impl` is real and must be evaluated before possibly merging.

knoan:
Left latest changes in a separate commit for your convenience. Will eventually squash.

hmottestad:
Note to self: Check performance of the sorting optimizations I made a while back to see if these changes have any negative effect.

jeenbroekstra:
> > Are we aiming to the make the entire Simple... stuff deprecated? Or are we aiming to have all Simple concrete classes derive from these new abstract classes? If so, what exactly can they inherit?
> 
> That was my original concern: the current `Simple*` implementation exposes lots of internals in the form of specialised literal classes, making it very difficult if ever possible to factor it to the `model.base` package without reimplementing the whole hierarchy, which would negate the initial goal.
> 
> As far as i can see, the new `AbstractValueFactory` and the hidden concrete implementations it serves could provide a useful starting point with a refactored code base, if the plan is to eventually deprecate the current implementations in `model.impl`, but that is imho definitely beyond the scope of this PR.

Agreed. 

> The current implementation could stil be useful in easing interoperability, as it would make it possible for third parties to depend only on `model.base` with no need to reimplement `ValueFactory` times and again.
> That said the partial overlapping with `model.impl` is real and must be evaluated before possibly merging.

True. And don't get me wrong: I would be fine with grandfathering out the Simple... implementations. We just need to be careful we keep things maintainable (and backward compatible, and performant, and... :)).

jeenbroekstra:
As an aside Ive disabled the Jenkins pr check again, don't let that red cross distract you.

jeenbroekstra:
I finally found some time to take another look at this. I think that overall I am happy to have this merged, but I will take another look and perhaps make one or two more tweaks before doing so, marking a few things deprecated and/or experimental. 

Given that we are not actually _replacing_ the SimpleValueFactory with any of this, there will be no direct impact on performance: this new AbstractValueFactory will, right now, only be used by users who deliberately choose it.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractBNode.java
jeenbroekstra: @knoan fwiw I put this up for discussion and seeing if there are some best practices we can agree upon on the rdf4j-dev mailinglist, feel free to chime in there. For now I'm fine with just some developer discretion, as in: feel free to use what you think makes the code easier to read / maintain, as long as nobody objects you're good :)

jeenbroekstra (approved):

knoan:
> Given that we are not actually replacing the SimpleValueFactory with any of this, there will be no direct impact on performance

`Simple*` values actually inherit from their `Abstract*` counterparts, but the original `.equals()/.hashCode()` implementations are preserved, so I don't expect either any direct impact on performance.

Issue #1942, opened 2020-02-22
Custom character encoding in RDFWriter/QueryResultWriter


erikgb:
**Background**
I am developing a simple REST service on top of our SPARQL endpoint, to provide our business people with some prepared SPARQL queries - with input parameters. The REST service supports various RDF serialization formats - based on the formats supported by RDF4J. It turns out that the CSV has become the most popular output format for tuple queries - because the users want the results in MS Excel. MS Excel will happily open any file with the csv file extension, but the problem is that MS Excel assumes CP-1250 (Windows-1250) character encoding. This becomes an issue when all character format writers in RDF4J produces UTF-8 characters....

**No solution**
"Transform" the character stream based on the _Accept-Encoding_ request HTTP header supplied to our REST service. Currently we are just supplying the HttpServletResponse#OutputStream to the selected writer, based on the _Accept_ request HTTP header. This is something that works good, and makes the code extremly simple.

**Minimal solution**
File a PR that allows me to supply a `java.io.Writer`, configured with the "correct" character encoding, to `SPARQLResultsCSVWriter` and `SPARQLResultsTSVWriter`. The problematic lines are here:
https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/queryresultio/text/src/main/java/org/eclipse/rdf4j/query/resultio/text/csv/SPARQLResultsCSVWriter.java#L48
https://github.com/eclipse/rdf4j/blob/d2a301dabcf9c051b019e3434327c15c29fc321e/core/queryresultio/text/src/main/java/org/eclipse/rdf4j/query/resultio/text/tsv/SPARQLResultsTSVWriter.java#L52

**New writer setting solution**
I have been looking a bit at the source code in RDF4J and the specifications for the RDF serialization formats, and I think that there are a few formats that allows non-UTF-8 character encoding. What do you think about introducing a new `BasicWriterSettings`/`BasicQueryWriterSettings`? I would suggest the name CHARACTER_ENCODING for this new hypothetical setting. And I think the following formats could potentially support this new setting:
- RDF/XML
- JSON-LD
- RDF/JSON
- RDFa
- SPARQL/XML
- SPARQL/JSON
- SPARQL/CSV
- SPARQL/TSV

I could eventually prepare a PR, but before spending time on this I would like to know what you think.

hmottestad:
[...]

PR #2106, opened 2020-04-19
GH-1942 Made sure all character format writers have Writer constructor

erikgb:
GitHub issue (partly) resolved: #1942

Briefly describe the changes proposed in this PR:

- Added `Writer` constructors to all character format `QueryResultWriter` classes
- Consistent handling of `Writer` (or `Writer` based) fields from `OutputStream` and `Writer` constructors in `QueryResultWriter` and `RDFWriter` classes

---- 
PR Author Checklist: 

 - [x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [ ] I've added tests for the changes I made
 - [x] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change
 - [x] every commit has been [signed off](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)

Note: we merge all feature pull requests using [squash and merge](https://help.github.com/en/github/administering-a-repository/about-merge-methods-on-github#squashing-your-merge-commits). See [RDF4J git merge strategy](https://rdf4j.org/documentation/developer/merge-strategy/) for more details.

PR #2617, opened 2020-10-23
GH-1942 Rio RDF and QueryResult Writer support for non-default character encoding

jeenbroekstra:
GitHub issue resolved: #1942  

- Added Writer constructors to all character format QueryResultWriter classes
- Consistent handling of Writer (or Writer based) fields from OutputStream and Writer constructors in QueryResultWriter and RDFWriter classes
- added roundtrip test for writers/parsers that support character encoding-based serializations

Contains contributions by @erikgb (from original PR #2106 ), with minor updates.

----

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.StringReader;
+import java.io.Writer;
 import java.net.HttpURLConnection;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
@@ -92,6 +93,8 @@ import org.eclipse.rdf4j.query.resultio.TupleQueryResultWriter;
 import org.eclipse.rdf4j.query.resultio.helpers.BackgroundTupleResult;
 import org.eclipse.rdf4j.query.resultio.helpers.QueryResultCollector;
 import org.eclipse.rdf4j.repository.RepositoryException;
+import org.eclipse.rdf4j.rio.ByteSink;
+import org.eclipse.rdf4j.rio.CharSink;
 import org.eclipse.rdf4j.rio.ParserConfig;
 import org.eclipse.rdf4j.rio.RDFFormat;
 import org.eclipse.rdf4j.rio.RDFHandler;
@@ -685,15 +688,24 @@ public class SPARQLProtocolSession implements HttpClientDependent, AutoCloseable
 						.orElseThrow(() -> new RepositoryException(
 								"Server responded with an unsupported file format: " + mimeType));
 
-				// Check if we can pass through to the output stream directly
+				// Check if we can pass through to the writer directly
 				if (handler instanceof TupleQueryResultWriter) {
 					TupleQueryResultWriter tqrWriter = (TupleQueryResultWriter) handler;
 					if (tqrWriter.getTupleQueryResultFormat().equals(format)) {
-						OutputStream out = tqrWriter.getOutputStream().orElse(null);
-						if (out != null) {
-							InputStream in = response.getEntity().getContent();
-							IOUtils.copy(in, out);
-							return;
+						if (tqrWriter instanceof CharSink) {
+							Writer w = ((CharSink) tqrWriter).getWriter();
+							if (w != null) {
+								InputStream in = response.getEntity().getContent();
+								IOUtils.copy(in, w, getResponseCharset(response).orElse(format.getCharset()));
+								return;
+							}
+						} else if (tqrWriter instanceof ByteSink) {
+							OutputStream out = ((ByteSink) tqrWriter).getOutputStream();
+							if (out != null) {
+								InputStream in = response.getEntity().getContent();
+								IOUtils.copy(in, out);
+								return;
+							}
 						}
 					}
 				}
@@ -853,15 +865,24 @@ public class SPARQLProtocolSession implements HttpClientDependent, AutoCloseable
 				RDFFormat format = RDFFormat.matchMIMEType(mimeType, rdfFormats)
 						.orElseThrow(() -> new RepositoryException(
 								"Server responded with an unsupported file format: " + mimeType));
-				// Check if we can pass through to the output stream directly
+				// Check if we can pass through to the writer directly
 				if (handler instanceof RDFWriter) {
 					RDFWriter rdfWriter = (RDFWriter) handler;
 					if (rdfWriter.getRDFFormat().equals(format)) {
-						OutputStream out = rdfWriter.getOutputStream().orElse(null);
-						if (out != null) {
-							InputStream in = response.getEntity().getContent();
-							IOUtils.copy(in, out);
-							return;
+						if (rdfWriter instanceof CharSink) {
+							Writer w = ((CharSink) rdfWriter).getWriter();
+							if (w != null) {
+								InputStream in = response.getEntity().getContent();
+								IOUtils.copy(in, w, getResponseCharset(response).orElse(format.getCharset()));
+								return;
+							}
+						} else if (rdfWriter instanceof ByteSink) {
+							OutputStream out = ((ByteSink) rdfWriter).getOutputStream();
+							if (out != null) {
+								InputStream in = response.getEntity().getContent();
+								IOUtils.copy(in, out);
+								return;
+							}
 						}
 					}
 				}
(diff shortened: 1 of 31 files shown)


jeenbroekstra:
Still WIP, I am looking into the idea that Erik had earlier, to introduce a CharSink / ByteSink interface for RDFWriters and QueryResultWriters. Sofar looking promising and we may even be able to do this in a backward-compatible fashion.

jeenbroekstra:
@erikgb I'd appreciate your review.

erikgb:
@jeenbroekstra This was an excellent start of my day! 🍾 Thanks a lot for picking up where I didn't have the bandwidth to finish what I started on. I will of course review! 👍

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -685,15 +688,24 @@ protected void getTupleQueryResult(HttpUriRequest method, TupleQueryResultHandle
...
+				// Check if we can pass through to the writer directly
 				if (handler instanceof TupleQueryResultWriter) {
 					TupleQueryResultWriter tqrWriter = (TupleQueryResultWriter) handler;
 					if (tqrWriter.getTupleQueryResultFormat().equals(format)) {
-						OutputStream out = tqrWriter.getOutputStream().orElse(null);
-						if (out != null) {
-							InputStream in = response.getEntity().getContent();
-							IOUtils.copy(in, out);
-							return;
+						if (tqrWriter instanceof CharSink) {
```
erikgb, line 695: I think the following block is quite long and a bit hard to read. Would it be possible to split it up with some private methods? Or replacing the `instanceof` conditionals with use of the Strategy/Visitor Pattern? I tend to prefer it to get better compile-time checks.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -853,15 +865,24 @@ protected void getRDF(HttpUriRequest method, RDFHandler handler, boolean require
...
+				// Check if we can pass through to the writer directly
 				if (handler instanceof RDFWriter) {
 					RDFWriter rdfWriter = (RDFWriter) handler;
 					if (rdfWriter.getRDFFormat().equals(format)) {
-						OutputStream out = rdfWriter.getOutputStream().orElse(null);
-						if (out != null) {
-							InputStream in = response.getEntity().getContent();
-							IOUtils.copy(in, out);
-							return;
+						if (rdfWriter instanceof CharSink) {
```
erikgb, line 872: I think the following block is quite long and a bit hard to read. Would it be possible to split it up with some private methods? Or replacing the instanceof conditionals with use of the Strategy/Visitor Pattern? I tend to prefer it to get better compile-time checks.

erikgb (changes requested):

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: Yeah good point, not sure about a visitor but I'll DRY it up a bit at least.

core/rio/trix/src/main/java/org/eclipse/rdf4j/rio/trix/TriXWriter.java
```diff
@@ -42,13 +43,13 @@
  *
  * @author Arjohn Kampman
  */
-public class TriXWriter extends AbstractRDFWriter implements RDFWriter {
+public class TriXWriter extends AbstractRDFWriter implements RDFWriter, CharSink {
 
 	/*-----------*
```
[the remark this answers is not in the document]
jeenbroekstra: I'm sort of removing these as I see them but missed this one. Will fix.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: I didn't quite use a strategy pattern in the end but I've managed to simplify things a little, have another look.

core/util/src/main/java/org/eclipse/rdf4j/common/io/Sink.java
```diff
@@ -0,0 +1,27 @@
...
+import org.eclipse.rdf4j.common.lang.FileFormat;
+
+/**
+ * 
+ * A Sink writes data in a particular {@link FileFormat}.
+ * 
+ * @author Jeen Broekstra
+ * @since 3.5.0
+ */
+public interface Sink {
```
jeenbroekstra, line 19: Fwiw I'm open to other/better naming suggestions. "Sink" is _very_ generic.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -1139,6 +1150,35 @@ protected String getResponseMIMEType(HttpResponse method) throws IOException {
...
+	/**
+	 * Gets the character encoding specified in the HTTP headers of the supplied response, if any. For example, if the
+	 * response headers contain <tt>Content-Type: application/xml;charset=UTF-8</tt>, this method will return
+	 * {@link StandardCharsets#UTF_8 UTF-8} as the character encoding.
+	 * 
+	 * @param response the response to get the character encoding from.
+	 * @return the response character encoding, {@link Optional#empty()} if it can not be determined.
+	 */
+	Optional<Charset> getResponseCharset(HttpResponse response) {
+		Header[] headers = response.getHeaders("Content-Type");
```
erikgb, line 1162: There are utilities in Http Client for obtaining the Charset! Something in the line of
```java
HttpEntity entity = response.getEntity();
ContentType contentType = ContentType.get(entity);
if (contentType == null) {
   return Optional.empty();
}
return Optional.ofNullable(contentType.getCharset());
```
Or is there a reason for not using it?

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
```diff
@@ -880,6 +866,31 @@ protected void getRDF(HttpUriRequest method, RDFHandler handler, boolean require
...
+	 * @return {@code true} if the content was passed through, {@code false} otherwise.
+	 * @throws IOException
+	 */
+	private boolean passThrough(HttpResponse response, FileFormat responseFormat, Sink sink)
+			throws IOException {
+		if (responseFormat.equals(sink.getFileFormat())) {
+			InputStream in = response.getEntity().getContent();
+			if (sink instanceof CharSink) {
+				IOUtils.copy(in, ((CharSink) sink).getWriter(),
+						getResponseCharset(response).orElse(responseFormat.getCharset()));
```
erikgb, line 884: Maybe introduce a variable for the used Charset?

erikgb (changes requested):
@jeenbroekstra Much better now! Great job! I added a few more comments, after being able to understand the code. 😉 I think we are getting out of WIP soon.....

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
erikgb: Much better, even if I try to avoid `instanceof` - if possible.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
erikgb: Great job! Much better now!

core/util/src/main/java/org/eclipse/rdf4j/common/io/Sink.java
jeenbroekstra: Thanks, all decent suggestions, but the core problem is that it should cover both RDF data and SPARQL Tuple Query results, so I'm a little uncomfortable prefixing it with "Rdf". I'll go the path of least resistance and stick with Sink for now I think.

core/queryresultio/sparqljson/src/main/java/org/eclipse/rdf4j/query/resultio/sparqljson/AbstractSPARQLJSONWriter.java
```diff
@@ -77,15 +75,26 @@
...
 	protected AbstractSPARQLJSONWriter(OutputStream out) {
-		super(out);
+		this(new OutputStreamWriter(out, StandardCharsets.UTF_8));
+	}
+
+	protected AbstractSPARQLJSONWriter(Writer writer) {
+		this.writer = writer;
 		try {
-			jg = JSON_FACTORY.createGenerator(new OutputStreamWriter(out, StandardCharsets.UTF_8));
+			jg = JSON_FACTORY.createGenerator(writer);
```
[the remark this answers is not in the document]
jeenbroekstra: Good catch - will fix.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: I'm aware of the utilities, but I figured I'd do the minimal change only, so I just copied / adapted this from existing internal methods dealing with the content type. A lot of that code is from the httpclient 3.x days when it didn't have these kinds of utilities (or perhaps we weren't aware of it at the time). 

You're right that it could probably be cleaned up a bit though, I'll take another look at it.

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: My personal rule of thumb is not to introduce a variable unless you refer to it at least twice and/or the code becomes unreadable without it. YMMV but I think this is still in the "easy enough to read" category :)

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
jeenbroekstra: I've had a quick look but I'm not confident about changing how content-type header processing is handled in general without introducing regressions. It needs some better test coverage to get certainty on that and I just don't want to spend the effort right now (it's also somewhat out of scope for this ticket).  So I'll stick with the current approach for now. I have logged a separate ticket to deal with some general cleanup of this part of the code: #2619.

erikgb (approved):
LGTM

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSession.java
erikgb: A programmer will always have a personal preference. 😉 I am fine with that, and agree that this case is in the category "easy enough to read".

Issue #2624, opened 2020-11-02
FedX: Support pass-through of query results to provided TupleQueryResultHandler


aschwarte10:
If a TupleQueryResultHandler is supplied for evaluating a query through `TupleQuery#evaluated(TupleQueryResultHandler)`, FedX should use this whenever possible.

Especially for single source queries (i.e. where the query can be sent to a single source) the results should be passed through. This is a huge performance improvement as the query result does not have to be materialized (and potentially converted) in-memory.

aschwarte10:
PR has been merged, closing this issue

PR #2625, opened 2020-11-02
GH-2624: support pass through of single source federated queries

aschwarte10:
GitHub issue resolved: #2624 

This change provides an optimization to the FedX federation engine to
allow pass-through of query results for single source queries, if a
TupleQueryResultHandler was supplied (i.e. if the query was executed
through TupleQuery#evaluate(handler).

The implementation is done through a helper tuple expression (to
propagate information into the federation evaluation strategy. Also we
use a number of marker interfaces.

The implementation is covered by unit tests whose assertions are very
strict and will ensure to find errors if other parts of the framework
change.

Note that if no pass-through is possible, the same logic as before is
applied.

----

 - [] I've [squashed](https://rdf4j.org/documentation/developer/squashing) my commits down to one or a few meaningful commits

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/structures/FedXTupleQuery.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/structures/FedXTupleQuery.java
@@ -7,15 +7,18 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.federated.structures;
 
+import org.eclipse.rdf4j.federated.algebra.PassThroughTupleExpr;
 import org.eclipse.rdf4j.federated.repository.FedXRepositoryConnection;
 import org.eclipse.rdf4j.federated.util.FedXUtil;
 import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.Dataset;
 import org.eclipse.rdf4j.query.QueryEvaluationException;
+import org.eclipse.rdf4j.query.QueryResults;
 import org.eclipse.rdf4j.query.TupleQueryResult;
 import org.eclipse.rdf4j.query.TupleQueryResultHandler;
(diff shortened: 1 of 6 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/structures/FedXTupleQuery.java
```diff
@@ -48,8 +51,26 @@ public TupleQueryResult evaluate() throws QueryEvaluationException {
...
+
+		if (!passThroughTupleExpr.isPassedThrough()) {
+			// if the result is not passed through to the handler directly,
+			// we need to make sure to report the result. Note that only
+			// SingleSourceQuery instances can be passed through
+			QueryResults.report(tqr, handler);
+		} else {
+			// to be absolutely sure that everything is closed
+			tqr.close();
+		}
```
jeenbroekstra, line 73: Should this be in a try-with-resources block rather than this if-else?

jeenbroekstra (commented):
Looks good to me, have only comment.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/structures/FedXTupleQuery.java
aschwarte10: I think this is not absolutely require, but it doesn't harm either.

In SailTupleQuery it is also

```
TupleQueryResult queryResult = evaluate();
QueryResults.report(queryResult, handler);
```

The only minimal difference here is the call to isPssedThrough which cannot throw an exception.

Do you still think we should have the close in a try with resources block?

jeenbroekstra: Nah, fair enough.

jeenbroekstra (approved):

Issue #2632, opened 2020-11-04
SHACL - RSX target shape bug when using sh:or


hmottestad:
Shape: 

```
ex:shape1
	a sh:NodeShape ;
	rsx:targetShape [
	        sh:property [
        		sh:path ex:info ;
        		sh:or ([ dash:hasValueIn ("blue") ; ] [ dash:hasValueIn ("green") ; ] [ dash:hasValueIn ("red") ; ]) ;
        	]
	];

	sh:property [
	   sh:path ex:info2 ;
	   sh:minCount 1;
	]
	 .
```

The following should match the `rsx:targetShape` and be considered invalid since it is not valid according to the `sh:minCount` restriction:

```
ex:a ex:b ex:validPerson1  .
```

PR #2633, opened 2020-11-04
GH-2632 rsx:targetShape bug when using sh:or or sh:and

hmottestad:
GitHub issue resolved: #2632  

Fixes a bug in rsx:targetShape when suing `sh:or` or `sh:and`.

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
@@ -11,11 +11,13 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.query.algebra.StatementPattern;
+import org.eclipse.rdf4j.query.algebra.Var;
 import org.eclipse.rdf4j.repository.sail.SailRepositoryConnection;
 import org.eclipse.rdf4j.sail.SailConnection;
 import org.eclipse.rdf4j.sail.shacl.ConnectionsGroup;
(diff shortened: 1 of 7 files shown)


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
```diff
@@ -288,9 +290,22 @@ public String buildSparqlValidNodes(String targetVar) {
 
 	@Override
 	public Stream<StatementPattern> getStatementPatterns() {
-		return and
-				.stream()
+		StatementPattern subject = new StatementPattern(
+				new Var("?this"),
```
jeenbroekstra, line 294: Typically variable names do not actually start with a question mark - in SPARQL they do of course but our query parsers all treat it like a marker prefix, not as part of the actual variable name.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
```diff
@@ -288,9 +290,22 @@ public String buildSparqlValidNodes(String targetVar) {
...
+				new Var(UUID.randomUUID().toString()),
+				new Var(UUID.randomUUID().toString())
+		);
+
+		StatementPattern object = new StatementPattern(
+				new Var(UUID.randomUUID().toString()),
```
jeenbroekstra, line 300: Are these randomly generated variable names ever serialized, for example as part of a SPARQL query? If so we'll need a more robust name assignement, because this will produce a variable name that won't be legal in SPARQL.

I'm pretty sure we have ready-made code somewhere (in the query parser utils perhaps) for this kind of thing. There should be a single place where we generate these kind of anonymous vars, so that if we need to change the name-generation logic for any reason, we don't have to change it all over the place.

jeenbroekstra (commented):
I kinda trust that with the tests you have this covered. It's not obvious from the PR itself - the actual shape being described in the issue is not in your diff, so I assume it was previously added. Clearly I don't know enough about the test structure that you have set up for this.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/AndPropertyShape.java
hmottestad: This PR is mostly to fix a bug in the old AST so that I can handel it properly in the new AST. This is essentially just copy-paste from other parts of the code that are built like this, so I don't want to start fixing it up now. I will take a look in the new AST and see if I prefix my vars with `?` or not. And also check the random UUID thing you commented.

Issue #2658, opened 2020-11-14
Add java.time support to Literal/ValueFactory 


knoan:
See https://github.com/eclipse/rdf4j/issues/2563#issuecomment-727184360.

Issue #2655, opened 2020-11-13
SimplValueFactory methods unintentionally show as deprecated


jeenbroekstra:
Caused by deprecation of old AbstractValueFactory in 3.5 M1.

Possible fix is to make SimpleValueFactory inherit from new AbstractValueFactory class directly, putting all old AbstractValueFactory code in SimpleValueFactory itself as overrides.

jeenbroekstra:
Problem: changing the superclass of `SimpleValueFactory` is not binary compatible (even if we copy over all methods).

PR #2659, opened 2020-11-14
GH-2658 Literal/ValueFactory java.time support

knoan:
GitHub issue resolved: #2658

- add Literal.temporalAccessor/AmountValue() methods
- add ValueFactory.createLiteral(TemporalAccessor/Amount) methods
- implement new methods in AbstractLiteral and AbstractValueFactory

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
@@ -8,15 +8,49 @@
 
 package org.eclipse.rdf4j.model.base;
 
+import static java.lang.Math.abs;
+import static java.time.temporal.ChronoField.DAY_OF_MONTH;
+import static java.time.temporal.ChronoField.HOUR_OF_DAY;
+import static java.time.temporal.ChronoField.MINUTE_OF_HOUR;
+import static java.time.temporal.ChronoField.MONTH_OF_YEAR;
+import static java.time.temporal.ChronoField.NANO_OF_SECOND;
+import static java.time.temporal.ChronoField.OFFSET_SECONDS;
+import static java.time.temporal.ChronoField.SECOND_OF_MINUTE;
+import static java.time.temporal.ChronoField.YEAR;
+import static java.time.temporal.ChronoUnit.DAYS;
+import static java.time.temporal.ChronoUnit.HOURS;
+import static java.time.temporal.ChronoUnit.MINUTES;
+import static java.time.temporal.ChronoUnit.MONTHS;
+import static java.time.temporal.ChronoUnit.NANOS;
+import static java.time.temporal.ChronoUnit.SECONDS;
+import static java.time.temporal.ChronoUnit.YEARS;
 import static java.util.Objects.requireNonNull;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.time.DateTimeException;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.time.format.SignStyle;
+import java.time.temporal.ChronoField;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalAccessor;
+import java.time.temporal.TemporalAmount;
+import java.time.temporal.TemporalUnit;
+import java.util.Collection;
+import java.util.EnumMap;
+import java.util.EnumSet;
 import java.util.GregorianCalendar;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import javax.xml.datatype.DatatypeConfigurationException;
 import javax.xml.datatype.DatatypeConstants;
@@ -68,22 +102,6 @@ public abstract class AbstractLiteral implements Literal {
 	private static final IRI XSD_DURATION_DAYTIME = new GenericIRI(XSD, "dayTimeDuration");
 	private static final IRI XSD_DURATION_YEARMONTH = new GenericIRI(XSD, "yearMonthDuration");
 
-	private static final String POSITIVE_INFINITY = "INF";
-	private static final String NEGATIVE_INFINITY = "-INF";
-	private static final String NAN = "NaN";
-
-	private static final ThreadLocal<DatatypeFactory> DATATYPE_FACTORY = ThreadLocal.withInitial(() -> {
-		try {
-
-			return DatatypeFactory.newInstance(); // not guaranteed to be thread-safe
-
-		} catch (DatatypeConfigurationException e) {
-
-			throw new RuntimeException("unable to create datatype factory", e);
-
-		}
-	});
-
 	static boolean reserved(IRI datatype) {
 		return RDF_LANGSTRING.equals(datatype);
 	}
@@ -114,17 +132,7 @@ public abstract class AbstractLiteral implements Literal {
 
 	@Override
 	public boolean booleanValue() {
-		return value(label -> Optional.of(label)
-
-				.map(String::trim)
-
-				.map(normalized -> (normalized.equals("true") || normalized.equals("1")) ? Boolean.TRUE
-						: (normalized.equals("false") || normalized.equals("0")) ? Boolean.FALSE
-								: null
-				)
-
-				.orElse(null)
-		);
+		return value(BooleanLiteral::parseBoolean);
 	}
 
 	@Override
@@ -149,20 +157,12 @@ public abstract class AbstractLiteral implements Literal {
 
 	@Override
 	public float floatValue() {
-		return value(label -> label.equals(POSITIVE_INFINITY) ? Float.POSITIVE_INFINITY
-				: label.equals(NEGATIVE_INFINITY) ? Float.NEGATIVE_INFINITY
-						: label.equals(NAN) ? Float.NaN
-								: Float.parseFloat(label)
-		);
+		return value(NumberLiteral::parseFloat);
 	}
 
 	@Override
 	public double doubleValue() {
-		return value(label -> label.equals(POSITIVE_INFINITY) ? Double.POSITIVE_INFINITY
-				: label.equals(NEGATIVE_INFINITY) ? Double.NEGATIVE_INFINITY
-						: label.equals(NAN) ? Double.NaN
-								: Double.parseDouble(label)
-		);
+		return value(NumberLiteral::parseDouble);
 	}
 
 	@Override
@@ -176,8 +176,18 @@ public abstract class AbstractLiteral implements Literal {
 	}
 
 	@Override
+	public TemporalAccessor temporalAccessorValue() throws DateTimeException {
+		return value(TemporalAccessorLiteral::parseTemporalAccessor);
+	}
+
+	@Override
+	public TemporalAmount temporalAmountValue() throws DateTimeException {
+		return value(TemporalAmountLiteral::parseTemporalAmount);
+	}
+
+	@Override
 	public XMLGregorianCalendar calendarValue() {
-		return value(label -> DATATYPE_FACTORY.get().newXMLGregorianCalendar(label));
+		return value(CalendarLiteral::parseCalendar);
 	}
 
 	@Override
@@ -234,7 +244,7 @@ public abstract class AbstractLiteral implements Literal {
 
 		TypedLiteral(String label, IRI datatype) {
 			this.label = label;
-			this.datatype = (datatype != null) ? datatype : XSD_STRING;
+			this.datatype = datatype != null ? datatype : XSD_STRING;
 		}
 
 		@Override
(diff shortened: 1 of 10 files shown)


knoan:
Mostly done. Comments welcome on tests and on the last commit.

`SimpleValueFactory` now transitively inherits `ValueFactory.createLiteral(TemporalAccessor)` from `oer.model.base.AbstractValueFactory`, so that tests don't break and existing code can be upgraded without migrating to `AbstractValueFactory`; as far as I can see, all other factory methods are overridden, so the existing code base should be safe.

Taking into account that we are adding new functionalities, everything should be ok, but I'd like an additional assessment about returning an `AbstractLiteral` from `SimpleValueFactory`.

knoan:
P.S. Hope this is not late for 3.5.0: if that's not the case, `@since` tags should be updated

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
```diff
@@ -175,6 +333,25 @@ public BigDecimal decimalValue() {
 		return value(BigDecimal::new);
 	}
 
+	@Override
+	public TemporalAccessor temporalValue() throws DateTimeException {
+		return value(label -> {
+
+			final TemporalAccessor value = label.startsWith("--") ? DASH_FORMATTER.parse(label)
+					: label.length() >= 8 && label.charAt(2) == ':' ? OFFSET_TIME_FORMATTER.parse(label)
+							: DATETIME_FORMATTER.parse(label);
```
jeenbroekstra, line 342: Can we not use nested/chained ternaries? YMMV, but I find them hard to read. Might be improved with linebreaks in different places, but I think an if-else is easier on the eye (if a little more verbose):

```java
TemporalAccessor value;
if (label.startsWith("--")) {
    value = DASH_FORMATTER.parse(label);
} else if (label.length() >= 8 && label.charAt(2) == ':' ) {
   value = OFFSET_TIME_FORMATTER.parse(label);
} else {
  value = DATETIME_FORMATTER.parse(label);
}
```

or perhaps more clean:

```java
final TemporalAccessor value = getFormatter(label).parse();

... 

private static DateFormatter getFormatter(String label) {
    if(label.startsWith("--")) {
       return DASH_FORMATTER;
   } 
   if (label.length() >= 8 && label.charAt(2) == ":") {
       return OFFSET_TIME_FORMATTER;     
   }
   return DATETIME_FORMATTER;
}
```

jeenbroekstra (commented):
This is definitely not too late for 3.5, we did a milestone build but we made it clear that it was not considered feature-complete, so there room for this. 

I think it looks very solid, thanks again @knoan. I only really have a few code organization comments (see inline).

core/model/src/main/java/org/eclipse/rdf4j/model/impl/AbstractValueFactory.java
```diff
@@ -39,7 +39,7 @@
  *             {@link org.eclipse.rdf4j.model.base.AbstractValueFactory}.
  */
 @Deprecated
-public abstract class AbstractValueFactory implements ValueFactory {
+public abstract class AbstractValueFactory extends org.eclipse.rdf4j.model.base.AbstractValueFactory {
```
jeenbroekstra, line 42: I actually need to look at this, see #2655. I will likely refactor SimpleValueFactory to directly inherit from the new AbstractValueFactory, copying all the old AbstractValueFactory code over and leaving it orphaned.

knoan: > I actually need to look at this, see #2655. I will likely refactor SimpleValueFactory to directly inherit from the new AbstractValueFactory, copying all the old AbstractValueFactory code over and leaving it orphaned.

As long as SimpleValueFactory inherits from the new AbstractValueFactor tests won't break.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
```diff
@@ -72,6 +87,149 @@
...
+			.appendLiteral(':')
+			.appendValue(MINUTE_OF_HOUR, 2)
+			.appendLiteral(':')
+			.appendValue(SECOND_OF_MINUTE, 2)
+
+			.optionalStart()
+			.appendFraction(NANO_OF_SECOND, 0, 9, true)
+
+			.toFormatter();
+
```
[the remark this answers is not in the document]
knoan: I refactor to migrated all support fields and parsing code to the relevant specialised nested classes: now we are in the position to easily promote them to top level, if we opt to do so.

For the time being, however, I'd leave them inside AbstractLiteral, in order not to relax encapsulation and symmetry with other generic implementations in other Abstract* value classes.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
knoan: Done.

knoan:
Before merging I'd like to add support for XSD duration datatypes: allot me a couple of days.

knoan:
> Before merging I'd like to add support for XSD duration datatypes: allot me a couple of days.

I was able to cram that into an unforeseen free slot, but it was quite nasty ;-P

`java.time` doesn't include `TemporalAmount` implementations able to represent XML schema duration values including both date and time components (e.g. `P1YT1H`): `Period` supports only date-based components and `Duration` supports only time-based ones.

I reckon that's conservatively done in order to prevent average/estimated year/month/day durations to creep into calendrical calculations involving also time-based-components, but that forced us to implement it by ourselves along with the companion parsing/formatting methods: some compatibility corner cases are likely to have done it through tests…

knoan:
P.S. I've already squashed the incremental changes, as they were very fragmented

knoan:
Handled new temporal methods also in delegating ValueFactories.

knoan:
Utilities methods in `Literals` could also be eventually completed, but I'd defer that after the whole class is reviewed for `Simple*` and other suspicious dependencies.

IMHO, most of these utilities could be factored to Optional-based methods on Value/Literal interfaces: hid `AbstractLiteral.value(Function<String, V> mapper)` method to avoid interferences with possible future developments in this direction.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/AbstractValueFactory.java
jeenbroekstra: I've merged my update for #2655. It shouldn't break anything for you but might be worth rebasing, just to make sure.

jeenbroekstra:
> I reckon that's conservatively done in order to prevent average/estimated year/month/day durations to creep into calendrical calculations involving also time-based-components, but that forced us to implement it by ourselves along with the companion parsing/formatting methods: some compatibility corner cases are likely to have done it through the tests…

Wow. Good stuff. So does this mean we are no longer depending on javax.xml.DatatypeFactory to process these kinds of datatype strings? If so that is actually quite a relief, because that thing was ancient and still wasn't fully updated for XSD 1.1.

knoan:
> So does this mean we are no longer depending on javax.xml.DatatypeFactory to process these kinds of datatype strings?

Correct. The relevant code is in the following methods:

- AbstractLiteral.TemporalAccessorLiteral.parseTemporalAccessor(String)
- AbstractLiteral.TemporalAmountLiteral.parseTemporalAmount(String)/toString(TemporalAmount)

There's still room for improvement, but the code is completely hidden, so we are completely at liberty to refactor it as we see fit.

Time permitting, it would also be worth extending parsing/formatting methods, as current codecs were hastily derived from java.time ISO formats, which XSD 2 restricts in a number of ways.

Given that we are dealing with RDF 1.1 literal expected to be well-formed, I'd stick to strict parsing, leaving lenient versions to eventual RIO upgrades.

knoan:
I'm realising that the current code base (which I took as a starting point for this PR) defines calendar values referring to XSD 2:

> A calendar representation can be given for literals whose label conforms to the syntax of the following <a href="http://www.w3.org/TR/xmlschema-2/">XML Schema datatypes</a> …

RDF 1.1 specs on the other hand are still based on XSD 1.1 (https://www.w3.org/TR/rdf11-concepts/#xsd-datatypes).

If we are not in a hurry, I'd review the whole matter to achieve full conformance with RDF 1.1, along the following lines:

- redefine new temporal methods wrt RDF 1.1 / XSD 1.1 (no impact on existing code as we are introducing them anew)
- add parsing/formatting tests based on XSD 1.1 specs

In order to provide better interoperability with XSD 2 (and an easier upgrade path), I'd still ignore temporal datatypes dropped from the latter version of the standard, that is `xsd:dateTimeStamp`, `xsd:yearMonthDuration` and `xsd:dayTimeDuration`.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/AbstractValueFactory.java
knoan: Rebased

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
jeenbroekstra: Great, thanks!

jeenbroekstra:
> I'm realising that the current code base (which I took as a starting point for this PR) defines calendar values referring to XSD 2:

I haven't looked in detail but I believe that to just be a mistake in the javadoc. AFAIK our actual calendar value support is based on XSD 1.1 specs (and in corner cases even only XSD 1.0).

jeenbroekstra:
> If we are not in a hurry, I'd review the whole matter to achieve full conformance with RDF 1.1, along the following lines:
> 
>  - redefine new temporal methods wrt RDF 1.1 / XSD 1.1 (no impact on existing code as we are introducing them anew)
> - add parsing/formatting tests based on XSD 1.1 specs

Making sure we are at least conformant with RDF 1.1 is of course a good idea. 

> In order to provide better interoperability with XSD 2 (and an easier upgrade path), I'd still ignore temporal datatypes dropped from the latter version of the standard, that is xsd:dateTimeStamp, xsd:yearMonthDuration and xsd:dayTimeDuration.

When you say "ignore", would that simply mean that if someone creates a literal with one of those datatypes (and valid lexical value for that type), we won't support `getTemporalValue` on such a literal? In other words: people are free to create, store, and query such literals, and even use the legacy Date/Calendar-based methods to process it, but the newly added support based on the java.time api will not handle such values. Correct?

knoan:
> When you say "ignore", would that simply mean that if someone creates a literal with one of those datatypes (and valid lexical value for that type), we won't support `getTemporalValue` on such a literal?

Legacy XSD 1.1 datatypes were defined as restriction of the base datatypes retained in XSD 2:

- assigning a more specific value to the base type is still formally correct (and IMHO more aligned with current LOD practice);
- if the literal is explicitly created using a label and one of the legacy datatypes (`ValueFactory.createLiteral(String, IRI)`),  `LIteral.getTemporal*Value()` will return the correct temporal value, as long as the label is well-formed;
- if the literal is created using a temporal value (`ValueFactory.createLiteral(TemporalAccessor/Amount)`), `LIteral.getTemporal*Value()` will return the supplied value, while `Literal.getDatatype()` will return the base XSD 1.1/2 datatype rather than the more specific XSD 1.1 datatype later dropped from XSD 2, i.e.:
    - xsd:dateTimeStamp >> xsd:dateTime
    - xsd:yearMonthDuration >> xsd:duration
    - xsd:dayTimeDuration >> xsd:duration

> In other words: people are free to create, store, and query such literals, and even use the legacy Date/Calendar-based methods to process it, but the newly added support based on the java.time api will not handle such values. Correct?

knoan:
> > I'm realising that the current code base (which I took as a starting point for this PR) defines calendar values referring to XSD 2:
> 
> I haven't looked in detail but I believe that to just be a mistake in the javadoc. AFAIK our actual calendar value support is based on XSD 1.1 specs (and in corner cases even only XSD 1.0). Or were you referring to your new implementation being based on XSD 2?

I was mislead by the reference to XSD 2 in the javadocs and derived new conformance tests from that specs rather than XSD 1.1: I don't really expect any deviation, but I'd rather review the whole matter.

knoan:
I reviewed XSD 1.1/2 conformance, confirming the overall scheme already discusessed in https://github.com/eclipse/rdf4j/pull/2659#issuecomment-729512460:

- `Literal.temporalAccessorValue()` parses lexical values according to XSD 1.1 rules for date/time datatypes 
- `Literal.temporalAmountValue()` parses lexical values according to XSD 2 rules for the `xsd:duration` datatype
- `ValueFactory.createLiteral(TemporalAccessor)` assigns an XSD 1.1 date/time datatype ignoring `xsd:dateTimeStamp`
- `ValueFactory.createLiteral(TemporalAmount)` always assigns the `xsd:duration` datatype ignoring more specific XSD 1.1 duration datatypes later dropped from XSD 2

While not strictly compliant with RDF 1.1, this approach is simpler, way easier to implement and (imho) more aligned with current LOD practice; known deviation are documented in javadocs.

Tests reviewed as well: some corner cases likely to be still slipping through, but probably not worth further immediate effort.

As far as I'm concerned ready for final reviews.

jeenbroekstra (approved):
All looks very solid to me. Great work @knoan ! 

I think we have to start considering a proper 3.5 release soon, so that we can get this out to the masses...

Issue #2654, opened 2020-11-13
Use Jetty instead of Tomcat


barthanssens:
Check if Google Jib maven plugin can be used to build and push a jetty image for the workbench and server wars, instead of tomcat

barthanssens:
Starting point: Jetty 9 and JRE 11-slim. 

If this works, it also reduces the image size from 385 MB (Tomcat) to 285 MB (Jetty with Jib)

```
<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>jib-maven-plugin</artifactId>
    <version>2.6.0</version>
    <configuration>
        <container>
           <user>jetty:jetty</user>
           <entrypoint>
                <arg>java</arg>
                <arg>-jar</arg>
                <arg>/usr/local/jetty/start.jar</arg>
           </entrypoint>
           <volumes>
                <volume>/var/lib/jetty/.RDF4J</volume>
           </volumes>
        </container>
        <from>
           <image>jetty:9-jre11-slim</image>
        </from>
        <to>
           <image>barthanssens/rdf4j-jetty:test</image>
        </to>
        <extraDirectories>
           <paths>
                <path>
                     <from>/home/bart/rdf4j/assembly/target/eclipse-rdf4j-${project.version}/war/</from>
                     <into>/var/lib/jetty/webapps</into>
                </path>
           </paths>
           <permissions>
                <permission>
                     <file>/var/lib/jetty/.RDF4J</file>
                     <mode>755</mode>
                </permission>
           </permissions>
        </extraDirectories>
    </configuration>
</plugin>
```

barthanssens:
Related to #2391

barthanssens:
Hmz, a bit out of luck since docker only provides official images for Jetty 9.4.26 and up for 9.4.x (and I would not go back to 9.3...), no 9.4.24....

So I'm looking into upgrading to 9.4.33/4/5 in #2666

barthanssens:
Closing this one, less of an issue now we have a spring application

PR #2665, opened 2020-11-17
GH-2654: initial pom for Jib/Jetty/JDK11

barthanssens:

GitHub issue resolved: #2654  

- use Google JIB to create docker images without docker daemon, and push images to hub.docker.io
- ARM and Intel 64-bit in one push
- version + latest tag
- use JRE 11 instead of JRE 8
- use Jetty 9.4 instead of Tomcat

----


barthanssens:
Current issues:

- no HTTPS yet
- for testing purposes, uses my hub docker instead of Eclipse's
-  uses stock Jetty 9.4.33 images, but RDF4J build has issues with Jetty 9.4.24+ (some tests fail)
- Jib does not work (by design) in a pom.xml with package "pom", hence the need for a dummy submodule / pom.xml and a dummy target/classes directory

jeenbroekstra:
> Current issues:
> 
> * no HTTPS yet

You mean for the actual jetty server being run in the docker container we produce? I don't think the currently manually produced one supports that either, right? Let's treat that as a separate feature request.

> * for testing purposes, uses my hub docker instead of Eclipse's

How does this plugin authenticate itself with dockerhub? Does it require that you are logged through a `docker login` command or something? 

Also, what maven commands do you use to actually build the docker images, and push them? It would be ideal if we could separate these out btw, so that, for example, we can build docker images locally for snapshots without pushing them to dockerhub. 

> * uses stock Jetty 9.4.33 images, but RDF4J build has issues with Jetty 9.4.24+ (some tests fail)

To separate that out, can we (temporarily) just use a slightly older image, until we've gotten to the root of the test failure issues?

> * Jib does not work (by design) in a pom.xml with package "pom", hence the need for a dummy submodule / pom.xml and a dummy target/classes directory

That's really strange. Why would the packaging type of the module have any influence on whether or not the plugin gets executed? 

That being said I don't object to a separate module for this, if that's the easiest way to get it set up, though I'm a little uncomfortable with it sitting at the project root level. We could have a quick look at reorganizing our assembly module a little, perhaps turn it into a set of submodules (sdk, onejar, docker).

barthanssens:
Indeed, currently there is no HTTPS version for our Tomcat image either.

mvn jib:build should do the trick, though it requires at least a dummy target/classes directory to be present (so mvn clean messes that up)

Pushing indeed relies on e.g. docker login, though other options are available: https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/README.md#authentication-methods

Jib not working when set to pom is indeed... weird... see also https://github.com/GoogleContainerTools/jib/issues/2533.

OTOH Jib seems to do some smarter stuff when packaging is war (using a embedded Jar), but not tested how this would work for two wars (server and workbench) ... https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/README.md#war-projects

jeenbroekstra:
Random thought: should we put this plugin into the rdf4j-workbench and/or rdf4j-server modules, instead of in its own assembly module?

barthanssens:
Hmz, to create separate images for workbench and server ? Could be very useful, especially for the server...

Haven't tested if Jib allows to bundle multiple war's when the packaging is set to war: looks like Jib uses a distroless Jetty, so I have to test it...

jeenbroekstra:
@barthanssens this has been sitting idle for a while, is it still on your radar?

barthanssens:
Well, given that the deployment is now much smoother thanks to the scripts Havard and you have written, it's not that useful anymore, so I'll close it

Issue #2667, opened 2020-11-19
Workbench + Console cannot load N-Triple file


tioannid:
Hi,
My configuarion is: VirtualBox v6.0.22 Guest Ubuntu 18.04LTS, 24GB Ram, 4 v-cores, RDF4J Server 3.4.3+7ffffb7, JDK 1.8.0_211, Tomcat v9.0.1.
I am trying to load the following GeoSPARQL N-Triple file (basic1.nt) which contains absolute URIs with Workbench and Console, but it gives error messages.
When I introduce namespace prefixes the same triples (basic1.ttl) which now contain relative IRIs are imported. Obviously this is not N-Triples format, it's Turtle.
In both cases I use the default NativeStore with default vaules.

[BasicTest.zip](https://github.com/eclipse/rdf4j/files/5568023/BasicTest.zip)
I am uploading both files.

For the basic1.nt file I get the following error from the workbench:
![WorkbenchLoadError_basic1_nt](https://user-images.githubusercontent.com/15914587/99691337-cc3b6080-2a91-11eb-911a-b8d9d06504eb.PNG)

For the basic.nt file I get the following error from the verify command of the console:
```
RDF Format is N-Triples
Verifying data...
FATAL ERROR: Not a valid (absolute) IRI: someroad  [line 1]
Unexpected RDFParseException: Not a valid (absolute) IRI: someroad [line 1]
```

For the basic.ttl file I get the following error from the verify command of the console:
```
RDF Format is Turtle
Verifying data...
Data verified, no errors were found
File contains 5 statements
```

Regards

tioannid:
Additionally, I found out that by adding "http://" prefix on the subject IRIs of the basic1.nt N-Triples file the records are imported!!

barthanssens:
Hi,

thanks for reporting.

I have to check the specs, but IIRC N-Triples can only contain absolute IRIs (or blank node) subjects, so basically the baseURI gets ignored when importing N-Triples files.

barthanssens:
Admittedly it may need some more attention in the documentation and tools, but indeed, N-Triple files may only contain absolute IRIs https://www.w3.org/TR/n-triples/#sec-iri, 2.2:

> IRIs may be written only as absolute IRIs

@tioannid  You mentioned that you were able to upload the file when changing the subject IRIs , so would it OK for you if we close the ticket ?

PR #2668, opened 2020-11-19
GH-2667 n-triples baseuri doc

barthanssens:

GitHub issue resolved: #2667 

Much of the N-Triples parser code is copy/paste of the Turtle parser code, however baseURI is never called / used (which makes sense, because N-Triples does not support relative IRIs)

* changed javadoc to mention that baseURI is ignored
* remove the not null check for baseURI, because it is not used anyway

----


jeenbroekstra (commented):
Looks fine to me - can you squash before merging?

I've also been thinking that we should add some overloads to the various parse / write methods that require specifying a base URI argument anyway. In 9 out of 10 cases you don't want to set a base URI at all, and it's annoying to always have to set it to the empty string. I'll log a followup for that.

Issue #2676, opened 2020-11-26
TurtleWriter: does not buffer an OutputStreamWriter


JervenBolleman:
As the [javadoc suggests](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/OutputStreamWriter.html) that is rather inefficient.

JervenBolleman:
I have a pull request incoming for this.

PR #2677, opened 2020-11-26
GH-2676 add BufferedWriter to avoid charset encoding overhead in OutputStreamWriter

JervenBolleman:
…to avoid needing

to encode latin1 strings

GitHub issue resolved: #2676

This buffers the outputwriter in the turtle writer which avoids a lot of Charset encoder overhead.
It also adds one method to the IndentingWriter that postpones when a String is converted into char[].
That might be useful if the String is an latin1 string and coversion to UTF-8 is a nop in that case.

In my testing it roughly doubles the speed at which a turtle file is writen.

----


jeenbroekstra (approved):
Good catch @JervenBolleman , thanks for the fix, looks good to me. 

We may need to do a similar fix for some of the other writers. I'll take a look.

Issue #1325, opened 2019-03-07
Provide static factory methods for Value creation


reckart:
The SparqlBuilder has some nice factory classes such as `Rdf` which allows for nicely compact code:

```
Rdf.iri("http://www.openrdf.org/contrib/lucenesail#")
```

or even 

```
import static org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf.iri;

iri("http://www.openrdf.org/contrib/lucenesail#")
```

This is much more convenient than the RDF4J approach which requires obtaining a value factory instance. It is also more compact because less verbose method names are chosen (`iri(...)` vs `createIri(...)`):

```
ValueFactory factory = SimpleValueFactory.getInstance();
factory.createIRI(iriString)
```

See also: https://github.com/eclipse/rdf4j/issues/1038#issuecomment-470449531

jeenbroekstra:
Valuefactory is a central component in the framework and it's unlikely that we can replace it. However, it should be possible to provide convenience adapers that make this possible.

jeenbroekstra:
Fwiw in which use cases/how often do you find yourself using the ValueFactory directly? 

Reason I ask is that in several places we have already provided convenience utilities or recommended patterns to make this easier. One example is the `ModelBuilder`, which has extensive support for minting new IRIs on the fly. It would be good to understand what angle we haven't covered.

reckart:
We have ~70 matches for `SimpleValueFactory` in INCEpTION. It's partially being used during manual query construction (a use-case which will be removed and replaced by using the `SparqlBuilder`now). The rest of the uses seem to be related to the fact that it was chosen to use RDF4J IRIs in some of our data objects in favor of plain old Strings. My comment relates mostly to the pre-SparqlBuilder query construction which looks somewhat like this:

```
            String QUERY = <SOME SPARQL QUERY>;
            ValueFactory vf = SimpleValueFactory.getInstance();
            TupleQuery tupleQuery = conn.prepareTupleQuery(QueryLanguage.SPARQL, QUERY);
            tupleQuery.setBinding("s", vf.createIRI(aIdentifier));
            ...
```

jeenbroekstra:
I'm thinking of a design where we provide a set of static functions that internally all use the SimpleValueFactory (though optionally injectable with another ValueFactory implementation), e.g. something like:

               public class Values {
                    public static IRI iri(String iriRef) {
                        return iri(SimpleValueFactory.getInstance(), iriRef);
                    }                
[...]

PR #2680, opened 2020-11-28
GH-1325 added Values convenience functions for quick IRI, Literal, BNode, and Triple creation

jeenbroekstra:
GitHub issue resolved: #1325  

- A new utility class `org.eclipse.rdf4j.model.util.Values` with static factory methods for quick creation of all value types. This allows easier code at the client end, e.g. instead of 

```java
   import org.eclipse.rdf4j.model.impl.SimpleValueFactory;

   IRI example =    SimpleValueFactory.getInstance().createIRI("http://www.example.org/");
```

We can now do 

```java
   import org.eclipse.rdf4j.model.util.Values;

   IRI example =    Values.iri("http://www.example.org/");
```

or even:

```java
   import static org.eclipse.rdf4j.model.util.Values.iri;

   IRI example =  iri("http://www.example.org/");
```

- The utility methods are deliberately quite strict about checking syntactic validity of input arguments (including checks for nulls). 

- Migrated several existing static methods from `Statements` and `Literals` into `Values` for consistency, marked existing methods deprecated.

- Updated documentation for Model API and Repository API, as well as (part of) the Tutorial
- As a side effect / cleanup thing, made the examples directory a proper maven module so we can more easily keep it up to date (and extend the set of code examples as well) 

----

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/Values.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/Values.java
@@ -10,7 +10,10 @@ package org.eclipse.rdf4j.model.util;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.time.temporal.TemporalAccessor;
+import java.util.Date;
 import java.util.Objects;
+
+import javax.xml.datatype.XMLGregorianCalendar;
 
 import org.eclipse.rdf4j.model.BNode;
 import org.eclipse.rdf4j.model.IRI;
@@ -25,7 +28,8 @@ import org.eclipse.rdf4j.model.impl.ValidatingValueFactory;
 import org.eclipse.rdf4j.model.vocabulary.XSD;
 
 /**
- * Convenience functions to quickly create {@link Value} objects without having to create a {@link ValueFactory} first.
+ * Factory methods to quickly create {@link Value} objects ( {@link IRI}, {@link Literal}, {@link BNode}, and
+ * {@link Triple}) without having to create a {@link ValueFactory} first.
  * <p>
  * Example usage:
  * 
@@ -39,6 +43,8 @@ import org.eclipse.rdf4j.model.vocabulary.XSD;
  * 
  * @author Jeen Broekstra
  * @since 3.5.0
+ * 
+ * @see Statements
  */
 public class Values {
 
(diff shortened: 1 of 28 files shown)


jeenbroekstra:
@hmottestad I know you only have very limited time at the moment, don't feel obligated to review this, merely adding you as a courtesy.

site/content/documentation/programming/model.md
```diff
@@ -49,7 +78,7 @@ The above interfaces and classes show how we can create the individual building
 
 ```java
 // create a new Model to put statements in
-Model model = new LinkedHashModel();
+Model model = DynamicModelFactory.createEmptyModel();
```
hmottestad, line 81: :)

hmottestad (approved):
This looks really sweet!

core/model/src/main/java/org/eclipse/rdf4j/model/util/Values.java
```diff
@@ -0,0 +1,694 @@
...
+	 */
+	private static Literal createLiteralFromObject(ValueFactory valueFactory, Object object,
+			boolean throwExceptionOnFailure)
+			throws IllegalArgumentException {
+		Objects.requireNonNull(valueFactory, "valueFactory may not be null");
+		Objects.requireNonNull(object, "object may not be null");
+
+		if (object instanceof Boolean) {
+			return valueFactory.createLiteral(((Boolean) object).booleanValue());
+		} else if (object instanceof Byte) {
```
aschwarte10, line 667: I am not a 100% sure but it may a good idea to also test for the primitive types here, i.e. `byte.class.equals(object.getClass()) `, similar for int, double etc

Might also be a good idea to cover this in tests if not already done so

aschwarte10 (commented):
This is a really good improvement and will help developers using RDF4J a lot to simplify their code.

Except for one comment the change looks good to me, please have a look at the inline comment and check if a change is required

core/model/src/main/java/org/eclipse/rdf4j/model/util/Values.java
jeenbroekstra: The argument here can never be a primitive type, since it is an `Object`. Instead, to cover primitives, the `literal` method has been overloaded with functions that _explicitly_ expect a primitive type. 

I think most of this is covered in the tests as well.

It's perhaps worth pointing out that due to autoboxing, we wouldn't even need the `literal(Object)` method if all we wanted was to support feeding in an object that corresponds to a primitive type: having the methods accept primitives is enough for that, the compiler would take care of the rest for us. 

What this `literal(Object)` method gives us extra is that you can feed it _any_ object, without needing to do type checking first, and you get back a literal.

Issue #2666, opened 2020-11-18
Jetty 9.4.24+ fails on federated service test


barthanssens:
When using Jetty 9.4.25 - 9 4.33 in the pom.xml, the eclipse.rdf4j.federated.ServiceTests.test11_errorHandling fails when running mvn clean install:

```The target server failed to respond. Caused by: org.apache.http.NoHttpResponseException: The target server failed to respond```

9.4.24 works just fine, so it looks like it is due to a chance in Jetty 9.4.25

barthanssens:
With jetty 9.4.33, "tools/federation/src/test/resources/tests/service/query11_error_b.rq" fails

barthanssens:
Or perhaps the connection gets closed upon error :-/

barthanssens:
Can be fixed by removing `disableAutomaticRetries` option in  `org.eclipse.rdf4j.http.client.SharedHttpClientSessionManager.createHttpClient`

barthanssens:
Looks like the combination of Apache HTTP Client 4.5.x and Jetty 9.4.24+ server is a bit problematic.

Current versions of Jetty  seem to be more aggressive at closing connections and/or HTTP Client does not notice this fast enough, resulting in a stale connection and errors when trying to (re)use such a connection.

a) In earlier versions of HTTP Client (4.3 ? 4.4 ?), it was checked if a connection was stale or not on every request, but this added a delay (tens of ms ?), which adds up when there are a lot of requests. So this was disabled by default.
It can be enabled again, but the method to do this is deprecated in favor of a more complex solution (see point c) below), and this would again add some delay.

b) We also disabled the more generic retry (see #647), which simply retries when a request fails for whatever reason. The downside is that this may cause issues in e.g non-idempotent POST requests. While it is possible to enable this for "safe" (idempotent) requests, this will still not fix the issue when e.g. a POST request wasn't sent at all in the first place.
A custom retry handler could be easily implemented, IF the error thrown by HTTP Client is specific enough (to be checked)

c) HTTP Client suggests to use a Basic or PoolingHttpClientConnectionManager and a separate monitoring thread to check after x ms (instead of on every request, so on average there is less delay) and evict stale connections, but this seems more complicated than it should IMHO

Issue #647, opened 2016-11-11
"Cannot retry request with a non-repeatable request entity" exception when uploading triples


rune66:
I have made a fresh install of rdf4j 2.1.1 on Tomcat 8.5.

Now I'm trying to add some triples using the workbench->add upload which doesn't work. I get the following exception:

org.apache.http.client.ClientProtocolException
	org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:186)
	org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:82)
	org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:55)
	org.eclipse.rdf4j.http.client.SparqlSession.execute(SparqlSession.java:1102)
	org.eclipse.rdf4j.http.client.SparqlSession.executeNoContent(SparqlSession.java:1083)
	org.eclipse.rdf4j.http.client.SesameSession.upload(SesameSession.java:845)
	org.eclipse.rdf4j.http.client.SesameSession.upload(SesameSession.java:672)
	org.eclipse.rdf4j.http.client.SesameSession.upload(SesameSession.java:661)
	org.eclipse.rdf4j.repository.http.HTTPRepositoryConnection.add(HTTPRepositoryConnection.java:466)
	org.eclipse.rdf4j.workbench.commands.AddServlet.add(AddServlet.java:101)
	org.eclipse.rdf4j.workbench.commands.AddServlet.doPost(AddServlet.java:52)
	org.eclipse.rdf4j.workbench.base.TransformationServlet.service(TransformationServlet.java:96)
	org.eclipse.rdf4j.workbench.base.AbstractServlet.service(AbstractServlet.java:125)
	org.eclipse.rdf4j.workbench.proxy.ProxyRepositoryServlet.service(ProxyRepositoryServlet.java:109)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchServlet.service(WorkbenchServlet.java:213)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchServlet.handleRequest(WorkbenchServlet.java:141)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchServlet.service(WorkbenchServlet.java:109)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchGateway.service(WorkbenchGateway.java:120)
	org.eclipse.rdf4j.workbench.base.AbstractServlet.service(AbstractServlet.java:125)
[...]

PR #2684, opened 2020-12-01
GH 2666 jetty federated fix

barthanssens:
GitHub issue resolved: #2666 

- bump Jetty and HttpClient version
- added custom retry handler to (only) retry stale connections, which became an issue for Jetty 9.4.24+
(doesn't seem to occur that often, but is happening in federated tests)

There are different ways to solve this, but 
- retries were explicitly disabled in #647 (including omnipotent requests)
- Apache HttpClient offers the option to retry stale connections at `RequestConfig` level, but this method `setStaleConnectionCheckEnabled` has been deprecated
- another HttpClient option is a ConnectionPool with a separate monitoring thread, but this seems to add more complexity and may need additional (sizing) configuration

----

### DIFF
--- a/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
+++ b/core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.http.client;
 
+import java.io.IOException;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
```diff
@@ -57,6 +62,39 @@
...
+			if (conn.isStale()) {
+				try {
+					logger.warn("Closing stale connection");
+					conn.close();
+					return true;
+				} catch (IOException e) {
+					logger.error("Error closing connection", e);
+				}
+			}
+			return false;
```
jeenbroekstra, line 94: I haven't looked in detail, but is this thread-safe? If (potentially) not, perhaps slap a synchronized block on it to be sure?

jeenbroekstra (commented):
Mostly looks good to me, good find @barthanssens. One minor inline remark only.

Can you squash your PR when you're happy with it?

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/SharedHttpClientSessionManager.java
barthanssens: Good point, I'll have to check if it is

barthanssens: added synchronize just to be sure

PR #2699, opened 2020-12-09
do not run api comparison checks in build phase

jeenbroekstra:
- should only happen in verify/test phase
- this avoids problems with module reorg across versions


jeenbroekstra:
Possibly a more structured fix is to make sure the japicmp plugin is only run as part of the 'compliance' profile, but right now I just want a quick fix to get the branches stable again, and the 3.5 release out the door.

PR #2712, opened 2020-12-10
Remove japicmp check for modules that won't build

hmottestad:
Quick fix to build issue while @jeenbroekstra manages to fix it properly.

### DIFF
--- a/.github/workflows/develop-status.yml
+++ b/.github/workflows/develop-status.yml
@@ -2,13 +2,13 @@ name: develop status
 
 on:
   push:
-    branches: 
+    branches:
       - develop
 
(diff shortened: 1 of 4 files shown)


hmottestad:
@jeenbroekstra if this builds I'm going to merge it so that anyone trying to work on rdf4j can continue to do so without getting build errors. I spent maybe an hour today trying to figure out what was wrong, and I would hate for others to have to go through the same.

jeenbroekstra:
Please don't.

jeenbroekstra:
See https://github.com/eclipse/rdf4j/pull/2713

Issue #2714, opened 2020-12-10
Benchmarks to show overhead of BindingSet


hmottestad:
We need benchmarks to show the overhead that BindingSet has. To do this we can measure the difference when doing a task in SPARQL and doing the same task on Statements directly.

PR #2715, opened 2020-12-10
GH-2714 benchmarks to show overhead of BindingSet

hmottestad:
GitHub issue resolved: #2714 

Benchmarks to show overhead of BindingSet.

----

### DIFF
--- a/core/model/src/test/java/org/eclipse/rdf4j/model/benchmark/ValueHashBenchmark.java
+++ b/core/model/src/test/java/org/eclipse/rdf4j/model/benchmark/ValueHashBenchmark.java
@@ -31,8 +31,6 @@ import org.openjdk.jmh.runner.Runner;
 import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 
-import sun.net.www.content.text.plain;
-
 @State(Scope.Benchmark)
 @BenchmarkMode(Mode.AverageTime)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
@@ -7,6 +7,12 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation;
 
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
 import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.query.AbstractBindingSet;
 import org.eclipse.rdf4j.query.Binding;
@@ -14,14 +20,6 @@ import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.impl.MapBindingSet;
 import org.eclipse.rdf4j.query.impl.SimpleBinding;
 import org.eclipse.rdf4j.util.iterators.ConvertingIterator;
-
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * An implementation of the {@link BindingSet} interface that is used to evalate query object models. This
@@ -32,35 +30,31 @@ public class QueryBindingSet extends AbstractBindingSet {
 
 	private static final long serialVersionUID = -2010715346095527301L;
 
-	private final Binding[] bindings;
+	private final Map<String, Value> bindings;
 
 	public QueryBindingSet() {
-		this(3);
+		this(8);
 	}
 
 	public QueryBindingSet(int capacity) {
 		// Create bindings map with some extra space for new bindings and
 		// compensating for HashMap's load factor
-		bindings = new Binding[capacity];
+		bindings = new HashMap<>(capacity * 2);
 	}
 
 	public QueryBindingSet(BindingSet bindingSet) {
+		this(bindingSet.size());
+		addAll(bindingSet);
+	}
+
+	public void addAll(BindingSet bindingSet) {
 		if (bindingSet instanceof QueryBindingSet) {
-			bindings = ((QueryBindingSet) bindingSet).bindings;
+			bindings.putAll(((QueryBindingSet) bindingSet).bindings);
 		} else {
-			bindings = new Binding[bindingSet.size()];
 			for (Binding binding : bindingSet) {
 				this.addBinding(binding);
 			}
 		}
-	}
-
-	public void addAll(BindingSet bindingSet) {
-
-		for (Binding binding : bindingSet) {
-			this.addBinding(binding);
-		}
-
 	}
 
 	/**
@@ -79,6 +73,7 @@ public class QueryBindingSet extends AbstractBindingSet {
 	 * @param value The binding's value.
 	 */
 	public void addBinding(String name, Value value) {
+		assert !bindings.containsKey(name) : "variable already bound: " + name;
 		setBinding(name, value);
 	}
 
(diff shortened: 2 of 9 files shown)


hmottestad:
```
Benchmark                        Mode  Cnt     Score     Error  Units
SortBenchmark.sortByQuery        avgt    5  2278.088 ± 254.732  ms/op
SortBenchmark.sortDirectly       avgt    5   635.099 ±  90.760  ms/op
SortBenchmark.sortGetStatements  avgt    5   660.591 ± 197.633  ms/op
```

Essentially using SPARQL to sort all the objects in all the statements takes ~3.5x more time than just sorting the statements yourself using getStatements and the ValueComparator.

hmottestad:
@jeenbroekstra this was what I actually wanted to work on today! Trying to get an idea of if we could maybe make query evaluation 2-3x faster by not using hash maps in BindingSet. Those hash maps aren't great for GC either.

hmottestad:
Question is how big of an overhead running a SPARQL query should take. I would think maybe 10-20% compared to rolling your own code for a specific scenario.

jeenbroekstra:
Interesting find. Note that we have several different BindingSet implementations. It should be possible to just introduce a new BindingSet implementation, and use it for comparison. If it turns out to be significantly faster, we can perhaps use it as a drop-in replacement for QueryBindingSet.

jeenbroekstra:
Though I wouldn't be surprised if a significant part of the overhead is just due to the fact that sparql query result ordering is inherently more complex than just ordering values directly. I mean sorting a getStatements call as you do here is just sorting statement objects by identity. SPARQL result ordering needs to dynamically look at individual elements _within_ each result. This is partly why the code in OrderComparator is so not-straightforward.

hmottestad:
Gets worse:

```
Benchmark                                         Mode  Cnt    Score     Error  Units
SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  348.053 ± 124.609  ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   64.502 ±   1.046  ms/op
```

That is the just simply retrieving all statements. SPARQL ~5x slower!

hmottestad:
Screenshot 2020-12-11 at 02 28 59

You cant really tell, but I went through it. 95% og everything above where I'm hovering is spent on various aspects of BindingSet. I can also see that the use of AtomicBoolean is costing us quite a lot during SPARQL execution within `AbstractCloseableIteration`.

jeenbroekstra:
> Gets worse:
> 
> ```
> Benchmark                                         Mode  Cnt    Score     Error  Units
> SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  348.053 ± 124.609  ms/op
> SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   64.502 ±   1.046  ms/op
> ```
> 
> That is the just simply retrieving all statements. SPARQL ~5x slower!

That includes parsing the query though, right? I believe you when you say a large part is to do with BindingSet, just saying that your benchmark as currently implemented also parses the query anew for every execution.

hmottestad:
Parsing the query isn't taking any attributable time. Dataset is large, 55 mb turtle file, specifically to reduce the % of time spent do other things. 

I honestly believe that this is all overhead due to BindingSet and the use of AtomicBoolean. 

It should be easy enough to test. Just remove the use of AtomicBoolean and see if the time changes appreciatively. That's obviously not a fix, just something to demonstrate. A fix would have to behave as an AtomicBoolean but be faster.

hmottestad:
```
Benchmark                                      Mode  Cnt    Score   Error  Units
SparqlOverheadBenchmark.getAllStatementSparql  avgt    5  283.903 ± 4.765  ms/op
```

hmottestad:
Swapped out AtomicBoolean with a volatile variable, seeing 8-10% better performance!

jeenbroekstra:
> Swapped out AtomicBoolean with a volatile variable, seeing 8-10% better performance!

Looks good. We only have a single write point for the boolean flag, and I see you've taken care of concurrent access there, so happy with this. 

As an aside, @JervenBolleman  just got in touch via gitter to say he's also been looking at performance of QueryBindingSet. It would be good to combine your efforts :)

hmottestad:
I would love to hear your take @JervenBolleman .

I was thinking about the following:
 1. Before executing the query we traverse the query plan and give each unique variable an index
 1. Then we create a binding set template with a fixed sized array so that it can always be used regardless of location in the query
 1. When the query runs it fills bindingsets like this: bindingset.set(var.index, val)
 1. We have a virtual binding set that wraps each one returned to the user with a shared hash map that translates the users variable name to the appropriate index

jeenbroekstra:
@hmottestad I like the general idea, though I'm not sure I fully understand points 3 and 4. I'd be more inclined to use the array indexing you create to initialize a new BindingSet implementation, that _internally_ just handles mapping the insertion and lookup operations based on variable names to index-based operations. There are an awful lot of places in the various Iteration implementation that just rely on the BindingSet interface to add or replace variable bindings, so if you encapsulate this you won't have to refactor all that code.

hmottestad:
I realize. If we had used Var with BindingSet instead of a string things would have been a lot simpler.

hmottestad:
And again. I'm still just highlighting the problem at the moment. I don't intend to work on it and I have a lot more work to do on the ShaclSail that I need to prioritize.

jeenbroekstra:
> I realize. If we had used Var with BindingSet instead of a string things would have been a lot simpler. 

That's actually an interesting idea, maybe something to consider for 4.0?

JervenBolleman:
@hmottestad it's a nice idea. I we indeed had used var objects instead of names we could do a lot more interesting optimizations.
In any case introducing a pre execution parse to figure out which variables we will bind would be very useful. I would love to chat about some ideas with someone who knows the engine better. I got into this because I kept on seeing a rehash of the map when just doing a `select ?s ?p ?o where ?s ?p ?o` in the query. 

So my first idea was to switch map implementations depending on having a few bindings or a lot. With the SmallMap being a plain array and we just do scan inserts etc.

If we used `Var` then we could tricks like.

````java
public Wrapped extends Var {
   private final Consumer<Value> inserter;
   private final Supplier<Value> retriever;
   public WrappedVar(String name, Consumer<Value> inserter){
      super(name);
   }

   @Override
   public getValue() {
     return retriever.get();
   }


   @Override
   public setValue(Value val) {
     return inserter.consume(val);
   }
}
```
and the inserter and consumer functions are constructed like this.
```java
ArrayBackedBindingSet abbs = ...;
// assume "s" is the first index.

new WrappedVar("s", (v) -> abbs.array[0], () -> abbs.array[0]);
```
This allows for some really sweet machine code. Especially if we could make this a Java Record (trusted final field optimization).

hmottestad:
I saw your comment on use of intern @JervenBolleman it gives around 5% faster query execution with my benchmark. We only really need to "intern" the variable names within a query, and not across queries. So we could easily just analyze the vars in the query and replace the variables names with a distinct string for that query. 

As for your idea about using an array for the binding set as a stop gap measure. Sounds good. Can I recommend you count the number of vars to start off, and then decide if the query should be executed with an array backed binding set or a hash map backed one based on how many distinct vars we have. If we can guarantee that all variable names are "interned" (not with .intern, but with our own logic) then we could always use == instead of .equals! Might actually end up being seriously fast that way, with no loss of backwards compatibility.

hmottestad:
Also. With regards to Java Records. Don’t get too happy. We are still on Java 8. Probably for a few more years I reckon. I presume most vendors will start pushing once Loom and maybe Valhalla are available. Those will push all the database providers and any networking library or performance oriented library to switch for the new features. There will probably still be lots of companies that say it’s cheaper to buy faster servers than upgrading. Those companies will probably not buy faster servers either, from my experience ;)

JervenBolleman:
@hmottestad I think that would be a first step from the volcano style query interpreter to a query compiler ;) Lot's of performance to be gained there. So this might trigger a bigger change.

For the record part, multi-release jar's might be worth investigating once java-17 is out.

hmottestad:
So I've made a test implementation of using an array. And for the one benchmark I ran about twice as fast. So that proves that it's a solution we may want to go after. 


It's still a far cry from the performance of getStatements, but it's a start.

jeenbroekstra:
@hmottestad where next? I believe you mentioned that you wouldn't have time to work on this, but these spikes show there is a lot to be gained here. Clearly, the test implementation you did here is not ready to be merged. Is this branch supposed to be throwaway code? Or is the idea that you continue working on this? If @JervenBolleman  wants to help contribute, can you coordinate with him what the best way to organize that is?

hmottestad:
I was hoping maybe to get Ontotext onboard and start working on this and maybe other performance aspects in Jan/Feb, once we have enough of the ShaclSail complete.

This branch started off as just benchmarks, but evolved a bit. I think I'll revert those changes and merge it in with only the benchmarks and the fix for AtomicBoolean.

hmottestad:
Looks like @JervenBolleman is a step ahead of us already: https://github.com/eclipse/rdf4j/pull/2722

JervenBolleman:
@hmottestad , @jeenbroekstra asked me to open a draft pull request so it is easier to compare the approaches.
I think I will open a issue+pull request just for the test case improvements to start with. Then I will wait for your benchmark to land, (or if already accepted) then I will open a pull request for adding the smallmap or constant map if it improves performance.

PR #2722, opened 2020-12-15
Smallmap

JervenBolleman:
GitHub issue resolved: # 

QueryBindingSet Map usage has a major overhead that we can avoid in the cases of small numbers of bindings by just using an array. 

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
@@ -7,9 +7,15 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation;
 
+import java.util.AbstractMap;
+import java.util.AbstractSet;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -25,12 +31,18 @@ import org.eclipse.rdf4j.util.iterators.ConvertingIterator;
  * An implementation of the {@link BindingSet} interface that is used to evalate query object models. This
  * implementations differs from {@link MapBindingSet} in that it maps variable names to Value objects and that the
  * Binding objects are created lazily.
+ *
+ * Also has an optimisation that it tries to use a very small map backed by a single array. And only switches to a
+ * heavier hash map once we have more contents.
+ * 
  */
 public class QueryBindingSet extends AbstractBindingSet {
 
 	private static final long serialVersionUID = -2010715346095527301L;
 
-	private final Map<String, Value> bindings;
+	private static final int SWITCH_TO_MAP = 16;
+
+	private Map<String, Value> bindings;
 
 	public QueryBindingSet() {
 		this(8);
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
I like this approach conceptually: it's not intrusive in the rest of the engine and also preserves existing functionality. 

However, I just ran the SparqlOverheadBenchmark on this change, and it seems to make no difference (in fact, performance looks to be slightly worse, even):

```
master branch

Benchmark                                         Mode  Cnt    Score   Error  Units
SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  220.660 ± 8.536  ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   24.353 ± 2.521  ms/op
SparqlOverheadBenchmark.queryWithManyVars         avgt    5    0.571 ± 0.174  ms/op

Smallmap PR
Benchmark                                         Mode  Cnt    Score    Error  Units
SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  239.299 ± 18.908  ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   25.701 ±  3.547  ms/op
SparqlOverheadBenchmark.queryWithManyVars         avgt    5    0.594 ±  0.109  ms/op
```

Can someone else confirm or disprove this? It's entirely possible I missed something in setting up the benchmark run.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
```diff
@@ -152,4 +190,220 @@ public boolean equals(Object other) {
...
+			for (int i = 0; i < keysValues.length; i = i + 2) {
+				if (keysValues[i] != null) {
+					size++;
+				}
+			}
+			return size;
+		}
+
+		private Iterator<Entry<String, Value>> iterator() {
+			return new Iterator<Entry<String, Value>>() {
```
hmottestad, line 218: Do you think the iterator can be swapped out with something like Arrays.stream(keysValues).filter(something or other not null).map(some sort of cast or something).iterator()?

hmottestad:
> I like this approach conceptually: it's not intrusive in the rest of the engine and also preserves existing functionality. 
> 
> 
> 
> However, I just ran the SparqlOverheadBenchmark on this change, and it seems to make no difference (in fact, performance looks to be slightly worse, even):
> 
> 
> 
> ```
> 
> master branch
> 
> 
> 
> Benchmark                                         Mode  Cnt    Score   Error  Units
> 
> SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  220.660 ± 8.536  ms/op
> 
> SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   24.353 ± 2.521  ms/op
> 
> SparqlOverheadBenchmark.queryWithManyVars         avgt    5    0.571 ± 0.174  ms/op
> 
> 
> 
> Smallmap PR
> 
> Benchmark                                         Mode  Cnt    Score    Error  Units
> 
> SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  239.299 ± 18.908  ms/op
> 
> SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   25.701 ±  3.547  ms/op
> 
> SparqlOverheadBenchmark.queryWithManyVars         avgt    5    0.594 ±  0.109  ms/op
> 
> ```
> 
> 
> 
> Can someone else confirm or disprove this? It's entirely possible I missed something in setting up the benchmark run. 
> 
> 

I'll see if I can run it locally.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
JervenBolleman: No, because we need to step by two to get the value. This is because the keys are interspersed with the values. This could work if it was an array of bindings. Not when we have a mixed array of String and Values.

JervenBolleman:
@hmottestad could you also run the query5 with the bindings ?s ?p ?o instead of ?a ?b ?c. One thing I saw was hashmap resizing in my traces, which could just be something unfortunate with ?s ?p ?o having a collision and ?a ?b ?c not.

JervenBolleman:
I will also try to run this benchmark locally.

hmottestad:
The hash map always resizes. Essentially, until you put something in it the map isn't allocated. So the first time you put something in it it will resize. It's not actually resizing, it's just that they've chosen to keep it simple so it's essentially a map with size 0 that gets resized to whatever is the initial size. I believe it's the case even if you provide an initial size or not.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
hmottestad: Probably better to have two arrays or an array of bindings then. Will be much easier for the next developer to understand.

hmottestad:
That there seems to be one of your problems. The use of the EmptyBindingSet triggers the use of hashmap!

hmottestad:
Adding this seems to help a bit: 
```
		if(bindingSet instanceof EmptyBindingSet){
			return new SmallMap();
		}
```

hmottestad:
I would also try for something simpler, like using Binding instead of an array that you have to iterate through on `i+=2`. 

Also, not that it seems to be an issue at the moment, but the use of exceptions to trigger expansion is usually quite expensive, but performance can be improved by not collecting a stack trace.

hmottestad:
Not that it's a part of this PR, but I noticed that the LockingIteration seems a bit too happy about synchronized methods.

JervenBolleman:
Yeah, the use of an exception to trigger the switch is not something that I would keep in any case! I will have another look at finishing this up and run with more performance tests.

JervenBolleman:
@hmottestad @jeenbroekstra I think  I am going to go for something more [invasive](https://github.com/JervenBolleman/rdf4j/tree/queryEvalSetBindings). Because it has more potential to lead to better performance downstream. At the same time I am playing around in the StrictEvaluationStrategy to see if there is more that can be done and for me to understand what different parts do.

Issue #2741, opened 2020-12-30
Reduce the overhead in evaluating a SPARQL query


JervenBolleman:
As shown by #2714 there is quite a bit of overhead in the interpretation of the SPARQL queries, compared to acting directly on a getStatements method. We should investigate different options to see if we can reduce this.

JervenBolleman:
Moving back to 09b8e28fc4daccb0f954a51c2d1caf1d81b31f63 in a new branch to add a bunch of test cases for the things that broke afterwards.

PR #2742, opened 2020-12-30
Use lambda's to avoid unneeded work in the core StatementPattern evaluation 

JervenBolleman:
GitHub issue resolved: #2741

This changes the StrictEvaluationStrategy core method to try and use invoke dynamic/lambda's to avoid un-necessary work.
It now uses an optimized array based bindingset. The code generates a direct setter into the array without further lookup of variables during iteration.
It also removes one iterator that was used to filter `?x a ?x` style equality. With as a side effect that this filter may now be totally
removed during execution.

This is a draft pull request as I want to further test, and document the logic in this rather invasive to the core engine change.
Also I think performance is better but my set-up is not good enough to be sure.

| Cold laptop original code |  | | |
--- | --- | ---- |---
SparqlOverheadBenchmark.getAllStatementSparql  |   avgt    5 | 330.215 ±  7.582  |ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly | avgt    5  | 36.324 ± 15.360 | ms/op
SparqlOverheadBenchmark.queryWithManyVars        | avgt   5   | 0.694 ±  0.187 | ms/op 

| Cold laptop original code |   Cnt    |Score    Error  |Units |
--- | --- | ---- |---
SparqlOverheadBenchmark.getAllStatementSparql     |avgt    5 |  301.934 ± 52.687 | ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly | avgt   5 |  32.173 ±  1.249 | ms/op
SparqlOverheadBenchmark.queryWithManyVars         |avgt   5 |   0.653 ±  0.017 | ms/op

So there might be a performance increase. But on my laptop I have seen wide divergence in this benchmark depending on how long I have used the laptop already.

@hmottestad how do you run this benchmark in isolation?

----


hmottestad:
When I swapped out the bindingset with arrays I saw a clear improvement. I did intern the variable name strings, so I could use `==`, that might make a difference. I also see that you are sorting and then using binary search, if I were you I would benchmark the difference between doing that and just running through the array and checking `.equals` instead. 

I would have expected more of an improvement, so I'm wondering what hot methods you are seeing when profiling now. There shouldn't be much activity for hashmap if your code works as expected, so that would be nice to verify.

I have a fairly old laptop (pre the Intel 14nm nightmare), so it doesn't really throttle much. Biggest impact I have seen is when Timemachine (backup) kicks in while I'm benchmarking. You can increase the number of warmup and measurement runs (20 - 20) and see if you see more stable results. I also ramp up my fans manually.

JervenBolleman:
@hmottestad the binary search is only when constructing the iterator. It should not be called once the iterator is running, if I did it right.

JervenBolleman:
@hmottestad I think it is because the QueryBindingSet is also used in the projection iterator and therefore we still have a high overhead of the hashmap etc..

hmottestad:
Something like this should work. With java 8. With >8 you need to remove the unlock commercial features flag. 

`@Fork(value = 1, jvmArgs = {"-Xms8G", "-Xmx8G", "-XX:+UseSerialGC", "-XX:+UnlockCommercialFeatures", "-XX:StartFlightRecording=delay=5s,duration=120s,filename=recording.jfr,settings=profile", "-XX:FlightRecorderOptions=samplethreads=true,stackdepth=1024", "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"})`

Make sure to only run the exact benchmark method you want. If not each method will override the previous one. 

Use Java Mission Control (JMC) to open the flight recorder data. 

If you have the commercial version of IntelliJ you can also make a main method that just calls the benchmark method in a loop, and use the built in profiler. It's got a nice visualizer and it's super easy to use.

JervenBolleman:
A new commit, using the results from looking in the Netbeans profiler.

SparqlOverheadBenchmark.getAllStatementSparql       avgt    5  217.877 ±  13.640  ms/op

Now I need to wait for my laptop to cool down ;) so I can get the right results for the other tests :D

```
Iteration   1: 20.129 ms/op
Iteration   2: 20.209 ms/op
Iteration   3: 21.591 ms/op
Iteration   4: 101.098 ms/op <!-- fans going crazy!
Iteration   5: 171.007 ms/op
```

JervenBolleman:
hmm, overdone optimism lot's of tests started failing :( not just the formatting...

JervenBolleman:
Ok, I am doing something that the SPIN magic property support does not like, surfaced in the ConstantOptimizer.
For which I would not mind some help in figuring out what the right behaviour should be.

Still performance improvement in the basic queries is good.

 Benchmark                                           | Mode |  Cnt |    Score |   Error | Units
 ----- | ---| --- | --- | --- | ---
SparqlOverheadBenchmark.getAllStatementSparql     |  avgt   |  5  | 187.797 | ± 3.953 | ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly   | avgt   | 5  | 34.304 | ± 1.478 | ms/op
SparqlOverheadBenchmark.getAllTypesStatementSparql | avgt |   5 |  19.479 | ± 1.157 | ms/op
SparqlOverheadBenchmark.queryWithManyVars          | avgt|    5|    0.761| ± 0.026|  ms/op

hmottestad:
> A new commit, using the results from looking in the Netbeans profiler.
> 
> SparqlOverheadBenchmark.getAllStatementSparql avgt 5 217.877 ± 13.640 ms/op
> 
> Now I need to wait for my laptop to cool down ;) so I can get the right results for the other tests :D
> 
> ```
> Iteration   1: 20.129 ms/op
> Iteration   2: 20.209 ms/op
> Iteration   3: 21.591 ms/op
> Iteration   4: 101.098 ms/op <!-- fans going crazy!
> Iteration   5: 171.007 ms/op
> ```

The SparqlOverheadBenchmark is GC sensitive. What you are seeing there could be GC kicking in. Essentially the iterations fill up the memory with garbage until GC kicks in and slows everything down. Or it could also be your laptop throttling. Your profiler should show you when GC starts ramping up. And you could use Intel Power Gadget to see if your CPU is throttling.


> Benchmark | Mode | Cnt | Score | Error | Units
> -- | -- | -- | -- | -- | --
> SparqlOverheadBenchmark.getAllStatementSparql | avgt | 5 | 187.797 | ± 3.953 | ms/op
> SparqlOverheadBenchmark.getAllStatementsDirectly | avgt | 5 | 34.304 | ± 1.478 | ms/op
> SparqlOverheadBenchmark.getAllTypesStatementSparql | avgt | 5 | 19.479 | ± 1.157 | ms/op
> SparqlOverheadBenchmark.queryWithManyVars | avgt | 5 | 0.761 | ± 0.026

Those numbers are looking very promising! If this translates for other queries (which I belive it does) we are essentially seeing a general 2x improvement for queries!

core/sail/memory/src/test/resources/benchmarkFiles/query7.qr
```diff
@@ -0,0 +1,18 @@
...
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+PREFIX dcat: <http://www.w3.org/ns/dcat#>
+PREFIX dc: <http://purl.org/dc/terms/>
+PREFIX skos:  <http://www.w3.org/2004/02/skos/core#>
+PREFIX foaf:  <http://xmlns.com/foaf/0.1/>
+PREFIX dct: <http://purl.org/dc/terms/>
+
+
+
+select * where {
```
hmottestad, line 15: You could try `distinct ?c`, which would limit the number of results returned, but the same number of statements would still need to be retrieved from the DB.

JervenBolleman: The test was for me to see if just selecting one variable would be cheaper than selecting three. Which it is :)

JervenBolleman:
Latest results. The breaking issue (in SPIN) was that in changing the projection iterator I did not take into account that one source variable could me mapped into multiple target variables.

Speed results are good.
```
Benchmark                                           Mode  Cnt    Score   Error  Units
SparqlOverheadBenchmark.getAllStatementSparql       avgt    5  170.399 ± 4.240  ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly    avgt    5   32.191 ± 2.217  ms/op
SparqlOverheadBenchmark.getAllTypesStatementSparql  avgt    5   18.158 ± 0.344  ms/op
SparqlOverheadBenchmark.queryWithManyVars           avgt    5    0.665 ± 0.068  ms/op
```

@hmottestad @jeenbroekstra would you please help me review my changes when you have time. It is not always elegant nor straightforward code. Also it really does mess with the the core of the sparql engine so more reviews than usual are needed.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ProjectionIterator.java
```diff
@@ -1,12 +1,16 @@
-/*******************************************************************************
+/** *****************************************************************************
```
jeenbroekstra, line 1: Looks like your code formatter mangles the copyright header a bit?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ProjectionIterator.java
```diff
@@ -18,32 +22,61 @@
...
 	public ProjectionIterator(Projection projection, CloseableIteration<BindingSet, QueryEvaluationException> iter,
 			BindingSet parentBindings) throws QueryEvaluationException {
 		super(iter);
-		this.projection = projection;
-		this.parentBindings = parentBindings;
-		this.isOuterProjection = determineOuterProjection();
+		ProjectionElemList pel = projection.getProjectionElemList();
+		if (parentBindings.size() == 0) {
+			if (pel.getElements().size() == 1) {
+				Function<BindingSet, ArrayBindingSet> oneVariableConversion = convertASingleVariable(pel);
```
jeenbroekstra, line 43: Can we fold the two function implementations into one - perhaps by just counting the number of elements inside the function?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
```diff
@@ -455,6 +461,88 @@ protected ZeroLengthPathIteration getZeroLengthPathIterator(final BindingSet bin
 		return visitor.boundVars;
 	}
 
+	/**
+	 * This class converts statements into BindingSets. It tries to do so by creating {@link BiConsumer}s for each no
+	 * null {@link Var}iable. That BiConsumer injects the Value directly into the new binding set.
+	 *
+	 * @param <T> the specific type of BindingSet being used at this stage.
+	 */
+	private static final class IntoBindingSetCoverter<T extends BindingSet>
```
jeenbroekstra, line 470: typo: `Coverter` -> `Converter`

jeenbroekstra (commented):
Thanks for looking into this Jerven. This is just a quick first pass with mostly editorial comments , I'll try and do a more thorough review soon. 

One thing I noticed that that there are several (empty) files called `.attach_pidXXXX` checked in. Can you remove those again? Not sure what they were produced by, if they're a regular thing feel free to add an entry to the project .gitignore.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ProjectionIterator.java
JervenBolleman: We could but then we can't avoid an array length check in the single variable conversion case. On the other hand the single variable case might not go through the projection iterator at all. So I am happy to drop the special one variable case.

hmottestad:
Note to self: Run the other sparql benchmarks and also the SHACL benchmarks to see if this improves performance other places too.

JervenBolleman:
@hmottestad I think the groupiterator introduces quite a bit of overhead. Both in the sets as well as in the aggregate evaluation. I suspect it would be worth opening a new issue for that, as changes there would make this pull request a monster one ;)

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
JervenBolleman: fixed

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ProjectionIterator.java
JervenBolleman: fixed

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ProjectionIterator.java
JervenBolleman: Third approach was taken. Always use a lambda, and prepare for reuse in the MultiProjectionIterator. I generate three lambda's for the three states possible at construction. No ParentBinding, include all the parentbinding, include the parentbinding value if the source value is null.

jeenbroekstra:
@JervenBolleman in case you haven't found out yet: you can run benchmarks in isolation from the command line. For example:

```bash
cd core/sail/memory
java -jar target/jmh-benchmarks.jar SparqlOverheadBenchmark   <!-- runs all sparql overhead benchmark tests
java -jar target/jmh-benchmarks.jar SparqlOverheadBenchmark .getAllStatementsSparql  <!-- runs one specific benchmark test 
```

You may need to run an `mvn clean install` beforehand ( use `-Pquick` to skip tests) so that the benchmark fat jar is up to date. There are also options in Eclipse and IntelliJ to run them directly from the IDE if you prefer that - not sure about Netbeans though.

jeenbroekstra (commented):
I'm really happy with this, it's a solid improvement, ran the SparqlOverheadBenchmark locally and seeing significant performance increases.

develop:
```
Benchmark                                         Mode  Cnt    Score     Error  Units
SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  364.988 ± 138.418  ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   37.382 ±   8.824  ms/op
SparqlOverheadBenchmark.queryWithManyVars         avgt    5    1.144 ±   0.863  ms/op
```

This PR:

```
Benchmark                                         Mode  Cnt    Score    Error  Units
SparqlOverheadBenchmark.getAllStatementSparql     avgt    5  226.642 ± 38.699  ms/op
SparqlOverheadBenchmark.getAllStatementsDirectly  avgt    5   34.069 ±  3.526  ms/op
SparqlOverheadBenchmark.queryWithManyVars         avgt    5    0.900 ±  0.244  ms/op
```

I also checked the query benchmarks we have for the native store. 

Your branch:
```
Benchmark                                               Mode  Cnt     Score     Error  Units
QueryBenchmark.complexQuery                             avgt   10    43.291 ±   2.522  ms/op
QueryBenchmark.distinctPredicatesQuery                  avgt   10  1008.616 ±   6.278  ms/op
QueryBenchmark.groupByQuery                             avgt   10    16.866 ±   0.038  ms/op
QueryBenchmark.removeByQuery                            avgt   10  1119.031 ±   8.159  ms/op
QueryBenchmark.removeByQueryReadCommitted               avgt   10  4447.119 ±  34.196  ms/op
QueryBenchmark.simpleUpdateQueryIsolationNone           avgt   10  4623.598 ±  29.948  ms/op
QueryBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt   10  5110.329 ± 119.463  ms/op
```

Develop:

```
Benchmark                                               Mode  Cnt     Score    Error  Units
QueryBenchmark.complexQuery                             avgt   10    36.764 ±  0.416  ms/op
QueryBenchmark.distinctPredicatesQuery                  avgt   10  1321.448 ±  5.030  ms/op
QueryBenchmark.groupByQuery                             avgt   10    18.705 ±  0.226  ms/op
QueryBenchmark.removeByQuery                            avgt   10  1164.537 ± 29.151  ms/op
QueryBenchmark.removeByQueryReadCommitted               avgt   10  4525.337 ± 67.505  ms/op
QueryBenchmark.simpleUpdateQueryIsolationNone           avgt   10  4697.181 ± 73.385  ms/op
QueryBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt   10  5181.044 ± 90.601  ms/op
```

At first glance it has a marked improvement on the distinct predicates query case (which is essentially just a "get all triples" query with a distinct projection on the predicate var), and a slight performance drop on the complex query case. Overall though, I think we can be fairly confident that this markedly improves performance.  

Code-wise I have very little to comment, from my point of view it's mainly about minor things like names and documentation. 

In particular: can you add a short explanation of the general principle behind the apporach, _why_ we use conversion lambdas in several relevant places, as javadoc and/or a comment? Some of these changes are not necessarily the most obvious way to write the code, so we should make it clear why it is done the way it is.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/ProjectionIterator.java
jeenbroekstra: Looks good to me. The names of the functions are a bit unwieldy, but I must admit I have no inspiration for anything better either, so :+1:

core/query/src/main/java/org/eclipse/rdf4j/query/impl/ArrayBindingSet.java
```diff
@@ -0,0 +1,161 @@
...
+import org.eclipse.rdf4j.query.AbstractBindingSet;
+import org.eclipse.rdf4j.query.Binding;
+import org.eclipse.rdf4j.query.BindingSet;
+
+/**
+ * An array implementation of the {@link BindingSet} interface.
+ *
+ * @author Jerven Bolleman
+ */
+public class ArrayBindingSet extends AbstractBindingSet {
```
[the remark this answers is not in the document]
hmottestad: I think it's good to start it off as internal use only, and then remove that in the future when we are sure we don't want to change anything else.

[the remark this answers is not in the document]
JervenBolleman: Done

hmottestad:
That is almost a 25% increase in execution time for the complex query. Probably worth checking what up with that.

JervenBolleman:
I got the regression in performance for the complex query down to 7%. I have ideas to lower this even further. Most of the performance loss can be regained (and more) by special casing some logic in the JoinIterator. Basically, a lot of work is redone in the joining to recalculate already known values.

develop
``` 
Benchmark                    Mode  Cnt   Score   Error  Units
QueryBenchmark.complexQuery  avgt   10  31.323 ± 0.246  ms/op
```

4ae9e6189545cb01dc433387eae298780e962991
```
Benchmark                    Mode  Cnt   Score   Error  Units
QueryBenchmark.complexQuery  avgt   10  29.970 ± 0.152  ms/op
```
In between  ( as noted by @hmottestad )
```
Benchmark                    Mode  Cnt   Score   Error  Units
QueryBenchmark.complexQuery  avgt   10  38.603 ± 0.136  ms/op
```

hmottestad:
What was causing the regression?

JervenBolleman:
> What was causing the regression?
For the JoinIterator, the fact that more work is done in the StrictEvaluationStrategy is done costs more, when the evaluation is called very often this leads to a performance cost. I think it can be avoided, by saving more work between calls to the right argument of the join evaluation.

hmottestad:
No, I mean...why did introducing the array based bindingset and the lambdas cause the complex query performance to regress?

JervenBolleman:
Because setting up the lambdas is expensive. It is called very often in the complex queries, as the joins evaluation calls strategy.evaluate often.

hmottestad:
I feel I really need to try to figure out how your code works. I get the array binding set part, but I'm still not understanding the lambda aspect. 

The join approach in RDF4J is one of the reasons why RDF4J is so fast and doesn't require sorted indexes. Essentially when joining two statement patterns we choose the smallest one as "master" (left) and use the results from that to lookup what data would be joined (right). It does have a lot of overhead though, so it's not all good. It also completely ignores things like disk blocks.

I think at this point we should try to make a bigger mix of queries so that we can see if the increased overhead nullifies the binding set improvements for more normal queries. If we create those benchmarks in a new branch we can merge them into master/develop to make it easier to compare before and after.

JervenBolleman:
Current code is really me also trying to explore how the evaluation strategy works in general. Trying to reduce the costs of object (lambda) creation as here a simple hashmap does seem faster if the iteration does not last long. But to be honest I am to tired to do so well today.

JervenBolleman:
Moving back to 09b8e28 in a new branch to add a bunch of test cases for the things that broke afterwards.

jeenbroekstra:
Sorry, I lost track a bit of where we are with this. @JervenBolleman anything specific that we can help you with to get this over the line? Or is it simply a matter of you finding the time to continue with this?

JervenBolleman:
A bit of time lack. 

I have also been trying out a few design alternatives. As well as dealing with the long feedback loop to make sure it is all correct. Now onto approach [4](https://github.com/JervenBolleman/rdf4j/tree/GH-2741.4). Mostly thinking about how to deal with the increase in overhead caused by the extra work to set up the core iterator. As well as me still getting used to getting good quality profile results that show where the over head happens.

Thinking only. I am wondering if it is wise to introduce a QueryEvaluationContext object which can hold things like references to threadpools and these lambda based settings. So I am investigating the multi-threaded story in the core engine. Which seems to be half done but not really used in the open-source side.

Issue #2745, opened 2021-01-01
ShaclSail follow-up after new AST


hmottestad:
This is a generic issue for any minor follow-up fixes for the ShaclSail after merging the new AST.

Currently:
 - fix for fast SchemaCachingRdfsInferencer instantiation (performance)
 - fix validation report when shapes use lists
 - use new Values.iri() methods
 - use .isIRI() instead of instance of

PR #2746, opened 2021-01-01
GH-2745 various shacl sail fixes

hmottestad:
GitHub issue resolved: #2745 

----

### DIFF
--- a/core/sail/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencer.java
+++ b/core/sail/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencer.java
@@ -54,8 +54,8 @@ import org.eclipse.rdf4j.sail.inferencer.InferencerConnection;
  */
 public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 
-	// The schema, or null
-	Repository schema;
+	// An optional predifinedSchema that the user has provided
+	Repository predefinedSchema;
 
 	// exclusive lock for modifying the schema cache.
 	private final ReentrantLock exclusiveWriteLock = new ReentrantLock(true);
@@ -100,7 +100,7 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 	 */
 	public SchemaCachingRDFSInferencer() {
 		super();
-		schema = null;
+		predefinedSchema = null;
 	}
 
 	/**
@@ -110,7 +110,7 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 	 */
 	public SchemaCachingRDFSInferencer(NotifyingSail data) {
 		super(data);
-		schema = null;
+		predefinedSchema = null;
 
 	}
 
@@ -119,13 +119,13 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 	 * other schema statements added will be ignored and no schema statements can be removed. Using a predefined schema
 	 * significantly improves performance.
 	 *
-	 * @param data   Base sail for storing data.
-	 * @param schema Repository containing the schema.
+	 * @param data             Base sail for storing data.
+	 * @param predefinedSchema Repository containing the schema.
 	 */
-	public SchemaCachingRDFSInferencer(NotifyingSail data, Repository schema) {
+	public SchemaCachingRDFSInferencer(NotifyingSail data, Repository predefinedSchema) {
 		super(data);
 
-		this.schema = schema;
+		this.predefinedSchema = predefinedSchema;
 
 	}
 
@@ -137,7 +137,7 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 	 */
 	public SchemaCachingRDFSInferencer(NotifyingSail data, boolean useAllRdfsRules) {
 		super(data);
-		schema = null;
+		predefinedSchema = null;
 
 		this.useAllRdfsRules = useAllRdfsRules;
 
@@ -148,15 +148,15 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 	 * other schema statements added will be ignored and no schema statements can be removed. Using a predefined schema
 	 * significantly improves performance.
 	 *
-	 * @param data            Base sail for storing data.
-	 * @param schema          Repository containing the schema.
-	 * @param useAllRdfsRules Usel all RDFS rules. If set to false rule rdf4a and rdfs4b will be ignore
+	 * @param data             Base sail for storing data.
+	 * @param predefinedSchema Repository containing the schema.
+	 * @param useAllRdfsRules  Usel all RDFS rules. If set to false rule rdf4a and rdfs4b will be ignore
 	 */
-	public SchemaCachingRDFSInferencer(NotifyingSail data, Repository schema,
+	public SchemaCachingRDFSInferencer(NotifyingSail data, Repository predefinedSchema,
 			boolean useAllRdfsRules) {
 		super(data);
 
-		this.schema = schema;
+		this.predefinedSchema = predefinedSchema;
 		this.useAllRdfsRules = useAllRdfsRules;
 
 	}
@@ -219,9 +219,9 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 
 			List<Statement> tboxStatments = new ArrayList<>();
 
-			if (schema != null) {
+			if (predefinedSchema != null) {
 
-				try (RepositoryConnection schemaConnection = schema.getConnection()) {
+				try (RepositoryConnection schemaConnection = predefinedSchema.getConnection()) {
 					schemaConnection.begin();
 					try (Stream<Statement> stream = schemaConnection.getStatements(null, null, null).stream()) {
 						tboxStatments = stream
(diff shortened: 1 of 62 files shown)


core/sail/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencer.java
```diff
@@ -636,4 +636,8 @@ public boolean isAddInferredStatementsToDefaultContext() {
 	public void setAddInferredStatementsToDefaultContext(boolean addInferredStatementsToDefaultContext) {
 		this.addInferredStatementsToDefaultContext = addInferredStatementsToDefaultContext;
 	}
+
+	boolean usesPredefinedSchema() {
+		return predefinedSchema != null || sharedSchema;
+	}
```
hmottestad, line 642: This is the key change here. I use a somewhat undocumented part of the RDFS reasoner that allows you to create a base reasoner that can be used to instantiate a new one much faster than building one from scratch by loading in the schema every time. There was a bug that caused the reasoner to think that new schema data was added, making everything fairly slow. Fixed that.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemEvaluationStatistics.java
```diff
@@ -86,27 +86,27 @@ public double getCardinality(StatementPattern sp) {
...
 
 			double cardinality;
 
-			if (listSizes.isEmpty()) {
+			if (minListSizes == Integer.MAX_VALUE) {
 				// all wildcards
 				cardinality = memStatementList.size();
 			} else {
-				cardinality = (double) Collections.min(listSizes);
+				cardinality = minListSizes;
```
hmottestad, line 109: I found that creating this arraylist was unneeded.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemEvaluationStatistics.java
```diff
@@ -60,14 +60,14 @@ public double getCardinality(StatementPattern sp) {
...
-			if (!(pred instanceof IRI)) {
+			if (!(pred != null && pred.isIRI())) {
 				// can happen when a previous optimizer has inlined a comparison
 				// operator. See SES-970 / SES-998
 				pred = null;
 			}
 			Value obj = getConstantValue(sp.getObjectVar());
 			Value context = getConstantValue(sp.getContextVar());
-			if (!(context instanceof Resource)) {
+			if (!(context != null && context.isResource())) {
```
hmottestad, line 70: Trying to use the new helper methods that are faster than instanceof checks.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementIterator.java
```diff
@@ -60,7 +60,7 @@
 	/**
 	 * The index of the last statement that has been returned.
 	 */
-	private volatile int statementIdx;
+	private int statementIdx;
```
hmottestad, line 63: Intellij says that the use of a volatile variable here is not needed since the code that uses the variable is not synchronized. If this class is supposed to support multi-threaded access then the code that uses this variable also needs to be synchronized.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/RdfsSubClassOfReasoner.java
```diff
@@ -50,12 +51,11 @@
...
 		}
 
-		SimpleValueFactory vf = SimpleValueFactory.getInstance();
 		if (statement.getPredicate().equals(RDF.TYPE)
-				&& forwardChainCache.containsKey(((Resource) statement.getObject()))) {
+				&& forwardChainCache.containsKey(statement.getObject())) {
 			return forwardChainCache.get(statement.getObject())
 					.stream()
-					.map(r -> vf.createStatement(statement.getSubject(), RDF.TYPE, r, statement.getContext()));
+					.map(r -> statement(statement.getSubject(), RDF.TYPE, r, statement.getContext()));
```
hmottestad, line 58: Trying my best here to use the new helper methods in Values.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -386,14 +388,17 @@ private ValidationReport performValidation(List<Shape> shapes, boolean validateE
...
 					.stream()
-					.map(shape -> shape.generatePlans(connectionsGroup, sail.isLogValidationPlans(),
-							validateEntireBaseSail))
-					.filter(Objects::nonNull)
-
-					.map(SingleCloseablePlanNode::new)
-
-					.map(planNode -> () -> {
-
+					.map(shape -> new ShapePlanNodeTuple(shape,
```
hmottestad, line 391: ShapePlanNodeTuple allows us to re-enable some of the logging that was previously commented out.

hmottestad: Eg. it gives us access to the underlying shape

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Shape.java
```diff
@@ -523,4 +530,17 @@ private static void calculateTargetChain(List<Shape> parsed) {
...
+		statements.setNamespace(SHACL.NS);
+		WriterConfig writerConfig = new WriterConfig();
+		writerConfig.set(BasicWriterSettings.PRETTY_PRINT, true);
+		writerConfig.set(BasicWriterSettings.INLINE_BLANK_NODES, true);
+
+		StringWriter stringWriter = new StringWriter();
+		Rio.write(statements, stringWriter, RDFFormat.TURTLE, writerConfig);
+		return stringWriter.toString().replace("@prefix sh: <http://www.w3.org/ns/shacl#> .", "").trim();
+	}
+
```
hmottestad, line 545: This is used for logging the shape, which is why I take some liberties with stripping out the namespace line.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/AndConstraintComponent.java
```diff
@@ -58,13 +58,14 @@ public AndConstraintComponent(AndConstraintComponent andConstraintComponent) {
...
-		HelperTool.listToRdf(and.stream().map(Shape::getId).collect(Collectors.toList()), getId(), model);
-
 		if (exported.contains(getId())) {
 			return;
 		}
 		exported.add(getId());
+
+		model.add(subject, SHACL.AND, getId());
+		HelperTool.listToRdf(and.stream().map(Shape::getId).collect(Collectors.toList()), getId(), model);
+
```
hmottestad, line 68: Found an issue when there were multiple violations for the same shape and that shape used an RDF list. Essentially the list would be printed multiple times, but would have the same head node and thus become corrupt. This code here is part of that solution.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
```diff
@@ -304,8 +304,8 @@ public static void afterClass() {
 				List<Object[]> collect = findTestCases(testCasePath, baseCase.name())
 						.stream()
 						.flatMap(path -> Stream
-//								.of(IsolationLevels.NONE, IsolationLevels.SNAPSHOT, IsolationLevels.SERIALIZABLE)// TODO: Uncomment this line!!!!!
-								.of(IsolationLevels.NONE)
+								.of(IsolationLevels.NONE, IsolationLevels.SNAPSHOT, IsolationLevels.SERIALIZABLE)
+//								.of(IsolationLevels.NONE)
```
hmottestad, line 308: Re-enabled testing the ShaclSail against multiple isolation levels. This gives us over 13 000 test cases, but when executed in parallel on my ageing macbook it only take around 40 seconds.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
```diff
@@ -439,7 +439,7 @@ private static void testValidationReport(String dataPath, Model validationReport
 			Model validationReportExpected = Rio.parse(resourceAsStream, "", RDFFormat.TURTLE);
 
 			if (!isIsomorphic(validationReportActual, validationReportExpected)) {
-				writeActualModelToExpectedModelForDevPurposes(dataPath, validationReportActual);
+//				writeActualModelToExpectedModelForDevPurposes(dataPath, validationReportActual);
```
hmottestad, line 442: As it says, this was for dev purposes and should have been commented out already. Fixed now though!

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/benchmark/ComplexLargeBenchmark.java
```diff
@@ -454,4 +488,12 @@ private static BufferedInputStream getData() {
...
 	}
 
+//	public static void main(String[] args) throws InterruptedException {
+//		ComplexLargeBenchmark complexLargeBenchmark = new ComplexLargeBenchmark();
+//		complexLargeBenchmark.setUp();
+//		while(true){
+//			complexLargeBenchmark.noPreloadingNonEmpty();
+//			System.out.println(".");
+//		}
+//	}
```
hmottestad, line 498: Could probably remove this.

hmottestad:
There is still one bug in the validation report, but I think I will do a M2 build first with these fixes here and anything else that has been merged since M1. Would also love to get @JervenBolleman 's PR merged into the final 3.6.0 release.

Issue #2752, opened 2021-01-09
SHACL - Improve caching of plan nodes


hmottestad:
The caching of plan nodes is broken after the upgrade to the new AST.

Fix the caching and also make the following cacheable:
 - BindSelect
 -  TargetChainRetriever

PR #2753, opened 2021-01-09
GH-2752 shacl plan node cache

hmottestad:
GitHub issue resolved: #2752 

Fix and improve caching of plan nodes in the ShaclSail.

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ClassConstraintComponent.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ClassConstraintComponent.java
@@ -11,6 +11,7 @@ import org.eclipse.rdf4j.model.vocabulary.SHACL;
 import org.eclipse.rdf4j.sail.shacl.ConnectionsGroup;
 import org.eclipse.rdf4j.sail.shacl.RdfsSubClassOfReasoner;
 import org.eclipse.rdf4j.sail.shacl.SourceConstraintComponent;
+import org.eclipse.rdf4j.sail.shacl.ast.HelperTool;
 import org.eclipse.rdf4j.sail.shacl.ast.StatementMatcher;
 import org.eclipse.rdf4j.sail.shacl.ast.paths.Path;
 import org.eclipse.rdf4j.sail.shacl.ast.planNodes.BulkedExternalInnerJoin;
@@ -76,7 +77,7 @@ public class ClassConstraintComponent extends AbstractConstraintComponent {
 
 				if (connectionsGroup.getStats().hasRemoved()) {
 					PlanNode deletedTypes = new UnorderedSelect(connectionsGroup.getRemovedStatements(), null, RDF.TYPE,
-							clazz, s -> new ValidationTuple(s.getSubject(), Scope.nodeShape, false));
+							clazz, HelperTool.SubjectScopedMapper.getFunction(Scope.nodeShape));
 
 					deletedTypes = getTargetChain()
 							.getEffectiveTarget("target_", Scope.nodeShape,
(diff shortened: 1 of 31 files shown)


hmottestad:
Intending to merge this tomorrow.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/Select.java
```diff
@@ -169,13 +171,28 @@ public boolean equals(Object o) {
 			return false;
 		}
 		Select select = (Select) o;
-		return connection.equals(select.connection) &&
-				mapper.equals(select.mapper) &&
-				query.equals(select.query);
+		if (connection instanceof MemoryStoreConnection && select.connection instanceof MemoryStoreConnection) {
```
jeenbroekstra, line 174: Why are we special-casing the MemoryStoreConnection here?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/targets/EffectiveTarget.java
```diff
@@ -273,4 +279,40 @@ public String getQueryFragment() {
 		}
 	}
 
+	class AllTargetsBindinsetMapper implements Function<BindingSet, ValidationTuple> {
```
jeenbroekstra, line 282: Minor typo: `AllTargetsBindinsetMapper` -> `AllTargetsBindingSetMapper`

jeenbroekstra (commented):
This is a bit deep into the guts of the SHACL Sail for me to fully follow what's going on. It looks ok to me from the description of the issue you're trying to solve, but a couple of questions/remarks.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/Select.java
hmottestad: Because every plan node that needs a connection to the two internal transaction memory stores (one for added one for removed) ask for a new connection every time instead of sharing one. This means that we need to check the base sail instead of the actual connection when comparing two plan nodes to see if they are equal.

hmottestad: And by equal I mean "will produce the same results".

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ClassConstraintComponent.java
```diff
@@ -179,7 +180,7 @@ public PlanNode getAllTargetsPlan(ConnectionsGroup connectionsGroup, Scope scope
 			// added type statements that match clazz could affect sh:not
 			if (connectionsGroup.getStats().hasAdded()) {
 				PlanNode addedTypes = new UnorderedSelect(connectionsGroup.getAddedStatements(), null, RDF.TYPE,
-						clazz, s -> new ValidationTuple(s.getSubject(), Scope.nodeShape, false));
+						clazz, HelperTool.SubjectScopedMapper.getFunction(Scope.nodeShape));
```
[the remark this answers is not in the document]
hmottestad: I can just move those new classes in The UnorderedSelect class.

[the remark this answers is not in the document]
hmottestad: The reason for not using lambdas anymore is that we can't hash them, so two plan nodes with lambdas that are functionally identical will still return different hash codes.

hmottestad:
Thanks for the review.

Issue #2727, opened 2020-12-17
Inverse property path with optional `?` does not work in some cases


damyan-ognyanov:
There is an issue with the Inverse property paths in conjunction with `?` optional modifier when grouping with brackets (forced precedense) are used.

To reproduce load following data:
```
insert data {
    <urn:1> <urn:prop> <urn:object> .
    <urn:2> <urn:prop> <urn:mid:1> .
    <urn:mid:1> <urn:prop> <urn:object> .
    <urn:3> <urn:prop> <urn:mid:2> .
    <urn:mid:2> <urn:prop> <urn:mid:3> .
    <urn:mid:3> <urn:prop> <urn:object> .
}
```

This query works as expected, returning 4 results:
```
select * where { 
    <urn:object> ^(<urn:prop>?) ?o .
}
```

This query, that uses the default operator precedense, also works correctly returning 4 results
```
select * where { 
    <urn:object> ^<urn:prop>? ?o .
} 
```

This query returns correctly 3 results:
```
select * where { 
    <urn:object> (^<urn:prop>) ?o .
}
```

but when we add `?` optional modifier, it is expected to return the above 3 results plus the so called `self` result but instead, it only return one result (`self`):

```
select * where { 
    <urn:object> (^<urn:prop>)? ?o .
}
```

which is unexpected.

the produced query model in the wrong case is not inversed (<urn:object> is on subject position and the var on object position, e.g. these are not swapped according to the `^` inverse modifier used.

Tried to find a solution/quick fix but the code around the transaltion of property paths is a bit of a mess. For simple inverse paths like the one used in the above queries the code could be patched somehow but if the path is more complex and involves either alternatives or other segments that are also inversed the number of `swaps` could not be deduced after the recursive step finishes converting the nested segments so it is hard to decide wheter to swap the subj/obj positions or not.

Hope this description is clear enough ...

jeenbroekstra:
> Tried to find a solution/quick fix but the code around the transaltion of property paths is a bit of a mess. 

I quite agree - we keep monkey-patching the code, but it really needs a good overhaul. 

> For simple inverse paths like the one used in the above queries the code could be patched somehow but if the path is more complex and involves either alternatives or other segments that are also inversed the number of swaps could not be deduced after the recursive step finishes converting the nested segments so it is hard to decide wheter to swap the subj/obj positions or not.

I think we should look at writing a dedicated class for handling of path expressions.

VladimirAlexiev:
[...]

PR #2754, opened 2021-01-09
GH-2727 simplified path expression processing and fixed nested inverse optional handling

jeenbroekstra:
GitHub issue resolved: #2727  

- refactor of property path processing to a cleaner tree-visitor approach instead of a massive if-else. 
- simplified path modifier processing code (old code was still written for {min,max} syntax which is no longer supported)
- added regression test for GH-2727

To explain the main ideas behind the refactor:

A SPARQL path expression is parsed into an AST tree structure. Each path expression is a PathAlternative, consisting of one more PathSequences. Each PathSequence, in turn, consists of one or more PathElements. PathElements have optional modifiers (?, + and *), and can be either negated or inverted (or both). Note that a PathElement itself can be a nested path expression again.

In the previous code, the entire path expression processing was handled in a single visitor method, the `visit(PathSequence)`.  It kept internal state, manually inspected its child elements (instead of letting the visitor pattern handle it), and was generally a mess of if-else conditions and special-case handling. What it also did was manipulate the global GraphPattern field directly, instead of just returning the produced TupleExpr. 

The refactor splits this out, so that processing logic is more local to the specific element: path sequences deal with introducing intermediate variables, keeping the context of what the "current" start and end variables of the sequence are. The details of processing each element of the path are now in `visit(PathElt)`. This checks if the element is nested (and if so defers back to the visitor to handle the nested expression), or negated, or inverse. It also handles all path modifiers. 

The handling of path modifiers has also been significantly changed. Previously it had convoluted logic that dealt with building up a tree of Unions to handle things like "at least 2, at most 5". This was introduced when SPARQL was still in draft and had this `{min, max}` syntax for path expressions. I removed this because the logic was hard to follow, essentially unused, and produced buggy behavior in several cases. 

----

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
@@ -1541,109 +1541,30 @@ public class TupleExprBuilder extends AbstractASTVisitor {
 	private TupleExpr handlePathModifiers(Scope scope, Var subjVar, TupleExpr te, Var endVar, Var contextVar,
 			long lowerBound, long upperBound) throws VisitorException {
 
-		TupleExpr result = te;
-
-		if (lowerBound >= 0L) {
-			if (lowerBound < upperBound) {
-				if (upperBound < Long.MAX_VALUE) {
-					// upperbound is fixed-length
-
-					// create set of unions for all path lengths between lower and upper bound.
-					Union union = new Union();
-					union.setVariableScopeChange(false);
-					Union currentUnion = union;
-
-					for (long length = lowerBound; length < upperBound; length++) {
-
-						TupleExpr path = createPath(scope, subjVar, te, endVar, contextVar, length);
-
-						currentUnion.setLeftArg(path);
-						if (length == upperBound - 1) {
-							path = createPath(scope, subjVar, te, endVar, contextVar, length + 1);
-							currentUnion.setRightArg(path);
-						} else {
-							Union nextUnion = new Union();
-							currentUnion.setRightArg(nextUnion);
-							currentUnion = nextUnion;
-						}
-					}
-
-					ProjectionElemList pelist = new ProjectionElemList();
-					for (String name : union.getAssuredBindingNames()) {
-						ProjectionElem pe = new ProjectionElem(name);
-						pelist.addElement(pe);
-					}
-
-					result = new Distinct(new Projection(union, pelist));
-				} else {
-					// upperbound is abitrary-length
-					result = new ArbitraryLengthPath(scope, subjVar.clone(), te, endVar.clone(), contextVar,
-							lowerBound);
-				}
-			} else {
-				// create single path of fixed length.
-				TupleExpr path = createPath(scope, subjVar, te, endVar, contextVar, lowerBound);
-				result = path;
-			}
+		// * and + modifiers
+		if (upperBound == Long.MAX_VALUE) {
+			// upperbound is abitrary-length
+			return new ArbitraryLengthPath(scope, subjVar.clone(), te, endVar.clone(), contextVar,
+					lowerBound);
 		}
 
-		return result;
-	}
-
-	private TupleExpr createPath(Scope scope, Var subjVar, TupleExpr pathExpression, Var endVar, Var contextVar,
-			long length) throws VisitorException {
-		if (pathExpression instanceof StatementPattern) {
-			Var predVar = ((StatementPattern) pathExpression).getPredicateVar();
-
-			if (length == 0L) {
-				return new ZeroLengthPath(scope, subjVar, endVar, contextVar);
-			} else {
-				GraphPattern gp = new GraphPattern();
-				gp.setContextVar(contextVar);
-				gp.setStatementPatternScope(scope);
-
-				Var nextVar = null;
-
-				for (long i = 0L; i < length; i++) {
-					if (i < length - 1) {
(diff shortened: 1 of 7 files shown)


core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
```diff
@@ -100,127 +99,7 @@
...
-import org.eclipse.rdf4j.query.parser.sparql.ast.ASTTz;
-import org.eclipse.rdf4j.query.parser.sparql.ast.ASTUUID;
-import org.eclipse.rdf4j.query.parser.sparql.ast.ASTUnionGraphPattern;
-import org.eclipse.rdf4j.query.parser.sparql.ast.ASTUpperCase;
-import org.eclipse.rdf4j.query.parser.sparql.ast.ASTVar;
-import org.eclipse.rdf4j.query.parser.sparql.ast.ASTYear;
-import org.eclipse.rdf4j.query.parser.sparql.ast.Node;
-import org.eclipse.rdf4j.query.parser.sparql.ast.SimpleNode;
-import org.eclipse.rdf4j.query.parser.sparql.ast.VisitorException;
+import org.eclipse.rdf4j.query.parser.sparql.ast.*;
```
jeenbroekstra, line 102: Heh cute, we went over 99 imports so Eclipse introduced a wildcard. I'll fix.

jeenbroekstra:
Due to the large amount of refactoring (I got rid of a ton of legacy code) the diff on this PR may be a bit hard to follow, apologies for that. It's not 100% cleaned up, but (IMO) a lot better than what it was.

hmottestad:
It's a bit complicated for me to be honest. Had a hard time grasping the original code, so it's tricky to understand what your changes entail. That's fine though. I don't need to understand everything :P

Only question I have is the changes to the two query plans. So you know why that has happened and also if those changes affect performance?

jeenbroekstra:
> It's a bit complicated for me to be honest. Had a hard time grasping the original code, so it's tricky to understand what your changes entail. That's fine though. I don't need to understand everything :P

Fair enough. I have added a bit more explanation to the description (see above), but I appreciate that even with that it's a fairly opaque piece of code.

> Only question I have is the changes to the two query plans. So you know why that has happened and also if those changes affect performance?

Presumably due to the way the algebra is now built up (how we're building the TupleExpr based on the GraphPattern object kept in the visitor), expressions are added to the algebra tree in a slightly different order sometimes. The testTupleQuery one simply produces a different join order in the unoptimized form, and for what it's worth the join order in the new version looks more correct to me than what was originally there (though presumably the query planner will optimize both forms to the same thing). 

As for the subquery one, that's a comparison of the plan after execution, and from what I can tell the main difference is that the filter condition has been pushed further down the tree, leading to a smaller intermediate result set (which should make it more efficient). I'm not 100% why it can now do this optimization when it couldn't before, but given that every other test we have in place greenlines I have assumed (after looking at the new plans and thinking they look conceptually correct for the query) that those query plan tests were brittle, and have just adjusted the expectation.  

I have not yet checked how any of this affects performance. I will try to at least run the parser benchmarks and the memory store query benchmarks later for a quick comparison.

hmottestad:
Pushing the filter down further is usually better.

The test were initially designed to check that the explain feature worked as expected, and then we added some extra cases for some poorly performing queries that you later fixed. They are kinda brittle, but then again they will only break if you actually change anything that affects the query plan - which means that if these failed but no other tests failed we might need better test coverage for the query optimisers?

Do we have any specific tests that test that the filter push optimisation works as expected?

hmottestad:
Actually. I think I remember something. Our optimisers ignore sub query plans, so since your changes affect queries before they are optimised then it would make sense that it would affect the subquery query too!

jeenbroekstra:
I ran the SparqlParseBenchmark against this vs master branch:

master:

```
Benchmark                                        Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.complexPathExpressionQuery  avgt   40  29.138 ± 0.898  ms/op
```

This branch:

```
Benchmark                                        Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.complexPathExpressionQuery  avgt   40  32.549 ± 1.307  ms/op
```

So we're seeing a marginal performance penalty, but to be honest given that we're talking about ~4ms difference on a pathologically complex path expression, I'm not too worried about this. 

More interesting will be to run some evaluation benchmarks with queries involving path expressions. We have nothing ready-made for this unfortunately, so I'll try and come up with a non-trivial example on the datagovbe dataset.

hmottestad:
This might help you understand that dataset if you haven't worked with DCAT before:

http://dcat.be/images/dcat-ap.png

damyan-ognyanov:
Looks good ...

jeenbroekstra:
I've added two benchmark queries on the memory store, using the dcat dataset from datagovbe. Performance comparison:

master:

```
Benchmark                            Mode  Cnt    Score    Error  Units
QueryBenchmark.pathExpressionQuery1  avgt   10  110.671 ± 12.691  ms/op
QueryBenchmark.pathExpressionQuery2  avgt   10    9.255 ±  1.041  ms/op
```

This branch:

```
Benchmark                            Mode  Cnt   Score    Error  Units
QueryBenchmark.pathExpressionQuery1  avgt   10  99.722 ± 16.703  ms/op
QueryBenchmark.pathExpressionQuery2  avgt   10   8.768 ±  0.207  ms/op
```

So certainly no peformance deterioration in evaluation performance, and if anything even a mild improvement (though again I think it kind of falls within the margin of error). 

Happy with this, will merge as soon as build succeeds.

Issue #2756, opened 2021-01-11
SHACL - ShaclSail + RDFS + NativeStore will sometimes deadlock


hmottestad:
Using the following `ShailSail(RDFS(NativeStore)` causes a deadlock in `org.eclipse.rdf4j.sail.shacl.MultithreadedTest#testLotsOfValidationFailuresSerializableValidation`.

After further testing this also causes a deadlock in the other tests in that class, but not quite as regularly.

PR #2757, opened 2021-01-11
GH-2756 ShaclSail + RDFS + NativeStore sometimes deadlocks

hmottestad:
GitHub issue resolved: #2756 

 - Remove shapes cache in ShaclSail
 - Clean up locking now that we don't have to keep a cache in sync
 - Disable serializable validation (with SNAPSHOT isolation) when used with the RDFS reasoner (since this causes deadlocks)

The shapes cache was initially created back when parsing shapes was slow. Since then we have tremendously improved the performance of parsing shapes so removing the cache causes very marginal performance impacts, but it obviously depends on the number of shapes. 

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -176,8 +176,6 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	private static final Logger logger = LoggerFactory.getLogger(ShaclSail.class);
 
-	private List<Shape> shapes = Collections.emptyList();
-
 	private static final Model DASH_CONSTANTS;
 
 	/**
@@ -351,7 +349,6 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 		try (SailRepositoryConnection shapesRepoConnection = shapesRepo.getConnection()) {
 			shapesRepoConnection.begin(IsolationLevels.NONE);
-			shapes = refreshShapes(shapesRepoConnection);
 			shapesRepoConnection.commit();
 		}
 
@@ -359,9 +356,8 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	}
 
-	@Experimental
-	public List<Shape> refreshShapes(RepositoryConnection shapesRepoConnection) throws SailException {
-
+	@InternalUseOnly
+	public List<Shape> getShapes(RepositoryConnection shapesRepoConnection) throws SailException {
 		SailRepository shapesRepoCache = new SailRepository(
 				SchemaCachingRDFSInferencer.fastInstantiateFrom(shaclVocabulary, new MemoryStore(), false));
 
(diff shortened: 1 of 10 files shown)


hmottestad:
There is a failing test, but I'm not too worried. Will take a look tonight maybe.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -768,6 +746,12 @@ public void setPerformanceLogging(boolean performanceLogging) {
 	 * @return <code>true</code> if serializable validation is enabled, <code>false</code> otherwise.
 	 */
 	public boolean isSerializableValidation() {
+		if (getBaseSail() instanceof SchemaCachingRDFSInferencer) {
+			if (serializableValidation) {
+				logger.error("SchemaCachingRDFSInferencer is not supported when using serializable validation!");
```
jeenbroekstra, line 751: Given that we're not actually stopping execution, I think this should be a warning rather than an error.

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/DeadlockTest.java
```diff
@@ -1,3 +1,10 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Eclipse RDF4J contributors.
```
jeenbroekstra, line 2: Thanks for catching this.

jeenbroekstra (commented):
Looks good - I like the fact that you're getting rid of so much custom locking code, that's a good sign I think. Only have one minor niggle, but other than that, assuming you can get the tests to go green, it looks fine to me.

hmottestad:
One of the failing tests checks that if one transaction modifies the shapes, then all other transactions should see those shapes once they are committed.

This solves the following scenario:

T1: data that is in conflict with shapes in T2 but otherwise valid
T2: shapes

Scenario 1:
T2: begin
T2: add shapes
T1: begin
T1: add data
T2: commit
T1: commit - T1 would pass since it can't see shapes from T2

Scenario 1 is not handled by SERIALIZABLE isolation. 

Scenario 2:
T2: begin
T2: add shapes
T1: begin
T1: add data
T1: commit
T2: commit - T2 would pass since it can't see data from T1

Scenario 2 should be handled by using SERIALIZABLE validation or by using SNAPSHOT isolation with SERIALIZABLE validation.

jeenbroekstra:
```
Scenario 1:
T2: begin
T2: add shapes
T1: begin
T1: add data
T2: commit
T1: commit - T1 would pass since it can't see shapes from T2

Scenario 1 is not handled by SERIALIZABLE isolation.
```

What do you mean with "it's not handled" exactly? If T2 is using serializable isolation, conceptually you should get a commit fail on either T1.commit or T2.commit (depending on which way around they are played). 

> Scenario 2 should be handled by using SERIALIZABLE validation or by using SNAPSHOT isolation with SERIALIZABLE validation.

What is "SERIALIZABLE validation"?

hmottestad:
I'll try to work on understanding the issue better. It's still confusing enough that I can't explain what's going on. 

SERIALIZABLE Validation is a setting that is enabled by default that gives you SERIALIZABLE isolation with regards to your validation when using SNAPSHOT isolation on your transaction.

This is how it works:

 - before the validation phase lock against the sail
 - start a new transaction and replay your changes onto that transaction
 - run validation on that transaction and if it succeeds roll it back and commit your original transaction
 - release the lock

hmottestad:
Found out I was missing a begin() call to the shapes repo connection. That fixed part of this issue.

hmottestad:
I now understand this issue much better.

1. The RDFS reasoner takes an exclusive lock while adding statements and during rollback
2. The ShaclSailConnection took an exclusive lock while adding statements to the shapes repo
3. The ShaclSailConnection uses exclusive locks in the prepare() -> commit()/rollback() phase which will deadlock with 1.

Ontotext also found an issue with their code (which I presume also uses locking in the prepare() -> commit()/rollback() phase), essentially due to 2. !

So I've now cleaned up the code in the ShaclSailConnection to only lock during the prepare() -> commit()/rollback() phase with no more locking when adding data.

Point 1. with far too early locking in the RDFS reasoner has not been fixed. The issue there is that we need to support some sort of versioning of the cache so we can have a local version during the transaction that is merged with the sail level cache during commit. This way we would reduce locking to only be needed during commit.

hmottestad:
We are starting to push the time our tests are taking. The ShaclSail tests now take 30 min in total, which if 75% of all the time. Most of that time I believe is used by the more fuzzy tests that catch issues like this one!

jeenbroekstra:
> TODO: Some of the commits are not signed off!

Just so you're aware: as you are a committer, you are not actually required to sign off your commits. You're on file as "legit" :)  Only third-party contributors need to use the commit sign off everywhere. 

Doesn't hurt to still do it of course, but you don't _have_ to.

Issue #2760, opened 2021-01-14
NPE in QueryEvaluationUtil.compareLiterals() when comparing with custom Literal implementations


damyan-ognyanov:
Thre is an issue with recent `3.5.0` release where `QueryEvaluationUril.compareLiterals()` throws na NPE whn one of the aguments is a custom literal (not an instance of SimpleLiteral created via simpleValueFactory)

```
Caused by: java.lang.NullPointerException
 at org.eclipse.rdf4j.query.algebra.evaluation.util.QueryEvaluationUtil.compareLiterals(QueryEvaluationUtil.java:301)
 at org.eclipse.rdf4j.query.algebra.evaluation.util.QueryEvaluationUtil.compare(QueryEvaluationUtil.java:95)
 at com.ontotext.trree.query.OwlimEvaluationStrategyImpl.evaluate(OwlimEvaluationStrategyImpl.java:1145)
 at org.eclipse.rdf4j.query.algebra.evaluation.impl.StrictEvaluationStrategy.evaluate(StrictEvaluationStrategy.java:1074)
 at com.ontotext.trree.query.OwlimEvaluationStrategyImpl.evaluate(OwlimEvaluationStrategyImpl.java:1426)
 at org.eclipse.rdf4j.query.algebra.evaluation.impl.StrictEvaluationStrategy.isTrue(StrictEvaluationStrategy.java:1805)
```

I've traced it down to the way `Literals.getXsdDatatype(Literal l)` is implemented which while the argument is not an instance of SimpleLiteral simply returns an Optional.empty(). That cause the code in QueryEvaluationUtil.compareLiterals() to hit a NPE while trying to match the XDS versions of the arguments datatypes (around line 301 in QueryEvaluationUtil.java).

PR #2761, opened 2021-01-14
GH-2760: fixes Literals.getXsdDatatype() to work also with custom Literal implementations

damyan-ognyanov:
Added test reproducing GH-2760 - an NPE in QueryEvaluationUtil.compareLiterals()

GitHub issue resolved: #2760 

Literals.getXsdDatatype() used to return Optional.empty() which the argument was not an instance of `SimpleLiteral` that cause the `QueryEvaluationUtil.compareLiterals()` to hit NPE while at least one of its arguments was not an instance of `SimpleLiteral`

Added also a test case reproducing #2760

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
@@ -191,6 +191,31 @@ public class QueryEvaluationUtil {
 				} else if (!strict && leftXsdDatatype.isDurationDatatype() && rightXsdDatatype.isDurationDatatype()) {
 					commonDatatype = XSD.DURATION;
 				}
+			} else if (commonDatatype == null && (leftXsdDatatype != null || rightXsdDatatype != null)) {
+				if (leftDatatype.equals(rightDatatype)) {
+					commonDatatype = leftDatatype;
+				} else if (XMLDatatypeUtil.isNumericDatatype(leftDatatype)
+						&& XMLDatatypeUtil.isNumericDatatype(rightDatatype)) {
+					// left and right arguments have different datatypes, try to find
+					// a
+					// more general, shared datatype
+					if (leftDatatype.equals(XSD.DOUBLE) || rightDatatype.equals(XSD.DOUBLE)) {
+						commonDatatype = XSD.DOUBLE;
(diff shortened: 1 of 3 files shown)


hmottestad:
I'm the developer of this datatype cache mechanisms. The point was to significantly speed up sorting of literals. I remember that this method was originally implemented somewhere else or in some other way, then Jeen had to move it when the basic model stuff was split out. 

I think we should look at this a bit to make sure that this is the correct approach. I'm wondering if maybe there should be a check higher up in the chain instead to stop us from getting to this code here.

hmottestad:
This is now more inline with what the documentation says, and how it was intended to work. That being said, we should consider simplifying the `compareLiterals(...)` method but first I would need to understand how it is supposed to work if we use custom datatypes (eg. not in the XSD namespace).

hmottestad:
@damyan-ognyanov does this fix the issue for you?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/util/QueryEvaluationUtil.java
```diff
@@ -191,6 +191,31 @@ public static boolean compareLiterals(Literal leftLit, Literal rightLit, Compare
 				} else if (!strict && leftXsdDatatype.isDurationDatatype() && rightXsdDatatype.isDurationDatatype()) {
 					commonDatatype = XSD.DURATION;
 				}
+			} else if (commonDatatype == null && (leftXsdDatatype != null || rightXsdDatatype != null)) {
```
jeenbroekstra, line 194: I'm not following the logic here. The `else if` condition checks if `leftXsdDatatype` or `rightXsdDatatype` are not null, but then the body of this condition not do anything with the enums at all, and just uses `leftDatatype` and `rightDatatype` (the IRI values).

jeenbroekstra: (as an aside that compareLiterals method is a monster of nested conditionals - out of scope to refactor for a quick patch release of course but it could do with a bit of cleaning up)

hmottestad: Might be a typo on my end. Wrote that code late last night. I'll fix it.

hmottestad: Fixed

jeenbroekstra (approved):
LGTM.

Issue #2457, opened 2020-08-22
Use the SHACL reference implementation to validate test cases


hmottestad:
It would be good to validate that all our test cases are compliant with the reference implementation by:

 - checking the shapes against the SHACL spec
 - checking each test case to make sure the validation result is the same

hmottestad:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22578

hmottestad:
I have a CQ pending here, but now I'm starting to wonder....do we need to do CQs for test dependencies anymore @jeenbroekstra ? The IP team hasn't done anything with my CQ since I opened it.

barthanssens:
Hmz, I think we still need CQ for test dependencies, but it's not required anymore to get approval before it can go into the dev branch (approval is of course still needed upon release)

https://www.eclipse.org/projects/handbook/#ip

You may want to add a kind reminder as comment in IPZilla (works for me :-))

hmottestad:
Would very much like to merge this so I can use it in one of my other branches.

PR #2763, opened 2021-01-17
GH-2457 shacl reference implementation

hmottestad:
GitHub issue resolved: #2457 

----


core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
```diff
@@ -538,6 +744,61 @@ static String modelToString(Model model) {
...
+
+				breadthFirstSearchBuffer = tempBuffer;
+
+			}
+
+			return emptyModel;
+		}
+
+	}
+
```
hmottestad, line 801: @jeenbroekstra do we have something like this already? It starts at a node and follows all outbound statements from that node using breadth first search. I use it to extract a subset of the validation report excluding the source shape.

If we don't have something like this I think it could be a useful addition to the Models helper class.

hmottestad: Jena has something similar: https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/GraphExtract.html

jeenbroekstra: I don't think we have something quite like that right now. I've been thinking that more generally I'd like a few more graph-traversal kind of operations available in RDF4J, and this fits right in with that.

So yes, nice idea. Want to open a ticket for adding this as a new feature?

jeenbroekstra:
Looks good to me. Please don't forget to squash.

pom.xml
```diff
@@ -963,18 +963,6 @@
...
-			<releases>
-				<enabled>false</enabled>
-			</releases>
-			<snapshots>
-				<enabled>true</enabled>
-			</snapshots>
-			<id>oss-sonatype-snapshots</id>
-			<url>https://oss.sonatype.org/content/repositories/snapshots</url>
-		</repository>
-	</repositories>
```
jeenbroekstra, line 977: Wait, this should not be removed.

jeenbroekstra (changes requested):

pom.xml
hmottestad: Ahh. Forgot. I remember now that I started working on this branch once when master was broken. Will take a good look and see if there is anything else I forgot.

hmottestad:
Also. I don't think I can squash since I merged develop into this branch. Can we have the squash and merge button back please? At least in the iOS app I now get asked what email I want squash and merge to use.

jeenbroekstra:
> Also. I don't think I can squash since I merged develop into this branch

You can still squash it. It's just a little harder to get right because you may be asked to resolve conflicts multiple times. If you want to avoid this, please use rebase instead of merge to keep your feature branch in sync with the main branches.

> Can we have the squash and merge button back please? At least in the iOS app I now get asked what email I want squash and merge to use.

I don't know what the iOS app does or how it does it, but the Github web interface squash and merge option still does not ask this, as far as I've been able to test: it just uses the author of the PR (not author of the commits, nor the person pushing the button) as the author of the squash commit.

I am not going to re-enable this just now, just for the sake of this PR. If you want to reopen the discussion on the merge strategy though, that's fine: I believe the ECA check has been modified to be more lenient with 'noreply.github.com' addresses and the like, so it shouldn't be as massive a problem as it was before. Perhaps we can run a few quick tests and eventually agree to go back to squash and merge.

hmottestad:
This is now squashed into two commits + two merge commits where I merged develop into this branch.

jeenbroekstra (approved):

Issue #2772, opened 2021-01-26
SHACL - Compressed target chain


hmottestad:
When reporting on a different level than what gets validated, multiple paths to the same target can cause the validation report to not report on invalid targets.

#### Example shape

```
ex:PersonShape
	a sh:NodeShape  ;
	sh:targetClass ex:Person, ex:SecondTarget ;
	sh:property [
		sh:path ex:knows1 ;
		sh:property [
		        sh:path ex:knows2;
                        sh:node [
                                sh:property [
		                        sh:path ex:knows3;
                                        sh:property [
		                                sh:path ex:knows4;
		                                sh:class ex:Person ;
                                        ]
                                ]
                        ]
		]
	] .

```

#### Example data

```
ex:validPerson1 a ex:Person ;
       	ex:knows1 ex:p1, ex:p1_1.

ex:p1 ex:knows2 ex:p2.
ex:p1_1 ex:knows2 ex:p2.

ex:p2 ex:knows3 ex:p3.
ex:p3 ex:knows4 ex:p4.
```

Should report both `ex:p1` and `ex:p1_1` as invalid.

### Solution

Introduce the concept of compressed target chain.

The validation target `ex:p3` will contain both target chains that lead to it. Eg. `[ex:p1, ex:p2, ex:p3] [ex:p1_1, ex:p2, ex:p3]`

PR #2773, opened 2021-01-26
GH-2772 Compressed target chain in ShaclSail

hmottestad:
GitHub issue resolved: #2772 

----


hmottestad:
# ! Remember to use squash when merging. !

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/DatatypeFilter.java
```diff
@@ -17,10 +17,12 @@
 public class DatatypeFilter extends FilterPlanNode {
 
 	private final Resource datatype;
+	private StackTraceElement[] stackTrace;
 
 	public DatatypeFilter(PlanNode parent, Resource datatype) {
 		super(parent);
 		this.datatype = datatype;
+//		stackTrace = Thread.currentThread().getStackTrace();
 	}
```
hmottestad, line 27: The stack trace is used while debugging, but I'm not sure how to make it so that it is only populated when we are debugging (or asserts enabled). So I've just commented them out instead.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/GroupByFilter.java
```diff
@@ -55,7 +55,7 @@ private void calculateNext() {
 						tempNext = parentIterator.next();
 					}
 
-					this.next = new ValidationTuple(tempNext);
+					this.next = tempNext;
```
hmottestad, line 58: This kind of efficiency is possible because ValidationTuple is now immutable.

hmottestad:
This is a bit complicated. Sorry for that.

The essence is that our SHACL implementation executes validation bottom-up. While the SHACL spec is written to support top-down.

```
ex:PersonShape
	a sh:NodeShape  ;
	sh:targetClass ex:Person ;
	sh:node [
	        sh:property [
		        sh:path ex:knows ;
		        sh:property [
			        sh:path ex:age ;
		                sh:datatype xsd:integer ;
		        ]
	        ]
	] .

```

Take the shape above. It says that for every ex:Person if it has a path `ex:knows / ex:age` then the values should all be integers. The typical way to implement this is to iterate through all people, then iterate through all their `ex:knows` predicates 
 and then through all their `ex:age` predicates. Then validate all the values for datatype.

Instead what we do is start with the validation of the datatype and retrieve all relevant values, then bubble up to the top of the shape and produce a validation result. To make sure we keep track of which values belong to which targets (eg. instances of ex:Person) we keep a chain of those values for every value we are validating, and as we bubble up to outer shape we pop values off of our chain. 

The issue is that sometimes there are two different chains to the same value that we want to validate. All our join operators assume that the left input has distinct targets, but if there are two different chains to the same value, then we would need to have two instances of the same value to keep track of both chains. To fix this we "compress" both chains into one chain, and when we pop the chain in the outer shape the compress chains become separate again.

hmottestad:
There is one thing I would like to do.

I have a lot of lists of values where I have to pop the last value off the list. Since I'm doing it immutably we end up with a lot of lists. What I thought I would do is to make a projection instead.

```
List<Value> chain = new ArrayList<>(this.chain);
chain.remove(chain.size() - 1);
return Collections.unmodifiableList(chain);
```

Would become:

```
return new ProjectedList<>(this.chain, 0, this.chain.size() - 1);
```

Has anyone come across something like this? So I don't need to implement it myself.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/DatatypeFilter.java
jeenbroekstra: This is fine for me, but for future reference you can do `if (logger.isDebugEnabled())`  to check that we're logging at debug level (which is usually a good proxy for actually doing debugging :)).

jeenbroekstra (approved):

jeenbroekstra:
> There is one thing I would like to do.
> 
> I have a lot of lists of values where I have to pop the last value off the list. Since I'm doing it immutably we end up with a lot of lists. What I thought I would do is to make a projection instead.
> 
> ```
> List<Value> chain = new ArrayList<>(this.chain);
> chain.remove(chain.size() - 1);
> return Collections.unmodifiableList(chain);
> ```
> 
> Would become:
> 
> ```
> return new ProjectedList<>(this.chain, 0, this.chain.size() - 1);
> ```
> 
> Has anyone come across something like this? So I don't need to implement it myself.

I think `ArraysList.subList()` might do what you're after: https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#subList-int-int-

hmottestad:
> > There is one thing I would like to do.
> > I have a lot of lists of values where I have to pop the last value off the list. Since I'm doing it immutably we end up with a lot of lists. What I thought I would do is to make a projection instead.
> > ```
> > List<Value> chain = new ArrayList<>(this.chain);
> > chain.remove(chain.size() - 1);
> > return Collections.unmodifiableList(chain);
> > ```
> > 
> > 
> > Would become:
> > ```
> > return new ProjectedList<>(this.chain, 0, this.chain.size() - 1);
> > ```
> > 
> > 
> > Has anyone come across something like this? So I don't need to implement it myself.
> 
> I think `ArrayList.subList()` might do what you're after: https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#subList-int-int-

Thanks, that was exactly what I was looking for :)

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -972,7 +971,8 @@ public IsolationLevel getDefaultIsolationLevel() {
 
 			Disabled("Disabled"),
 			Auto("Auto"),
-			Bulk("Bulk");
+			Bulk("Bulk"),
+			BulkParallelCache("BulkParallelCache");
```
hmottestad, line 975: @jeenbroekstra What do you think about this name. `Bulk` is used for typical bulk operations like loading a lot of data. Since that much data could make us run out of memory, but parallel validation and caching is disabled. But if you do have the memory, then running the bulk validation in parallel with cache will be up to 2.5x faster (in our benchmark). Question is, what do you think of the naming?

jeenbroekstra: I have no immediate objection to the naming itself, but I am a little confused how this combines with `setParallelValidation`, and more generally: instead of creating a new enums for combinations of settings (Bulk + Parallel + Caching), shouldn't we perhaps make Parallel and Caching enum options themselves?

hmottestad: Might make for sense to make separate enums yes. The `Bulk` enum currently overrides any sail level settings like `setParallelValidation(...)`.

hmottestad:
@jeenbroekstra I've ended up doubling the amount of memory for our tests (1G to 2G). I've also filed an issue https://github.com/eclipse/rdf4j/issues/2829 . Essentially we are creating a lot more observations with the new AST, which is causing us to run out of memory when using SERIALIZABLE transactions on the NativeStore.

jeenbroekstra:
> @jeenbroekstra I've ended up doubling the amount of memory for our tests (1G to 2G). I've also filed an issue #2829 . Essentially we are creating a lot more observations with the new AST, which is causing us to run out of memory when using SERIALIZABLE transactions on the NativeStore.

First of all: fixing this OOM error is not really related to #2772, but is a separate issue (#2806). Can you remove it from this PR and submit it separately? It wouldn't be such a problem if we could merge-commit this PR, but since you need to squash-and-merge it, this will lead to a single commit on the root pom with the wrong issue number and a message that doesn't make sense from the POV of the root pom.

Second, and more to the point of the actual change: I'm a little hesitant about upping our memory limit in this fashion for the entire test suite, but given that you've created a followup ticket to look into the underlying problem I guess this is workable for now (at least the build no longer fails :) ) - though I am a little concerned that it seems to blow out our verification build time by a good 10-15 minutes. We'll have to keep an eye on that.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
hmottestad: Decided to not solve this here. So I've removed BulkParallelCache.

Issue #2774, opened 2021-01-26
Contention in AbstractCloseableIteration synchronization of `closed` causes query timeout to fail


hmottestad:
In `AbstractCloseableIteration` we synchronize wirtes to `closed` in order to only allow an iteration to be closed once. This is a good approach, however subclasses can also use synchronization and when they do it would contend with the synchronization of `closed`. 

This issue has been reported in how the query timeout mechanism handles slow remote endpoints. If the endpoint hangs, then the code that is waiting on the endpoint would block the timeout mechanism from working if the code is within a critical section.

Propose to simply synchronize on our own monitor instead of on `this`.

PR #2775, opened 2021-01-26
GH-2774 introduce our own monitor instead of using _this_ in AbstractCloseableIteration

hmottestad:
GitHub issue resolved: #2774 

Fix GH-2774 by introducing our own object to monitor on.

----


hmottestad:
@jeenbroekstra would appreciate if you could take a look. The changes shouldn't have any negative effect since the code wasn't originally synchronized (we used to use atomic Boolean). I made a good description of the problem in the issue.

jeenbroekstra (approved):
Good find. Looks good to me. 

I'm fine with documenting instead of having a test for this (concurrency is difficult to write tests for and it doesn't seem worth the effort in this case). Were you able to reproduce the original issue in a manual setting and check if this fix addresses it?

damyan-ognyanov:
@jeenbroekstra I was able to check the changes from the PR on the set of tests that we used to spot the concurency issue and can confirm it fixes it.

Issue #2765, opened 2021-01-18
Utility functions for easy mapping between prefixed names and IRIs in SPARQL queries


jeenbroekstra:
It would be useful to be able to quickly convert from a prefixed name to a full IRI, and the other way around, when working with SPARQL queries in code.

Example uses:

```java
Namespace ex = new SimpleNamespace("ex", "http://example.org");

IRI fullIRI = Values.iri("ex:foo", ex); // fullIRI is http://example.org/foo
String prefixedName = Values.prefixedName(fullIRI, ex); // prefixedName is "ex:foo"
```

Working with a collection of namespaces (e.g. from a Model or a RepositoryConnection):

```java
IRI fullIRI = Values.iri("ex:foo", connection.getNamespaces()); // fullIRI is http://example.org/foo
```

A side utility that would also be useful is a quick way to convert that same collection of namespaces to a SPARQL query prefix declaration header, so that we could for example do something like this:

```java
try(var conn = rep.getConnection) {
   String query = SPARQLQueries.prefixes(conn.getNamespaces) 
                         + "SELECT ?s ?o "
                         + "WHERE {  ?s ex:property ?o . }");
}
```

Up for discussion is exact naming and location of these functions. 

One possibility is  to abstract away from the specifics of SPARQL and instead offer something based on the CURIE spec (https://www.w3.org/TR/curie/). The problem with that is that SPARQL is slightly more restrictive, so not every legal CURIE is also a legal SPARQL prefixed name.

See also user group discussion at https://groups.google.com/d/msgid/rdf4j-users/a8a97b0a-5aae-48ed-a356-15f76ef0d41bn%40googlegroups.com .

hmottestad:
@jeenbroekstra did you ever implement: `String prefixedName = Values.prefixedName(fullIRI, ex); // prefixedName is "ex:foo"` ?

jeenbroekstra:
Huh. No don't think so, completely overlooked that. 

I think its practical use would be limited but if you want we can add a follow-up issue for it.

PR #2776, opened 2021-01-27
GH-2765 namespace utils

jeenbroekstra:
GitHub issue resolved: #2765  

- added utility method to `Values` class to quickly create an IRI for a given prefixed name + set of Namespaces
- added semi-related method `SPARQLQueries.getPrefixClauses` for getting a SPARQL prefix declaration string for one or more namespace objects.
- some refactoring/cleanup of some older related utilities

----


core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * Escape the supplied string with backslashes for any special characters, so it can be used as a string literal
+	 * value in a SPARQL query.
+	 * 
+	 * @see https://www.w3.org/TR/sparql11-query/#grammarEscapes
+	 * 
+	 * @since 3.6.0
+	 */
+	public static String escape(String s) {
```
jeenbroekstra, line 51: I'm a little bit uncertain about keeping this method as-is: while it handles string literal escaping as describe in section 19.4 (https://www.w3.org/TR/sparql11-query/#grammarEscapes), it does not handle the more general unicode code point escaping (see https://www.w3.org/TR/sparql11-query/#codepointEscape) .  We have functionality for that elsewhere but we may want to make it more obvous which method(s) to apply where. It goes a bit beyond the scope of this PR to fully fix that though.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+public class SPARQLQueries {
+
+	/**
+	 * Creates a string representing of the supplied {@link Namespace namespaces} as SPARQL prefix declarations. This
+	 * can be used when composing a SPARQL query string in code, for example:
+	 * 
+	 * <pre>
+	 * <code>
+	 * String query = SPARQLQueries.getPrefixClauses(connection.getNamespaces()) + "SELECT * WHERE { ?s ex:myprop ?o }";
```
hmottestad, line 26: Looks nice. Would be even nicer if we could do something like `sail.prependNamespaces(true)` or something like that which would essentially do this automatically.

hmottestad: Thought that is unrelated to this PR obviously. Just my wishful thinking :P

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
hmottestad: Up to you.

hmottestad (approved):
LGTM

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
barthanssens: Well, if it's documented in the code that it may not work in all cases... then I'd say "don't let perfect be the enemy of the good"

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
```diff
@@ -0,0 +1,131 @@
...
+		while (backSlashIdx != -1) {
+			sb.append(s.substring(startIdx, backSlashIdx));
+
+			if (backSlashIdx + 1 >= sLength) {
+				throw new IllegalArgumentException("Unescaped backslash in: " + s);
+			}
+
+			char c = s.charAt(backSlashIdx + 1);
+
+			if (c == 't') {
```
barthanssens, line 96: Probably a matter of personal taste: wouldn't a switch instead of if-else-if be slightly better ?

barthanssens (approved):

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
jeenbroekstra: Thanks, we'll keep as-is then.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
jeenbroekstra: I just copy-pasted the old code without looking at it to be honest. But I want to add some test coverage anyway so I'll also clean this up. Good catch.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLQueries.java
jeenbroekstra: There's some functionality already for automatically inserting "well-known" namespaces when preparing a SPARQL query, though it currently uses a hardcoded list of constants rather than whatever is available in the repository (this happens in the PrefixDeclProcessor which is part of the SPARQL parser). I wouldn't be against adding a configuration option at the level of the Repository API to automatically let it insert prefix declarations from the underlying store into SPARQL queries. As you said though: it's out of scope for this issue :)

Issue #2768, opened 2021-01-20
Reimplement graph isomorphism


jeenbroekstra:
The current implementation of graph isomorphism in `Models.isomorphic` is a naive depth-first search that has very poor average-case performance and even potential non-completing (or at least not in reasonable time) performance on even fairly small test cases.

In a technical paper by Aidan Hogen (see http://aidanhogan.com/docs/rdf-canonicalisation.pdf) a more optimized algorithm for isomorphism is proposed. Bonus advantage is that it also offers (sub)algorithms for canonicalizing blank nodes that would be useful for things like skolemization (see also #2734). 

I've done a quick scratchpad spike of comparitive performance of this algorithm against what we have in RDF4J, and it's looking promising. This task is about translating that spike into a full new reimplementation.

hmottestad:
Ooohhhh :)

Would be awesome. I've noticed one of the SHACL tests I have that uses isomorphism to compare results will not complete (AFIK) when the models are sorted. 

I've also in a previous project ended up using Jena specifically for isomorphic, since it's pretty much always faster and always seems to complete. 

Do you know what algorithm Jena uses?

jeenbroekstra:
> Do you know what algorithm Jena uses?

I don't (I do not, as a rule, look at Jena, as I'd rather not be accused of plagiarism). But I imagine it's close to Hogen's algorithm. The notion of using a canonical blank node mapping as a basis for comparison is a fairly well-known one, even from before this paper I believe. And you need a solid theoretical foundation for an algorithm that actually does well on non-trivial examples (graph isomorphism in general is an NP problem).

jeenbroekstra:
> Would be awesome. I've noticed one of the SHACL tests I have that uses isomorphism to compare results will not complete (AFIK) when the models are sorted.

Could you share that with me as an example? I could perhaps use it as one of the tests for the new implementation.

jeenbroekstra:
So the annoying thing is I have this functionally correct now... But it's slightly slower on most benchmarks. The good news is that I suspect I can optimise quite a bit in how data structures are handled internally. 

I also _suspect_ that it's less prone to endless loops depending on sorting of statements. But I need a test case that reproduces that problem to verify that.

hmottestad:
Sorry for being slow @jeenbroekstra . Here is the example that is super slow for me: https://github.com/eclipse/rdf4j/pull/2786

I don't actually know how long time it takes.

hmottestad:
[...]

PR #2787, opened 2021-01-30
GH-2768 new isomorphism algorithm

jeenbroekstra:
GitHub issue resolved: #2768  

A basic reimplementation of structural graph isomorphism using blank node canonicalization as per Hogan's paper

----


jeenbroekstra:
Some performance benchmarks:

# GH-2768 algorithm (initial)
```
Benchmark                                   Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes              avgt   10    7.193 ±  0.553  ms/op
IsomorphicBenchmark.bsbm                    avgt   10  209.855 ± 11.531  ms/op
IsomorphicBenchmark.bsbmArrayList           avgt   10  362.381 ± 11.750  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic       avgt   10  208.463 ± 11.631  ms/op
IsomorphicBenchmark.bsbmTree                avgt   10  203.189 ±  2.621  ms/op
IsomorphicBenchmark.empty                   avgt   10    0.001 ±  0.001  ms/op
IsomorphicBenchmark.emptyNotIsomorphic      avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.internallyIsomorphic    avgt   10  415.244 ± 23.781  ms/op
IsomorphicBenchmark.list                    avgt   10   13.223 ±  0.158  ms/op
IsomorphicBenchmark.listEquals              avgt   10    0.608 ±  0.002  ms/op
IsomorphicBenchmark.longChain               avgt   10    0.151 ±  0.001  ms/op
IsomorphicBenchmark.manyProperties          avgt   10   55.516 ±  0.287  ms/op
IsomorphicBenchmark.manyProperties2         avgt   10   38.477 ±  0.111  ms/op
IsomorphicBenchmark.sameModel               avgt   10   ≈ 10⁻⁶           ms/op
IsomorphicBenchmark.shacl                   avgt   10    6.870 ±  0.106  ms/op
IsomorphicBenchmark.sparqlTestCase          avgt   10    0.359 ±  0.003  ms/op
IsomorphicBenchmark.spinFullForwardchained  avgt   10  137.883 ±  0.633  ms/op
IsomorphicBenchmark.uuid                    avgt   10   55.814 ±  0.296  ms/op
```
## GH-2768 optimized + validation report test (commit fb61072)

Benchmark                                   Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes              avgt   10    5.972 ±  0.380  ms/op
IsomorphicBenchmark.bsbm                    avgt   10   55.922 ±  5.075  ms/op
IsomorphicBenchmark.bsbmArrayList           avgt   10  196.037 ± 11.573  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic       avgt   10   36.147 ±  0.409  ms/op
IsomorphicBenchmark.bsbmTree                avgt   10   82.633 ±  1.872  ms/op
IsomorphicBenchmark.empty                   avgt   10   ≈ 10⁻⁴           ms/op
IsomorphicBenchmark.emptyNotIsomorphic      avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.internallyIsomorphic    avgt   10  384.797 ± 31.194  ms/op
IsomorphicBenchmark.list                    avgt   10   11.516 ±  0.133  ms/op
IsomorphicBenchmark.listEquals              avgt   10    0.924 ±  0.005  ms/op
IsomorphicBenchmark.longChain               avgt   10    0.023 ±  0.001  ms/op
IsomorphicBenchmark.manyProperties          avgt   10  137.631 ±  0.205  ms/op
IsomorphicBenchmark.manyProperties2         avgt   10  133.255 ±  4.994  ms/op
IsomorphicBenchmark.sameModel               avgt   10   ≈ 10⁻⁶           ms/op
IsomorphicBenchmark.shacl                   avgt   10    7.309 ±  0.876  ms/op
IsomorphicBenchmark.shaclValidationREport   avgt   10  126.804 ±  8.151  ms/op
IsomorphicBenchmark.sparqlTestCase          avgt   10    0.373 ±  0.029  ms/op
IsomorphicBenchmark.spinFullForwardchained  avgt   10  158.539 ±  8.004  ms/op
IsomorphicBenchmark.uuid                    avgt   10   13.516 ±  1.241  ms/op

# old Models algorithm

```
Benchmark                                   Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes              avgt   10    0.097 ±  0.007  ms/op
IsomorphicBenchmark.bsbm                    avgt   10   46.783 ±  0.069  ms/op
IsomorphicBenchmark.bsbmArrayList           avgt   10  139.373 ±  1.298  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic       avgt   10   29.215 ±  0.150  ms/op
IsomorphicBenchmark.bsbmTree                avgt   10   76.278 ±  4.570  ms/op
IsomorphicBenchmark.empty                   avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.emptyNotIsomorphic      avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.internallyIsomorphic    avgt   10   19.885 ±  0.621  ms/op
IsomorphicBenchmark.list                    avgt   10    4.340 ±  0.008  ms/op
IsomorphicBenchmark.listEquals              avgt   10    0.675 ±  0.001  ms/op
IsomorphicBenchmark.longChain               avgt   10    0.013 ±  0.001  ms/op
IsomorphicBenchmark.manyProperties          avgt   10  175.772 ±  2.484  ms/op
IsomorphicBenchmark.manyProperties2         avgt   10  181.687 ±  2.452  ms/op
IsomorphicBenchmark.sameModel               avgt   10   ≈ 10⁻⁶           ms/op
IsomorphicBenchmark.shacl                   avgt   10    2.540 ±  0.009  ms/op
IsomorphicBenchmark.shaclValidationREport   avgt   10 5138.554 ± 46.644  ms/op
IsomorphicBenchmark.sparqlTestCase          avgt   10    0.040 ±  0.001  ms/op
IsomorphicBenchmark.spinFullForwardchained  avgt   10  753.499 ±  2.776  ms/op
IsomorphicBenchmark.uuid                    avgt   10    4.293 ±  0.013  ms/op
```

As can be seen, the new algorithm performs slightly poorer on most non-complex cases, but unless I'm reading the data wrong we're talking a comparitive difference in the order of 10-20ms in most cases. For slightly more complex comparisons though, it makes some big improvements (see the new shaclValidationReport comparison, and that is even without taking into account that if input ordering is different, the old model sometimes doesn't terminate in reasonable time at all). 

There's probably further optimizations to be made, I'll run a few tests with a profiler to spot any obvious bottlenecks.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
```diff
@@ -0,0 +1,415 @@
...
+							context
+					);
+				}
+
+				Model contextInlModel1 = model1.filter(null, null, null, context);
+				Model contextInModel2 = model2.filter(null, null, null, context);
+				if (contextInlModel1.size() != contextInModel2.size()) {
+					return false;
+				}
+				final Model canonicalizedContext1 = isoCanonicalize(contextInlModel1);
```
hmottestad, line 112: FYI. Final variables inside methods don't make any difference for performance.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
```diff
@@ -0,0 +1,415 @@
...
+
+	protected static Set<BNode> getBlankNodes(Model m) {
+		final Set<BNode> blankNodes = new HashSet<>();
+
+		m.subjects().forEach(s -> {
+			if (s.isBNode()) {
+				blankNodes.add((BNode) s);
+			}
+		});
+		m.objects().forEach(o -> {
```
hmottestad, line 175: Does calling m.objects and m.subjects not cause the model to do a lot of work? Wouldn't it be faster to just iterate through all the statements once and check the subject and object at the same time instead?

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
```diff
@@ -0,0 +1,415 @@
...
+		final List<Collection<BNode>> sortedPartitions = new ArrayList<>(partitionMapping.asMap().values());
+		Collections.sort(sortedPartitions, new Comparator<Collection<BNode>>() {
+			public int compare(Collection<BNode> a, Collection<BNode> b) {
+				int result = a.size() - b.size();
+				if (result == 0) {
+					// break tie by comparing value hash
+					HashCode hashOfA = blankNodeMapping.get(a.iterator().next());
+					HashCode hashOfB = blankNodeMapping.get(b.iterator().next());
+					BigInteger difference = new BigInteger(1, hashOfA.asBytes())
+							.subtract(new BigInteger(1, hashOfB.asBytes()));
```
hmottestad, line 198: If we used a 64bit hash I suppose we could `long` instead for added performance. What would be the trade offs?

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
```diff
@@ -0,0 +1,415 @@
...
+	private static BigInteger mappingSize(Map<BNode, HashCode> mapping) {
+		BigInteger size = mapping.values()
+				.stream()
+				.map(hashCode -> new BigInteger(1, hashCode.asBytes()))
+				.reduce(BigInteger.ZERO, (v1, v2) -> v1.add(v2));
+		return size;
+	}
+
+	private static Model labelModel(Model original, Map<BNode, HashCode> hash) {
+		Model result = new LinkedHashModel(original.size());
```
hmottestad, line 298: Worth benchmarking against the dynamic model. Kinda depends what you end up doing with the model.

hmottestad (commented):
Haven't read the paper, so it's no wonder I can't really follow the code. So these are just a couple of random comments on stuff that may or may not make a difference.

hmottestad:
Benchmark results are nice :) but some seem very poor. 5-10 ms is a big difference when it's amounts to 60x slower like in the first case (if I'm reading things correctly). 

A general hit on performance is very acceptable if it means we don't have anymore infinite loops!

compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicBenchmark.java
```diff
@@ -59,6 +64,7 @@
 	private Model empty_2 = getModel("empty.ttl");
 	private Model blankNodes_2 = getModel("blankNodes.ttl");
 	private Model shacl_2 = getModel("shacl.ttl");
+	private Model shaclValidationReport_2 = getModel("shaclValidationReport.ttl");
```
hmottestad, line 67: The sorting I was doing was what was causing the isomorphic check on this file to take so long. Thought 6 seconds was kinda slow, it was nothing in comparison to how slow it was when I sorted on object.

jeenbroekstra: I'm aware, which is why I added the sorted version as a unit test in IsomorphicTest. It has a timeout of 2 seconds on it.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
jeenbroekstra: Good point. I thought I had already changed that, must have accidentally rolled it back. Will fix.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
jeenbroekstra: The main reason for using a 128 bit hash is that it gives us a hash that is, for all practical purposes, globally unique (so not just unique within the graph). This means we can use these values as unique identifiers for skolemization.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
jeenbroekstra: I tend to use final as a statement of intent rather than (purely) for performance reasons - plus in code where there's a lot juggling of (in this case) Model objects going on, it is helpful to make sure a reference is not overridden later on down.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
jeenbroekstra: Note that this model creation only happens once, right at the end, so I suspect the performance difference will not be major - but yeah we can try.

jeenbroekstra:
The fact that models encapsulate multiple (named) graphs and that we allow blank node identifiers for named graphs is doing my head in a bit. I thought I could get away with just logging a warning and otherwise ignoring it, but we're relying on being able to compare across blank node contexts for some of our  RDF format parser tests (e.g. TriG - which allows blank node graph identifiers).

jeenbroekstra:
I may have found a solution for the blank node contexts thing. Let's see what the build check says, and then we can focus on further optimizing again.

hmottestad:
Build check seems to agree with you :)

hmottestad:
I'm very excited about this PR btw. In a project a few years ago where we used RDF4J I had to specifically serialize our data and transfer it over to Jena for our isomorphic check. 

Also in the ShaclSail I use isomorphic checks as part of the test suite and I've had to jump through a bunch of hoops to get performance where it doesn't slow down everything. I think I actually ended up checking the turtle output first so I could skip isomorphic. There are some 15000 test cases, so every millisecond counts.

hmottestad:
I made a minor performance improvement to the LinkedHashModel, doesn't really make much of a difference. I still think the biggest impact on performance is coming from the large use of HashMaps and HashMultiMaps. If there is a way you can think of to not use hash maps in general, then I think your performance will increase dramatically for the trivial benchmarks.

jeenbroekstra:
> I made a minor performance improvement to the LinkedHashModel, doesn't really make much of a difference.

I don't mind the performance tweak you made (though I didn't check in detail), however this has little or nothing to do with what we're trying to solve here. The use of LinkedHashModel in the current version of this algorithm is incidental (and in fact I was planning to swap it out for a DynamicModel, though I suspect the performance different will be negligable). It is not surprising to me that making LinkedHashModel more performant has no impact on the algorithm's performance.

> I still think the biggest impact on performance is coming from the large use of HashMaps and HashMultiMaps. If there is a way you can think of to not use hash maps in general, then I think your performance will increase dramatically for the trivial benchmarks.

Being able to pass collections of constantly-changing pairs of blank nodes and hashcode, where we can quickly look up a bnode given a hashcode and vice versa, is kind of intrinsic to the approach. What this algorithm does is iterate and constantly recompute hashcodes - and we need to be able to compare the hashcodes from a previous iteration against the ones computed in the current one, as well as be able to quickly see which blank nodes map to which hashcodes.

I suspect that we may be able to take some shortcuts on that if we move from the current stateless utility to an object method on Model itself: it will allows us to keep some internal state on the model being compared, which may lead to less passing and copying of data (and may also allow us to do optimizations that are specific to the model type). But that's not a line of optimization I want to pursue right now, we'll save that for a followup.

Profiling shows that the biggest performance hotspot in general is the actual hashing algorithm itself (though hashmap lookups are of course also up there, as there is a _lot_ of that going on). That is at some level unavoidable, though a possible optimization could be to make the algorithm configurable, using a weaker/faster hash when globally unique hashes are not a requirement. I'm not confident on that though: earlier experiments gave me disappointing results when switching to a weaker hash, and it's actually not trivial to pick a good algorithm that has the required characteristics.

Another possible hotspot is the computation of compatible automorphisms, we're doing a lot of arraylist manipulation there that shows up as expensive. Taking a closer look at that now. 

Having said all that: even if we don't find any further optimizations right now this algorithm is already more useful than the previous one, as at least we now have something that doesn't unpredictably get stuck if things are in the wrong order.

jeenbroekstra:
Another data point: I just did an isomorphic comparison using two copies of a large OWL ontology (~68,000 lines of Turtle). The new algorithm takes (on my laptop) 22 seconds to complete (single run, no warmups), which sounds slightly disappointing until you compare it against the old algorithm, which crashes with an OutOfMemoryError.

hmottestad:
Have to fix the little things when we see them. 

I'm happy this works well on larger models. Still a bit worried about the smaller ones like the `blankNodes()` benchmark. Also `internallyIsomorphic()` and `sparqlTestCase()`. Those are considerably slower now.

jeenbroekstra:
> I'm happy this works well on larger models. Still a bit worried about the smaller ones like the `blankNodes()` benchmark. Also `internallyIsomorphic()` and `sparqlTestCase()`. Those are considerably slower now.

I'll doublecheck again, but most of the performance differences come down to computing hashcodes being a bit pricy - which is something that we cannot avoid doing. Also keep in mind that internallyIsomorphic in the old approach was only faster if you hadn't had the sorting done differently. It's one of those cases that actually hangs if you change the order in one of the models. I'll do a few profile runs and post the hotspots.

jeenbroekstra:
I may consider dropping the notion of using strong hashing to reduce risk of global collision for now, I'll run a few tests with cheaper hashes to see if the performance gain is significant enough.

jeenbroekstra:
I've reduced  the complexity of the hashing function to 32 bits, and replaced the BigInteger maths with longs. This is possible because we currently only use this hash for isomorphic comparison purposes, and not for assign persistent public canonical identifiers to blank nodes (in which case we'd need a stronger hash to avoid potential clashes between different models). 

I haven't re-run all benchmarks, but a subset, giving these numbers:

```
Benchmark                                 Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes            avgt   10    4.851 ±  0.033  ms/op
IsomorphicBenchmark.internallyIsomorphic  avgt   10  340.253 ± 23.828  ms/op
IsomorphicBenchmark.sparqlTestCase        avgt   10    0.104 ±  0.001  ms/op
```

So quite a significant performance increase. Also, the big ontology run I previously did now executes in 16 seconds instead of 22, so that's not bad either (though of course this is only measured as a single run so it's more a smoke test than a real benchmark).

jeenbroekstra:
I am happy as-is. There are followups, including exploring even faster hashes, and working on making the canonicalize function publicly available (with a stronger hash) for skolemization purposes. But as a first cut this is good, and quite an improvement over the original algorithm.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
jeenbroekstra: Fwiw (see also below) I've switched it down to a 32 bit hash and using long instead of BigInteger. We can get away with this for isomorphism purposes (but not for skolemization). Oh and long instead of int (in case you're wondering) to handle potential overflows when we combine hashes.

somanythings:
Big thanks @jeenbroekstra !! 

I'd missed this PR, hence the commit comments (which I'd only seen through a link in a comment). I'd taken a copy of this and tried in our project where we had non-completing comparisons, and things all run in seconds, now. 


One minor thing was I had to remove the jdk1.10 use of orElseThrow but that's because for a reason I couldn't fathom our project is targeting 1.8. I've been in scala and kotlin world too long to know whether that's a java or jvm enhancement, and whether I'll need to worry once it's in main.

hmottestad:
### `blankNodes()`

Screenshot 2021-02-06 at 12 50 10

#### Zoomed in on `hashBNodes(...)`

Screenshot 2021-02-06 at 12 54 16

hmottestad:
@jeenbroekstra I'm now getting an error on one of the benchmarks.

```
# JMH version: 1.21
# VM version: JDK 15, OpenJDK 64-Bit Server VM, 15+36
# VM invoker: /Library/Java/JavaVirtualMachines/adoptopenjdk-15.jdk/Contents/Home/bin/java
# VM options: -Xms1G -Xmx1G -XX:+UseSerialGC
# Warmup: 20 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport

# Run progress: 0.00% complete, ETA 00:05:00
# Fork: 1 of 1
# Warmup Iteration   1: 13:20:08.602 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.RDFParserRegistry - Registered service class org.eclipse.rdf4j.rio.turtle.TurtleParserFactory
13:20:08.611 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.RDFParserRegistry - Registered service class org.eclipse.rdf4j.rio.turtlestar.TurtleStarParserFactory
13:20:08.638 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.DatatypeHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.datatypes.XMLSchemaDatatypeHandler
13:20:08.640 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.DatatypeHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.datatypes.RDFDatatypeHandler
13:20:08.641 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.DatatypeHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.datatypes.DBPediaDatatypeHandler
13:20:08.642 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.DatatypeHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.datatypes.VirtuosoGeometryDatatypeHandler
13:20:08.643 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.DatatypeHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.datatypes.GeoSPARQLDatatypeHandler
13:20:08.649 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.LanguageHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.languages.RFC3066LanguageHandler
13:20:08.651 [org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport-jmh-worker-1] DEBUG org.eclipse.rdf4j.rio.LanguageHandlerRegistry - Registered service class org.eclipse.rdf4j.rio.languages.BCP47LanguageHandler
<failure>

java.lang.IllegalStateException: Not isomorphic
	at org.eclipse.rdf4j.model.util.IsomorphicBenchmark.isomorphic(IsomorphicBenchmark.java:271)
	at org.eclipse.rdf4j.model.util.IsomorphicBenchmark.shaclValidationReport(IsomorphicBenchmark.java:135)
	at org.eclipse.rdf4j.model.util.generated.IsomorphicBenchmark_shaclValidationReport_jmhTest.shaclValidationReport_avgt_jmhStub(IsomorphicBenchmark_shaclValidationReport_jmhTest.java:264)
	at org.eclipse.rdf4j.model.util.generated.IsomorphicBenchmark_shaclValidationReport_jmhTest.shaclValidationReport_AverageTime(IsomorphicBenchmark_shaclValidationReport_jmhTest.java:208)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.openjdk.jmh.runner.BenchmarkHandler$BenchmarkTask.call(BenchmarkHandler.java:453)
	at org.openjdk.jmh.runner.BenchmarkHandler$BenchmarkTask.call(BenchmarkHandler.java:437)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)

```

hmottestad:
### `spinFullForwardchained()`

Screenshot 2021-02-06 at 13 29 46

This is one of the benchmarks where the new algorithm is faster. Notice how it is missing the long recursive tail of `distinguish`. Not sure what this actually signifies since I don't understand the algorithm.

jeenbroekstra:
> Big thanks @jeenbroekstra !! 
> 
> I'd missed this PR, hence the commit comments (which I'd only seen through a link in a comment). I'd taken a copy of this and tried in our project where we had non-completing comparisons, and things all run in seconds, now. 

Good to know, thanks. 
>
> One minor thing was I had to remove the jdk1.10 use of orElseThrow but that's because for a reason I couldn't fathom our project is targeting 1.8. 

That was actually a mistake on my end: orElseThrow without arguments is java 9 (iirc), but rdf4j still targets java 8 as well, so that should not have been in the code. It's been fixed.

hmottestad:
Would a `VersionedHashedBNode` be a concept that would capture the current use of hashmaps?

Essentially:
```
public class VersionedHashedBnode implements BNode {

	private final BNode delegate;
	List<HashCode> hashCodes = new ArrayList<>();


	public VersionedHashedBnode(BNode delegate) {
		this.delegate = delegate;
	}

	public HashCode getHashCode(int version){
		return hashCodes.get(version);
	}
	
	public int appendHashCode(HashCode hashCode){
		hashCodes.add(hashCode);
		return hashCodes.size()-1;
	}
	
	public void setHashCode(int version, HashCode hashCode){
		hashCodes.set(version, hashCode);
	}
	
	/*
	Delegate all BNode methods to the delegate BNode.
	 */

}
```

jeenbroekstra:
> Would a `VersionedHashedBNode` be a concept that would capture the current use of hashmaps?

I doubt it, because it doesn't address the fact that we have to juggle two collections of such nodes, nor that we need an efficient way to look up a node given a hashcode. I'm also not convinced that replacing a datastructure holding refereneces to N existing items with a datastucture holding references to N _newly created_ items will gain us much.

The problem is not necessarily that we're using Maps, but I think there is something to be gained by using a different (perhaps even custom) Map implementation (or a totally bespoke datastructure that indexes in both directions). One thing in particular stands out to me: we know the size of the Map in advance since we know the number of BNodes, so there should be no need for the Map to ever resize (which shows up in your hot methods list).

jeenbroekstra:
> @jeenbroekstra I'm now getting an error on the `IsomorphicBenchmark.shaclValidationReport` benchmark.

I can reproduce it here. Odd that it happens in the benchmark but not in the test case in IsomorphicTest (the only difference is the sorting). It may have something to do with using long instead of BigInteger to compute the mapping size. Taking a look.

jeenbroekstra:
I'm not entirely sure I understand why but apparently my assumption that a 32-bit hash is good enough for isomorphism is incorrect - the benchmark case consistently fails when using that, while reverting back to 128-bit hash (and BigInteger math) fixes the problem. The annoying thing is that it _only_ happens on the benchmark test, I somehow can't reproduce it in a unit test that uses the exact same structure. 

I'm reverting the change to a 32-bit hash to be certain we have a correct algorithm, and will instead have a look at the Map datastructure to see what we can optimize there.

jeenbroekstra:
I've also noticed that we do a bit more copying/cloning of hashmaps than is necessary. I'll try eliminating some of that.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
```diff
@@ -0,0 +1,447 @@
...
+			do {
+				Map<BNode, HashCode> temp = currentHash;
+				currentHash = new HashMap<>(previousHash);
+				previousHash = temp != null ? temp : initialHash;
+
+				for (BNode b : blankNodes) {
+					for (Statement st : m.getStatements(b, null, null)) {
+						HashCode c = hashTuple(hashForValue(st.getObject(), previousHash, staticValueMapping),
+								hashForValue(st.getPredicate(), previousHash, staticValueMapping), outgoing);
+						currentHash.put(b, hashBag(c, currentHash.get(b)));
```
hmottestad, line 375: If BNodes could hold their own hashes, then this line would be something like `b.addHash(hashBag(c, b.getHashVersion(version)))`.

jeenbroekstra:
I am working on an abstraction / code cleanup to make the handling of partitionions and node mappings a little more transparent - the idea is also to make carrying of state a little easier. I will need a littl more time than anticipated to get his done though.

core/model/src/main/java/org/eclipse/rdf4j/model/util/GraphComparisons.java
jeenbroekstra: I'm considering something along these lines as a test, but it will require quite a bit of rewiring and is unlikely to be done before your milestone build is due.

jeenbroekstra:
benchmark after introduction of Partitioning class:

```
Benchmark                                   Mode  Cnt    Score    Error  Units
IsomorphicBenchmark.blankNodes              avgt   10    4.915 ±  0.070  ms/op
IsomorphicBenchmark.bsbm                    avgt   10   43.839 ±  0.543  ms/op
IsomorphicBenchmark.bsbmArrayList           avgt   10  170.881 ±  8.771  ms/op
IsomorphicBenchmark.bsbmNotIsomorphic       avgt   10   26.754 ±  2.378  ms/op
IsomorphicBenchmark.bsbmTree                avgt   10   69.740 ±  0.214  ms/op
IsomorphicBenchmark.empty                   avgt   10   ≈ 10⁻⁴           ms/op
IsomorphicBenchmark.emptyNotIsomorphic      avgt   10   ≈ 10⁻⁵           ms/op
IsomorphicBenchmark.internallyIsomorphic    avgt   10  344.443 ±  5.124  ms/op
IsomorphicBenchmark.list                    avgt   10   11.433 ±  0.063  ms/op
IsomorphicBenchmark.listEquals              avgt   10    0.973 ±  0.004  ms/op
IsomorphicBenchmark.longChain               avgt   10    0.013 ±  0.001  ms/op
IsomorphicBenchmark.manyProperties          avgt   10  133.871 ±  0.376  ms/op
IsomorphicBenchmark.manyProperties2         avgt   10  133.015 ±  1.358  ms/op
IsomorphicBenchmark.sameModel               avgt   10   ≈ 10⁻⁶           ms/op
IsomorphicBenchmark.shacl                   avgt   10    5.720 ±  0.020  ms/op
IsomorphicBenchmark.shaclValidationReport   avgt   10  101.355 ±  0.327  ms/op
IsomorphicBenchmark.sparqlTestCase          avgt   10    0.245 ±  0.001  ms/op
IsomorphicBenchmark.spinFullForwardchained  avgt   10  137.998 ±  0.482  ms/op
IsomorphicBenchmark.uuid                    avgt   10    7.965 ±  0.048  ms/op
```

hmottestad:
Looks like a great improvement for almost everything.

jeenbroekstra:
I think I have one last short-circuit that may improve things a little. Running some benchmarks now to see if it makes a difference.

jeenbroekstra:
The last update is especially more efficient detecting when two models are _not_ isomorphic: instead of labeling both models with canonical mapping and detecting equality, we only check the non-bnode statements to see if they appear in both models (this is obviously after establishing that the canonical bnode mappings are compatible). New benchmark test added for the case, which shows quite an improvement. 

Before:

```
Benchmark                                          Mode  Cnt   Score   Error  Units
IsomorphicBenchmark.shaclValidationReport_changed  avgt   10  90.310 ± 0.540  ms/o
```

After:

```
Benchmark                                          Mode  Cnt   Score   Error  Units
IsomorphicBenchmark.shaclValidationReport_changed  avgt   10  55.683 ± 1.215  ms/op
```

jeenbroekstra:
Unless there are any objections, I am ready to merge this and mark it done. Further improvements can be logged as followup tasks.

hmottestad:
Since some of the benchmarks are so much slower I would appreciate if we kept the old algorithm around so our users can choose to use the legacy algorithm if it’s faster for them. Other than that maybe some comments in the code to reference which algorithm you are using and where to find out more?

hmottestad:
I can hold off on the M3 build so this can be included. If you want?

jeenbroekstra:
> Since some of the benchmarks are so much slower I would appreciate if we kept the old algorithm around so our users can choose to use the legacy algorithm if it’s faster for them.

I don't think that's a good idea, especially since we have no reasonable guidelines on which algorithm to use in which case. We are not going to start maintaining two algorithms and making our users choose, it's impossible to explain and maintain.  Once we swap over, that's it.

As for performance: yes, it's slower on some benchmarks, and I'd like to see us make further improvements. But in its current state it's already objectively better than the old algorithm: on at least one of those benchmarks it's _only_ slower if the input happens to be sorted correctly - if if's sorted differently the old algorithm doesn't even terminate. I also note that in most cases where it's slower, we are still talking a total execution time in the order of 10-50 ms, which is completely acceptable IMO.

> Other than that maybe some comments in the code to reference which algorithm you are using and where to find out more?

There's a big javadoc reference with a link to the paper on the class already, but I can go over it again and expand if necessary. There's also a few minor code cleanups warranted perhaps. Anything in particular you'd like clarified? 

> I can hold off on the M3 build so this can be included. If you want?

I probably won't have time the rest of this week to work on this further - nor do I expect that we will make further significant performance improvements in the short term. I'll do a javadoc/cleanup pass, but after that we'll merge this before the M3 build, so that it's included and we can get early feedback, and we can further improve/fix things in a followup before the actual release.

jeenbroekstra:
Btw current hotspot analysis shows that actual equality checking on the Guava HashCode objects is currently the major bottleneck:

Screenshot from 2021-02-08 20-22-04

I'd still like to check if we can make that simpler somehow.

jeenbroekstra:
Huh. I just halved the execution time of that big ontology comparison by doing this:

```java
	        Set<HashCode> values1 = Sets.newHashSet(mapping1.values());
		Set<HashCode> values2 = Sets.newHashSet(mapping2.values());

		if (!(values1.equals(values2))) {
			return true;
		}
```

instead of this:

```java
		Collection<HashCode> values1 = mapping1.values();
		Collection<HashCode> values2 = mapping2.values();

		if (!(values1.containsAll(values2) && values2.containsAll(values1))) {
			return true;
		}
```

hmottestad:
> As for performance: yes, it's slower on some benchmarks, and I'd like to see us make further improvements. But in its current state it's already objectively better than the old algorithm: on at least one of those benchmarks it's _only_ slower if the input happens to be sorted correctly - if if's sorted differently the old algorithm doesn't even terminate. I also note that in most cases where it's slower, we are still talking a total execution time in the order of 10-50 ms, which is completely acceptable IMO.

I've extended the synthetic blank node benchmark so show that the difference between the new and the old algorithm scales so that the 10-50ms difference becomes > 6 seconds. 

https://github.com/eclipse/rdf4j/pull/2805

If you can keep the old algorithm around, maybe as a @deprecated and @experimental or something like that, so that if users complain about the new algorithm being slow then we can point them to the old one.

somanythings:
> > As for performance: yes, it's slower on some benchmarks, and I'd like to see us make further improvements. But in its current state it's already objectively better than the old algorithm: on at least one of those benchmarks it's _only_ slower if the input happens to be sorted correctly - if if's sorted differently the old algorithm doesn't even terminate. I also note that in most cases where it's slower, we are still talking a total execution time in the order of 10-50 ms, which is completely acceptable IMO.
> 
> I've extended the synthetic blank node benchmark so show that the difference between the new and the old algorithm scales so that the 10-50ms difference becomes > 6 seconds. 
> 
> https://github.com/eclipse/rdf4j/pull/2805
> 
> If you can keep the old algorithm around, maybe as a @deprecated and @experimental or something like that, so that if users complain about the new algorithm being slow then we can point them to the old one.
> 

As much as I find the current / old algorithm insane, and not useful for so many of my cases, I agree with this on principal! Arguably, the new one should be under a new name. Per Hickey's "growth, not change". But I wouldn't want this to be hard to find :-)

somanythings:
@hmottestad somewhere, on a commit, I made a comment, and you responded, about isBnode being a significant proportion of profile runs I'd done on the old algo. I misremembered as I wrote that from my phone, it was actually the 'isBlank'. Sorry if I caused alarm!

Likely a non-issue, in the new world! 😄 

image

```java
	private static boolean isBlank(Value value) {
		if (value instanceof BNode) {
			return true;
		}

		if (value instanceof IRI) {
			boolean skolemizedBNode = value.stringValue().contains("/.well-known/genid/");
			return skolemizedBNode;
		}
		return false;
	}
```

jeenbroekstra:
> somewhere, on a commit, I made a comment, and you responded, about isBnode being a significant proportion of profile runs I'd done on the old algo. I misremembered as I wrote that from my phone, it was actually the 'isBlank'. Sorry if I caused alarm!

@somanythings thanks for bringing this up actually: it does point to a difference between the old algorithm and the new that sofar I have just ignored: in the old approach it treated skolemized node IRIs as equivalent to blank nodes, which meant it would consider a skolemized model isomorphic to the original blank node model. I've changed that approach for two reasons: 1) it performs very poorly (as your analysis shows) and 2) I disagree with the premise in principle: once blank nodes have been converted to IRIs, they're IRIs, not "almost-IRIs-but-also-still-a-bit-bnodes". A possible third reason is that I suspect actually no-one relies on this (undocumented) feature. 

But perhaps that is a reason to keep the old algorithm around in deprecated form, as much as possible corner case performance issues. So what we'll do is this: `Models.isomorphic` will implement the new algorithm. We will also add a `Models.legacyIsomorphic` (open to better names) which will be marked deprecated from the start, that will implement the old algorithm. We will document that it is not recommended for continued use as we willl phase it out from the code base in the next major release.

jeenbroekstra:
> I've extended the synthetic blank node benchmark so show that the difference between the new and the old algorithm scales so that the 10-50ms difference becomes > 6 seconds.

To be honest I don't find that particularly compelling. It is a synthetic worst-case scenario. The main reason it performs poorly is that the algorithm assumes it can compute unique identifiers for blank nodes by "anchoring" bnode identifiers to the IRIs and literals with which they are connected. In this case, there are no distinct IRIs or literals to connect it to (because _everything_ is either a bnode or foaf:knows, a single relation - so there is nothing to easily distinguish blank nodes) which leads to it having to compute through a massive amount of automorphic variations before finding a canonical mapping. 

A possible shortcut would be to check if all subjects and objects are blank nodes, and if so, do some special-case handling. But it's not worth the effort because this is a situation that would never occur in a real-world RDF graph, and the check itself would, in most scenarios, be a useless performance burden.

hmottestad:
> > I've extended the synthetic blank node benchmark so show that the difference between the new and the old algorithm scales so that the 10-50ms difference becomes > 6 seconds.
> 
> 
> 
> To be honest I don't find that particularly compelling. It is a synthetic worst-case scenario. The main reason it performs poorly is that the algorithm assumes it can compute unique identifiers for blank nodes by "anchoring" bnode identifiers to the IRIs and literals with which they are connected. In this case, there are no distinct IRIs or literals to connect it to (because _everything_ is either a bnode or foaf:knows, a single relation - so there is nothing to easily distinguish blank nodes) which leads to it having to compute through a massive amount of automorphic variations before finding a canonical mapping. 
> 
> 
> 
> A possible shortcut would be to check if all subjects and objects are blank nodes, and if so, do some special-case handling. But it's not worth the effort because this is a situation that would never occur in a real-world RDF graph, and the check itself would, in most scenarios, be a useless performance burden. 
> 
> 
> 
> Possibly a better idea would be to add a "leaning" step to the algorithm. A lot of those blank nodes can probably be eliminated as duplicates before doing isomorphic comparison. We could make that optional as people may want to choose between "structural" isomorphism (same graph structure, keeping duplicate bnodes) and "semantic" isomorphism (lean the graph to eleminate bnodes that add no additional information). 

Thank you for explaining the reason why the algorithm was slow on this particular benchmark. My point was just to show that the new algorithm is just as much slower on some benchmarks as it is faster on others.

jeenbroekstra:
> Thank you for explaining the reason why the algorithm was slow on this particular benchmark. My point was just to show that the new algorithm is just as much slower on some benchmarks as it is faster on others.

That's fair enough and I agree with you that it needs further investigation - but I don't think it's a blocker at this time.

> I presume you still want this PR included in 3.6.0-M3. Will you be able to get it merged today or tomorrow?

I don't really have the time to do much work on this right now, but I'll push a small extension to have the old algorithm available (as a deprecated method). Then we should be good to go I think, at least for the milestone build.

jeenbroekstra:
I've logged followup issues for further improvements in general (GH-2809) and specifically for the notion of introducing graph leaning functionality (GH-2808). @hmottestad I've merged your larger benchmark file with this PR, as soon as the thing is green this can be merged, and the issue marked as done.

Issue #2815, opened 2021-02-09
Gracefully handle DBPedia returning a LANGSTRING without a language


reckart:
Some SPARQL endpoints may not include language information on a LANGSTRING. The result parser currently fails in such situations. To gracefully handle this, the parser should downgrade the LANGSTRING to a simple STRING.

E.g. when querying DBPedia (http://de.dbpedia.org/sparql) for:

```
          SELECT DISTINCT ?lc ?subj
          WHERE { { { OPTIONAL { ?pLabel <http://www.w3.org/2000/01/rdf-schema#subPropertyOf>* <http://www.w3.org/2000/01/rdf-schema#label> . }
          { ?subj ?pLabel ?lc .
          ?lc <bif:contains> '''"lala"''' .
          FILTER ( ( ( REGEX( ?lc, "^.+lala$", "i" ) && LANGMATCHES( LANG( ?lc ), "de" ) ) || ( REGEX( ?lc, "^.+lala$", "i" ) && LANGMATCHES( LANG( ?lc ), "" ) ) ) ) } } } }
          LIMIT 6
```

[...]

PR #2816, opened 2021-02-09
GH-2815 Gracefully handle DBPedia returning a LANGSTRING without a language

reckart:
Signed-off-by: Richard Eckart de Castilho <richard.eckart@gmail.com>

GitHub issue resolved: #2815

- Added a unit test illustrating the problem
- Added a fix which silently falls back to using STRING as the data-type if no language is given despite the result claiming a langString type
- 
----


core/queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/SPARQLResultsSAXParser.java
```diff
@@ -132,12 +134,21 @@ public void startTag(String tagName, Map<String, String> atts, String text) thro
...
+					datatypeIri = valueFactory.createIRI(datatype);
 				} catch (IllegalArgumentException e) {
 					// Illegal datatype URI
 					throw new SAXException(e.getMessage(), e);
 				}
+
+				// For broken SPARQL endpoints which return LANGSTRING without a language, fall back
+				// to using STRING as the datatype
+				if (RDF.LANGSTRING.equals(datatypeIri) && xmlLang == null) {
+					datatypeIri = XSD.STRING;
```
jeenbroekstra, line 148: We should perhaps log a warning here.

jeenbroekstra (commented):
Thanks for the fix, this looks good to me. However, this only fixes it for the SPARQL XML result format. We'll need to add a similar fix for the SPARQL JSON result parser.  Could you add that as part of this PR so we don't have to open a separate ticket for it? 

I don't think we need to fix this in any other formats btw: binary is pretty much RDF4J-only, so unlikely to ever contain malformed literals, and csv/tsv do not even record datatypes I think.

core/queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/SPARQLResultsSAXParser.java
reckart: Added a log message. Since this may happen often and there's not much one can do about it, I only log at debug level. Not sure how to properly contextualize the message though - maybe you could extend the message so that the user has some reasonable idea as to in relation to which result triple this is happening. Since a triple may have an arbitrary large value (e.g. for long description) some cutting down of the information might be called for. Probably you have best practices for this from other places in the code - but I am not familiar with them. Would be great if you could fix this up accordingly as required.

reckart:
Ok, the JSON parser now also has the test and the fix.

core/queryresultio/sparqlxml/src/main/java/org/eclipse/rdf4j/query/resultio/sparqlxml/SPARQLResultsSAXParser.java
jeenbroekstra: Thanks. I'm actually fine with this as a log message (minus the small typo), though we could stick a fixed-length substring of the lexical value in there. I'll take a closer look.

jeenbroekstra (approved):
Thanks, looks good! I'll push a small fix for the debug log message to this PR, and then will merge.

jeenbroekstra:
Turns out I can't push to your fork (or at least `gh` CLI won't let me) - closing this and using a new branch / PR (#2825) instead.

Issue #2812, opened 2021-02-09
move dependency scope declaration to actual dependency instead of management


reckart:
When a downstream project imports the RDF4J management, it also gets the dependency management for commons-codec bound as a "runtime" dependency from `org.eclipse.rdf4j:rdf4j:pom:3.5.1`

```
			<dependency>
				<groupId>commons-codec</groupId>
				<artifactId>commons-codec</artifactId>
				<version>1.13</version>
				<scope>runtime</scope>
			</dependency>
```

This implicitly turns all the dependencies that the downstream project may have on commons-codec from compile scope (default) to the runtime scope and breaks compilation.

In my experience, it is generally a good idea not to specify scopes in the dependency management section at all and use it only in the dependencies section.

jeenbroekstra:
Why would a downstream project ever import our root pom directly?

jeenbroekstra:
It's easy enough to fix btw, I'll put up a PR in a sec. But I'd appreciate it if you could give me some background on when this is a problem.

reckart:
Well, yes - I was as surprised as you are. I didn't import the root POM - just the BOM. But still Maven picks up the managed dependency from the root POM. Here is an example:

```
<project xmlns="http://maven.apache.org/POM/4.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>test</groupId>
  <artifactId>rdf4j-dependeny-import-test</artifactId>
  <version>0.0.1-SNAPSHOT</version>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.eclipse.rdf4j</groupId>
        <artifactId>rdf4j-bom</artifactId>
        <version>3.5.1</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>commons-codec</groupId>
      <artifactId>commons-codec</artifactId>
    </dependency>
  </dependencies>
</project>
```

Result:

<img width="550" alt="2021-02-10_09-10-36" src="https://user-images.githubusercontent.com/1410238/107482526-18250f80-6b80-11eb-9434-7f64f7e0f3b6.png">

reckart:
[...]

PR #2818, opened 2021-02-10
GH-2812 set scope for dependencies in actual dependency, not management

jeenbroekstra:
GitHub issue resolved: #2812  

- removed scope declaration for commons-codec from root pom dependencyManagement section, moved to specific dependency usages in module poms
- removed old version number for commons-codec 1.11 from tools module

----

### DIFF
--- a/core/http/client/pom.xml
+++ b/core/http/client/pom.xml
@@ -59,6 +59,7 @@
 			<!-- include newer codec version for httpclient -->
 			<groupId>commons-codec</groupId>
 			<artifactId>commons-codec</artifactId>
+			<scope>runtime</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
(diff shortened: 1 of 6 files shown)


jeenbroekstra:
This change looks trivial enough but I'm a little panicky about it unintentionally breaking things. I know we've had some difficulties with incompatible versions of various commons libraries in the past. @barthanssens I believe you may have some experience with that?

jeenbroekstra:
We may want to do this more generally for the managed dependencies in the root pom: get rid of the scope tags, and move them to the appropriate places in the dependencies.

jeenbroekstra:
@hmottestad @barthanssens I have done a general cleanup of dependency management: a lot of our managed dependencies were declared  twice (both in the root pom and the tools pom) - probably a leftover from merging the multi-repo setup. I've also removed the scope declaration from all managed dependencies. As @reckart pointed out it is better practice to only manage versions, and let scope handling be done at the point where the _actual_ dependency is declared (to avoid weird surprises in downstream projects). 

I've ran the project locally and this all works as expected as far as I can tell. I do want to run a quick distribution build to make sure no dependencies that shouldn't be there have now sneaked into the SDK archive or the onejar.

jeenbroekstra:
If there are no objections I will merge this today.

Issue #2834, opened 2021-02-16
Shacl target extend needs to handle input of varying cardinality


hmottestad:
At the moment the `extend(...)` functionality of the EffectiveTarget doesn't support input of varying cardinality. It assumes that all input tuples have the same number of items as the first one.

PR #2835, opened 2021-02-16
GH-2834 shacl extend fix

hmottestad:
GitHub issue resolved: #2834 

 - fixes unclosed resources in TargetChainRetriever
 - fixes a bug in InnerJoin that could lead to discarderLeft or discarderRight missing the last ValidationTuple
 - fixes a bug in BindSelect where the input tuples are of different cardinalities

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/QualifiedMaxCountConstraintComponent.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/QualifiedMaxCountConstraintComponent.java
@@ -9,6 +9,7 @@ package org.eclipse.rdf4j.sail.shacl.ast.constraintcomponents;
 
 import static org.eclipse.rdf4j.model.util.Values.literal;
 
+import java.math.BigInteger;
 import java.util.List;
 import java.util.Set;
 
@@ -17,7 +18,6 @@ import org.eclipse.rdf4j.model.Model;
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.model.vocabulary.SHACL;
-import org.eclipse.rdf4j.model.vocabulary.XMLSchema;
 import org.eclipse.rdf4j.repository.RepositoryConnection;
 import org.eclipse.rdf4j.sail.shacl.ConnectionsGroup;
 import org.eclipse.rdf4j.sail.shacl.ShaclSail;
@@ -84,7 +84,7 @@ public class QualifiedMaxCountConstraintComponent extends AbstractConstraintComp
 		}
 
 		if (qualifiedMaxCount != null) {
-			model.add(subject, SHACL.QUALIFIED_MAX_COUNT, literal(qualifiedMaxCount + "", XMLSchema.INTEGER));
+			model.add(subject, SHACL.QUALIFIED_MAX_COUNT, literal(BigInteger.valueOf(qualifiedMaxCount)));
 		}
 
 		qualifiedValueShape.toModel(null, null, model, cycleDetection, rdfListDedupe);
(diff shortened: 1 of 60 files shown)


hmottestad:
Will be wanting to merge this tomorrow.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/QualifiedMaxCountConstraintComponent.java
```diff
@@ -83,7 +84,7 @@ public void toModel(Resource subject, IRI predicate, Model model, Set<Resource>
 		}
 
 		if (qualifiedMaxCount != null) {
-			model.add(subject, SHACL.QUALIFIED_MAX_COUNT, literal(qualifiedMaxCount));
+			model.add(subject, SHACL.QUALIFIED_MAX_COUNT, literal(qualifiedMaxCount + "", XMLSchema.INTEGER));
```
jeenbroekstra, line 87: I don't understand what's going on here. You have a `Long` value, which you then concatenate with an empty string, and you turn it into an integer? Why do this? 

Also: I'd suggest using `XSD` instead of `XMLSchema` (I think it's deprecated?)

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/Sort.java
```diff
@@ -39,9 +40,25 @@ public Sort(PlanNode parent) {
...
 			@Override
 			public void close() throws SailException {
-				iterator.close();
+				if (closed) {
+					throw new IllegalStateException("Already closed");
+				}
+				closed = true;
+
+				if (sortedTuples != null && iterator.hasNext()) {
+					throw new AssertionError("All tuples from parent iterator where not retrieved when sorting!");
```
jeenbroekstra, line 53: This isn't right, we shouldn't be explicitly throwing AssertionErrors.

jeenbroekstra (changes requested):

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/QualifiedMaxCountConstraintComponent.java
hmottestad: It's to make the output look identical to the input. This code is currently only used while running the tests, so we can change it up later if we want.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/Sort.java
hmottestad: My bad, I was doing this to make it easier to debug, will turn it back into an assert statement.

hmottestad: Fixed it.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/QualifiedMaxCountConstraintComponent.java
jeenbroekstra: Can you maybe add a code comment to indicate that? These kinds of things tend to stick around longer than you think, and it's not really clear just from looking at the code what it's for.

hmottestad: Will do :)

jeenbroekstra: Other than that happy with this being merged btw

hmottestad: Im taking a break now to go to a cafe and relax a bit. So I'll fix things up and merge this later today.

hmottestad: Found out I could clean this up because of: https://github.com/eclipse/rdf4j/issues/2843

hmottestad: Actually read the spec again. So I'm converting long to BigInteger on the way out. In the future we could consider using BigInteger directly.

Issue #2843, opened 2021-02-18
Isomorphism ignores datatypes and language tags


hmottestad:
I'm wondering if the following two models, one with an untyped number and the other with an xsd:long are supposed to be isomorphic.

```
@prefix ex: <http://example.com/ns#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix rsx: <http://rdf4j.org/shacl-extensions#> .

ex:PersonShape sh:not [
      sh:not [
          sh:maxCount 3;
          sh:path ex:ssn
        ]
    ];
  sh:targetClass ex:Person .
```

```
@prefix ex: <http://example.com/ns#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix rsx: <http://rdf4j.org/shacl-extensions#> .

ex:PersonShape sh:not [
      sh:not [
          sh:maxCount "3"^^xsd:long;
          sh:path ex:ssn
        ]
    ];
  sh:targetClass ex:Person .

```

hmottestad:
@jeenbroekstra I'm wondering if this is a bug or if it is supposed to be like this?

jeenbroekstra:
This is a bug in the new isomorphism algorithm, caused by the fact that for non-bnode values (IRIs and Literals) we rely on `Object.hashCode` to give us a unique hash identifier. That assumption is generally incorrect, and in particular in the case of Literals: `Literal.hashCode` uses only the lexical label, not the language tag or datatype (this is for performance reasons, see #665).

Fix is to calculate unique hashes for IRIs and Literals ourselves in the algorithm, instead of relying on Object.hashCode.

jeenbroekstra:
[...]

PR #2844, opened 2021-02-18
GH-2843 hashcode for literals needs to include lang tag and datatype

jeenbroekstra:
GitHub issue resolved: #2843  

- hash code calculation for literals now includes language tag and datatype
- added regression test

----

### DIFF
--- a/compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicTest.java
+++ b/compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicTest.java
@@ -9,23 +9,19 @@
 package org.eclipse.rdf4j.model.util;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.eclipse.rdf4j.model.util.Values.iri;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.List;
 
-import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Model;
 import org.eclipse.rdf4j.model.Statement;
 import org.eclipse.rdf4j.model.impl.TreeModel;
-import org.eclipse.rdf4j.model.vocabulary.RDF;
-import org.eclipse.rdf4j.model.vocabulary.SHACL;
 import org.eclipse.rdf4j.rio.RDFFormat;
 import org.eclipse.rdf4j.rio.Rio;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Test;
 
 public class IsomorphicTest {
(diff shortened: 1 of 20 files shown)


jeenbroekstra:
@hmottestad I'm getting test failures on the ShaclTest suite and I can't make heads nor tails of it - though it looks to be related to the SHACL engine making some assumptions about language tag equality that aren't quite  correct. Could you have a look?

hmottestad:
Screenshot 2021-02-19 at 06 26 32

It's failing in the last step of the test where it compares the expected and actual validation report. I'll fix it.

hmottestad:
Btw. Does this mean that the new isomorphism algorithm is reliant on the hash so much that it wouldn't detect a collision?

jeenbroekstra:
> Btw. Does this mean that the new isomorphism algorithm is reliant on the hash so much that it wouldn't detect a collision?

I'm not sure what you mean with "detect a collision", but it relies on the hash being globally unique for different values. This is one reasone why we use a 128-bit hash, because collision is for all practical purposes impossible at that size. 

The reason we got problems with literals was that we forgot to include the datatype and language tag in the hash calculation, so two literals with the same lexical value but different datatypes would get the same hashcode.

hmottestad:
> > Btw. Does this mean that the new isomorphism algorithm is reliant on the hash so much that it wouldn't detect a collision?
> 
> I'm not sure what you mean with "detect a collision", but it relies on the hash being globally unique for different values. This is one reason why we use a 128-bit hash, because collision is for all practical purposes impossible at that size.

That's what I meant. That if there were a collision where two different values hashed to the same hash we would not be able to detect that this happened. As you say though, astronomically unlikely. 

Btw. This is what Oracle does for their triple store. Every IRI, blank node and literal is hashed to a 128 bit (or possibly higher) value that is stored on disk. That way you can have constant size blocks on disk combined with a one way dictionary. We instead have a two way dictionary, since we use an incrementing ID instead. So when we have a value and what to know if it's in the database we have to look it up in the reverse dictionary to get the ID, while Oracle just has to hash it. This also leads to an optimization we could do in our codebase. Virtual IRI/BNode/Literal where the actual value isn't retrieved from disk until you need to do more complex operations. Instead for simple things like joins it would just use the ID, and since a lot of data is discarded during typical joins then for that data we wouldn't need to retrieve the actual values from the dictionary.

hmottestad (approved):
lgtm 👍

jeenbroekstra:
> Btw. This is what Oracle does for their triple store. Every IRI, blank node and literal is hashed to a 128 bit (or possibly wider) hash that is stored on disk. That way you can have constant size blocks on disk (for the index) combined with a one way dictionary. We instead have a two way dictionary, since we use an incrementing ID instead. So when we have a value and what to know if it's in the database we have to look it up in the reverse dictionary to get the ID, while Oracle just has to hash it. This also leads to an optimization we could do in our codebase. Virtual IRI/BNode/Literal where the actual value isn't retrieved from disk until you need to do more complex operations. Instead for simple things like joins it would just use the ID, and since a lot of data is discarded during typical joins then for that data we wouldn't need to retrieve the actual values from the dictionary.

It's an interesting idea - we instead went with marker blocks in the native store, probably because (and don't hold me to this it's a long time ago) at the time disk space was at a premium so we were looking for compactness and making optimal use of java nio page sizing. 

Unfortunately none of us really have the time and resources to implement a new triplestore from scratch (or even to fundamentally refactor the native store). If we did, this would be an interesting approach I think.

PR #2853, opened 2021-02-21
Move to 3.7.0 as the next release for the develop branch

hmottestad:
Move to 4.0.0 for the next release for the develop branch.

Alternatively we can stick to 3.x and move to 3.7.0 instead.


jeenbroekstra:
I suggest keeping it at 3.7.0 for now. Once we start actually putting up PRs that require a major release, we can always bump the version number.

Issue #2829, opened 2021-02-15
Changeset should use less memory hungry objects for keeping track of observations


hmottestad:
This is particularly evident in `org.eclipse.rdf4j.sail.shacl.MultithreadedTest#testLotsOfValidationFailuresSerializable` where the creation of `Var` in `Changeset#observe` is the largests memory hog.

Mem-based profiling:
<img width="1346" alt="Screenshot 2021-02-15 at 11 36 57" src="https://user-images.githubusercontent.com/797185/107936138-915aa300-6f82-11eb-9192-6fb0e31f66b6.png">

PR #2865, opened 2021-02-23
GH-2829 Optimized observations in Changeset

hmottestad:
GitHub issue resolved: #2829  

 - new POJO class for keeping observations
 - conversion to old StatementPattern for legacy code

----

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
@@ -15,6 +15,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


hmottestad:
The changeset could actually use a more optimized datastructure for storing observations.

Since we allow observations with nulls, we know that "null, rdf:type, null" overlaps with "ex:pete, rdf:type, foaf:Person, ex:graph1". Don't really know what this datastructure would would be.

jeenbroekstra (commented):
It all looks technically ok, but what exactly is optimized about this? Is it the fact that you've cut out the `Var` objects and keep direct references to the Values?

hmottestad:
We use less memory.

hmottestad:
Less memory usage also means less GC. So technically it's also faster.

hmottestad:
Was also debating making the observations set synchronized. With the ShaclSail we do do parallel reads from a single connection.

jeenbroekstra (approved):

Issue #2799, opened 2021-02-04
OOM in MemoryOverflowModel


mannyrivera2010:
Given following code.  using http://purl.obolibrary.org/obo/ncit.owl
 ```
    private static ValueFactory vf = SimpleValueFactory.getInstance();

    public static void main(String[] args) throws IOException {
        private static String indexes = "spoc,posc,cosp"; 

         private static String data_directory = "data2";

         File dataDir = new File(data_directory);
         Repository repo1 = new SailRepository(new NativeStore(dataDir, indexes));
         repo1.init();


           try (RepositoryConnection conn = repo1.getConnection();
                     FileInputStream fis = new FileInputStream("ncit.owl");
                ) {
                    conn.begin();
                    conn.add(fis, "", RDFFormat.RDFXML, vf.createIRI(g+i));
                    conn.commit();
                }
                
```

You get following issue: 
```
Exception in thread "main" java.lang.OutOfMemoryError: GC overhead limit exceeded
	at org.eclipse.rdf4j.model.impl.LinkedHashModel$ModelNode.<init>(LinkedHashModel.java:344)
	at org.eclipse.rdf4j.model.impl.LinkedHashModel.asNode(LinkedHashModel.java:554)
	at org.eclipse.rdf4j.model.impl.LinkedHashModel.add(LinkedHashModel.java:150)
	at org.eclipse.rdf4j.model.impl.LinkedHashModel.add(LinkedHashModel.java:48)
	at org.eclipse.rdf4j.sail.nativerdf.MemoryOverflowModel.add(MemoryOverflowModel.java:129)
	at org.eclipse.rdf4j.sail.nativerdf.MemoryOverflowModel.add(MemoryOverflowModel.java:42)
	at org.eclipse.rdf4j.sail.base.Changeset.approve(Changeset.java:276)
	at org.eclipse.rdf4j.sail.base.SailSourceBranch.flush(SailSourceBranch.java:493)
	at org.eclipse.rdf4j.sail.base.SailSourceBranch.compressChanges(SailSourceBranch.java:325)
	at org.eclipse.rdf4j.sail.base.SailSourceBranch.merge(SailSourceBranch.java:302)
	at org.eclipse.rdf4j.sail.base.SailSourceBranch$1.flush(SailSourceBranch.java:175)
```

previous version of RDF4J like 2.5.5 did not have this outOfMemory Issue.

hmottestad:
[...]

PR #2872, opened 2021-02-24
GH-2799 avoid OOM through HashMap resizing by setting initial size

jeenbroekstra:
GitHub issue resolved: #2799 

- initialize the LinkedHashModel with a fixed size equal to the threshold we use for overflow to disk - this should avoid unexpected resizing before overflow is triggered
- note that LinkedHashModel itself internally already uses `size * 2` for the statement set to compensate for the load factor, so using the exact block size should be fine for our purposes

----

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowModel.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowModel.java
@@ -62,7 +62,7 @@ abstract class MemoryOverflowModel extends AbstractModel {
 	private long maxBlockSize = 0;
 
 	public MemoryOverflowModel() {
-		memory = new LinkedHashModel();
+		memory = new LinkedHashModel(LARGE_BLOCK);
 	}
 
 	public MemoryOverflowModel(Model model) {
@@ -76,7 +76,7 @@ abstract class MemoryOverflowModel extends AbstractModel {
 	}
 
 	public MemoryOverflowModel(Set<Namespace> namespaces) {
-		memory = new LinkedHashModel(namespaces);
+		memory = new LinkedHashModel(namespaces, LARGE_BLOCK);
 	}
 
 	@Override
@@ -282,7 +282,7 @@ abstract class MemoryOverflowModel extends AbstractModel {
 				}
 			};
 			disk.addAll(memory);
-			memory = new LinkedHashModel(memory.getNamespaces());
+			memory = new LinkedHashModel(memory.getNamespaces(), LARGE_BLOCK);
 			logger.debug("overflow synced to disk");
 		} catch (IOException | SailException e) {
 			String path = dataDir != null ? dataDir.getAbsolutePath() : "(unknown)";
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowModel.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowModel.java
@@ -244,8 +244,8 @@ abstract class MemoryOverflowModel extends AbstractModel {
 						maxBlockSize = blockSize;
 					}
 					// Sync if either the estimated size of the next block is larger than remaining memory, or
-					// if less than 10% of the heap is still free (this last condition to avoid GC overhead limit)
-					if (freeToAllocateMemory < Math.min(0.1 * maxMemory, maxBlockSize)) {
+					// if less than 15% of the heap is still free (this last condition to avoid GC overhead limit)
+					if (freeToAllocateMemory < Math.min(0.15 * maxMemory, maxBlockSize)) {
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
There's a bit of a false assumption in this change I just realized, due to poor variable/constant naming: although the `LARGE_BLOCK` constant plays a roll in the overflow trigger, it is not as simple as "overflow when we have more statements than this fixed constant". 

Alternative is that we move the "10% of the heap is still free" requirement up a bit, to, say, 15%?

hmottestad:
I think we should start with a test, or a benchmark if that's easier. Essentially limit the memory available and load in a big file. I'm also curious if this scales as it should, eg. does loading in a bigger file while giving more memory trigger the overflow as consistently as loading a smaller file with less memory available?

jeenbroekstra:
Not at my desk right now, but I believe there's an existing compliance test for the overflow that we could extend if necessary.

jeenbroekstra:
I'm not aware of a simple way to control the available heap space for a single junit test - so we may want to "abuse" a jmh benchmark for this purpose. I'm currently out of spare time, so won't continue on this immediately. @hmottestad if you have an idea and time/will to set up such a test feel free to add it on to this branch.

hmottestad:
After I commented here I set out to google if maybe junit 5 could fork the JVM for a test to configure the amount of memory. Could not find anything. I guess abusing JMH might be the only way unfortunately.

jeenbroekstra:
To be fair even with benchmarking in place there's only so much we can do to make  this work better: at the end of the day we are relying on an _estimate_ of available heap space and predicted usage to trigger disk overflow. In real life situations that estimate could be way off because of circumstances beyond our control (e.g. some different Java object suddenly consuming a lot of memory or something as basic as the data currently being uploaded suddenly containing a few massive literal values). We make a best effort, but we can't _guarantee_ the process won't run out of memory.

jeenbroekstra:
Btw a reason disk syncing on memory overflow is giving us such a performance hit was given by Arjohn Kampman on the mailinglist:

> The reason that syncing the data to disk when the overflow is triggered takes so long is also related to a call to dataset(). When overflowToDisk() calls disk.addAll(memory), this triggers a call to SailSourceModel.add(s,p,o,c...) for each statement. This method then calls both contain(s,p,o,c...) and sink().approve(s,p,o.c) for for each statement. The latter call starts a new transaction and updates the txn-status file, but the contains() call then commits the transaction for the previous statement via a call to dataset(), again updating the txn-status file. So for every cached statement, rdf4j does two I/O calls. On a spinning disk with a an average write time of 10 ms, this limits the overflow process to at most 50 statements per second.

That is not _directly_ related to this fix, but if we can address some of that, we can have more confidence that setting a lower overflow threshold won't cause a massive performance penalty.

hmottestad:
Do you mind @jeenbroekstra if I force push this branch? I would like to add a commit with a benchmark as the first commit in this branch so that we can easily test before and after.

jeenbroekstra:
Go for it

hmottestad:
ok, I've just force pushed, but will need to force push at least one more time before I'm done.

hmottestad:
I've added two benchmarks, one synthetic and one with a real world file.

The synthetic one I've got failing with very low memory. The real world file I've got failing with higher memory. Both are failing at different places. And to make things even more complicated the real world file is failing only within a certain memory range (500-700 MB), outside of that range it either overflows correctly or doesn't need to overflow.

hmottestad:
The synthetic benchmark only fails on java 8 for me.

jeenbroekstra:
Related: #2998

hmottestad:
Not going to dig any further now. I've spent like 3 hours on this because IntelliJ was acting up and running the benchmarks takes a long time.

jeenbroekstra:
I'm becoming more and more convinced that we should invest in replacing the current MemoryOverflowModel with a completely different implementation, something that just relies on standard Java Object Serialization or some simple abstraction like MapDB for disk syncing. I will try and spike something, see what we get.

hmottestad:
Maybe for the time being just try to adjust the point where the model overflows to disk. Perhaps you can add some logging to show what it currently thinks is going on memory wise so you can see what numbers it's comparing. Also maybe try without Xms in the benchmark. 

The real world benchmark is best to start with I think.

hmottestad:
I did some logging and found out that sometimes the algorithm would decide that the max block size was something like 10 MB. This is when it would overflow. So I've just added a hard limit of 32MB of required available memory. This works fine for both my tests (benchmarks) and I feel it's a reasonable tradeoff. How many users give their NativeStore 64MB or 128MB and then load large files?

jeenbroekstra:
> I did some logging and found out that sometimes the algorithm would decide that the max block size was something like 10 MB. This is when it would overflow. So I've just added a hard limit of 32MB of required available memory. This works fine for both my tests (benchmarks) and I feel it's a reasonable tradeoff. How many users give their NativeStore 64MB or 128MB and then load large files?

This seems reasonable to me as well, nice workaround.

jeenbroekstra (commented):
LGTM, happy to have this merged. I'll leave it to you.

Issue #2828, opened 2021-02-13
RDFLoader doesn't follow redirections from HTTP to HTTPS


manuelfiorelli:
The following piece of code should resolve the IRI of a DBpedia resource into its RDF description.

```java
RDFLoader rdfLoader = new RDFLoader(new ParserConfig(), SimpleValueFactory.getInstance());
RDFHandler rdfHandler = new StatementCollector();
rdfLoader.load(new URL("http://dbpedia.org/resource/Rome"), null, null, rdfHandler);
```

Unfortunately, the code throws this exception:

```
Exception in thread "main" org.eclipse.rdf4j.rio.UnsupportedRDFormatException: Could not find RDF format for URL: /resource/Rome
	at org.eclipse.rdf4j.repository.util.RDFLoader.lambda$null$1(RDFLoader.java:137)
	at java.util.Optional.orElseThrow(Optional.java:290)
	at org.eclipse.rdf4j.repository.util.RDFLoader.lambda$load$2(RDFLoader.java:136)
	at java.util.Optional.orElseGet(Optional.java:267)
	at org.eclipse.rdf4j.repository.util.RDFLoader.load(RDFLoader.java:135)
	at TestRDFLoader.main(TestRDFLoader.java:17)
```

The issue can be replicated using RDF4J 3.6.0-M3 and AdoptOpenJDK 1.8.0_252

The issue seems to be caused by the fact that when asked to resolve a resource IRI (e.g. 'http://dbpedia.org/resource/Rome'), DBpedia first redirects the client to an address using `HTTPS` (e.g. 'https://dbpedia.org/resource/Rome'), which in turn returns the appropriate representation (e.g. HTML, Turtle, RDF/XML, etc.) by content negotiation.

Under the hoods, `RDFLoader` uses `java.net.URL`, which seems not to follow such redirects, seemingly because of security concerns.

My workaround was to copy the implementation of `RDFLoader`, while switching to `HttpClient`: https://bitbucket.org/art-uniroma2/semantic-turkey/src/7708c555f5b86e2628a1084cdebb7233ac6e4cb2/core-services/src/main/java/it/uniroma2/art/semanticturkey/services/core/ResourceView.java#lines-1108

manuelfiorelli:
If you are interested I can upstream my change.

jeenbroekstra:
[...]

PR #2884, opened 2021-03-03
GH-2828 Make RDFLoader follow redirections from HTTP to HTTPS

manuelfiorelli:
The proposed solution follows redirects manually.

Signed-off-by: Manuel Fiorelli <manuel.fiorelli@gmail.com>

GitHub issue resolved: #2828 

The class `java.net.URL` doesn't follow redirects from HTTP to HTTPS (i.e. insecure request upgrades), seemingly for security reasons. In the discussion associated with the issue has been decided that it should be secure enough to change this behavior and follow such redirects (as Apache's HttpClient does by default).

To accomplish its goal, this PR disables automatic redirects by `java.net.URL` and manually follows redirects. The code honors the system property `http.maxRedirects`.


core/repository/api/src/main/java/org/eclipse/rdf4j/repository/util/RDFLoader.java
```diff
@@ -87,7 +93,10 @@ public void load(File file, String baseURI, RDFFormat dataFormat, RDFHandler rdf
 	}
 
 	/**
-	 * Parses the RDF data that can be found at the specified URL to the RDFHandler.
+	 * Parses the RDF data that can be found at the specified URL to the RDFHandler. This method uses the class
+	 * {@link URL} to resolve the provided <tt>url</tt>. While this class doesn't follow redirects upgrading the
+	 * protocol from HTTP to HTTPS, this method implements its own strategy to follow redirects that also supports this
+	 * kind of insecure request upgrades.
```
jeenbroekstra, line 99: I'd reformulate slightly, to something like this:

> This method honors HttpUrlConnection.getFollowRedirects to determine if redirects are followed and if set to true will also follow redirects from http to https. The maximum number of redirects can be controlled using system property `http.maxRedirects`.

jeenbroekstra (commented):
Looks good to me, just one suggestion on the javadoc description. Also: would it be possible to add a unit test or two in, just for some common cases?

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/util/RDFLoader.java
manuelfiorelli: I can edit the Javadoc according to your suggestions. Concerning the unit tests:
* where should I add the unit tests (i.e. module, and fully qualified class name)?
* I had in mind to use something like [Mock Server](https://www.mock-server.com/). Can I use it, or do you already use something else? In the latter case, can you point me to some example test

jeenbroekstra: Great, thanks! 

Unit tests go in the same module, in src/test/java, the only exception being tests where other parts of the framework are needed that would introduce a circular dependency: in those cases, the test is more like an integration tests, and goes into the compliance modules. In this case, I think you should be fine with just a unit test in the repository-api module itself. Name it `org.eclipse.rdf4j.repository.util.RDFLoaderTest`. 

As for mocking, I actually had in mind to just use Mockito to mock out the connection behavior at the client side, but if you prefer to mock the server-side instead, we use [WireMock](https://www.baeldung.com/introduction-to-wiremock) in several places for this purpose. Have a look at [SPARQLProtocolSessionTest](https://github.com/eclipse/rdf4j/blob/main/core/http/client/src/test/java/org/eclipse/rdf4j/http/client/SPARQLProtocolSessionTest.java) or [RemoteRepositoryManagerTest](https://github.com/eclipse/rdf4j/blob/main/core/repository/manager/src/test/java/org/eclipse/rdf4j/repository/manager/RemoteRepositoryManagerTest.java) for some examples.

manuelfiorelli:
Hi @jeenbroekstra I applied the change you have proposed to the JavaDoc and I wrote a few tests cases using WireMock. Unfortunately, I have still to figure out how to stub the upgrade of insecure requests, which is actually the focus of this PR.

jeenbroekstra:
> Hi @jeenbroekstra I applied the change you have proposed to the JavaDoc and I wrote a few tests cases using WireMock. Unfortunately, I have still to figure out how to stub the upgrade of insecure requests, which is actually the focus of this PR.

Thanks Manuel. I'll try and take a closer look this weekend, see if I can help figuring it out (I'm no expert on WireMock either to be honest).

jeenbroekstra:
@manuelfiorelli I've pushed a commit to your branch that addresses working with a https redirect in the wiremock test (it's a bit of a dirty hack but I think we're fine with this in a test environment). 

I just noticed, however, that one of your commits is not signed off. Could you fix that? Please make sure you pull the latest changes first - I pushed my fix directly to your branch.

core/repository/api/pom.xml
```diff
@@ -34,6 +34,16 @@
...
 		</dependency>
+		<dependency>
+			<groupId>com.github.tomakehurst</groupId>
+			<artifactId>wiremock-jre8</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>ch.qos.logback</groupId>
+			<artifactId>logback-classic</artifactId>
+			<scope>test</scope>
```
hmottestad, line 45: I didn't think we needed to add a logback dependency just to support a logback test config.

hmottestad:
I'm not sure if there are any tests that test multiple redirects.

core/repository/api/pom.xml
jeenbroekstra: How else would you expect the unit test runner to find a logger implementation?

jeenbroekstra:
> I'm not sure if there are any tests that test multiple redirects.

Good point. The current test only checks a single redirect (and only one redirect type: 301 moved permanently). I'm fine with only covering the one redirect type for now, but some additional tests with multiple redirects shouldn't be too hard to add.

@manuelfiorelli I won't do any further pushes for now until you've had the opportunity the fix the commit sign off (let me know if you need any tips for that - you'll probably need to do an interactive rebase which can be tricky). Let's get that out of the way first. Then, if you feel up for it, could you add an extra few test cases with multiple redirects as well? I think we need to cover two things: 1) multiple redirects are followed and 2) multiple redirects beyond the specified limit are aborted (should be easy enough to check, just configure the limit to 1 and then do a 2-jump redirect).

manuelfiorelli:
Hi @jeenbroekstra I think that I can implement multiple redirects and the limit on redirects as well. Concerning the problem with the sign off, any tips is welcome :-D

jeenbroekstra:
Ok, regarding the sign-off, the particular commit that is the problem is 4814cf1. Probably the easiest way to fix this is to do an interactive rebase, and squash this commit into your previous one, taking care to sign off the newly produced commit.

1. make sure your local main branch is up to date (do a `git pull`), then switch to your PR branch (again, making sure that that is up to date as well). 
2. Run `git rebase -i main --signoff` (the `--signoff` bit here is important, as that makes sure that the new squashed commit will be signed off!). It will show you a list of the commits in an editor. 
3. In this editor, change  the line for the second commit (4814cf1) from 'pick' to 'squash'. Leave everything else as-is, save and quit the editor. 
4. It will probably open a second editor to allow you to edit the commit message for the squashed commit - you can do so if you wish, or leave as-is in this case. Save and quit again. 
5. Finally, force-push your local branch. 

For more info on interactive rebase and squashing, have a look at https://rdf4j.org/documentation/developer/squashing/.

manuelfiorelli:
Hi @jeenbroekstra thank you very much for your support. In the end, I avoided the `--sign-off` because it added the sign off also to your commits. Therefore, I decided to edit the commit message by hand yo only sign-off the squashed commits. Not sure if you have to rebase your own commits to include your sign-off

jeenbroekstra:
> Hi @jeenbroekstra thank you very much for your support. In the end, I avoided the `--sign-off` because it added the sign off also to your commits.  Therefore, I decided to edit the commit message by hand yo only sign-off the squashed commits.

Ah, I didn't realize it would do that - I though it would only apply to the new squashed commit. Good to know!

> Not sure if you have to rebase your own commits to include your sign-off

That's the good thing about being a committer rather than a contributor: I don't have to sign off :) 

Thanks for sorting this out, looks like we're back on track.

core/repository/api/pom.xml
hmottestad: I found out why this isn't needed. The pom file in the core module already contains one.

jeenbroekstra: Huh, I missed that. Fair enough. It doesn't hurt to have it here, but you're correct that it's not necessary.

hmottestad:
I would really like to release 3.6.1 on Thursday. So if it's at all possible to have this PR merged by then, that would be great!

manuelfiorelli:
Hi @hmottestad I didn't read you message. Unfortunately, I wasn't able to work on this PR during the last days. I'll try to add the missing tests by this week

manuelfiorelli:
I've added the requested unit tests in the new commit, together with an additional test checking the ability to load a resource using `Class.getResource(String)`.
It seems to me that `http.maxRedirects` is not inclusive, meaning that if it is set to `N` than `N` redirects are sufficient to raise the exception.
I looked at sources related to `java.net.URL` (e.g. http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/484e16c0a040/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java). Given the nature of the problem, it is perhaps inevitable that the two pieces of code are quite similar. Do you think that my code is sufficiently different to be considered original? Actually, my code might be an oversimplification given the large amount of checks implemented by the JDK code base.

jeenbroekstra (commented):
This looks fine to me, just one minor cleanup detail. Code looks original enough to me as well, and as for additional security checks: unless I misunderstood the only thing we're bypassing here is the redirect handling itself - everything else related to request processing (including checks) is still done by HTTPUrlConnection. So I see no real risk.

manuelfiorelli:
I moved the test resources into `src/test/resources`.
Concerning the "additional security checks", I meant that the code from the JDK does more things than mine. For example, this method http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/484e16c0a040/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#l843 does things including
* check that the status code is a specific value (rather than a generic `3xx`)
* check the host is the same
* check on the protocol (which is what we want to bypass :-D )

Actually, I'm unsure whether this method is related to a separate (more constrained manner) to open an URL connection.

jeenbroekstra (approved):
I think I'm fine with this as-is. @manuelfiorelli thanks again for contributing this.

Issue #2887, opened 2021-03-04
SHACL - simplify shape -> model code


hmottestad:
Simplify the `toModel(...)` code in Shape so that we don't have to supply the rdf list dedupe set.

PR #2888, opened 2021-03-05
GH-2887 simplify toModel(...) code 

hmottestad:
GitHub issue resolved: #2887 

 - We don't need the dedupe set if we have stable bnode id minting in the rdf list builder

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Cache.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Cache.java
@@ -1,3 +1,10 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
 package org.eclipse.rdf4j.sail.shacl.ast;
 
 import java.util.HashMap;
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Exportable.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Exportable.java
@@ -1,3 +1,10 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
 package org.eclipse.rdf4j.sail.shacl.ast;
 
 import java.util.Set;
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Exportable.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Exportable.java
@@ -8,6 +8,5 @@ import org.eclipse.rdf4j.model.Resource;
 
 public interface Exportable {
 
-	void toModel(Resource subject, IRI predicate, Model model, Set<Resource> cycleDetection,
-			Set<Resource> rdfListDedupe);
+	void toModel(Resource subject, IRI predicate, Model model, Set<Resource> cycleDetection);
 }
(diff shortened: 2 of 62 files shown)


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/HelperTool.java
```diff
@@ -21,13 +21,15 @@ public static void listToRdf(Collection<? extends Value> values, Resource head,
...
+		int counter = 0;
+
 		Iterator<? extends Value> iter = values.iterator();
 		while (iter.hasNext()) {
 			Value value = iter.next();
 			model.add(head, RDF.FIRST, value);
 
 			if (iter.hasNext()) {
-				Resource next = bnode();
+				Resource next = bnode(head.stringValue() + "_rdf_list_item_" + (++counter));
```
hmottestad, line 32: Here we mint stable identifiers for the blank nodes based on the identifier of the first blank node in the list.

jeenbroekstra: Can you use a better class name than "HelperTool"? Also: this method duplicates existing functionality for creating RDF collections from a Java lists. Any reason you're rolling your own instead of just reusing/improving/extending the existing utilities?

jeenbroekstra (changes requested):

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/HelperTool.java
hmottestad: I think I was initially struggling with pretty printing. Then I also found out I wanted type coercion. And now I've found out I want stable identifiers for each blank node in the list.

hmottestad: Does our one use stable blank node identifiers for each item in the list?

jeenbroekstra: I'm not sure. But this all sounds potentially generally useful, so why not improve the existing utilities instead of adding a duplicate?

jeenbroekstra: Also depends on what exactly you mean by stable, of course.

hmottestad: The reason for this is that the toModel method may be called multiple times with the same model. So it's important that all identifiers are stable so we don't create a new list every time. 

I think it's a bad idea to add this to the general rdf list code.

jeenbroekstra: > The reason for this is that the toModel method may be called multiple times with the same model. So it's important that all identifiers are stable so we don't create a new list every time.

So... correct me if I'm wrong, you are using stable node identifiers so that if the same list of items is added to the output model twice, it does not result in a second list if there already is one (because they will merge based on bnode identity)? I don't understand why you're solving that in this fashion instead of just quickly checking if there is an existing list or not before adding.

> I think it's a bad idea to add this to the general rdf list code.

This specific feature probably, but apart from being uncomfortable with the use to which you put this here, I'm left wondering if there is a better way to do it (e.g. some way we can let a user inject a lambda into the generic utility which can handle these kinds of bespoke conversions). I also think that the type coercion you mentioned might be relevant more generally.

Refactoring all of this is probably out of scope for this fix, but at the very least, I think we should make sure that:

1. `HelperTool` is renamed (to something meaningful like `ShaclRdfListUtil` perhaps);
2. It should be package-private or if that is not technically possible, marked as `@InternalUseOnly`;
3. It needs to clearly have the assumptions you're making in the javadoc (so that it is clear why we need this bespoke rdf collection function instead of the general one).

jeenbroekstra:
@hmottestad apart from the class rename and a bit of javadoc I asked is there anything holding this up? Anything I can do to help get this merged?

hmottestad:
Feel free to rename the class.

jeenbroekstra:
I've also just noticed that a lot of classes in `org.eclipse.rdf4j.sail.shacl` are missing copyright headers. I'll do the class rename and fix a few headers, but would appreciate if you also went through it to make sure we didn't miss any.

jeenbroekstra (approved):
I'm fine with this now. I'd still appreciate it if you could take a look at possible missing copyright headers, but that's not necessarily something that needs fixing in _this_ PR.

jeenbroekstra:
Btw I didn't squash my two commits because of branch sharing and possible rebasing, but we really should. Let me know when you are done adding further commits and I'll squash and force-push.

hmottestad:
Looks good @jeenbroekstra you can squash your commits and merge :)

I swear I had fixed those headers. I remember they went missing at some point and I went through all the classes and had to put them back in and now they are apparently gone again. I will make a follow up issue to fix it again. And this time I will actually fix it and not just apparently imagine that I did!

hmottestad:
https://github.com/eclipse/rdf4j/issues/2982

Issue #2890, opened 2021-03-06
Add Path interface to rdf4j-model-api module


knoan:
In support of use cases like data modelling and linked data navigation, it would be very useful to be able to describe RDF property paths in a consistent way.

`rdf4j-sparqlbuilder` and `rdf4j-queryalgebra` modules contain path-related classes, but they are deeply ingrained with the respective operational internals and not immediately available for declarative path description.

Submitting a PR for `Path` interface in the `rdf4j-model-api` module.

jeenbroekstra:
Can you give an example/use case to illustrate what this is  about?

knoan:
> Can you give an example/use case to illustrate what this is about?

See https://github.com/eclipse/rdf4j/pull/2891#issuecomment-791961433

jeenbroekstra:
@knoan I hope I didn't discourage you... I'm happy to have a chat about any improvements for linked data architectures you have in mind.

knoan:
Never mind… I wasn't that happy with this approach and your feedback motivated me to look for something cleaner.

PR #2891, opened 2021-03-06
GH-2890 Add Step/Path interfaces to model-api module

knoan:
Signed-off-by: Alessandro Bollini <22@metreeca.com>

GitHub issue resolved: #2890

- Add Step/Path interfaces and default implementations to `model-api` module, to support use cases like data modelling and linked data navigation

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/IRI.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/IRI.java
@@ -31,7 +31,7 @@ package org.eclipse.rdf4j.model;
  *           specs.
  */
 @SuppressWarnings("deprecation")
-public interface IRI extends URI, Resource {
+public interface IRI extends URI, Resource, Step, Path.Hop {
 
 	@Override
 	default boolean isIRI() {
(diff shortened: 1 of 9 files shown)


hmottestad:
Consider adding default interface methods so that this becomes backwards compatible.

knoan:
> Consider adding default interface methods so that this becomes backwards compatible.

New methods inherited by `IRI` from `Step` and `Path.Hop` are already defaulted: do you see other specific places where defaults are missing?

jeenbroekstra:
Woa. Ok so this looks involved, and like a lot of work went into it @knoan. Before we can start thinking about accepting it in though, I'll need a slightly better understanding of what this is for. I have to say, at first glance this looks misplaced, as this introduces SPARQL query concepts into the model api.

knoan:
> Woa. Ok so this looks involved, and like a lot of work went into it @knoan. Before we can start thinking about accepting it in though, I'll need a slightly better understanding of what this is for. I have to say, at first glance this looks misplaced, as this introduces SPARQL query concepts into the model api.

Over the past few years I've been working on a model-driven linked data API engine (https://github.com/metreeca/link) and concepts like direct/inverse links and property chains keep popping up everywhere; for instance:

- when describing a model for a framed JSON-LD representation, you need to say things like "this JSON fields maps to an inverse property of the target resource"
- when querying for possible values to be included in a facet for a nested resource property, you need the concept of "the values reachable from the collection resources through a chain of direct/inverse properties"

The code I'm submitting is a refactored version of what we developed internally in support of such use cases: I though it could be useful elsewhere.

Finally, my feeling is that property paths are more of a linked data concept borrowed by SPARQL than the other way around: for instance, you could think of applying paths for compact linked data traversal and value retrieval on the `Model` interface.

knoan:
> The code I'm submitting is a refactored version of what we developed internally in support of such use cases

P.S. The critical bit here is the `Step` interface; `Path` is a generalisation of the concept compliant with SPARQL specs

knoan:
If we don't care about `IRI/Step` interoperability we could also consider migrating `Link/Step/Path` to a dedicate module.

knoan:
Rev

> If we don't care about `IRI/Step` interoperability, migrating `Link/Step/Path` to a dedicate module is likely less confusing.

Reviewing the proposal to this end.

jeenbroekstra:
Thanks @knoan . I'm still struggling a little to wrap my head around it to be honest. Could you give a simple example of how you'd use this in code?

I like the idea separating it out a little. IRI inheriting from those new interfaces seems odd to me (though that may because I don't fully grok it yet).

knoan:
After reviewing once more the whole matter, I can't but agree with you that the use case is moot…

Back to the drawing board to see if we can handle the core use cases in a more straightforward way.

In the meantime just put the thing on hold.

Issue #2876, opened 2021-02-24
Upgrade JUnit to Junit5/Jupiter


JervenBolleman:
We should try to upgrade to Junit5. Just a small maintenance task, not urgent.

hmottestad:
Same as https://github.com/eclipse/rdf4j/issues/1563

JervenBolleman:
@hmottestad thanks, I had done an issue search but it had not shown up :( I did notice that a lot of tasks where already done. So I will have a look and see if I can finish it.

jeenbroekstra:
> @hmottestad thanks, I had done an issue search but it had not shown up :( I did notice that a lot of tasks where already done. So I will have a look and see if I can finish it.

I see you've already started using this issue, so I'll just close the older issue as a duplicate.

jeenbroekstra:
CQ for wiremock version bump: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23133

jeenbroekstra:
CQ approved

PR #2894, opened 2021-03-07
GH-2876 Upgrade to JUnit5

JervenBolleman:
GitHub issue resolved: #2876

This is a JUnit5 upgrade most changes are in the POMs as we are still using the JUnit4 APIs via junit vintage.
It does update a number of dependencies to newer versions which might need special review.

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -560,19 +560,16 @@
 				<groupId>org.junit.jupiter</groupId>
 				<artifactId>junit-jupiter-engine</artifactId>
 				<version>${junit.version}</version>
-				<scope>test</scope>
 			</dependency>
 			<dependency>
 				<groupId>org.junit.vintage</groupId>
 				<artifactId>junit-vintage-engine</artifactId>
 				<version>${junit.version}</version>
-				<scope>test</scope>
 			</dependency>
 			<dependency>
 				<groupId>org.junit.jupiter</groupId>
 				<artifactId>junit-jupiter-params</artifactId>
 				<version>${junit.version}</version>
-				<scope>test</scope>
 			</dependency>
 			<dependency>
 				<groupId>org.mockito</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -351,6 +351,7 @@
 		<guava.version>29.0-jre</guava.version>
 		<jmhVersion>1.21</jmhVersion>
 		<servlet.version>3.1.0</servlet.version>
+		<junit.version>5.7.1</junit.version>
 	</properties>
 	<dependencyManagement>
 		<dependencies>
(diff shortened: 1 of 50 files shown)


pom.xml
```diff
@@ -556,24 +557,43 @@
 			</dependency>
 			<!-- Testing: JUnit -->
 			<dependency>
-				<groupId>junit</groupId>
-				<artifactId>junit</artifactId>
-				<version>4.13.1</version>
+				<groupId>org.junit.jupiter</groupId>
+				<artifactId>junit-jupiter-engine</artifactId>
+				<version>${junit.version}</version>
+				<scope>test</scope>
```
jeenbroekstra, line 563: Can you remove the scope declaration here (and for the other new dependencies)? We've recently stepped away from declaring scope in the dependency management section of the root pom, because it may have some unexpected effects in downstream projects (see GH-2812). 

In this case this is of course very unlikely to ever be a problem (I doubt many projects would use a scope other than test for junit), but let's stick to the principle: dependencyManagement declares only versions, scope is declared where the actual dependency is needed.

pom.xml
```diff
@@ -556,24 +557,43 @@
...
+				<groupId>org.mockito</groupId>
+				<artifactId>mockito-junit-jupiter</artifactId>
+				<version>3.8.0</version>
+				<scope>test</scope>
 			</dependency>
 			<dependency>
 				<groupId>com.github.tomakehurst</groupId>
 				<artifactId>wiremock-jre8</artifactId>
-				<version>2.23.2</version>
+				<version>2.27.2</version>
```
jeenbroekstra, line 591: Unfortunately, the clearlydefined score for this library is too low, so we'll have to log a CQ for this. I'll take care of it.

jeenbroekstra (commented):
This is great work @JervenBolleman, thanks! I just have one minor remark inline, plus we'll need a CQ for the wiremock version bump (the other version bumps are fine without btw). But other than that this looks fine.

pom.xml
JervenBolleman: I think I can remove the scope, but that would make it the default scope compile wouldn't it?
Anyway the compliance suites still use the original junit4 approach and in compile scope if I figured that out.
Considering the pain I had with spring dependencies in the tools module I can understand the problem.

jeenbroekstra: > I think I can remove the scope, but that would make it the default scope compile wouldn't it?

It would, but the idea is that wherever we actually use the dependency, we declare the scope (typically test, though as you found in the compliance suites, sometimes compile). So the root pom defines the version of the library, but not the scope.

jeenbroekstra:
CQ has been approved. Test failure was an intermittent failure in one of the SHACL tests, I've restarted the CI build to hopefully get rid of this.

Issue #2906, opened 2021-03-10
SHACL - parallel validation during SERIALIZABLE can cause stack overflow


hmottestad:
Essentially we end up with a lot of delegating sail datasets with an observation for each query that's been run. So in the end we get a stack overflow trying to query more data.

We do intend to reduce the number of queries being generated, but since this particular issue here only happens when running parallel validation during SERIALIZABLE isolation it's simplest to disable parallel validation in those cases.

PR #2907, opened 2021-03-10
GH-2906 disable parallel validation during SERIALIZABLE isolation

hmottestad:
GitHub issue resolved: #2906 

----


core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/TransactionSettingsTest.java
```diff
@@ -167,6 +167,32 @@ public void testDefaultOverride() {
...
+		try (SailRepositoryConnection connection = sailRepository.getConnection()) {
+
+			connection.begin(IsolationLevels.SERIALIZABLE,
+					ShaclSail.TransactionSettings.PerformanceHint.ParallelValidation);
+
+			ShaclSailConnection sailConnection = (ShaclSailConnection) connection.getSailConnection();
+			ShaclSailConnection.Settings transactionSettings = sailConnection.getTransactionSettings();
+
+			assert transactionSettings.getValidationApproach() == ShaclSail.TransactionSettings.ValidationApproach.Auto;
+			assert !transactionSettings.isParallelValidation();
```
jeenbroekstra, line 195: Wrong kind of assert. Should be using junit `assertTrue` (or even better, AssertJ `assertThat`) here.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -938,6 +946,16 @@ void applyTransactionSettings(Settings transactionSettingsLocal) {
 				validationApproach = transactionSettingsLocal.validationApproach;
 			}
 
+			assert transactionSettingsLocal.isolationLevel == null;
```
jeenbroekstra, line 949: I'm not sure I follow the logic of this assertion. Why are you assuming that the isolationLevel in the supplied `Settings` object will always be `null`?

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/TransactionSettingsTest.java
```diff
@@ -167,6 +167,32 @@ public void testDefaultOverride() {
 			assert transactionSettings.getValidationApproach() == ShaclSail.TransactionSettings.ValidationApproach.Auto;
 			assert !transactionSettings.isCacheSelectNodes();
 			assert !transactionSettings.isParallelValidation();
+			assert transactionSettings.getIsolationLevel() == IsolationLevels.SNAPSHOT_READ;
```
jeenbroekstra, line 170: Wrong kind of assert. These should be junit or assertj assertions, not java asserts.

jeenbroekstra (commented):
Looks generally ok to me, bar a few remarks. Also: test failure, not immediately obvious to me that it is related though.

hmottestad:
The test failure is related to: https://github.com/eclipse/rdf4j/issues/2908

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
hmottestad: It's because we don't support overriding the isolation level.

Issue #2918, opened 2021-03-12
Possible deadlock when processing query results from an RDF4J server


ledsoft:
Hello all,
I've encountered a situation that feels like a bug in query result processing.

**The problem:** I have a select that I execute against a repository on an RDF4J server. When going through the results, a follow-up select query is run for each item. The problem is that RDF4J hangs when processing the follow-up query results. 

**Steps to reproduce:** Setup an RDF4J server, create a repository (type doesn't matter, no inference is needed). Load the [SKOS](https://www.w3.org/TR/skos-reference/skos.rdf) vocabulary as sample data. The following test (JUnit 5) hangs when attempting to get the `ranges`:

```java
public class DeadlockTest {

    private static final String REPO_URI = "http://localhost:8080/rdf4j-server/repositories/skos";

    private static final String MAIN_QUERY = "SELECT ?property WHERE { ?property a rdf:Property . }";

    private static final String SUB_QUERY = "SELECT ?range WHERE { ?property rdfs:range ?range . }";

    private Repository repository;

    @BeforeEach
    void setUp() {
        this.repository = RepositoryProvider.getRepository(REPO_URI);
    }

    @AfterEach
    void tearDown() {
        repository.shutDown();
    }

    @Test
    void testRDF4JQueryDeadlock() {
        final Map<URI, Set<URI>> propertyRanges = new HashMap<>();
        try (final RepositoryConnection conn = repository.getConnection()) {
            final TupleQuery main = conn.prepareTupleQuery(MAIN_QUERY);
            final TupleQueryResult mainResult = main.evaluate();
            while (mainResult.hasNext()) {
                final BindingSet current = mainResult.next();
                final URI u = URI.create(current.getValue("property").stringValue());
                // This works just fine
                assertTrue(conn.getStatements(conn.getValueFactory().createIRI(u.toString()), null, null).hasNext());
                final TupleQuery subQuery = conn.prepareTupleQuery(SUB_QUERY);
                subQuery.setBinding("property", conn.getValueFactory().createIRI(u.toString()));
                final TupleQueryResult sqResult = subQuery.evaluate();
                // This hangs indefinitely
                final Set<URI> ranges = sqResult.stream().map(bs -> URI.create(bs.getValue("range").stringValue())).collect(Collectors.toSet());
                propertyRanges.put(u, ranges);
            }
        }
        assertFalse(propertyRanges.isEmpty());
    }
}
```

**Additional notes:**
[...]

PR #2923, opened 2021-03-16
GH-2918 switch to CachedThreadPool to avoid deadlocks in interdependent queries

jeenbroekstra:
GitHub issue resolved: #2918 

- switch to CachedThreadPool executor since it has better behavior for spawning new threads as needed
- allow core threads to be terminated when keepAlive timeout has passed to conserve resources
- RDF4JProtocolSession uses a separate scheduled executor specifically for its transaction ping functionality
- added regression test

----

### DIFF
--- a/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/http/HTTPRepositoryTest.java
+++ b/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/http/HTTPRepositoryTest.java
@@ -7,10 +7,29 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.repository.http;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.rdf4j.model.util.Values.iri;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.vocabulary.FOAF;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.query.BindingSet;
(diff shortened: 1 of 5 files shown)


hmottestad:
Can you make a test that has 8 "subqueries" (nested 8 levels) so we know that we are fixing the underlying issue of how our code handles asynchronous execution?

jeenbroekstra:
> Can you make a test that has 8 "subqueries" (nested 8 levels) so we know that we are fixing the underlying issue of how our code handles asynchronous execution?

No. Because we already know that that will fail, unless the core pool size is further upped.

jeenbroekstra:
I had another look at this and I'm starting to think that switching to a CachedThreadPoolExecutor might actually be a better solution after all.

jeenbroekstra:
Unless I receive any objections I'll merge this by the end of the day.

hmottestad:
Does it fix the underlying design issue so that users can have unlimited nesting of queries?

jeenbroekstra:
> Does it fix the underlying design issue so that users can have unlimited nesting of queries?

Not unlimited no - there is an upper limit to the number of concurrent threads a CachedThreadPool can support (at most Integer.MAX_VALUE though in practice you're more likely to hit resource boundaries first of course). But for all practical purposes that is good enough. 

In contrast to a ScheduledExecutor service, a Cached executor service does not queue jobs to wait for an existing thread to become available - it either grabs an existing available thread or immediately spawns a new one to execute the job. So it will never block to wait, and the deadlock situation can never occur, even with a large number of nested queries.

hmottestad:
> > Does it fix the underlying design issue so that users can have unlimited nesting of queries?
> 
> Not unlimited no - there is an upper limit to the number of concurrent threads a CachedThreadPool can support (at most Integer.MAX_VALUE though in practice you're more likely to hit resource boundaries first of course). But for all practical purposes that is good enough.
> 
> In contrast to a ScheduledExecutor service, a Cached executor service does not queue jobs to wait for an existing thread to become available - it either grabs an existing available thread or immediately spawns a new one to execute the job. So it will never block to wait, and the deadlock situation can never occur, even with a large number of nested queries.

That’s the definition of unlimited in computer science.

Issue #2925, opened 2021-03-16
Clean up test suite to reduce build time


jeenbroekstra:
The current full validation test cycle runs at close to an hour, 22 minutes of which are taken up by ShaclSail-related tests:

```
[INFO] Reactor Summary for Eclipse RDF4J 3.6.2-SNAPSHOT:
[INFO] 
[INFO] Eclipse RDF4J ...................................... SUCCESS [  1.318 s]
[INFO] RDF4J Assembly Descriptors ......................... SUCCESS [  0.960 s]
[INFO] RDF4J Core ......................................... SUCCESS [  0.533 s]
[INFO] RDF4J: Model API ................................... SUCCESS [  5.406 s]
[INFO] RDF4J: util ........................................ SUCCESS [  7.859 s]
[INFO] RDF4J: RDF Vocabularies ............................ SUCCESS [  2.246 s]
[INFO] RDF4J: Model ....................................... SUCCESS [  5.992 s]
[INFO] RDF4J: SparqlBuilder ............................... SUCCESS [  3.259 s]
[INFO] RDF4J: Rio ......................................... SUCCESS [  0.474 s]
[INFO] RDF4J: Rio - API ................................... SUCCESS [  3.328 s]
[INFO] RDF4J: Rio - Languages ............................. SUCCESS [  1.262 s]
[INFO] RDF4J: Rio - Datatypes ............................. SUCCESS [  1.960 s]
[INFO] RDF4J: Query ....................................... SUCCESS [  1.760 s]
[INFO] RDF4J: Rio - Binary ................................ SUCCESS [  3.095 s]
[INFO] RDF4J: Rio - N-Triples ............................. SUCCESS [  4.262 s]
[INFO] RDF4J: Rio - HDT ................................... SUCCESS [  2.421 s]
[INFO] RDF4J: Rio - JSON-LD ............................... SUCCESS [  4.545 s]
[INFO] RDF4J: Rio - Turtle ................................ SUCCESS [  7.678 s]
[INFO] RDF4J: Rio - N3 (writer-only) ...................... SUCCESS [  2.055 s]
[INFO] RDF4J: Rio - N-Quads ............................... SUCCESS [  4.135 s]
[INFO] RDF4J: Rio - RDF/JSON .............................. SUCCESS [  3.405 s]
[INFO] RDF4J: Rio - RDF/XML ............................... SUCCESS [  6.165 s]
[...]

PR #2926, opened 2021-03-16
GH-2925 reduce test time by disabling some long-running tests

jeenbroekstra:
GitHub issue resolved: #2925  

----


jeenbroekstra:
Just disabling these two shacl test suites shaved off nearly 15 minutes.

hmottestad:
Please don't disable these. They have detected various issues in both the ShaclSail and the NativeStore. These are some of the only fuzzing tests we have against the NativeStore. 

I've previously had success with using file name filters to split tests between the surefire and the other maven test plugin. We could do something similar so that only the core tests are blocking merging while the remaining tests still run with the GitHub CI so you can always wait on those if you want to be extra sure.

jeenbroekstra:
> Please don't disable these. They have detected various issues in both the ShaclSail and the NativeStore. 

I know you find them useful, but frankly they're in everybody else's way. I am not prepared to spend an extra 15 minutes on every single build just for these tests. 

We need a mechanism to either run them on manual request only, or as part of a separate "do all the slow tests" check for the main branch. Or like you propose, as part of a separate "be extra" sure step of the build process at the very least.

>  These are some of the only fuzzing tests we have against the NativeStore.

We have full set of compliance tests for the native store. And these tests are about Shacl, they're not in the right place for running tests on the native store. 

> I've previously had success with using file name filters to split tests between the surefire and the other maven test plugin. We could do something similar so that only the core tests are blocking merging while the remaining tests still run with the GitHub CI so you can always wait on those if you want to be extra sure.

We used have a separation between compliance tests and unit tests, precisely for this purpose: it allowed us to control which test plugins were executed by means of activating profiles. The unit tests used surefire, the compliance tests used failsafe. Unfortunately it looks as if that setup has been compromised somewhat with all the migrations we've done over the years. And of course for Shacl you have not made a clear separation between unit and compliance tests, instead keeping everything in the one module. I'll take a closer look at compliance test setup.

Another alternative is that we look into Junit 5's conditional execution mechanism (see the comment on the issue).

jeenbroekstra:
@hmottestad got a couple of ideas for approaches to this, see comments on GH-2925. Would appreciate your input (especially since the ShaclSail tests are such a large part of the problem here).

hmottestad:
@jeenbroekstra I've split out our integration tests into their own github action and turned the slow shacl tests into integration tests.

hmottestad:
TODO: Edit the printTestResults.sh script to also print integration test failures (failsafe).

hmottestad:
Just going to admit it right here and now. I'm terrible at configuring maven, and maven is terrible at telling me why it's doing something. I have no idea why it's suddenly running the integration tests during the build even though tests are disabled.

barthanssens:
Not sure, probably I'm not interpreting this correctly, but looking at the pom it looks like while the surefire plugin is set to false, the failsafe plugin is configured to run files ending on either Test or IT ?

```
<include>**/*Test.java</include>
<include>**/*IT.java</include>
```

.github/workflows/pr-integration-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-jdk${{ matrix.jdk }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-jdk${{ matrix.jdk }}-maven-
+    - name: Build
+      run: mvn -B clean install -Pquick,\!formatting
+    - name: Run tests
+      run: mvn -B integration-test -Pintegration -Dskip.surefire.tests
```
jeenbroekstra, line 27: Just as an FYI: never execute the 'integration-test' goal directly like this, because doing so makes maven skip the post-integration test phase (which is where the cleanup happens, e.g.  spinning down jetty servers and other resources). Always use `mvn verify` to execute integration testing.

jeenbroekstra:
> Just going to admit it right here and now. I'm terrible at configuring maven, and maven is terrible at telling me why it's doing something. I have no idea why it's suddenly running the integration tests during the build even though tests are disabled.

Don't worry too much about the maven side of things right now, I'll try and spend some time on reorganizing things along the lines I proposed on the ticket, which I'll try and get done over the weekend. The renaming/reorganizing of tests you did here is a good start, thanks.

jeenbroekstra:
@hmottestad @barthanssens I've made some progress in a better setup for this. If I've got it right, it's now set up as follows:

1. the pr-verify github action first runs all unit tests (the surefire plugin): only test classes ending on *Test.java. Idea is to have fast failure if unit tests fail anywhere in the project
2. as a second testing step it runs the integration tests (using the failsafe-plugin)
3. github skips any tests annotated with `@Tag("slow")` (be aware that this only works on Junit 5 unit tests, I had to migrate a few tests to get that working)
4. you can easily run slow tests in your own evironment by just executing them in your IDE (the tag should not make your IDE skip the test), or alternatively by locally disabling the "skipSlowTests" maven profile: `mvn -P\!skipSlowTests verify` (if you don't want to do this from the command line every time, you can also configure your local maven settings.xml file).
5. we have a separate github action for running the slow tests against a PR - the idea being that you typically do not have to wait on that action to complete before merging. 
 
Of course, what is a unit test and what an integration test (and indeed what is a _slow_ test) is a bit of a judgment call. I've focused on the SHACL testsuite sofar, there may be some other tests elsewhere that we can tag in this way. Of course we also want to be careful not to mark too many tests as slow, as that may cause undetected regressions.

jeenbroekstra:
Oh drat, for some reason the additional github action is automatically marked as a required check. I'll remove it for now. If we find we really need this we can reconsider, but for now we should be good with just using local test execution and a bit of discretion.

jeenbroekstra:
There's a couple of problems I need to weed out, most likely related to the version bump of the surefire/failsafe plugins (to accomodate junit 5).

I'm also not 100% convinced that separating out the unit test phase actually makes things quicker. We may be better off just doing the unit tests as part of the build, and only keeping the integration tests separate.

hmottestad:
Great work @jeenbroekstra. 

To make things quicker try making another action for just the integration tests. That action can also skip formatting and all the other checks we have. 

`mvn -T 2 install -DskipTests -Dmaven.javadoc.skip=true -Dformatter.skip=true -Dimpsort.skip=true -Dxml-format.skip=true  -Djapicmp.skip -Denforcer.skip=true -Dbuildnumber.plugin.phase=none -Danimal.sniffer.skip=true
`

Also can you add back the action for the slow tests so I can see how it works? How long time do you think the eclipse admins would take to change that action to be not required?

pom.xml
```diff
@@ -773,25 +771,42 @@
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-surefire-plugin</artifactId>
-					<version>2.22.2</version>
+					<version>3.0.0-M5</version>
 					<configuration>
 						<encoding>UTF-8</encoding>
 						<argLine>-Xmx1G</argLine>
+						<skipTests>${skipUnitTests}</skipTests>
```
hmottestad, line 778: Btw. This make it so that the regular `-DskipTests` doesn't work. I wonder if there is some way to set `skipUnitTests` to true if `skipTests` is true. For the sanity of other developers I think it's best that the default maven properties still work, so they don't need to hunt in our documentation to figure out they should use the quick profile.

jeenbroekstra:
> Great work @jeenbroekstra.
> 
> To make things quicker try making another action for just the integration tests. That action can also skip formatting and all the other checks we have.

Another action won't be quicker. The quickest way to get things verified is to stick to the maven build lifecycle. And we already skip formatting for the verification step. 
 
> This is the fastest install I have managed: `mvn -T 2 install -DskipTests -Dmaven.javadoc.skip=true -Dformatter.skip=true -Dimpsort.skip=true -Dxml-format.skip=true -Djapicmp.skip -Denforcer.skip=true -Dbuildnumber.plugin.phase=none -Danimal.sniffer.skip=true `

Well yes, of course, if you skip everything, it will be very quick. But that's not very useful.
 
> Also can you add back the action for the slow tests so I can see how it works? How long time do you think the eclipse admins would take to change that action to be not required?

I'm not going to add that into the mix for now, I want to get this PR running stable first. Adding this in as a separate github action is something we'll consider in a followup.

pom.xml
jeenbroekstra: Good point, I'll take a look at this.

jeenbroekstra:
> Well yes, of course, if you skip everything, it will be very quick. But that's not very useful.

That came out crankier than I intended. The idea of using the -T 2 flag is definitely worth considering.

hmottestad:
@jeenbroekstra I intended to skip everything. Here is the setup i had in mind:

## Actions:

### PR-verify Unit tests (required)
 1. check formatting
 1. clean install skip tests (single threaded so the log is easy to read)
 1. unit tests

### PR-verify Integration tests (required)
 1. check formatting (since it's a common thing to get wrong, so we can fail the build fast)
 1. fast install where we skip everything and use multiple cores
 1. run integration tests

### PR-verify Slow tests (optional)
 1. check formatting (since it's a common thing to get wrong, so we can fail the build fast)
 1. fast install where we skip everything and use multiple cores
 1. run the slow tests

hmottestad:
Each of those actions will run up a new VM, so all the actions should run in parallel! Meaning that the time spent will only be whichever of the required actions is the slowest. Should bring the wall clock down to ~20 min.

jeenbroekstra:
Ah, I see, parallelize by means of separate actions. It's possible though I'm not sure that we win by parallelizing won't be lost by the fact that even a fast install still takes a while. But it's worth trying.

I want to make sure that I can get a green build again first, with all the changes I've made to the poms. Once we have that, putting the separate actions in won't be too hard.

jeenbroekstra:
I've hit a snag. Apparently the newer version of the failsafe plugin sets its directory context differently. This cause the integration tests that spin up an RDF4J Server in a Jetty container to fail, because they cannot locate the war file when maven is executed from the project root. 

Current options I'm considering:

1. Find a way to refer to the absolute path in Java code (I'll need some way to figure out the maven project root at runtime) instead of the current relative path we use for the webapp (./target/rdf4j-server). 
2. Look into using the jetty maven plugin instead of spinning up a server from code.

jeenbroekstra:
As an aside we seem to have three separate implementations for spinning up a Jetty container - we really should clean that up (though perhaps not as part of this PR).

hmottestad:
Doesn't help much now, but spring-boot lets you spinn up the entire server by just annotating the class with `@SpringBootTest`. No need for starting your own webserver and uploading the war file.

jeenbroekstra:
As usual, the next thing to give me grief is the integration tests for solr, which somehow now end up with NoClassDefFound errors. Not immediately spotting what might be causing this.

hmottestad:
I really appreciate that you are willing to make this work so my tests can still keep running :) I know by now you probably wished you had just disabled my tests and left it at that. So thank you @jeenbroekstra I really appreciate you putting in the effort for me!

jeenbroekstra:
> I really appreciate that you are willing to make this work so my tests can still keep running :) I know by now you probably wished you had just disabled my tests and left it at that. So thank you @jeenbroekstra I really appreciate you putting in the effort for me!

Oh, no worries, a lot of this was overdue maintenance anyway. I think a lot of the brittleness in the suite is caused by certain testsuites still running on Junit 3. 

That's not to say it isn't a little like pulling teeth to get it all to work :)

jeenbroekstra:
Ah, this is an interesting data point: the build on jdk 1.8 now succeeds, but on JDK 15 fails. That might indicate the problem is in one of the profile configs we have in there that activate only for higher java versions.

pom.xml
jeenbroekstra: This should be fixed now btw.

jeenbroekstra:
It's painful but we're getting there I think

jeenbroekstra:
The build errors are getting really exotic now:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-jar-plugin:2.3.1:jar (default-jar) on project rdf4j-geosparql-compliance: Error assembling JAR: Failed to read filesystem attributes for: /home/runner/work/rdf4j/rdf4j/compliance/geosparql/pom.xml: Failed to retrieve numeric file attributes using: '/bin/sh -c ls -1nlad /home/runner/work/rdf4j/rdf4j/compliance/geosparql/pom.xml': Error while executing process. Cannot run program "/bin/sh": error=13, Permission denied -> [Help 1]
```

hmottestad:
Maybe a version bump will help? You can choose between 2.6 or 3.2.0 .

jeenbroekstra:
I'll check. the good thing is I can reproduce locally.

jeenbroekstra:
:partying_face: now I can look at splitting into several actions.

jeenbroekstra:
First attempt at splitting up. I've left each action to use both JDK 8 and 15, which is probably overkill. @hmottestad any suggestions on which jobs you think should check both java versions, and which ones can do only java 8?

jeenbroekstra:
I think I may have been a bit overzealous in disabling the japicmp and the animal-sniffer. At least _one_ of the jobs should run those of course.

jeenbroekstra:
Taking a look at putting multiple jobs in a single action script, might make it a little easier to manage.

jeenbroekstra:
Ok, so I think that we only need the matrix build on Java 15 for the compilation + unit tests. If we do something that's not Java 15-compatible it will be at that level, rather than something that will only show up in the integration tests. So I'll take out the matrix strategies for integration tests and slow tests in the PR verify flow.

Note, by the way, that for the main and develop branch check flows we still run all tests on both Java 8 and 15, so we have a backup. But this should limit the number of concurrent jobs somewhat (and from what I understand, that number is capped per organization, so we should be nice to our fellow Eclipse projects and not hog too many concurrent jobs :))

jeenbroekstra:
I wonder if we're hitting Github at a bad time, or if this is a consequence of putting the jobs in a single action, but I've never seen checks queued for this long before... Also rather annoyed at the fact that it chooses to run the most expensive jobs first :/

jeenbroekstra:
I think I'm happy with this as-is for now. Let's for the moment assume that the job queuing I saw is a temporary thing at Github's end. If it turns out that with this setup it consistently queues jobs for a long time we can try going back to mutliple workflow files. @hmottestad have a look at the current state, tell me if you see any flaws in the plan, if not I think we can merge this.

hmottestad:
Two things I noticed.

1. slow-tests is actually running all tests and not just the slow tests
2. There are some strange references to google collections:

`2021-03-25T08:24:32.7257248Z [INFO] Running com.google.common.collect.testing.testers.CollectionStreamTester`

jeenbroekstra:
> Two things I noticed.
> 
> 1. slow-tests is actually running all tests and not just the slow tests

Yes, good point. That was easier to set up, but I'll take a quick look, see if I can tweak it.

> 2. There are some strange references to google collections:
> 
> `2021-03-25T08:24:32.7257248Z [INFO] Running com.google.common.collect.testing.testers.CollectionStreamTester`

These are caused by `org.eclipse.rdf4j.model.impl.ModelCollectionTest`, which uses a Guava test library to run a suite of tests on our Model implementations to make sure they adhere to the Java Collection API contracts. I'm not sure why these log messages have started showing up to be honest (it might be a forking thing in the version-bumped surefire runner), but I don't think it's cause for alarm - the entire suite runs in less than a second.

jeenbroekstra:
Update on the running of _only_ slow tests: in theory, this should be a simple as just adding `-P\!skipSlowTests -Dgroups=slow` to the maven command. However, I've tried this locally and it causes a NullPointerException in the running of the tests for serql compliance (and, I suspect, for others as well). Speculatively I'm thinking this is caused by the fact that it still uses the now very obsolete JUnit 3 TestSuite mechanism.

jeenbroekstra:
I suggest that we merge this now, and treat making it possible to run _only_ the slow tests as a followup. I want to avoid this growing stale.

Issue #2932, opened 2021-03-19
Blank nodes lost in Turtle output when using INLINE_BLANK_NODES


edwardsph:
The following code, using v3.6.1 demonstrates how nodes are lost in the output:
```Java
InputStream is = getClass().getClassLoader().getResourceAsStream("blank-node-example.ttl");
Model model = Rio.parse(is, RDFFormat.TURTLE);
assertEquals(43, model.size());
StringWriter sw = new StringWriter();
RDFWriter rdfWriter = Rio.createWriter(RDFFormat.TURTLE, sw);
rdfWriter.getWriterConfig().set(BasicWriterSettings.PRETTY_PRINT, true).set(BasicWriterSettings.INLINE_BLANK_NODES, true);
Rio.write(model, rdfWriter);
logger.debug("blank-node-example.ttl\n{}", sw.toString());
```
The input file is:
```
@prefix test-harness: <http://example.org/> .
@prefix earl: <http://www.w3.org/ns/earl#> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

test-harness:DISPLAY_NAME a earl:TestCriterion, earl:TestRequirement;
                          dcterms:title "FEATURE NAME";
                          earl:assertions _:node1f152do7hx3;
                          dcterms:hasPart _:node1f152do7hx5, _:node1f152do7hx8 .

_:node1f152do7hx3 a earl:Assertion;
                  earl:assertedBy <http://example.org/>;
                  earl:test test-harness:DISPLAY_NAME;
                  earl:subject test-harness:test;
                  earl:mode earl:automatic;
                  earl:result _:node1f152do7hx4 .

_:node1f152do7hx4 a earl:TestResult;
                  earl:outcome earl:failed;
                  dcterms:date "1970-01-01T01:00:00.999+01:00"^^xsd:dateTime .

_:node1f152do7hx5 a earl:TestCriterion, earl:TestCase;
                  dcterms:title "SCENARIO 1";
                  dcterms:isPartOf test-harness:DISPLAY_NAME;
                  earl:assertions _:node1f152do7hx6 .

_:node1f152do7hx6 a earl:Assertion;
                  earl:assertedBy <http://example.org/>;
                  earl:test _:node1f152do7hx5;
                  earl:subject test-harness:test;
                  earl:mode earl:automatic;
                  earl:result _:node1f152do7hx7 .

[...]

PR #2934, opened 2021-03-21
GH-2932 blank node inlining with circular references

hmottestad:
GitHub issue resolved: #2932 

----

### DIFF
--- a/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
+++ b/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
@@ -845,45 +845,25 @@ public class TurtleWriter extends AbstractRDFWriter implements RDFWriter, CharSi
 	}
 
 	private Optional<Resource> nextSubject(Model contextData, Set<Resource> processedSubjects) {
+		// first try finding a subject that has not yet been processed and is not the object of another
+		// unprocessed subject
 		for (Resource subject : contextData.subjects()) {
 			if (processedSubjects.contains(subject)) {
 				continue;
 			}
 			if (subject.isBNode() && inlineBNodes) {
 				Set<Resource> otherSubjects = contextData.filter(null, null, subject).subjects();
-				if (otherSubjects.stream()
-						.anyMatch(s -> !processedSubjects.contains(s)
-								&& !isInBlankNodeCycle((BNode) subject, contextData, processedSubjects))) {
-					// Other unprocessed subject bnode used by this subject is present, and current subject is not part
-					// of a cycle. We should not yet pick this subject.
+				if (otherSubjects.stream().anyMatch(s -> !processedSubjects.contains(s))) {
+					// Other unprocessed subject using this subject as an object is present. Skip this one for now.
 					continue;
 				}
 			}
 			return Optional.of(subject);
 		}
-		return Optional.empty();
-	}
 
-	private boolean isInBlankNodeCycle(BNode subject, Model contextData, Set<Resource> processedSubjects) {
-		final Set<BNode> visited = new HashSet<>();
-
-		final ArrayDeque<BNode> nodeStack = new ArrayDeque<>();
-		nodeStack.push(subject);
-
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
@hmottestad since this is a PR opened by you I can't add you as a reviewer, but have a look nevertheless please.

hmottestad:
I'll try to take a look tonight.

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleWriter.java
```diff
@@ -860,6 +865,21 @@ private void processBuffer() throws RDFHandlerException {
...
+	private boolean isInBlankNodeCycle(BNode subject, Model contextData, Set<Resource> processedSubjects,
+			Set<BNode> visited) {
+		visited.add(subject);
+		for (Statement st : contextData.filter(null, null, subject)) {
+			if (st.getSubject().isBNode()) {
+				BNode otherSubject = (BNode) st.getSubject();
+				if (visited.contains(otherSubject)) {
+					return true;
+				}
+				return isInBlankNodeCycle(otherSubject, contextData, processedSubjects, visited);
```
hmottestad, line 877: I presume that this will break down for deeply nested blank nodes due to stack overflow. I also noticed that you are checking if the subject blank node leads to a cycle, rather than that the subject is part of a cycle. Eg. a would return true for: a->b->c->b. 

For:

```
				"<http://example.org/DISPLAY_NAME>\n" +
 				"  a <http://www.w3.org/ns/earl#TestCriterion> ;\n" +
 				"  dc:hasPart _:genid3 .\n" +
 				"\n" +
 				"_:genid3\n" +
 				"  a ns0:TestCase ;\n" +
 				"  ns0:assertions [\n" +
 				"    a ns0:Assertion ;\n" +
 				"    ns0:test _:genid3 \n" + // direct circular reference between two blank nodes
 				"  ] .";
```

Why is it not enough to check if count of (null, null, _:genid3) <= 1 before inlining it?

jeenbroekstra: > I presume that this will break down for deeply nested blank nodes due to stack overflow. 

Yes. As with any recursive method. Though if your blank nodes are nested that deep you have different problems I think (note by the way that this code path is not followed for RDF collections, which is the most likely place to have deeply nested blank nodes). 

I always find recursive methods easier to think about, but we could probably rewrite it to an iterative method, keeping a manual stack.

> Why is it not enough to check if count of (null, null, _:genid3) <= 1 before inlining it?

I'm not sure I follow what you're suggesting, but the problem here is not inlining, at least not directly. 

The bug is in the selection of the next subject to write statements for. In the pre-existing code we skip a blank node subject if there are other statements with that subject in the object-position that have not yet been written - the reasoning being that we can likely inline this current subject if we write that _other_ subject first. The bug in that reasoning is that if there is a cycle in the blank nodes, none of the nodes in that cycle will ever get chosen as the next subject to write. So this cycle check is just there to "break the loop", and pick a point in the cycle to start writing.

jeenbroekstra: I've rewritten the recursive check to an iterative one to avoid potential stack overflow.

hmottestad: How about then instead of using cycle detection we instead have a check before we return optional empty? If there are still subjects left to print, select a subject (any will do) and return that. If there are really no more subjects left then we return optional empty.

jeenbroekstra: Ah. Good idea. I'll take a look, I think we can solve that with a simple lambda.

jeenbroekstra: Done.

hmottestad:
Looks very simple and clean :)

Issue #2940, opened 2021-03-26
NativeStore constructor to default to temporary data directory


barthanssens:
For a few projects I'm using the NativeStore (instead of the MemoryStore, since the number of statements exceeds the 150k triples the MemoryStore is designed for) to load triples, perform some queries, extract the results and delete the NativeStore files.
I.e. in this case I'm not really interested where the store is kept, nor do I want to reuse the store afterwards.

Suggestion is to add a convenience method `setTempDir` (perhaps in sail.helpers.AbstractSail) that uses `Files.createTempDirectory` and (recursively) deletes it on `shutdown()`

Admittedly this only saves a few lines of code, but IMHO it's a nice little touch.

jeenbroekstra:
I get the use case but I'm not comfortable with sticking that directly in AbstractSail. 

I've been thinking that instead, we offer a utility class somewhere that does a one-line configuration of a Repository with a native store, backed up with a temp dir. We could call it "CommonConfig" or something like that. Also gives us the option of offering quick convenience methods for other config stacks.

jeenbroekstra:
Perhaps another alternative is to let the native store handle this internally, directly. It currently craps out with an exception if you haven't set a datadir, we could change its behavior to automatically create a tempDir (and clear it on shutDown) if you initialize it without a datadir.

barthanssens:
That was actually my initial idea, so works for me.

jeenbroekstra:
I'm just happy I managed to make a suggestion _other_ than "stick it in a utility class" - I'm aware I say that a lot :)

PR #2954, opened 2021-03-29
Gh 2940 nativestore tempdir

barthanssens:
GitHub issue resolved: #2940 

- set temp data dir instead of throwing an exception (will be deleted on shutdown)
- added tests for tmp datadir, including one when re-initializing after shutdown
- corrected minor typo
- slightly improved the nativestore/memstore website documentation on amount triples + explaine use case

----

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeStore.java
@@ -9,6 +9,9 @@ package org.eclipse.rdf4j.sail.nativerdf;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Comparator;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.commons.io.FileUtils;
@@ -78,6 +81,9 @@ public class NativeStore extends AbstractNotifyingSail implements FederatedServi
 	// used to decide if store is writable, is true if the store was writable during initialization
 	private boolean isWritable;
 
+	// indicates if a datadir is temporary (i.e. will be deleted on shutdown)
+	private boolean isTmpDatadir = false;
+
 	/**
 	 * Data directory lock.
 	 */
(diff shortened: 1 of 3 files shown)


barthanssens:
Might be worth mentioning this on the documentation website

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeStore.java
```diff
@@ -311,14 +335,33 @@ protected void shutDownInternal() throws SailException {
...
 		} finally {
 			dirLock.release();
 			if (dependentServiceResolver != null) {
 				dependentServiceResolver.shutDown();
 			}
-			logger.debug("NativeStore shut down");
+		}
+
+		if (isTmpDatadir) {
+			File tmpDatadir = getDataDir();
```
jeenbroekstra, line 346: Above, in setDatadir, you're checking if the supplied datadir is null to determine if `isTmpDatadir` should be true. I think that means that it's possible to get at this poin with the result `getDataDir()` being null? If so you need a null-check here somewhere.

site/content/documentation/programming/repository.md
```diff
@@ -59,9 +59,13 @@ memStore.setSyncDelay(1000L);
...
 ```
 
+The MemoryStore is designed for datasets with fewer than 100,000 triples.
+
 ### Native RDF Repository
 
-A Native RDF Repository does not keep its data in main memory, but instead stores it directly to disk (in a binary format optimized for compact storage and fast retrieval). It is an efficient, scalable and fast solution for RDF storage of datasets that are too large to keep entirely in memory.
+The NativeStore saves data to disk in a binary format which is optimized for compact storage and fast retrieval. If there is sufficient physical memory, the Native store will act like the MemoryStore on most operating systems because the read/write commands will be cached by the OS.
+
+It is therefore an efficient, scalable and fast solution for datasets with up to 100 million triples.
```
jeenbroekstra, line 68: We really need to do some scale benchmarking, I believe with some of the changes that we've made recently we can probably go higher than that.

jeenbroekstra (commented):
Already looks good to me, just one remark. Thanks for also including documentation changes.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/NativeStore.java
barthanssens: I don't think this is the case (though probably by coincidence: this could be the case when `setDatadir(null)` is called and `shutdownInternal()` is called without initialize... but currently this is not the case because `shutdown()` exits early - without calling `shutdownInternal()` - when not initialized)
 
but I'll add a check just to be sure

PR #2955, opened 2021-03-30
GH-2957 Optimize RDFJSONWriter to stream output

tokovach:
GitHub issue resolved: #2957 

Reimplemented RDFJSONWriter to stream output rather than storing all statements consumed in a TreeModel before writing output.

### DIFF
--- a/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFJSONWriterSettings.java
+++ b/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFJSONWriterSettings.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.rio.helpers;
+
+import org.eclipse.rdf4j.rio.RioSetting;
+
+/**
+ * A selection of writer settings specific to RDF/JSON parsers.
+ * <p>
+ * Settings can be overridden by means of a system property, but only if specified at JVM startup time.
+ *
+ * @author Tomas Kovachev t.kovachev1996@gmail.com
+ */
+public class RDFJSONWriterSettings {
+
+	/**
+	 * Boolean setting for RDF/JSON writer to determine whether it should stream output which can result in duplicate
+	 * object keys. By default, the writer sorts output in-memory first to avoid duplicate keys.
+	 * <p>
+	 * Defaults to false.
+	 * <p>
+	 * Can be overridden by setting system property {@code org.eclipse.rdf4j.rio.rdfjson.allow_multiple_object_values}.
+	 */
+	public static final RioSetting<Boolean> ALLOW_MULTIPLE_OBJECT_VALUES = new BooleanRioSetting(
+			"org.eclipse.rdf4j.rio.rdfjson.allow_multiple_object_values", "Allow multiple object values",
+			Boolean.FALSE);
+
+	private RDFJSONWriterSettings() {
+	}
+
+}
--- a/core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONWriter.java
+++ b/core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONWriter.java
@@ -13,6 +13,7 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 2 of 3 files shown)


jeenbroekstra (commented):
Thanks @tokovach! Looks good to me, code-wise. One tiny book-keeping request: I've created an issue for this fix (GH-2957). Could you change your commit message (with a `git commit --amend` perhaps) to use that issue number rather than the internal GDB JIRA number?

tokovach:
Thank you for reviewing the PR @jeenbroekstra! Just updated the commit message to use the issue number. Please let me know should there be any other issue.

hmottestad (approved):
Looks very clean. Easy to follow :)

hmottestad:
One question actually, does this require the input stream of statements to be sorted?

tokovach:
Nope, it is not required. The reason why a TreeModel was used was to reduce the potential size of the JSON file as you would not have duplicate subject JSON objects and subsequently predicate array JSON objects if there are any.

hmottestad:
Ok. So now we will get duplicate subject json objects?

tokovach:
That would depend on the order of statements passed to the writer. If two or more consecutive statements share the same subject, it would not create duplicate subject JSON objects. The same goes for consecutive statements with identical subjects and predicates, where the objects would share the same predicate array JSON object. This design is similar to the one used for the Turtle writer.

hmottestad:
Should our RDF JSON output follow the W3C Working Group Note? https://www.w3.org/TR/rdf-json/

That note specifically states that we need to group all statements by subject and for each unique subject we should create a JSON object.

tokovach:
Both approaches have their pros/cons. With the streaming approach, we still group the statements but only when they are consecutive. The streaming approach broadens the use case of the writer (as now you can process tens of millions of statements without needing more than 4GB of heap) and improves the performance as the number of statement iterations is halved and we do not build an entire TreeModel.

hmottestad:
Would we not end up with duplicate keys in our JSON?

jeenbroekstra:
I see the attraction of being able to do streaming serialization but unfortunately, like @hmottestad says, the spec for the format prohibits this, as it requires that all keys are unique. 

That being said, perhaps we can enable this by means of a configuration flag? So it wouldn't be the default behavior, but it can be explicitly enabled if needed. 

Plus I would like to see how our own RDF/JSON parser actually handles ingesting data with duplicate keys...

tokovach:
@hmottestad , @jeenbroekstra should this configuration flag be used as part of BasicWriterSettings or should it be considered as a general property that is set beforehand? I will add test cases to cover the Parser and make it a flag-set functionality. Thank you for the review!

jeenbroekstra:
> @hmottestad , @jeenbroekstra should this configuration flag be used as part of BasicWriterSettings or should it be considered as a general property that is set beforehand? I will add test cases to cover the Parser and make it a flag-set functionality. Thank you for the review!

I'd make it part of the WriterSettings, but perhaps not BasicWriterSettings as this is very specific to the RDF-JSON format - let's introduce a new class `RDFJSONWriterSettings` (the counterpart to `RDFJSONParserSettings`) for this. Let's call it `ALLOW_MULTIPLE_OBJECT_VALUES` or something along those lines. Defaults to `false`, but if set to true allows using the new streaming approach.

Btw the corresponding parser setting is `RDFJSONParserSettings.FAIL_ON_MULTIPLE_OBJECT_VALUES`, which by default is true (to be spec-conformant), but which can be disabled to deal with files with duplicate keys - so it looks as if we already have this covered at the parser end :)

tokovach:
Great! I will make the necessary changes, thanks for the input :)

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFJSONWriterSettings.java
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: That copyright header needs updating. First line should read

> Copyright (c) 2021 Eclipse RDF4J contributors.

jeenbroekstra (commented):
Looks good, just one minor editorial comment

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFJSONWriterSettings.java
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: Um, no, sorry, it's "Copyright (c) 2021 Eclipse RDF4J contributors." 

Remove the bit about Aduna and others - that's only used in legacy code :)

core/rio/rdfjson/src/main/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONWriter.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * Copyright (c) 2021 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: Same here.

core/rio/rdfjson/src/test/java/org/eclipse/rdf4j/rio/rdfjson/RDFJSONStreamingWriterTest.java
```diff
@@ -0,0 +1,50 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: ...and here.

jeenbroekstra:
Seeing a test failure, but it's in a place that at first glance is not influenced by your changes - might be a temporary glitch.

core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/RDFJSONWriterSettings.java
jeenbroekstra: See also https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md#code-formatting

jeenbroekstra (approved):
Very nice, thanks @tokovach !

Issue #2963, opened 2021-04-03
SHACL - SHACL validation using only SPARQL for simple constraint components


hmottestad:
For something like: 

```
ex:PersonShape
	a sh:NodeShape  ;
	sh:targetClass ex:Person ;
	sh:property  ex:PersonPropertyShape .

ex:PersonPropertyShape sh:path ex:age ;
                       		sh:datatype xsd:integer .
```

We would query for invalid data with the following query:

```
select distinct ?target_0000000000 ?value {
    ?target_0000000000 a <http://example.com/ns#Person> .
    ?target_0000000000 <http://example.com/ns#age> ?value .
    FILTER(!isLiteral(?value) || datatype(?value) != <http://www.w3.org/2001/XMLSchema#integer>)
}
```

#### Scope
 - Support all simple constraint components (those implemented by `SimpleAbstractConstraintComponent` where it's enough to validate a single value at a time
 - Support nested `sh:property` and `sh:node`

hmottestad:
Performance before:

```
Benchmark                              Mode  Cnt     Score      Error  Units
ClassBenchmarkEmpty.shaclBulk          avgt   10  2947.908 ±   79.024  ms/op
DatatypeBenchmarkEmpty.shaclBulk       avgt   10  1355.371 ±   79.050  ms/op
HasValueBenchmarkEmpty.shaclBulk       avgt   10  5169.209 ±  113.416  ms/op
LanguageInBenchmarkEmpty.shaclBulk     avgt   10  3654.121 ±  212.149  ms/op
MaxCountBenchmarkEmpty.shaclBulk       avgt   10  1289.769 ±   40.286  ms/op
MinCountBenchmarkEmpty.shaclBulk       avgt   10  1502.615 ±   50.833  ms/op
NotClassBenchmarkEmpty.shaclBulk       avgt   10  8816.199 ±  330.436  ms/op
NotMaxCountBenchmarkEmpty.shaclBulk    avgt   10  1330.465 ±    8.435  ms/op
NotUniqueLangBenchmarkEmpty.shaclBulk  avgt   10  6372.431 ± 1108.434  ms/op
ValueInBenchmarkEmpty.shaclBulk        avgt   10  6418.861 ±  389.547  ms/op
```

PR #2964, opened 2021-04-03
GH-2963 SPARQL based SHACL validation

hmottestad:
GitHub issue resolved: #2963 

 - generate SPARQL queries for simple constraint components

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -19,7 +19,6 @@ import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.Callable;
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -177,6 +177,9 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	private static final Model DASH_CONSTANTS;
 
+	// Temporary field used to control if the new SPARQL based validation should be enabled!
+	final boolean experimentalSparqlValidation;
+
 	/**
 	 * an initialized {@link Repository} for storing/retrieving Shapes data
 	 */
@@ -242,6 +245,9 @@ public class ShaclSail extends NotifyingSailWrapper {
 		ReferenceQueue<ShaclSail> objectReferenceQueue = new ReferenceQueue<>();
 		startMonitoring(objectReferenceQueue, new PhantomReference<>(this, objectReferenceQueue), initialized,
 				executorService);
+		this.experimentalSparqlValidation = "true"
+				.equalsIgnoreCase(System.getProperty("org.eclipse.rdf4j.sail.shacl.experimentalSparqlValidation"));
+
 	}
 
 	public ShaclSail() {
@@ -249,6 +255,9 @@ public class ShaclSail extends NotifyingSailWrapper {
 		ReferenceQueue<ShaclSail> objectReferenceQueue = new ReferenceQueue<>();
 		startMonitoring(objectReferenceQueue, new PhantomReference<>(this, objectReferenceQueue), initialized,
 				executorService);
+		this.experimentalSparqlValidation = "true"
+				.equalsIgnoreCase(System.getProperty("org.eclipse.rdf4j.sail.shacl.experimentalSparqlValidation"));
+
 	}
 
 	// This is used to keep track of the current connection, if the opening and closing of connections is done serially.
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ConstraintComponent.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ConstraintComponent.java
@@ -1,6 +1,5 @@
 package org.eclipse.rdf4j.sail.shacl.ast.constraintcomponents;
 
-import java.util.Set;
 import java.util.stream.Stream;
 
 import org.eclipse.rdf4j.sail.shacl.ConnectionsGroup;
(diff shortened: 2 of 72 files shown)


jeenbroekstra (commented):
I've taken a look and I don't notice any major problems, but i'm having a hard time following what's going on from just looking at the diff. I'll try and find some time later in the week to check out the branch and look at it locally.

hmottestad:
The core concept is that each constraint component returns a ValidationQuery with the full query body, a list of variables and two integers to keep track of which variable is the target and which is the value. 

As the ValidationQuery gets passed up the chain of PropertyShapes and NodeShapes we decrement the integers. At some point we decide which variables will be used for the validation report and set these as a sort of snapshot. This can also be changed later on as we "change our mind" on which variables should be part of the report. 

I'm not sure how this will extend to things like sh:not or sh:or. However I'm not too worried right now. Next steps will be to support sh:minCount, sh:maxCount and sh:class.

hmottestad:
Should really make some benchmarks though and rebase them to be the first commit. The benchmarks I have currently don't cover this, but could probably easily be extended to cover it.

hmottestad:
Todo:

- [ ] Check if we are handling sh:deactivated correctly

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ConstraintComponent.java
```diff
@@ -11,21 +10,22 @@
...
-	PlanNode generateSparqlValidationPlan(ConnectionsGroup connectionsGroup, boolean logValidationPlans,
+	ValidationQuery generateSparqlValidationQuery(ConnectionsGroup connectionsGroup, boolean logValidationPlans,
 			boolean negatePlan, boolean negateChildren, Scope scope);
 
 	PlanNode generateTransactionalValidationPlan(ConnectionsGroup connectionsGroup,
 			boolean logValidationPlans, PlanNodeProvider overrideTargetNode,
 			Scope scope);
 
-	ValidationApproach getPreferedValidationApproach();
+	ValidationApproach getPreferredValidationApproach(ConnectionsGroup connectionsGroup);
```
hmottestad, line 26: Needs docs

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/ConstraintComponent.java
```diff
@@ -11,21 +10,22 @@
...
 
-	Set<ValidationApproach> getSupportedValidationApproaches();
+	ValidationApproach getOptimalBulkValidationApproach();
```
hmottestad, line 28: Needs docs

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/SimpleAbstractConstraintComponent.java
```diff
@@ -76,69 +76,58 @@ public PlanNode generateTransactionalValidationPlan(ConnectionsGroup connections
...
-		String query = effectiveTarget.getQuery(false);
+		return new ValidationQuery(query, allTargetVariables, value, scope, getConstraintComponent(), null, null);
 
-		StatementMatcher.Variable targetVar = effectiveTarget.getTargetVar();
+	}
 
-		Optional<String> pathQuery = targetChain.getPath()
-				.map(p -> p.getTargetQueryFragment(effectiveTarget.getTargetVar(), value,
-						connectionsGroup.getRdfsSubClassOfReasoner()));
+	private String getSparqlFilter(boolean negatePlan, StatementMatcher.Variable variable) {
```
hmottestad, line 115: Needs docs

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/SimpleAbstractConstraintComponent.java
```diff
@@ -76,69 +76,58 @@ public PlanNode generateTransactionalValidationPlan(ConnectionsGroup connections
...
 
-		if (pathQuery.isPresent()) {
-			query += "\n" + pathQuery.get();
-			query += "\n FILTER(" + getSparqlFilterExpression(value.getName(), negated) + ")";
-		} else {
-			query += "\n FILTER(" + getSparqlFilterExpression(targetVar.getName(), negated) + ")";
-		}
+		String varName = variable.getName();
+		String tempVar = "?" + UUID.randomUUID().toString().replace("-", "");
```
hmottestad, line 118: Can we use something other than uuid here?

hmottestad:
### Performance before:

```
Benchmark                              Mode  Cnt     Score      Error  Units
ClassBenchmarkEmpty.shaclBulk          avgt   10  2947.908 ±   79.024  ms/op
DatatypeBenchmarkEmpty.shaclBulk       avgt   10  1355.371 ±   79.050  ms/op
HasValueBenchmarkEmpty.shaclBulk       avgt   10  5169.209 ±  113.416  ms/op
LanguageInBenchmarkEmpty.shaclBulk     avgt   10  3654.121 ±  212.149  ms/op
MaxCountBenchmarkEmpty.shaclBulk       avgt   10  1289.769 ±   40.286  ms/op
MinCountBenchmarkEmpty.shaclBulk       avgt   10  1502.615 ±   50.833  ms/op
NotClassBenchmarkEmpty.shaclBulk       avgt   10  8816.199 ±  330.436  ms/op
NotMaxCountBenchmarkEmpty.shaclBulk    avgt   10  1330.465 ±    8.435  ms/op
NotUniqueLangBenchmarkEmpty.shaclBulk  avgt   10  6372.431 ± 1108.434  ms/op
ValueInBenchmarkEmpty.shaclBulk        avgt   10  6418.861 ±  389.547  ms/op
```


### After:

```
Benchmark                              Mode  Cnt     Score     Error  Units
ClassBenchmarkEmpty.shaclBulk          avgt   10    97.436 ±   0.405  ms/op  <--
DatatypeBenchmarkEmpty.shaclBulk       avgt   10    50.704 ±   0.421  ms/op  <--
HasValueBenchmarkEmpty.shaclBulk       avgt   10   130.049 ±   2.503  ms/op  <--
LanguageInBenchmarkEmpty.shaclBulk     avgt   10   144.706 ±   1.860  ms/op  <--
MaxCountBenchmarkEmpty.shaclBulk       avgt   10    66.679 ±   1.089  ms/op  <--
MinCountBenchmarkEmpty.shaclBulk       avgt   10    43.574 ±   0.419  ms/op  <--
NotClassBenchmarkEmpty.shaclBulk       avgt   10  8257.425 ± 140.511  ms/op
NotMaxCountBenchmarkEmpty.shaclBulk    avgt   10  1400.417 ±   8.859  ms/op
NotUniqueLangBenchmarkEmpty.shaclBulk  avgt   10  6193.996 ±  33.848  ms/op
ValueInBenchmarkEmpty.shaclBulk        avgt   10  3287.959 ±  15.009  ms/op
```

hmottestad:
Constraints that are left:
 - unique lang
 - dash values in
 - all the logical constraints like or, and, not (but these I don't think I'll fit in this PR)

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/SimpleAbstractConstraintComponent.java
hmottestad: I don't think I will try to deal with this.

jeenbroekstra:
Er, wow. Those benchmark improvements are massive. 

> I think it would also be best to make this feature experimental before we release it, so it would need to be enabled. It would suck if I had made a mistake and it allowed invalid data to pass validation when validated with the SPARQL part.

Makes sense.

hmottestad:
There are still two things that I have not completed. Fix for high minCount and high maxCount. Implement SPARQL based validation for dash:valuesIn. 

I don't think I will fix these two things here now, especially since I made a system property you need to set to enable the SPARQL based validation anyway.

I'm getting ready to merge this now @jeenbroekstra do you have any comments before I do?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -243,13 +245,19 @@ public ShaclSail(NotifyingSail baseSail) {
 		ReferenceQueue<ShaclSail> objectReferenceQueue = new ReferenceQueue<>();
 		startMonitoring(objectReferenceQueue, new PhantomReference<>(this, objectReferenceQueue), initialized,
 				executorService);
+		this.experimentalSparqlValidation = "true"
+				.equalsIgnoreCase(System.getProperty("org.eclipse.rdf4j.sail.shacl.experimentalSparqlValidation"));
```
hmottestad, line 249: The system property I choose. I prefer using a system property rather than adding a new message, since this is very temporary. That way it's easy to just remove without breaking anything for anyone.￼￼

PR #2971, opened 2021-04-07
main shacl tests are not integration tests

hmottestad:
GitHub issue resolved: # 

Main SHACL test suite has been marked as integration tests even though they are the main unit tests for the ShaclSail that checks that constraints work.

----

### DIFF
--- a/core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ShaclTest.java
+++ b/core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ShaclTest.java
@@ -18,9 +18,9 @@ import org.junit.runners.Parameterized;
  * @author Håvard Ottestad
  */
 @RunWith(Parameterized.class)
-public class ShaclIT extends AbstractShaclTest {
+public class ShaclTest extends AbstractShaclTest {
 
-	public ShaclIT(String testCasePath, String path, ExpectedResult expectedResult, IsolationLevel isolationLevel) {
+	public ShaclTest(String testCasePath, String path, ExpectedResult expectedResult, IsolationLevel isolationLevel) {
 		super(testCasePath, path, expectedResult, isolationLevel);
 	}
 
@@ -31,6 +31,10 @@ public class ShaclIT extends AbstractShaclTest {
 
 	@Test
 	public void testSingleTransaction() {
+		// we don't need to run this test for every isolation level
+		if (isolationLevel != IsolationLevels.NONE) {
+			return;
+		}
 		runWithAutomaticLogging(() -> runTestCaseSingleTransaction(testCasePath, path, expectedResult, isolationLevel));
 	}
 


jeenbroekstra:
Arguably they _are_ integration tests in that they execute multiple test cases against a real MemoryStore - they're not proper unit tests in that respect. But semantics aside (and we aren't always very careful about this in other place either to be honest), the main reason to mark them as integration tests is that they take over 2 minutes to complete, and moving that time to the integration testing phase makes the parallel CI faster. 

It's only 2 minutes so I have no massive objections to naming them back, but: what problem are you having with executing these as integration tests?

hmottestad:
Integration tests aren't required to pass before merging, and this is the core test suite for the ShaclSail.

hmottestad:
In this run now they took 1:40 to run.

jeenbroekstra:
> Integration tests aren't required to pass before merging, and this is the core test suite for the ShaclSail.

So? You know when you've made changes to the ShaclSail, you'll have to wait for the integration tests to succeed, right? 

The point of the setup is that for those PRs where nothing was changed in the ShaclSail, we can (optionally) choose to merge quickly.

jeenbroekstra:
As an aside I'm perfectly fine with making the integration testing job required before merging as well if you prefer. I treat it as required for pretty much anything except the most trivial editorial changes already anyway.

jeenbroekstra:
> In this run now they took 1:40 to run.

Yes, in the grand scheme of things it's not a lot either way. But there's still the fact that technically, these tests _are_ integration tests, not unit tests. That doesn't make them less important though (quite the reverse I'd say). It was never my intent that integration tests can be safely skipped before merging.

hmottestad:
That's not really a definition of integration and unit tests that I would use. The intention of the ShaclTest class is to test various shapes and transaction scenarios. It only indirectly tests the integration with the MemoryStore, the SPARQL engine and Rio. The intention is to test shapes. Mocking the entire Sail backend, SPARQL engine and Rio code would be overkill when we can just simply use that code.

If we are going to be that strict on the difference between integration and unit tests then things like `SparqlRegexTest`, `StoreSerializationTest`, `FederationConnectionTest` and a lot more should be marked as integration tests.  

Things like `MultithreadedMemoryStoreIT` though on the other hand specifically test how the ShaclSail acts when integrated with the MemoryStore under a multithreaded workload. That is why we also have `MultithreadedNativeStoreIT` and `MultithreadedNativeStoreRDFSIT`.

hmottestad:
I could split out the different isolation levels into their own suite marked as slow, so that the main tests only run with NONE level.

jeenbroekstra:
> That's not really a definition of integration and unit tests that I would use. The intention of the ShaclTest class is to test various shapes and transaction scenarios. It only indirectly tests the integration with the MemoryStore, the SPARQL engine and Rio. The intention is to test shapes. Mocking the entire Sail backend, SPARQL engine and Rio code would be overkill when we can just simply use that code.

I'm not worried about the use of Rio to read in fixtures - that's fine. It's specifically the use of the MemoryStore that in my eyes makes this an integration test suite, though I'll grant you that if you squint a bit and say "but the memorystore is only used in the role of a stub component so that we don't have to _really_ mock/stub things", you could get away with it. In fact we do that, in several places.

> If we are going to be that strict on the difference between integration and unit tests then things like SparqlRegexTest, StoreSerializationTest, FederationConnectionTest and a lot more should be marked as integration tests

Probably, if we're strict. I'm not really suggesting that we really should always be this strict. But the title of your pull request is "main shacl tests are not integration tests", which is a strong statement that I wanted to examine a little.

> I could split out the different isolation levels into their own suite marked as slow, so that the main tests only run with NONE level.

That might be an option worth keeping in reserve if this suite starts to grow larger (and more costly) again, but for now let's keep things simple.

You've about halfway convinced me that these should "count" as unit tests, so just go ahead and merge this - could you fix the commit  message with an issue number though? No need for a new ticket, the original issue on which we did the test suite/CI refactor might be a good one (even if it's closed).

jeenbroekstra:
For what it's worth, after I've collected my thoughts on this a bit more I'll write something up about how we should generally treat the distinction between different test types (and why it matters), I think this may be useful to other contributors as well.

jeenbroekstra (approved):

Issue #2967, opened 2021-04-06
NativeStore SailSourceModel NullPointerException on remove in long transaction


jeenbroekstra:
Reported on user list (see https://groups.google.com/g/rdf4j-users/c/NaPslvPoS6k/m/1HVFFyN2CQAJ ): on a long-running transaction a NullPointerException can occur in SailSourceModel:

```
java.lang.NullPointerException: null
    at org.eclipse.rdf4j.sail.nativerdf.SailSourceModel.remove(SailSourceModel.java:291)
    at org.eclipse.rdf4j.sail.nativerdf.MemoryOverflowModel.remove(MemoryOverflowModel.java:134)
    at org.eclipse.rdf4j.model.impl.AbstractModel.remove(AbstractModel.java:171)
    at org.eclipse.rdf4j.sail.base.Changeset.deprecate(Changeset.java:286)
    at org.eclipse.rdf4j.sail.base.SailSourceBranch.flush(SailSourceBranch.java:487)
    at org.eclipse.rdf4j.sail.base.SailSourceBranch.compressChanges(SailSourceBranch.java:325)
    at org.eclipse.rdf4j.sail.base.SailSourceBranch$2.close(SailSourceBranch.java:215)
    at org.eclipse.rdf4j.sail.base.SailSourceConnection.endUpdateInternal(SailSourceConnection.java:646)
    at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.endUpdate(AbstractSailConnection.java:535)
    at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.flush(AbstractSailConnection.java:351)
    at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.flushPendingUpdates(AbstractSailConnection.java:827)
    at org.eclipse.rdf4j.sail.helpers.AbstractSailConnection.addStatement(AbstractSailConnection.java:443)
    at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.addWithoutCommit(SailRepositoryConnection.java:383)
    at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.add(AbstractRepositoryConnection.java:428)
    ...
```

To quote:

[...]

PR #2977, opened 2021-04-08
GH-2967 fix nulllpointerexception in SailSourceModel.remove

jeenbroekstra:
GitHub issue resolved: #2967  

- added unit test that reproduces the bug
- added additional unit tests on SailSourceModel to have better coverage/compliance
- fixed by making sure private `sink` field is not directly referenced, but only via the `sink()` method

----

### DIFF
--- a/core/sail/nativerdf/pom.xml
+++ b/core/sail/nativerdf/pom.xml
@@ -85,6 +85,12 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-model-testsuite</artifactId>
+			<version>${project.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.openjdk.jmh</groupId>
 			<artifactId>jmh-core</artifactId>
 			<version>${jmhVersion}</version>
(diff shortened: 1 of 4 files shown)


hmottestad:
Should the sink variable be volatile?

jeenbroekstra:
> Should the sink variable be volatile?

I don't think so, as far as I can tell all access to it is via synchronized methods, so making it volatile won't add anything - unless I missed something. And anyway the issue we're solving here has nothing to do with multithreading.

hmottestad:
You are right. I had to read up more on how synchronized works on the memory model in java and you are correct.

ArjohnKampman:
SailSourceModelTest.testRemoveTermIteration() always fails for me due to a ClosedByInterruptException. Thread's interrupted status prevents the thread from writting the txn-status file.

jeenbroekstra:
> SailSourceModelTest.testRemoveTermIteration() always fails for me due to a ClosedByInterruptException. Thread's interrupted status prevents the thread from writting the txn-status file.

That's peculiar - the test succeeds for me (and on our CI as well). How are you executing it?

ArjohnKampman:
Running from IntelliJ. Will try from command line now.

jeenbroekstra:
@hmottestad could you check this out and run the test in IntelliJ, see if you get the same problem? I'd like to get this merged but if there's something flaky about the test....

jeenbroekstra:
I will merge this fix by the end of the day unless someone has a good reason I shouldn't.

ArjohnKampman:
This change seems to have fixed the NPE that I ran into. I'm still running into out-of-memory-errors, but I guess that's a topic for another ticket.

Issue #2985, opened 2021-04-11
SPARQLParser adds unnecessary extension elements into algebra tree


jeenbroekstra:
Consider this query:

```sparql
select ?b1 ?b2 ?b3
where {  bind(1 as ?b1) }
```

The algebra tree produced for this by the parser is this:

```
Projection
   ProjectionElemList
      ProjectionElem "b1"
      ProjectionElem "b2"
      ProjectionElem "b3"
   Extension
      ExtensionElem (b1)
         Var (name=b1)
      ExtensionElem (b2)
         Var (name=b2)
      ExtensionElem (b3)
         Var (name=b3)
      Extension
         ExtensionElem (b1)
            ValueConstant (value="1"^^<http://www.w3.org/2001/XMLSchema#integer>)
         SingletonSet
```

The first three extension elements in this are redundant. The correct algebra tree should be :

```
Projection
   ProjectionElemList
      ProjectionElem "b1"
      ProjectionElem "b2"
      ProjectionElem "b3"
   Extension
      ExtensionElem (b1)
         ValueConstant (value="1"^^<http://www.w3.org/2001/XMLSchema#integer>)
      SingletonSet
```

While the additional extension elements make no difference for the result of the query, they are redundant and slow down processing. They also make it harder to interpret the algebra tree and to implement things like re-rendering into a SPARQL string.

PR #2986, opened 2021-04-11
GH-2985 remove unnecessary addition of extension elems

jeenbroekstra:
GitHub issue resolved: #2985  

- removed unnecessary creation of extension elements for variables that do not appear in WHERE clause
- added test case
- also did some deprecation cleanup

----

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
@@ -601,15 +601,6 @@ public class TupleExprBuilder extends AbstractASTVisitor {
 				Var projVar = (Var) child.jjtAccept(this, null);
 				ProjectionElem elem = new ProjectionElem(projVar.getName());
 				projElemList.addElement(elem);
-
-				VarCollector whereClauseVarCollector = new VarCollector();
-				result.visit(whereClauseVarCollector);
-
-				if (!whereClauseVarCollector.collectedVars.contains(projVar)) {
-					ExtensionElem extElem = new ExtensionElem(projVar, projVar.getName());
-					extension.addElement(extElem);
-					elem.setSourceExpression(extElem);
-				}
 			} else {
 				throw new IllegalStateException("required alias for non-Var projection elements not found");
 			}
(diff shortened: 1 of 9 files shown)


jeenbroekstra:
@avataar @damyan-ognyanov I'd appreciate your eyes on this as well. The change breaks nothing in our compliance tests as far as I can see, but I'd like confirmation that this won't mess anything up in GraphDB either.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
```diff
@@ -601,15 +601,6 @@ public TupleExpr visit(ASTSelect node, Object data) throws VisitorException {
...
-
-				VarCollector whereClauseVarCollector = new VarCollector();
-				result.visit(whereClauseVarCollector);
-
-				if (!whereClauseVarCollector.collectedVars.contains(projVar)) {
-					ExtensionElem extElem = new ExtensionElem(projVar, projVar.getName());
-					extension.addElement(extElem);
-					elem.setSourceExpression(extElem);
-				}
 			} else {
```
jeenbroekstra, line 604: Removing the lines above here is the core of this fix. Everything else in the diff (apart from the added test) is just cleanup.

jeenbroekstra:
Of course, it breaks a few tests seconds after me saying it wouldn't. The failing tests are specific to SPARQL-star and also _specifically_ seem to be about expecting redundant/nested extension elements. I think we might just adjust those test to remove the redundancy.

hmottestad (commented):
Not really sure what this entails. Would this potentially affect any of the SPARQL optimizers? (eg. do any of them rely on this extension?)

jeenbroekstra:
> Not really sure what this entails. Would this potentially affect any of the SPARQL optimizers? (eg. do any of them rely on this extension?)

It should not. First of all the extension elements eliminated here are all redundant, basically binding a variable to itself. Second as far as I'm aware we have no optimisers in the pipeline that actually do anything with extensions, specifically.

hmottestad:
Was it there to handle that you can use AS in the variable clause in select? Eg. SELECT ?a as ?b ...; or something like that.

jeenbroekstra:
> Was it there to handle that you can use AS in the variable clause in select? Eg. SELECT ?a as ?b ...; or something like that.

No, that's handled by a different part of the parser. As an example, this query:

    select (?foo as ?bar) where { ?foo <foo:bar> "bar" }

Translates to the following algebra:

```
Projection
   ProjectionElemList
      ProjectionElem "foo" AS "bar"
   Extension
      ExtensionElem (bar)
         Var (name=foo)
      StatementPattern
         Var (name=foo)
         Var (name=_const_d75d935f_uri, value=foo:bar, anonymous)
         Var (name=_const_17c13_lit_e2eec718_0, value="bar", anonymous)
```

This hasn't changed by this fix.

jeenbroekstra:
Unless I get any further feedback on this, I will merge this by the end of the day.

Issue #2993, opened 2021-04-12
Improve performance of NTriples and Nquads parsers


tokovach:
The current implementation of NTriples and NQuads uses PushbackReader when the format allows for the much more efficient use of BufferedReader and readLine(). To enhance performance, this should be modified to use BufferedReader and readLine().

tokovach:
With the benchmarks introduced from the second PR, here is the comparison for NQuads parser with 1m statements:
BEFORE:
```
Benchmark                                  Mode  Cnt  Score   Error  Units
NQuadsParserBenchmark.parseBufferedReader  avgt    3  3.824 ± 0.128   s/op
NQuadsParserBenchmark.parseStream          avgt    3  6.423 ± 0.200   s/op
```
AFTER (same test twice as new impl uses buffered reader anyway):
```
Benchmark                                  Mode  Cnt  Score   Error  Units
NQuadsParserBenchmark.parseBufferedReader  avgt    3  2.149 ± 0.333   s/op
NQuadsParserBenchmark.parseStream          avgt    3  2.241 ± 0.254   s/op 
```

PR #2994, opened 2021-04-12
GH-2993 Refactored NTriples/NQuads Parser to use buffering

tokovach:
Signed-off-by: Tomas Kovachev <t.kovachev1996@gmail.com>

GitHub issue resolved: #2993 

Refactored NTriples/NQuads Parser to use a buffered reader for processing input to improve performance.

### DIFF
--- a/core/rio/nquads/src/main/java/org/eclipse/rdf4j/rio/nquads/NQuadsParser.java
+++ b/core/rio/nquads/src/main/java/org/eclipse/rdf4j/rio/nquads/NQuadsParser.java
@@ -7,17 +7,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.rio.nquads;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.PushbackReader;
-import java.io.Reader;
-import java.io.UnsupportedEncodingException;
-import java.nio.charset.StandardCharsets;
-
-import org.apache.commons.io.input.BOMInputStream;
 import org.eclipse.rdf4j.model.Resource;
-import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.ValueFactory;
 import org.eclipse.rdf4j.rio.RDFFormat;
 import org.eclipse.rdf4j.rio.RDFHandlerException;
 import org.eclipse.rdf4j.rio.RDFParseException;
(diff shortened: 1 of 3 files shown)


hmottestad:
The Eclipse ECA check is complaining about the following: 

> Anonymous (tomas.kovachev@****.com) is not covered by the necessary legal agreements in order to proceed.

core/rio/nquads/src/main/java/org/eclipse/rdf4j/rio/nquads/NQuadsParser.java
```diff
@@ -35,143 +26,82 @@
...
-
-	@Override
-	public synchronized void parse(final InputStream inputStream, final String baseURI)
-			throws IOException, RDFParseException, RDFHandlerException {
-		if (inputStream == null) {
-			throw new IllegalArgumentException("Input stream can not be 'null'");
-		}
+	/*--------------*
+	 * Constructors *
+	 *--------------*/
```
jeenbroekstra, line 31: Fwiw we don't really use these kinds of markers anymore as they do more harm than good. Not a big deal, you can leave it in here, but no need to add this in for any future contributions :)

jeenbroekstra (commented):
I haven't looked in detail yet but at first glance a good improvement.

@tokovach are you familiar with JMH benchmarks at all? It might be good to use this opportunity to add one or two micro-benchmarks for parser performance and measure some before- and after figures, just so we have numbers to back up that this is more performant.

core/rio/nquads/src/main/java/org/eclipse/rdf4j/rio/nquads/NQuadsParser.java
tokovach: I don't like these markers either, I just noticed they are present for a number of parsers/writers and wanted to be consistent. But I will remove them :)

tokovach:
I am familiar with JMH, I have created an entire module for parser benchmarking, which as discussed with @hmottestad I will submit from a separate PR. Should I link the said PR to this issue?

jeenbroekstra:
> I am familiar with JMH, I have created an entire module for parser benchmarking, which as discussed with @hmottestad I will submit from a separate PR. Should I link the said PR to this issue?

Ah excellent! Yeah, you might as well reuse this ticket for that.

hmottestad (approved):
I've had a look, but it's quite a lot of code. I've tried to look at everything, and I can't find anything to fault!

LGTM :)

jeenbroekstra (approved):
LGTM - good improvement, thanks!

PR #2995, opened 2021-04-13
GH-2993 Introduced JMH Benchmarks for main parsers

tokovach:
Signed-off-by: Tomas Kovachev <t.kovachev1996@gmail.com>

GitHub issue resolved: #2993 

Introduced JMH Benchmarks for main parsers as well as a synthetic dataset generator and black hole rdf handler.


testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/ParserBenchmark.java
```diff
@@ -0,0 +1,90 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: Same comment as in the other PR: please change to "Copyright (c) 2021 Eclipse RDF4J contributors."

Sorry to be a bit pedantic about this, but for IP governance this is kinda important.

tokovach: Sorry, I thought I had just the year wrong, will fix it now :)

hmottestad: FYI. The correct header is in the formatting section here: https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md

jeenbroekstra (commented):
Thanks, looks good now!

Could you run them and post some results here? Then we can use that as a baseline to judge your parser optimizations against.

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/ParserBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+	public static final String XMS = "-Xms4g";
+	public static final String XMX = "-Xmx4g";
+	public static final int WARM_UPS = 3;
+	public static final int EXECUTIONS = 3;
+	public static final int TOTAL_STATEMENTS = 1_000_000;
+	public static final int MIN_STRING_LENGTH = 100;
+	public static final int MAX_STRING_LENGTH = 300;
+	public static final int PERCENT_BNODE = 30;
+	public static final int PERCENT_LITERALS = 50;
+	public static final boolean TEXT_ONLY = false;
```
hmottestad, line 48: Could you use annotations instead? I like to run the benchmarks with the JMHack plugin in IntelliJ instead of with a Runner.

Here is how I do annotations: https://github.com/eclipse/rdf4j/blob/main/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/QueryBenchmark.java

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/ParserBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+ * {@link BlackHoleRDFHandler}.
+ *
+ * @author Tomas Kovachev t.kovachev1996@gmail.com
+ */
+@State(Scope.Benchmark)
+public abstract class ParserBenchmark {
+	public static final String XMS = "-Xms4g";
+	public static final String XMX = "-Xmx4g";
+	public static final int WARM_UPS = 3;
+	public static final int EXECUTIONS = 3;
```
hmottestad, line 42: Could you use annotations instead? I like to run the benchmarks with the JMHack plugin in IntelliJ instead of with a Runner.

Here is how I do annotations: https://github.com/eclipse/rdf4j/blob/main/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/QueryBenchmark.java

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/util/DataSetGenerator.java
```diff
@@ -0,0 +1,430 @@
...
+		this.randomResource = new Random(64);
+		this.randomProbability = new Random(64);
+		this.avgWordSize = 0;
+		fetchWords();
+		generatePredicateList();
+	}
+
+	public File generateStatementsFile(RDFFormat format, String fileName, int percentBNodes, int percentLiterals,
+			int textMinLength, int textMaxLength, int totalStatements, boolean isTextOnly,
+			boolean withContext) throws IOException {
```
hmottestad, line 75: This is very nice! Thank you :)

Btw. If you want to parse a real world file I regularly use the two following files:
 - https://github.com/eclipse/rdf4j/blob/main/core/sail/memory/src/test/resources/benchmarkFiles/datagovbe-valid.ttl (remember to also add https://github.com/eclipse/rdf4j/blob/main/core/sail/memory/src/test/resources/benchmarkFiles/LICENSE)
 -  https://github.com/eclipse/rdf4j/blob/75b97f6106e6a73776d7f0f663f18a3edfb0a4ae/compliance/model/src/test/resources/benchmarkFiles/bsbm-100.ttl

You don't have to use these files though. It was just to let you know what we currently have.

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/util/DataSetGenerator.java
```diff
@@ -0,0 +1,430 @@
...
+		return file.exists();
+	}
+
+	private File createFile(RDFFormat format, String fileName) throws IOException {
+		Path path = Paths.get(this.getClass().getProtectionDomain().getCodeSource().getLocation().getPath());
+		path = path.getParent().getParent();
+		File file = new File(path.toString(), DIR + fileName + "." + format.getDefaultFileExtension());
+		assert file.createNewFile();
+		return file;
+	}
```
hmottestad, line 128: Where does this create the file? Would it be better to use `Strings.append(File.separator).to(System.getProperty("java.io.tmpdir"))` ?

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/ParserBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+		if (toReadFrom == null) {
+			// If format supports graphs, they will be included in the dataset
+			toReadFrom = generator.generateStatementsFile(format, fileName, PERCENT_BNODE, PERCENT_LITERALS,
+					MIN_STRING_LENGTH, MAX_STRING_LENGTH, TOTAL_STATEMENTS, TEXT_ONLY, true);
+		}
+	}
+
+	@TearDown(Level.Trial)
+	public void tearDown() {
+		toReadFrom.deleteOnExit();
```
hmottestad, line 69: Wouldn't this make sense to put in the setup method?

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/ParserBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	@Benchmark
+	@BenchmarkMode(Mode.AverageTime)
+	public void parseStream(Blackhole blackhole) throws IOException {
+		parser.setRDFHandler(rdfHandler);
+		rdfHandler.setBlackHoleConsumer((blackhole::consume));
+		parser.parse(new FileInputStream(toReadFrom), DataSetGenerator.NAMESPACE);
```
hmottestad, line 77: We should close our input stream instead of relying on the parse to close it.

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/ParserBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	@Benchmark
+	@BenchmarkMode(Mode.AverageTime)
+	public void parseBufferedReader(Blackhole blackhole) throws IOException {
+		parser.setRDFHandler(rdfHandler);
+		rdfHandler.setBlackHoleConsumer((blackhole::consume));
+		parser.parse(new BufferedReader(new InputStreamReader(new FileInputStream(toReadFrom))),
+				DataSetGenerator.NAMESPACE);
```
hmottestad, line 86: We should close our input stream instead of relying on the parse to close it.

tokovach:
With the benchmarks introduced from the second PR, here is the comparison for NQuads parser with 1m statements:
BEFORE:

```
Benchmark                                  Mode  Cnt  Score   Error  Units
NQuadsParserBenchmark.parseBufferedReader  avgt    3  3.824 ± 0.128   s/op
NQuadsParserBenchmark.parseStream          avgt    3  6.423 ± 0.200   s/op
```
AFTER (same test twice as new implementation uses buffered reader anyway):

```
Benchmark                                  Mode  Cnt  Score   Error  Units
NQuadsParserBenchmark.parseBufferedReader  avgt    3  2.149 ± 0.333   s/op
NQuadsParserBenchmark.parseStream          avgt    3  2.241 ± 0.254   s/op 
```

jeenbroekstra:
That's seconds per op. So quite significant!

tokovach:
Indeed, after resolving the issues I will run the same test again but with 5m statements.

hmottestad:
Just want to test this branch out locally and if everything works we should be good to merge.

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/util/DataSetGenerator.java
```diff
@@ -0,0 +1,420 @@
...
+		randomProbability = new Random(64);
+		return file;
+	}
+
+	private File createFile(RDFFormat format, String fileName) throws IOException {
+		File file = new File(Strings.append(File.separator).to(System.getProperty("java.io.tmpdir")) + fileName + "."
+				+ format.getDefaultFileExtension());
+		assert file.createNewFile();
+		return file;
+	}
```
hmottestad, line 118: Just one last thing @tokovach . Could you create a temp file with a UUID instead of the filename. Since you are passing the file back up the stack all the way to the jmh benchmark code, we don't really need to care what the file is called (eg. you can remove the fileName argument). By using a UUID we will also remove any chance of clashing with an existing file. This dataset generator is very useful and we might end up using it other places, so would be great if it was resilient :)

jeenbroekstra (approved):
This looks good to me, no further comments from my side, but I'll wait with merging until @hmottestad also approves.

jeenbroekstra:
@hmottestad can you indicate if you approve merging this? Or do you need a little more time to review?

testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/util/DataSetGenerator.java
```diff
@@ -0,0 +1,422 @@
...
+		randomResource = new Random(64);
+		randomProbability = new Random(64);
+		return file;
+	}
+
+	private File createFile(RDFFormat format) throws IOException {
+		File file = new File(
+				Strings.append(File.separator).to(System.getProperty("java.io.tmpdir")) + UUID.randomUUID() + "."
+						+ format.getDefaultFileExtension());
+		assert file.createNewFile();
```
hmottestad, line 118: This will only create a new file if asserts are enabled. Asserts should probably be left disabled during benchmarking so we don't skew performance.

hmottestad (changes requested):

hmottestad:
Yeah, give it a shot and see if it helps.

jeenbroekstra:
> I see that there is an unrelated dependency issue, should I rebase the branch?

I'm not seeing any problems, the build succeeds. What am I missing?

hmottestad:
That was on the third rebuild.

Issue #2959, opened 2021-04-01
NativeStore and ES tests: cleanup temp dirs


barthanssens:
Some of the NativeStore / ElasticSearchStore tests create temp dirs which are not deleted when the test ends.
(i.e. not created in a temp dir managed by junit Rule)

The result is a series of leftover directories in temp dir when tests are finished

PR #2997, opened 2021-04-14
GH-2959 use junit4 rule for temporary dirs

barthanssens:

GitHub issue resolved: #2959  

- use a junit4 rule to create temp dirs (instead of relying on custom ways to achieve the same thing) that should be removed after the test
- use random temp dirs, instead of fixed onces (e.g. "nativestore"): this should help running tests in parallel (though currently all tests are performed in series), otherwise different tests may end up using the same subdirectory

Not using junit5 (yet), because it looks like some tests inherit from other junit4 tests outside their own package.

----

### DIFF
--- a/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeStoreTest.java
+++ b/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeStoreTest.java
@@ -35,6 +35,7 @@ public class NativeStoreTest extends RDFNotifyingStoreTest {
 
 	@Rule
 	public TemporaryFolder tempDir = new TemporaryFolder();
+	private File dataDir;
 
 	/*---------*
 	 * Methods *
@@ -43,7 +44,8 @@ public class NativeStoreTest extends RDFNotifyingStoreTest {
 	@Override
 	protected NotifyingSail createSail() throws SailException {
 		try {
-			NotifyingSail sail = new NativeStore(tempDir.newFolder("nativestore"), "spoc,posc");
+			dataDir = tempDir.newFolder();
+			NotifyingSail sail = new NativeStore(dataDir, "spoc,posc");
 			sail.init();
 			return sail;
 		} catch (IOException e) {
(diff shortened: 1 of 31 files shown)


barthanssens:
Nativestore should be ok, next is checking ES store

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeStoreTest.java
```diff
@@ -43,7 +43,7 @@
 	@Override
 	protected NotifyingSail createSail() throws SailException {
 		try {
-			NotifyingSail sail = new NativeStore(tempDir.newFolder("nativestore"), "spoc,posc");
+			NotifyingSail sail = new NativeStore(tempDir.newFolder(), "spoc,posc");
```
jeenbroekstra, line 46: I'm not sure why you changed this, but this change causes a test failure (because the test assumes there is a 'nativestore' subdir)

barthanssens:
Yeah, still somewhat working on it :-/

The reason I've removed this is that the test will indeed create nativestore dir (not a randomized one with this prefix, just "nativestore"), just like other test with nativestore. So if at one point these tests are performed in parallel, they will all use the same subdir...

Alternative would be to use "name-of-test/nativestore", but this may or may not be necessary when we switch to junit5 TempDir (which is still marked "experimental" in Junit 5.7)

jeenbroekstra:
> Alternative would be to use "name-of-test/nativestore", but this may or may not be necessary when we switch to junit5 TempDir (which is still marked "experimental" in Junit 5.7)

I'd be fine with that as a fix, even if in a later Junit 5 migration it's no longer necessary, it won't break anything either. But if you want to explore other possibilities first I have no objection.

barthanssens:
Should fix a few 'leaking' tests, there are still some temp directories left after running the tests:
 
- some directories starting with elasticsearch (probably fixed when upgrading to a new ES - or more precise: using the ES test runner in the PR containing the ES upgrade)
- and some temp dirs created by SHACL tests: these tests may benefit from Junit5 TempDir, since the SHACL tests already use Junit5 (no need to go back to Junit4 Rule)

Issue #3004, opened 2021-04-17
SHACL - rsx:targetShape has severe performance degradation in 3.6.x


hmottestad:
Example shape:

```
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .


<http://example.org/shape/_Character> a sh:NodeShape;
  <http://rdf4j.org/shacl-extensions#targetShape> [ a sh:NodeShape;
      sh:and ([ a sh:NodeShape;
            sh:or ([ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:or ([ a sh:NodeShape;
                        sh:hasValue <https://starwars.org/vocabulary/Droid>
                      ] [ a sh:NodeShape;
                        sh:hasValue <https://starwars.org/vocabulary/DroidAlternative>
                      ])
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/AllHuman>
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/Human>
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/Clone>
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/Slave>
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/CloneSoldier>
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/CloneSlave>
                ] [ a sh:PropertyShape;
                  sh:path rdf:type;
                  sh:hasValue <https://starwars.org/vocabulary/CloneFriend>
                ])
          ])
    ];
  sh:property [ a sh:PropertyShape;
      sh:path <https://starwars.org/vocabulary/bestfriend>;
      sh:maxCount 2
    ] .
```

Even with a small dataset of valid data this shape takes 3 seconds to evaluate on my laptop with `main` while it takes just a few hundred millis in 3.5.1.

PR #3007, opened 2021-04-19
GH-3004 fixed performance issue with rsx:targetShape

### LINKED ISSUE #3013: SHACL - rsx:targetShape bug when using node shapes in sh:and/or
```
ex:shape1
	a sh:NodeShape ;
	rsx:targetShape [
	        sh:property [
        		sh:path ex:info ;
        		sh:hasValue "blue" ; 
        		sh:and ( [ sh:hasValue "blue" ; ] ) ; 
        	]
	];

	sh:property [
	   sh:path ex:info2 ;
	   sh:minCount 1;
	]

	 .
```

Two issues here. First is that `sh:and ( [ sh:hasValue "blue" ; ] )` does not account for objects as targets, only for subjects. Second issue is that `sh:hasValue "blue"` creates a `BIND` clause that must be written to the SPARQL query before the SPARQL generated by `sh:and`.

hmottestad:
GitHub issue resolved: #3004 

 - join the code that generates rsx:targetShape SPARQL and statement matchers
 - add a "previous" field in TargetChainRetriever so that if a statement matchers matches the same target (in a row) we can skip running a SPARQL query
 - reduce statement matchers so that statement matchers that are already covered can be eliminted
 - also resolve #3013 since it was discovered while fixing the performance issues

----


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
```diff
@@ -44,6 +46,73 @@ public StatementMatcher(Variable subject, Variable predicate, Variable object) {
...
+				covers = false;
+			}
+		} else {
+			if (!Objects.equals(objectName, s.objectName)) {
+				covers = false;
+			}
+			if (!Objects.equals(objectValue, s.objectValue)) {
+				covers = false;
+			}
+		}
```
boboetuk, line 110: Hi, Not sure how often this is called but once the `covers` is determined to be false the rest of the checks does not makes sense.

hmottestad: Good point!

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/PropertyShape.java
```diff
@@ -282,44 +283,36 @@ public SparqlFragment buildSparqlValidNodes_rsx_targetShape(StatementMatcher.Var
 		StatementMatcher.Variable someObject = new StatementMatcher.Variable(
 				UUID.randomUUID().toString().replace("-", ""));
 
-		boolean isFilterCondition = constraintComponents.stream()
-				.map(o -> o.buildSparqlValidNodes_rsx_targetShape(object, someObject, rdfsSubClassOfReasoner,
+		List<SparqlFragment> sparqlFragments = constraintComponents.stream()
```
rdstn, line 286: Is it worth pulling those changes into a single class to reduce code duplication?

hmottestad:
@rdstn 

One of the general issues is that for the following shape:

```
ex:shape1
	a sh:NodeShape ;
	rsx:targetShape [
	        sh:property [
        		sh:path ex:info ;
        		sh:and ([ sh:hasValue "blue" ; ]) ;
        	]
	];

	sh:property [
	   sh:path ex:info2 ;
	   sh:minCount 1;
	]

	 .
```

The targetShape matches and subject/object that does not have a value for `ex:info`. Essentially the `sh:and` means "for every value of `ex:info`", so if there are no values for `ex:info` then the `sh:and` clause is actually valid.

You can try the following shape in the SHACL playground (https://shacl.org/playground/)

```
@prefix dash: <http://datashapes.org/dash#> .
@prefix schema: <http://schema.org/> .

schema:PersonShape
    a sh:NodeShape ;
    sh:targetNode "a", schema:test ;
    sh:property [
        sh:path schema:givenName ;
        sh:and ([sh:hasValue "abc"])
    ].
```

hmottestad:
Currently struggling with a more complicated target shape:

```
ex:shape1
	a sh:NodeShape ;
	rsx:targetShape [
	        sh:property [
        		sh:path ex:info ;
        		sh:hasValue "blue" ; 
        		sh:and ( [ sh:hasValue "blue" ; ] ) ; have to be "blue"
        	]
	];

	sh:property [
	   sh:path ex:info2 ;
	   sh:minCount 1;
	]

	 .
```

Which produces the following query:
```
select ?_target0000000000 where {
        
        BIND("blue" as ?9926930321be40178cca7830f83fe037)
        
        ?_target0000000000 <http://example.com/ns#info> ?9926930321be40178cca7830f83fe037 .
        
        {
         	?_target0000000000 <http://example.com/ns#info> ?9926930321be40178cca7830f83fe037 .
        	 FILTER (
        	        ! EXISTS {
        		        ?_target0000000000 <http://example.com/ns#info> ?c0d94ad39ee941e8857511559d54c61c .
        		        FILTER(!( ?9926930321be40178cca7830f83fe037 = "blue" ) )
        	        }
                )
        	
        } UNION {
        
        	 ?_target0000000000 ?1257e2b27eb248b1a9a0e043e40fe210 ?e7cad390b77446e58b6d4d87b48fa758.
        	 FILTER(
        	        NOT EXISTS {
         	                ?_target0000000000 <http://example.com/ns#info> ?5cbb22a8abc146d1a0b4c9e0d67dbe94 .
                        }
                 )
                 
        } UNION {
        
        	 ?a7f66c8e8cb441b0be5c6b36a0bd04ea ?9f9bb841e45947d9ae4d9225a6277205 ?_target0000000000.
        	 FILTER(
        	        NOT EXISTS {
         	                ?_target0000000000 <http://example.com/ns#info> ?5cbb22a8abc146d1a0b4c9e0d67dbe94 .
                        }
                )
                
        }
}

```

But should instead be producing something like:

```
select ?_target0000000000 where {

        {
                BIND("blue" as ?9926930321be40178cca7830f83fe037)
        
                ?_target0000000000 <http://example.com/ns#info> ?9926930321be40178cca7830f83fe037 .

         	?_target0000000000 <http://example.com/ns#info> ?9926930321be40178cca7830f83fe037 .
        	 FILTER (
        	        ! EXISTS {
        		        ?_target0000000000 <http://example.com/ns#info> ?c0d94ad39ee941e8857511559d54c61c .
        		        FILTER(!( ?9926930321be40178cca7830f83fe037 = "blue" ) )
        	        }
                )

        } UNION {

                BIND("blue" as ?9926930321be40178cca7830f83fe037)
        
                ?_target0000000000 <http://example.com/ns#info> ?9926930321be40178cca7830f83fe037 .

        	 ?_target0000000000 ?1257e2b27eb248b1a9a0e043e40fe210 ?e7cad390b77446e58b6d4d87b48fa758.
        	 FILTER(
        	        NOT EXISTS {
         	                ?_target0000000000 <http://example.com/ns#info> ?5cbb22a8abc146d1a0b4c9e0d67dbe94 .
                        }
                 )

        } UNION {
        
                BIND("blue" as ?9926930321be40178cca7830f83fe037)
        
                ?_target0000000000 <http://example.com/ns#info> ?9926930321be40178cca7830f83fe037 .        

        	 ?a7f66c8e8cb441b0be5c6b36a0bd04ea ?9f9bb841e45947d9ae4d9225a6277205 ?_target0000000000.
        	 FILTER(
        	        NOT EXISTS {
         	                ?_target0000000000 <http://example.com/ns#info> ?5cbb22a8abc146d1a0b4c9e0d67dbe94 .
                        }
                )

        }
}

```

hmottestad:
So I was wrong about the unions being the issue. I had actually messed up this part:

```
        	 FILTER (
        	        ! EXISTS {
        		        ?_target0000000000 <http://example.com/ns#info> ?c0d94ad39ee941e8857511559d54c61c .
        		        FILTER(!( ?9926930321be40178cca7830f83fe037 = "blue" ) )
        	        }
                )
```

I use a new object variable for the path query but forgot to change the one in the filter condition. I've made a somewhat hacky string replace solution, which seems to work.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/SparqlFragment.java
```diff
@@ -7,29 +7,175 @@
...
+		if (USE_UNION_PRESERVING_JOIN && sparqlFragments.stream().anyMatch(s1 -> s1.union)) {
+			return unionPreservingJoin(sparqlFragments);
+
+		} else {
+			String collect = sparqlFragments.stream()
+					.peek(s -> {
+						assert !s.filterCondition;
+					})
+					.map(SparqlFragment::getFragment)
+					.map(s -> s.replaceAll("(?m)^", "\t"))
```
boboetuk, line 97: If this is going to stay long term, maybe precompile the pattern

hmottestad: It's to make the query a bit more readable when debugging. I'll compile it, that'll help a lot! Thanks.

hmottestad: Btw. The thing that is currently using the most time is that the generated SPARQL query needs to be optimized and evaluated. That's about 99% of the execution time.

core/sail/shacl/src/test/resources/test-cases/hasValue/targetShapeAnd/invalid/case2/report.ttl
```diff
@@ -4,17 +4,31 @@
 @prefix sh: <http://www.w3.org/ns/shacl#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix rsx: <http://rdf4j.org/shacl-extensions#> .
 
 [] a sh:ValidationReport;
-  sh:conforms false;
   <http://rdf4j.org/schema/rdf4j#truncated> false;
+  sh:conforms false;
```
jeenbroekstra, line 11: What's the reason all these test case files change with almost every PR related to SHACL? I assume there's good reason but I find it a bit disconcerting to have a PR that is supposed to be a single bug fix touch 123 files, and (apparently) adjust test cases. Was the test wrong?

jeenbroekstra:
Do you have any benchmark figures that show how much these changes improve performance?

core/sail/shacl/src/test/resources/test-cases/hasValue/targetShapeAnd/invalid/case2/report.ttl
hmottestad: Tests were wrong yes..

hmottestad:
I didn't make a separat commit for the benchmark this time so I can run a before and after comparison. I do remember the numbers though, used to take around 20 seconds and now down to 250 ms.

core/sail/shacl/src/test/resources/test-cases/hasValue/targetShapeAnd/invalid/case2/report.ttl
hmottestad: Essentially:

```
rsx:targetShape [
	        sh:property [
        		sh:path ex:info ;
        		sh:and ([ sh:hasValue "blue" ; ]) ;
        	]
	];
```

Matches `ex:a` and `"red"` in `ex:a rdfs:label "red"` since neither of those nodes have an `ex:info` predicate. Previously I had only managed to match `ex:a`, even though I had intended to also match `"red"`.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/PropertyShape.java
hmottestad: Since I cleaned up this code quite a bit it has gotten very short and concise. I did split out the code from `sh:and` and `sh:or` so that it could be used by both classes.

jeenbroekstra (approved):

Issue #3008, opened 2021-04-19
Values in path expressions


fragosog:
Hello, we ran into sparql performance issues during testing of rdf4j 3.6.2.  A query like the one below using inlined ```values``` completes quickly and comparable to sesame 2.7.12 (where we are upgrading from),

```
prefix foo:  <http://example.com/voc#>
select (count(distinct ?subject) as ?total_concept_count)
where 
{
  values ?superclass { foo:bar }
  ?subject (rdfs:subClassOf|(owl:equivalentClass/owl:intersectionOf/rdf:rest*/rdf:first) )* ?superclass . 
}
``` 

However, adding more triple patterns to the above query, such as

```
{
  values ?superclass { foo:bar }
  ?subject (rdfs:subClassOf|(owl:equivalentClass/owl:intersectionOf/rdf:rest*/rdf:first) )* ?superclass ;
     owl:equivalentClass ?an .
}
```

results in the query running longer in rdf4j in comparison to sesame 2.7.12, in this particular instance 35 secs vs 190 ms in sesame using a memory store with 2.2 million triples (with 8.5 million triples the query terminates in sesame - just as quickly - but not in rdf4j).

If we eliminate "values" and instead use foo:bar as the object in the query

```
{
  ?subject (rdfs:subClassOf|(owl:equivalentClass/owl:intersectionOf/rdf:rest*/rdf:first) )* foo:bar ;
     owl:equivalentClass ?an .
}
```

 
this query (and others like it that include the same path expression) then terminates in about 7 secs with 8.5 million triples in the memory store.  The use of ```values``` in queries without a path expression finish in about the same time as sesame, so the issue appears to be restricted to its use in a path expression like the ones above.  Hopefully the treatment of ```values``` can be reviewed?

Please let me know if you need additional details.
  
Thank you!

jeenbroekstra:
It sounds like a glitch in the query planner. Could you your query through the [explanation](https://rdf4j.org/documentation/programming/repository/#explaining-queries) tool (using `Optimized` level) on your store and post the outcome here?

fragosog:
Yes, I'm attaching the output of explain for the query with the ```values``` keyword at the ```Unoptimized``` and ```Timed``` levels:  

[values-unoptimized.txt](https://github.com/eclipse/rdf4j/files/6372196/values-unoptimized.txt)
[values-timed.txt](https://github.com/eclipse/rdf4j/files/6372197/values-timed.txt)

[...]

PR #3024, opened 2021-04-27
GH-3008 inline VALUES clause values if possible

jeenbroekstra:
GitHub issue resolved: #3008 

- special-case fix for VALUES clause with a single value for a variable: inline where possible
- added regression test

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/BindingSetAssignmentInliner.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/BindingSetAssignmentInliner.java
@@ -1,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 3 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/BindingSetAssignmentInliner.java
```diff
@@ -0,0 +1,78 @@
...
+
+		private BindingSet bindingSet;
+
+		@Override
+		public void meet(BindingSetAssignment bsa) {
+			List<BindingSet> bindingSets = new ArrayList<>();
+			bsa.getBindingSets().forEach(bindingSets::add);
+
+			if (bindingSets.size() == 1) {
+				bindingSet = bindingSets.get(0);
```
hmottestad, line 47: Can't you just iterate over the bindingsets, cache the first one and then check if there are anymore? Instead of creating an array list every time.

hmottestad (commented):
Would be nice with some more test cases. Maybe a union, nested optional, a filter not exists and maybe also a minus clause?

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/BindingSetAssignmentInliner.java
jeenbroekstra: Yep, good catch, done.

Issue #3076, opened 2021-05-27
RDFWriter benchmarks


frensjan:
It would be good to add benchmarks for RDFWriters next to the benchmarks for RDFParsers that were recently introduced by @tokovach in `testsuites/benchmark`. There was some overlap with https://github.com/eclipse/rdf4j/issues/3010 so I dropped the benchmarks from my work there. I'd be happy to take up this work.

frensjan:
I have created a PR for this: https://github.com/eclipse/rdf4j/pull/3077.

PR #3077, opened 2021-05-27
GH-3076 rdfwriter benchmarks

frensjan:
GitHub issue resolved: #3076

Adds JMH benchmarks for writing various RDF formats.

----


testsuites/benchmark/src/main/java/org/eclipse/rdf4j/benchmark/rio/WriterBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+			"jsonld",
+			"n3",
+			"nq",
+			"nt",
+			"rdf",
+			"rj",
+			"trig",
+			"trix",
+			"ttl",
+	})
```
frensjan, line 66: Note that this is a deliberate deviation from the setup of the ParserBenchmark. The setup of a class per file type was a bit heavy for my test. Let me know what you think. I'd be happy to take the same route as the ParserBenchmark or move the ParserBenchmark towards this setup.

jeenbroekstra: The main advantage I see to the parser benchmark approach is that you can run a standalone test for one particular format more easily. But presumably this is not a benchmark test that takes a tremendous amount of time to run anyway, so am fine with this more "lightweight" approach. No need to also change ParserBenchmark though.

jeenbroekstra (approved):
I'm fine with this as-is. I'll give others a bit more time to have a look, but unless something comes up I'll merge this in 24 hours or so. Thanks again @frensjan !

Issue #3101, opened 2021-06-05
SHACL - BindSelect should use optimised binding set


hmottestad:
BindSelect has a very narrow use of binding sets that we can utilise to design a more efficient binding set.

Issue #3100, opened 2021-06-05
SHACL - Datatype filter should use XSD enums


hmottestad:
Datatype filter currently uses the datatype IRI but should use the new enum support if applicable.

Issue #3099, opened 2021-06-05
SHACL - Unique plan node should detect if it's wrapping another Unique plan node


hmottestad:
Unique is a potentially expensive plan node so we can gain a bit of performance by detecting if it's wrapping another Unique plan node and then remove the superfluous node.

Issue #3098, opened 2021-06-05
SHACL - Bulk validation should allow shapes for retrieve all targets directly instead of extending the provided targets


hmottestad:
Bulk validation works by overriding the targets that a shape should validate. Since the provided targets use the root shape to find its targets, the child shape needs to extend them to account for paths in property shapes.

Instead of extending the provided targets, we could instead detect that we are using bulk validation and then retrieve the targets directly at the child shape. This would allow us to retrieve targets with their respective paths in one query. We can also filter the results before sorting them if we want.

PR #3104, opened 2021-06-05
GH-3088 various optimizations for sh:or

hmottestad:
GitHub issue resolved: #3088 
GitHub issue resolved: #3097 
GitHub issue resolved: #3098 
GitHub issue resolved: #3099 
GitHub issue resolved: #3100 
GitHub issue resolved: #3101 

 - Implement .equals and .hashCode for all plan nodes so that we can more easily reduce duplicates. We also add some code that simplifies nested unions, which also makes it easier to deduplicate. Deduplication makes sh:or faster since we can reduce the number of plan nodes used to calculate the relevant targets for a transaction. #3088 
 - Related to #3088 is performance or sh:or during bulk validation. The problem boils down to skipping the target extension phase by being able to detect when we are using bulk validation and retrieving all the targets directly. We introduce a new plan node to signal bulk validation and also add some support for filtering data earlier. #3098 
 - Similar to the simplification of nested unions, we can also simplify nested Unique (distinct) plan nodes #3099
 - The new XSD enum support allows us to optimise the DatatypeFilter plan node #3100 
 - The ListBindingSet mints a new hash set every time the getBindingNames method is called, this is now cached on the first call. #3097 
 - Even with the optimisations in #3097 it was still better to be able to provide the binding names as a set, so we introduced a new binding set with this support (in the ShaclSail code base) #3101 

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/ShaclProperties.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/ShaclProperties.java
@@ -49,7 +49,7 @@ public class ShaclProperties {
 	private Long minCount;
 	private Long maxCount;
 
-	private Resource datatype;
+	private IRI datatype;
 	private Resource in;
 	private final List<Value> hasValue = new ArrayList<>();
 	private final List<Resource> hasValueIn = new ArrayList<>();
@@ -148,7 +148,7 @@ public class ShaclProperties {
 					if (datatype != null) {
 						throw new IllegalStateException(predicate + " already populated");
 					}
-					datatype = (Resource) object;
+					datatype = (IRI) object;
 					break;
 				case "http://www.w3.org/ns/shacl#minCount":
 					if (minCount != null) {
@@ -335,7 +335,7 @@ public class ShaclProperties {
 		return maxCount;
 	}
 
-	public Resource getDatatype() {
+	public IRI getDatatype() {
 		return datatype;
 	}
 
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Shape.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Shape.java
@@ -23,7 +23,6 @@ import org.eclipse.rdf4j.model.Statement;
 import org.eclipse.rdf4j.model.impl.DynamicModel;
 import org.eclipse.rdf4j.model.impl.LinkedHashModelFactory;
 import org.eclipse.rdf4j.model.util.ModelBuilder;
-import org.eclipse.rdf4j.model.util.Values;
 import org.eclipse.rdf4j.model.vocabulary.DASH;
 import org.eclipse.rdf4j.model.vocabulary.RDF;
 import org.eclipse.rdf4j.model.vocabulary.RSX;
@@ -396,7 +395,7 @@ abstract public class Shape implements ConstraintComponent, Identifiable, Export
 				return Shape.this.generateTransactionalValidationPlan(connectionsGroup, logValidationPlans, null,
 						Scope.none);
 			} else {
-				return new EmptyNode();
+				return EmptyNode.getInstance();
 			}
 
 		} else {
@@ -430,7 +429,8 @@ abstract public class Shape implements ConstraintComponent, Identifiable, Export
 	 */
 	public SparqlFragment buildSparqlValidNodes_rsx_targetShape(StatementMatcher.Variable subject,
 			StatementMatcher.Variable object,
-			RdfsSubClassOfReasoner rdfsSubClassOfReasoner, Scope scope) {
+			RdfsSubClassOfReasoner rdfsSubClassOfReasoner, Scope scope,
+			StatementMatcher.StableRandomVariableProvider stableRandomVariableProvider) {
 		throw new UnsupportedOperationException(this.getClass().getSimpleName());
 	}
 
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
@@ -170,6 +170,24 @@ public class StatementMatcher {
 		return Objects.hash(subjectName, subjectValue, predicateName, predicateValue, objectName, objectValue);
 	}
 
+	public static class StableRandomVariableProvider {
+		private static final String BASE = UUID.randomUUID().toString().replace("-", "") + "_"; // We just need a random
+																								// base that isn't used
+																								// elsewhere in the
+																								// ShaclSail, but we
+																								// don't want it to be
+																								// stable so we can
+																								// compare the SPARQL
+																								// queries where these
+																								// variables are used
+		private int counter = 0;
+
+		public Variable next() {
+			return new Variable(BASE + counter++);
+		}
+
+	}
+
 	public static class Variable {
 		String name;
 		Value value;
@@ -199,8 +217,11 @@ public class StatementMatcher {
 			return name == null && value == null;
 		}
 
-		public static Variable getRandomInstance() {
-			return new Variable(UUID.randomUUID().toString().replace("-", ""));
+		public String asSparqlVariable() {
+			if (value != null)
+				throw new IllegalStateException(
+						"Can not produce SPARQL variable for variables that have fixed values!");
+			return "?" + name.replace("-", "__");
 		}
 
 		@Override
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
@@ -171,15 +171,11 @@ public class StatementMatcher {
 	}
 
 	public static class StableRandomVariableProvider {
-		private static final String BASE = UUID.randomUUID().toString().replace("-", "") + "_"; // We just need a random
-																								// base that isn't used
-																								// elsewhere in the
-																								// ShaclSail, but we
-																								// don't want it to be
-																								// stable so we can
-																								// compare the SPARQL
-																								// queries where these
-																								// variables are used
+
+		// We just need a random base that isn't used elsewhere in the ShaclSail, but we don't want it to be stable so
+		// we can compare the SPARQL queries where these variables are used
+		private static final String BASE = UUID.randomUUID().toString().replace("-", "") + "_";
+
 		private int counter = 0;
 
 		public Variable next() {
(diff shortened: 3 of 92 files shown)


hmottestad:
This became quite big, but it was hard to separate out everything since all this work is based on performance explorations I've been doing for the ShaclSail.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/ShaclProperties.java
```diff
@@ -148,7 +148,7 @@ public ShaclProperties(Resource id, RepositoryConnection connection) {
 					if (datatype != null) {
 						throw new IllegalStateException(predicate + " already populated");
 					}
-					datatype = (Resource) object;
+					datatype = (IRI) object;
```
hmottestad, line 151: SHACL spec says IRI, so this is in line with that.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/Shape.java
```diff
@@ -430,7 +429,8 @@ public boolean requiresEvaluation(ConnectionsGroup connectionsGroup, Scope scope
 	 */
 	public SparqlFragment buildSparqlValidNodes_rsx_targetShape(StatementMatcher.Variable subject,
 			StatementMatcher.Variable object,
-			RdfsSubClassOfReasoner rdfsSubClassOfReasoner, Scope scope) {
+			RdfsSubClassOfReasoner rdfsSubClassOfReasoner, Scope scope,
+			StatementMatcher.StableRandomVariableProvider stableRandomVariableProvider) {
```
hmottestad, line 433: Plan nodes that run SPARQL queries also need to support `.equals` and `.hashCode`. That only works if the code that generates the SPARQL queries generate the same query for every run (eg. functional). The `StatementMatcher.StableRandomVariableProvider` gives sequential variables prefixed by a static UUID. Name is a bit long I might admit.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
```diff
@@ -170,6 +170,24 @@ public int hashCode() {
...
+	public static class StableRandomVariableProvider {
+		private static final String BASE = UUID.randomUUID().toString().replace("-", "") + "_"; // We just need a random
+																								// base that isn't used
+																								// elsewhere in the
+																								// ShaclSail, but we
+																								// don't want it to be
+																								// stable so we can
+																								// compare the SPARQL
+																								// queries where these
+																								// variables are used
```
hmottestad, line 182: Code formatter went a bit haywire here I reckon.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/constraintcomponents/SimpleAbstractConstraintComponent.java
```diff
@@ -179,13 +225,13 @@ private PlanNode generateTransactionalValidationPlan(ConnectionsGroup connection
...
-				return filterAttacher.apply(targets).getTrueNode(UnBufferedPlanNode.class);
-			} else {
-				return filterAttacher.apply(targets).getFalseNode(UnBufferedPlanNode.class);
-			}
+			return effectiveTarget.getPlanNode(connectionsGroup, scope, false, p -> {
+				if (negatePlan) {
+					return filterAttacher.apply(p).getTrueNode(UnBufferedPlanNode.class);
+				} else {
+					return filterAttacher.apply(p).getFalseNode(UnBufferedPlanNode.class);
+				}
```
hmottestad, line 233: Could probably extract this so I don't have to repeat it everywhere.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/AllTargetsPlanNode.java
```diff
@@ -0,0 +1,171 @@
...
+import org.eclipse.rdf4j.sail.SailException;
+import org.eclipse.rdf4j.sail.shacl.ConnectionsGroup;
+import org.eclipse.rdf4j.sail.shacl.ast.StatementMatcher;
+import org.eclipse.rdf4j.sail.shacl.ast.constraintcomponents.ConstraintComponent;
+import org.eclipse.rdf4j.sail.shacl.ast.targets.EffectiveTarget;
+
+/**
+ * Used for adding a custom log statement to tuples as they pass through. Should only be used for debugging.
+ */
+public class AllTargetsPlanNode implements PlanNode {
```
hmottestad, line 28: Copy pasted from another class, gotta update that comment there!

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/AllTargetsPlanNode.java
```diff
@@ -0,0 +1,171 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
```
hmottestad, line 2: 2021

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/DatatypeFilter.java
```diff
@@ -32,11 +38,35 @@ boolean checkTuple(ValidationTuple t) {
 		}
 
 		Literal literal = (Literal) t.getValue();
-		return literal.getDatatype() == datatype || literal.getDatatype().equals(datatype);
+		if (xsdDatatype != null && literal instanceof SimpleLiteral) {
+			return xsdDatatype == ((SimpleLiteral) literal).getXsdDatatype().orElse(null);
+		} else {
+			return literal.getDatatype() == datatype || literal.getDatatype().equals(datatype);
```
hmottestad, line 44: XSD enum code.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/SimpleBindingSet.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Override
+	public Value getValue(String bindingName) {
+		for (Binding binding : bindings) {
+			if (binding.getName().equals(bindingName)) {
+				return binding.getValue();
+			}
+		}
+		return null;
+	}
```
hmottestad, line 80: This can call `getBinding()` instead.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/SimpleBindingSet.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Override
+	public boolean hasBinding(String bindingName) {
+		for (Binding binding : bindings) {
+			if (binding.getName().equals(bindingName)) {
+				return true;
+			}
+		}
+		return false;
+	}
```
hmottestad, line 70: We can assume that all binding names are in use, then we can use the binding names set here instead of looping through the bindings.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/StatementMatcher.java
jeenbroekstra: Huh, weird. Least effort is probasbly to put this comment above instead of after the line. More reabable anyway.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/BindSelect.java
```diff
@@ -212,7 +215,7 @@ public void calculateNext() {
 					updateQuery(parsedQuery, bindingSets, targetChainSize);
 
 					bindingSet = connection.evaluate(parsedQuery.getTupleExpr(), parsedQuery.getDataset(),
-							new MapBindingSet(), true);
+							new EmptyBindingSet(), true);
```
jeenbroekstra, line 218: You might want to use the `EmptyBindingSet.getInstance()` singleton here.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/planNodes/SimpleBindingSet.java
```diff
@@ -0,0 +1,86 @@
+/*******************************************************************************
+ * .Copyright (c) 2021 Eclipse RDF4J contributors.
```
jeenbroekstra, line 2: Accidental '.' added to copyright statement

core/sail/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/benchmark/OrDatatypeBenchmark.java
```diff
@@ -0,0 +1,168 @@
...
+
+		try (SailRepositoryConnection connection = repository.getConnection()) {
+			for (List<Statement> statements : allStatements) {
+				connection.begin(IsolationLevels.SNAPSHOT);
+				connection.add(statements);
+				connection.commit();
+			}
+		}
+
+//		repository.shutDown();
```
jeenbroekstra, line 164: Should this be commented out?

jeenbroekstra (commented):
LGTM for the most part, couple of minor inline remarks. It's quite massive and not strictly speaking a bug fix, but as it only really addresses internal code and doesn't change any public API contract we can get away with doing this as part of a patch release. 

This fixes a number of only very loosely-related issues, so we definitely shouldn't squash this. In fact, we may choose to use Github rebase to merge this, so that after merge they stand alone (your individual commits are all nicely self-contained AFAICT).

One last thing I noticed. Although the package `org.eclipse.rdf4j.sail.shacl.ast` is marked InternalUseOnly (in package-info.java),  its 'subpackages' (planNodes, paths, etc) are not. Not a huge deal as all of this is hidden from the javadoc anyway, but could you copy over the packag-info.java into each subpackage just so we're nicely covered?

hmottestad:
I'll add the package info. 

We've included performance fixes in patch releases before as long as they didn't change or add any user facing code. With the size of this I think I would normally have considered a minor release though, but I didn't feel that it justified a special 3.8 release. 

Was thinking a merge commit would be nice. What do you think?

jeenbroekstra:
> Was thinking a merge commit would be nice. What do you think?

That's fine too, but we can also use github rebase. This has no negative consequences AFAIK because the main branch is protected, so it can't introduce any inconsistencies. With an eye on possibly wanting to do cherry picking later (if we decide we do need a 3.8 release) it would be useful to have a more linear history (and yes, I've given up on trying to come with One Merge Policy to Rule Them All).

Issue #3106, opened 2021-06-08
FedX cannot handle large joins with > 65k intermediate result


aschwarte10:
In the concurrency facilities of FedX we make use of the Java Phaser.
This is limited to 65535 registered parties.

In case of large joins we were seeing errors that the the phaser has reached limits.

Example trace

```
Caused by: java.lang.IllegalStateException: Attempt to register more than 65535 parties for java.util.concurrent.Phaser@3ab90002[phase = 0 parties = 65535 arrived = 0]

	at java.base/java.util.concurrent.Phaser.doRegister(Phaser.java:429)

	at java.base/java.util.concurrent.Phaser.register(Phaser.java:581)

	at org.eclipse.rdf4j.federated.evaluation.join.ControlledWorkerLeftJoin.handleBindings(ControlledWorkerLeftJoin.java:61)

	at org.eclipse.rdf4j.federated.evaluation.join.JoinExecutorBase.performExecution(JoinExecutorBase.java:51)
```

aschwarte10:
@jeenbroekstra I have a fix for this locally and will prepare a PR. Should this go as a bugfix for 3.7.x targeted to the master branch?

PR #3108, opened 2021-06-09
GH-3106: fix large joins in FedX engine

aschwarte10:
GitHub issue resolved: #3106 

In the concurrency facilities of FedX we make use of the Java Phaser.
This is limited to 65535 registered parties.

In case of large joins (see the example test case) we were seeing errors
that the the phaser has reached limits.

This is now avoided by using tiering of phasers as suggested by the
Phaser javadocs. Note that we pass on a delegate executor control now to
call the arrive and deregister on the correct phaser instance.

----

### DIFF
--- a/tools/federation/src/test/java/org/eclipse/rdf4j/federated/LargeJoinTest.java
+++ b/tools/federation/src/test/java/org/eclipse/rdf4j/federated/LargeJoinTest.java
@@ -1,0 +1,130 @@
+package org.eclipse.rdf4j.federated;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.eclipse.rdf4j.common.iteration.Iterations;
+import org.eclipse.rdf4j.federated.FedXWithLocalRepositoryManagerTest.TestLocalRepositoryManager;
+import org.eclipse.rdf4j.federated.repository.FedXRepository;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.util.Values;
+import org.eclipse.rdf4j.model.vocabulary.FOAF;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.query.BindingSet;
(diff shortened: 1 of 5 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/PhaserHandlingParallelExecutor.java
```diff
@@ -0,0 +1,65 @@
...
+import org.eclipse.rdf4j.query.BindingSet;
+import org.eclipse.rdf4j.query.QueryEvaluationException;
+
+/**
+ * A delegating {@link ParallelExecutor} which arrives and de-registers on the phaser on completion of a task.
+ * 
+ * @author Andreas Schwarte
+ *
+ */
+public class PhaserHandlingParallelExecutor implements ParallelExecutor<BindingSet> {
```
jeenbroekstra, line 23: Any chance we can make this package-private? If not, perhaps slap an `@InternalUseOnly` annotation on it.

jeenbroekstra (commented):
Good find, thanks for the fix. Looks good to me, just two minor inline remarks.

tools/federation/src/test/java/org/eclipse/rdf4j/federated/LargeJoinTest.java
```diff
@@ -0,0 +1,130 @@
...
+		repoManager.shutDown();
+	}
+
+	@Override
+	protected FederationContext federationContext() {
+		throw new UnsupportedOperationException("Not available in this context.");
+	}
+
+	@Test
+	@Ignore
```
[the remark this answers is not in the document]
aschwarte10: good spot, done

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/PhaserHandlingParallelExecutor.java
aschwarte10: I have moved it to the `org.eclipse.rdf4j.federated.evaluation.join` package and made it package private.

jeenbroekstra (approved):

Issue #3071, opened 2021-05-26
Missing results in cartesian product of values blocks


tayloj:
I discovered this in RDF4j 3.2.2, but can confirm it with 3.7.0 as well.  When using multiple `values` blocks in a query to bind different variables, I'm not getting as many combinations as I'm expecting.  As an example, this query should have four results:

```sparql
select ?x ?y where {
  values ?x { undef 67 }
  values ?y { undef 42 }
}
```

Executing that with Jena, I get:

```
-----------
| x  | y  |
===========
|    |    |
|    | 42 |
| 67 |    |
| 67 | 42 |
-----------
```

But in RDF4j, I only get three results.  Printing the binding sets from Java:

```
[]
[y="42"^^<http://www.w3.org/2001/XMLSchema#integer>]
[x="67"^^<http://www.w3.org/2001/XMLSchema#integer>;y="42"^^<http://www.w3.org/2001/XMLSchema#integer>]
```

Here's small, failing, unit test that reproduces the issue and asserts the expected behavior (full file attached: 
[RDF4jTests.java.txt](https://github.com/eclipse/rdf4j/files/6547039/RDF4jTests.java.txt)):

```java
   @Test
   public void testValuesCartesianProduct() {
      final Repository repository = new SailRepository(new MemoryStore());
      try (RepositoryConnection repositoryConnection = repository.getConnection()) {
         final String queryString = ""
               + "select ?x ?y where { "
               + "  values ?x { undef 67 } "
               + "  values ?y { undef 42 } "
               + "}";
         final TupleQuery tupleQuery = repositoryConnection.prepareTupleQuery(queryString);

         final List<BindingSet> bindingSets = new ArrayList<>();
         try (TupleQueryResult tupleQueryResult = tupleQuery.evaluate()) {
            tupleQueryResult.forEach(bs -> bindingSets.add(bs));
         }
         assertThat("should have four results", bindingSets.size(), equalTo(4));
      }
   }
```

which fails with:

```
java.lang.AssertionError: should have four results
Expected: <4>
     but: was <3>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at RDF4jTests.testValuesCartesianProduct(RDF4jTests.java:33)
```

PR #3113, opened 2021-06-12
GH-3071 ensure bsa result iter returns result when assignment is UNDEF

jeenbroekstra:
GitHub issue resolved: #3071 

- fix corner case of binding set assignment result processing when binding is UNDEF.
- added regresssion test
- some variable renaming and code cleanup 

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
@@ -730,52 +730,48 @@ public class StrictEvaluationStrategy implements EvaluationStrategy, FederatedSe
 
 	public CloseableIteration<BindingSet, QueryEvaluationException> evaluate(BindingSetAssignment bsa,
 			BindingSet bindings) throws QueryEvaluationException {
-		final Iterator<BindingSet> iter = bsa.getBindingSets().iterator();
-		if (bindings.size() == 0) { // empty binding set
-			return new CloseableIteratorIteration<>(iter);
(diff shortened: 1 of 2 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
```diff
@@ -730,52 +730,48 @@ protected Value getVarValue(Var var, BindingSet bindings) {
...
-					final BindingSet assignedBindings = iter.next();
+				QueryBindingSet nextResult = null;
+				while (nextResult == null && assignments.hasNext()) {
+					final BindingSet assignedBindings = assignments.next();
+
 					for (String name : assignedBindings.getBindingNames()) {
+						if (nextResult == null) {
+							nextResult = new QueryBindingSet(bindings);
+						}
+
```
jeenbroekstra, line 754: This is the crux of the fix: we initialize `nextResult` as soon as we have a binding name (previously it was only initialized if that binding name also had a value in the BSA)

hmottestad:
@jeenbroekstra tried to look at this several times but couldn’t really understand what’s going on. Sorry for my tardiness.

jeenbroekstra:
> @jeenbroekstra tried to look at this several times but couldn’t really understand what’s going on. Sorry for my tardiness.

No worries, that's fine. For what it's worth as long as I know someone has taken a look and doesn't spot anything obviously wrong (even if they don't fully understand the entire fix), that is useful feedback to me. So don't feel you need to refrain from commenting even if you only have a few minutes to take a quick glance :)

hmottestad:
@jeenbroekstra This fix here is impacting one of the SHACL tests, 3-4x slower than before. TruncatedValidationReportTest.

Not sure what's going on, just wanted to let you know.

Issue #3131, opened 2021-06-25
FedX iteration closing improvements


damyan-ognyanov:
During integration of FedX over local repositories we stumbled on some concurency issues, which lead to connection references leaks that cause some connections to be kept open due to unclosed iterations after query evaluation has been completed.

I've prepared a PR addressing some of the issues I've found  (workarounds.) The PR alos includes a new configuration parameter to specify the number of solutions processed by ConsumingIteration when it is created.

PR #3132, opened 2021-06-25
wip: GH-3131 FedX cancel tasks. configurable ConsumingIteration parameter

damyan-ognyanov:

GitHub issue resolved: #3131

 - address issues with connection leaks when tasks are being canceled
 - add new configuration parameter that specify number of results collected by ConsumingIteration initially

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
@@ -97,7 +97,10 @@ public abstract class ParallelExecutorBase<T> extends LookAheadIteration<T, Quer
 		}
 
 		try {
-			rightQueue.put(res);
+			if (!closed)
+				rightQueue.put(res);
+			else
+				res.close();
 		} catch (InterruptedException e) {
 			throw new RuntimeException("Error adding element to right queue", e);
 		}
@@ -153,6 +156,7 @@ public abstract class ParallelExecutorBase<T> extends LookAheadIteration<T, Quer
 	@Override
 	public void handleClose() throws QueryEvaluationException {
 
+		closed = true;
 		try {
 			rightQueue.close();
 		} finally {
@@ -162,7 +166,6 @@ public abstract class ParallelExecutorBase<T> extends LookAheadIteration<T, Quer
 				rightIter = null;
 			}
 		}
-		closed = true;
 		super.handleClose();
 	}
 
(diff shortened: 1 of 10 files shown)


aschwarte10 (commented):
Good findings. Thanks for providing the improvements and fixes.

I think this PR looks generally good and clean. I don't see any problems with the changes.

One question to you: are you able to reproduce the dangling connections in some test environment? I think we have seen also some behavior (rarely) which we could not explain - maybe this was also caused by the dangling connection (and thereby lack of resources).

Looking forward to have this in

damyan-ognyanov:
@aschwarte10 still working on the PR because the build randomly fails with java 15. The test/setup I'm using to reproduce it is relatively simple but a bit heavy - we load a large dataset (50m statements) in one repository and relatively small dataset in another repo (both are local repositories, so no remote http requests). Then just running few hundreds times a single query that does join between both local repositories. 
first issue we spotted is that when a thread is interrupted during read operation from a file channel obtained from a RandomAccessFile instance, the channel is force closed which made the repo unusable for further operations. The other issue is the one with the connections that was left open these keep references to some of our internal structures and after few hundred evaluations we hit OOM exceptions because these were not closed.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
```diff
@@ -97,7 +97,10 @@ public void addResult(CloseableIteration<T, QueryEvaluationException> res) {
 		}
 
 		try {
-			rightQueue.put(res);
+			if (!closed)
+				rightQueue.put(res);
+			else
+				res.close();
```
jeenbroekstra, line 103: Doesn't this introduce a possible race condition? The `closed` flag seems to signal if `rightQueue` has been closed or not, but there is no mutex guarding that access. It looks like it might be possible for `rightQueue` to be closed in between the check of the if-condition and the adding of the result. 

Also, minor editorial: can you add curly braces to your if-else statements? We try to consistently use this everywhere, even if the block is a one-liner.

aschwarte10:
@damyan-ognyanov we have also seen quite some errorneuos cases recently and I have followed up on your work. I have borrowed some of your fixes, and added solutions for quite a number of further problems that I encountered while testing to https://github.com/eclipse/rdf4j/pull/3255

Please let me know if you are OK in combining the fixes in that PR and how you want to handle attribution

Note: the improvement of making the size in consuming iteration configurable I have left out. I thing this is a good isolated improvement.

aschwarte10:
I think we should close this PR as most of the relevant bits are merged as part of other work in the mean time

@damyan-ognyanov do you agree? Alternatively please propose a way forward

PR #3134, opened 2021-06-27
Release notes 4.0.0

jeenbroekstra:
I've opened a branch with a draft release notes document. I suggest we all add upgrade points to this as we go, preferable as soon as we merge our fixes. It's easier to update this when the change you've made is still fresh than it is if we have to scramble to get the release done in a few months from now :)


jeenbroekstra:
I am merging this in directly now: I've realized it's easier to updates these notes as part of the PR that actually does the relevant change.

Issue #3140, opened 2021-06-30
Remove unused FileUtil methods


barthanssens:
The org.eclipse.rdf4j.common.io.FileUtil class seems to contain a few unused (but public) methods, check if they can be removed

PR #3141, opened 2021-06-30
GH-3140: remove unused methods

barthanssens:

GitHub issue resolved: #3140 

- remove unused (public) methods
- replace deprecated createTempDir method with Files.createTempDirectory

----


core/util/src/main/java/org/eclipse/rdf4j/common/io/FileUtil.java
```diff
@@ -9,307 +9,12 @@
...
-	private static final String ILLEGAL_FILE_NAME_CHARS = "\\/:*?\"<>|";
-
-	/**
-	 * Gets the relative representations of a file compared to another.
-	 *
-	 * @param subj       The File to find the relative form for.
-	 * @param relativeTo The File 'subj' should be made relative to.
-	 * @return The relative representation of subj.
-	 */
-	public static String getRelativePath(File subj, File relativeTo) {
```
jeenbroekstra, line 32: There are several public methods here that were never marked deprecated. I know we're aiming for a major so backward compatibility is less of an issue, but perhaps we should keep these in place marked deprecated.

jeenbroekstra (commented):
Good work on removing these use of these outdated utility methods. I think completely removing the methods themselves from the code base might a bit harsh though. Let's stick some deprecation warnings on them instead (in fact, on the entire FileUtil class).

jeenbroekstra:
Actually, thinking about this a bit more: perhaps we should just go ahead and rip them out. If we get complaints, we can stick (some) back in, and/or document our recommended alternatives.

barthanssens:
Yeah, there's some more code in other util classes that could also be removed.

Would it help to add the ideprectated or nternaluseonly  annotation on these methods in the 3.7.2 release (or even on all util classes / methods ? I'm assuming only people writing their own sails / parsers may have a good use case for these methods)
or is that not appropriate on a patch release ?

jeenbroekstra:
> Yeah, there's some more code in other util classes that could also be removed.

Just keep in mind that just because we don't use it ourselves, doesn't mean others don't. For StringUtil and FileUtil there are clearly better alternatives, but some utilities are genuinely useful (even if they're not strictly RDF-related).

> Would it help to add the ideprectated or nternaluseonly annotation on these methods in the 3.7.2 release (or even on all util classes / methods ? I'm assuming only people writing their own sails / parsers may have a good use case for these methods)
> or is that not appropriate on a patch release ?

I have no objections to marking them deprecated in a patch release. InternalUseOnly is not really appropriate here I think (quite a lot of this _is_ intended for public use - it's just a bit out of date and no longer necessary).

core/util/src/main/java/org/eclipse/rdf4j/common/io/FileUtil.java
barthanssens: As mentioned below, removing them (and add them back into the code if we receive complaints)

Issue #3146, opened 2021-07-02
Use try-with-resources


barthanssens:
Use try-with-resources instead of try-finally on autocloseable resources.

Since RDF4J is going to require JDK11 in 4.0, we can now also use try-with-resources on (effectively) final variables (introduced in  JDK9)

PR #3147, opened 2021-07-02
GH-3146: use try-with-resources (via Netbeans Inspect and Transform)

barthanssens:

GitHub issue resolved: #3146 

- use try-with-resources instead of try-finally

----

### DIFF
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -195,10 +195,8 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 
 	@Override
 	public void close() throws RepositoryException {
-		try {
+		try (client) {
 			super.close();
-		} finally {
-			client.close();
 		}
 	}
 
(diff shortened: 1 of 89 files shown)


core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -195,10 +195,8 @@ public void setHttpClient(HttpClient httpClient) {
 
 	@Override
 	public void close() throws RepositoryException {
-		try {
+		try (client) {
```
jeenbroekstra, line 198: I've not seen try-with-resources used in this way before (autoclosing something that was defined out of scope of the actual try block). Could we do a quick smoke test that that actually works as expected?

jeenbroekstra (commented):
Mostly looks good to me, nice bit of (auto) refactoring. One question about a bit of an odd pattern I noticed.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
barthanssens: TBH I haven't tested all of them, and I'm thinking to roll back those (auto-generated) changes.
I assume it (mostly ?) works, but at the very least it is confusing...

jeenbroekstra:
This has grown a bit stale I think @barthanssens - should we close this or do you want to take another stab at it?

barthanssens:
Thanks for the reminder, I'll look into it

Issue #3150, opened 2021-07-06
UnionIteration can cause stack overflow


hmottestad:
UnionIteration uses recursion when retrieving the next iterator. There doesn't seem to be any particular need to use recursion and it doesn't make the code cleaner or easier to understand.

https://github.com/eclipse/rdf4j/blob/75b97f6106e6a73776d7f0f663f18a3edfb0a4ae/core/util/src/main/java/org/eclipse/rdf4j/common/iteration/UnionIteration.java#L82

PR #3151, opened 2021-07-06
GH-3150 use while loop instead of recursion

hmottestad:
GitHub issue resolved: #3150 

 - Use while loop instead of recursion in UnionIteration

----

### DIFF
--- a/core/util/src/main/java/org/eclipse/rdf4j/common/iteration/UnionIteration.java
+++ b/core/util/src/main/java/org/eclipse/rdf4j/common/iteration/UnionIteration.java
@@ -62,24 +62,26 @@ public class UnionIteration<E, X extends Exception> extends LookAheadIteration<E
 		if (isClosed()) {
 			return null;
 		}
-		Iteration<? extends E, X> nextCurrentIter = currentIter;
-		if (nextCurrentIter != null && nextCurrentIter.hasNext()) {
-			return nextCurrentIter.next();
-		}
 
-		// Current Iteration exhausted, continue with the next one
-		Iterations.closeCloseable(nextCurrentIter);
+		while (true) {
 
-		synchronized (this) {
-			if (argIter.hasNext()) {
-				currentIter = argIter.next();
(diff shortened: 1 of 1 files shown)


core/util/src/main/java/org/eclipse/rdf4j/common/iteration/UnionIteration.java
```diff
@@ -62,24 +62,26 @@ protected E getNextElement() throws X {
...
 			return null;
 		}
-		Iteration<? extends E, X> nextCurrentIter = currentIter;
-		if (nextCurrentIter != null && nextCurrentIter.hasNext()) {
-			return nextCurrentIter.next();
-		}
 
-		// Current Iteration exhausted, continue with the next one
-		Iterations.closeCloseable(nextCurrentIter);
+		while (true) {
```
jeenbroekstra, line 66: I don't particularly like the notion of an infinite loop. Is there a way we can move the conditional to jump out of the loop in here? `while(argIter.hasNext())` perhaps? I'm not sure how that would play out with the synchronized block to be honest (not even sure why we're synchronizing inside an iterator etiher, tbh, but that's a separate issue).

hmottestad: It's hard to make sure all the conditions are met correctly, especially with that synchronized block. Not sure why there is a synchronized block to be honest...is the code assuming that there will be some sort of parallel access to this iteration?

Don't like the while(true) either, but it's essentially what was there already.

jeenbroekstra: Fair enough. The synchronized block might be a leftover from the past, when there was a bit of a misguided "better safe than sorry" philosophy towards concurrency. I'm not convinced it's useful, but ripping it out here might be a bit beyond the scope of your intended fix.

jeenbroekstra (approved):

Issue #3153, opened 2021-07-07
Null Pointer exception on adding RDF data


volkerjaenisch:
Dear RDF4J-Developers !

I noticed that RDF4J throws a null pointer exception in the following case:

* Utilizing the Web API PUT Method 
* Request:
  - Content-Type: 'application/x-turtle'
  - Payload: some TTL RDF data [without a rep:Repository structure]
  - URI: /repositories/test1 [not already existing]

Response is an empty <500> status-code. 

I am aware that I did not use the WebAPI correctly. But IMHO more context in the response would really be useful in this case. Something like "No repository config supplied".

```
[ERROR] 2021-07-06 14:00:52,449 [repositories/test1] Error while handling request
java.lang.NullPointerException: null
        at org.eclipse.rdf4j.repository.manager.LocalRepositoryManager.addRepositoryConfig(LocalRepositoryManager.java:385)
        at org.eclipse.rdf4j.repository.manager.LocalRepositoryManager.addRepositoryConfig(LocalRepositoryManager.java:381)
        at org.eclipse.rdf4j.http.server.repository.RepositoryController.handleRequestInternal(RepositoryController.java:170)
        at org.springframework.web.servlet.mvc.AbstractController.handleRequest(AbstractController.java:177)
        at org.springframework.web.servlet.mvc.SimpleControllerHandlerAdapter.handle(SimpleControllerHandlerAdapter.java:52)
        at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1060)
        at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:962)
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1006)
        at org.springframework.web.servlet.FrameworkServlet.doPut(FrameworkServlet.java:920)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:663)
        at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:883)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:741)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:231)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
[...]

PR #3177, opened 2021-07-17
GH-3153 better error handling/reporting on PUT repo config operations

jeenbroekstra:
GitHub issue resolved: #3153  

- check for missing config data in RepositoryController and return an explicit /  helpful error (400 Bad Request)
- also catch other config validation errors  and also return a 400 error with useful message in body

----

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/RepositoryController.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/RepositoryController.java
@@ -64,6 +64,7 @@ import org.eclipse.rdf4j.repository.Repository;
 import org.eclipse.rdf4j.repository.RepositoryConnection;
 import org.eclipse.rdf4j.repository.RepositoryException;
 import org.eclipse.rdf4j.repository.config.RepositoryConfig;
+import org.eclipse.rdf4j.repository.config.RepositoryConfigException;
 import org.eclipse.rdf4j.repository.config.RepositoryConfigUtil;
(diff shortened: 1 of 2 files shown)


tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/RepositoryController.java
```diff
@@ -167,8 +167,15 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpSer
 								.orElseThrow(() -> new HTTPException(HttpStatus.SC_BAD_REQUEST,
 										"unrecognized content type " + request.getContentType())));
 				RepositoryConfig config = RepositoryConfigUtil.getRepositoryConfig(model, repId);
```
hmottestad, line 169: Kinda wish that this method couldn't return null. Throwing an exception would be best, RepositoryConfigNotFound, or at the least use optional. 

I presume it's harder to change that in a bug fix though.

hmottestad (approved):
Looks good

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/RepositoryController.java
jeenbroekstra: Yeah, I'd rather leave it as-is for now. Perhaps worth opening a more general refactoring issue on for 4.0 though, going through the public API and replacing some "may return null" signatures with an optional return type?

Issue #3199, opened 2021-07-18
Use issue forms 


jeenbroekstra:
The use of [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) may be a nice compromise between the need to help guide/structure creation of issues and the somewhat overloaded approach of having a template with a lot of unnecessary bloat in it. 

Forms could help us to automatically apply some relevant default labels ("bug", "enhancement") as well as reminding submitters to tell us sufficient detail (e.g. the version of RDF4J they're observing a problem in etc).

All of this should be done with the intent not to restrict people when reporting issues, but to help them do it right without us having to mail after them to ask for clarification.

jeenbroekstra:
A side note: for some reason the github action we have to automatically put a new issue in the triage column of our [Progress Board](https://github.com/eclipse/rdf4j/projects/19) is having trouble with the new discussion forum. It looks like this started when I migrated a bunch of issues from the list into the forum. It was so bad that our more essential github issues got stuck in a queue. I have temporarily disabled the action. It would be nice if we could get rid of it completely by replacing it with a form that automatically sets it.

jeenbroekstra:
Unfortunately the issue template thing does not seem to have support to automatically set the project....

jeenbroekstra:
Am considering just getting rid of the project progress board altogether. From the fact that nobody ever bothers to update their issue status I get the impression I'm the only one who actually uses it.

PR #3201, opened 2021-07-18
GH-3199 issue templates for bug reports and feature requests

jeenbroekstra:
GitHub issue resolved: #3199  

- adds issue templates with forms for bug reports and feature requests
- adds an obvious direct link to the discussion forum for questions, comments, etc.
- still possible to create a "blank" new issue if none of the templates fit (e.g. for our own internal tasks).
- bumped version of the new-issue github action (hopefully fixes the problems I ran into earlier, but this is as yet untested)

----

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,57 @@
+name: Bug Report
+description: File a bug report
+labels: [ "🐞 bug", "triage" ]
+body:
+    - type: markdown
+      attributes:
+          value: |
+              Thanks for taking the time to fill out this bug report!
+    - type: textarea
+      attributes:
+        label: Current Behavior
+        description: A concise description of what you're experiencing.
+      validations:
+        required: true
+    - type: textarea
+      attributes:
+        label: Expected Behavior
+        description: A concise description of what you expected to happen.
+      validations:
+        required: true
+    - type: textarea
+      attributes:
+        label: Steps To Reproduce
+        description: Steps to reproduce the behavior.
+        placeholder: |
+          1. In this environment...
+          2. With this config...
+          3. Run '...' (include code example if possible)
+          4. See error...
+        validations:
+          required: false
+    - type: dropdown
+      id: version
+      attributes:
+        label: Version
+        description: What version of RDF4J are you using?
+        options:
+        - 3.7.1
+        - 3.7.0
+        - 3.6.3
+        - 3.6.2
+        - 3.6.1
+        - 3.6.0
+        - 3.5.1
+        - 3.5.0
+        - older (please specify in a comment)
+      validations:
+        required: true
+    - type: textarea
+      attributes:
+        label: Anything else?
+        description: |
+          Links? References? Anything that will give us more context about the issue you are encountering!
+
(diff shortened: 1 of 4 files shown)


jeenbroekstra:
@hmottestad @barthanssens @aschwarte10  what do we think? I want to encourage people to report issues with sufficient information for us to act on them, so this is very intended as "guide rails". The intent is to make it _easier_ to report a problem, rather than harder.

I personally think this is an improvement, but it's less "free format" than just a blank issue of course (although if you really want you can also still do that).

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,36 @@
+name: Feature Request
+description: Suggest an idea for improving RDF4J
+labels: [ "📶 enhancement", "triage" ]
```
jeenbroekstra, line 3: If we can get the github action to put new issues on the project board working again, I'll remove the "triage" tag from these templates again, as it would be redundant with how we use the "Inbox" column.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,57 @@
...
+          4. See error...
+        validations:
+          required: false
+    - type: dropdown
+      id: version
+      attributes:
+        label: Version
+        description: What version of RDF4J are you using?
+        options:
+        - 3.7.1
```
hmottestad, line 38: I think it will be easier on us if we just have a text field instead of a drop down. That way we don't need to keep this updated when we make a new release.

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,36 @@
...
+      attributes:
+        label: Problem description
+        description: Provide a concise description of the problem your improvement will address.
+        placeholder: "Performing task X would be a lot easier if RDF4J could do..." 
+      validations:
+        required: true
+    - type: textarea
+      attributes:
+        label: Preferred solution
+        description: Provide a clear and concise description of what you want to happen.
```
hmottestad, line 19: Any way we could encourage code contribution. I want to keep expectations aligned with our dev capacity without making people feel that they can't ask for a feature while at the same time gently reminding them that we are a small open source project.

.github/ISSUE_TEMPLATE/bug-report.yml
jeenbroekstra: Excellent point. I was thinking about adding an option "latest" or something but that doesn't age well of course. A text field is much simpler.

.github/ISSUE_TEMPLATE/feature-request.yml
jeenbroekstra: Good point. I'm not sure if we want that in the placeholder text or perhaps more prominent.

jeenbroekstra: @hmottestad I've fixed this by actually ading a (non-required) extra field to ask if they're interested in contributing themselves.

.github/ISSUE_TEMPLATE/bug-report.yml
jeenbroekstra: Done.

Issue #2710, opened 2020-12-10
Reorganize the rdf4j-util module: split into separate modules for easier dependency mgmt


jeenbroekstra:
The `@Experimental` and `@InternalUseOnly` annotations are currently part of rdf4j-model-api. They were moved there (from rdf4j-util) as part of an effort to make rdf4j-model-api as lightweight as possible (and not have to include all the iterator and utility classes from rdf4j-util). 

This has introduced a problem in the dependency hierarchy, in particular for rdf4j-util: it now has to depend on rdf4j-model-api, just to be able to refer to those two annotation classes. Note that even if we removed their actual use from rdf4j-util code, we'd still get api comparison errors, because those two classes no longer appear (making it binary incompatible).

We could move them back and make rdf4j-model-api depend on rdf4j-util again, but that increases its footprint again. 

A better fix is to reorganize the rdf4j-util module, splitting it into separate "commons utilities" modules. That way the annotation classes can be moved back into a (leaner) commons module that rdf4j-model-api can then depend on again.

Possible new Maven structure (`oer` = `org.eclipse.rdf4j`):

- `rdf4j-commons` (POM only, contains submodules for various 'commons' functionality)
    - `rdf4j-commons-utilities` -> annotations and other classes (for which no more specific module can be found) for use throughout the rdf4j codebase. Intended to be as small as possible to decrease footprint. 
    - `rdf4j-commons-iterators`  -> everything in `oer.common.iterator`, `oer.common.iteration` and `oer.util.iterators`
    - `rdf4j-commons-transaction` -> everything in `oer.common.transaction` (at some point we'll migratate `IsolationLevel` there as well)
    - `rdf4j-commons-io` -> everything in `oer.common.io` and `oer.common.net`
    - `rdf4j-commons-xml` -> everything in `oer.common.xml`
    - `rdf4j-commons-text` -> everything in `oer.common.text`

Note that rdf4j-util would be removed. Effectively we get six additional modules, all relatively small. Downside is more modules to juggle, but possible upside is that it gives us more control when wanting to share functionality across the code base (_especially_ when sharing code across both client and server). It also gives us a better handle on what code in that "catch-all" bucket is out of date and no longer used, and sets us up nicely for the goal of making sure each package is only in a single module.

Issue #1494, opened 2019-08-09
Ensure each package is contained in a single module


jeenbroekstra:
Several packages in the `org.eclipse.rdf4j.*` namespace have classes and interfaces spread over multiple maven modules (and hence multiple jar files). To prepare our code base for a Java 9 modular architecture, we should migrate code such that each package is contained in _exactly_ one module. 

Examples: 

- [x]  `org.eclipse.rdf4j` has classes in both rdf4j-model and rdf4j-config
- [x] `org.eclipse.rdf4j.sail.lucene` has classes in both rdf4j-sail-lucene and rdf4j-sail-lucene-api
- [ ] all reusable testsuite modules reuse packages from the core components they're intended for.

jeenbroekstra:
I'm not confident we can do this before the 3.0 release. Especially the lucene packages are fairly intertwined, and there's cross-repository migrations that need to happen.

It will be a lot easier to do this once we're back to a monorepo.

PR #3225, opened 2021-07-24
GH-2710 module structure reorganization

jeenbroekstra:
GitHub issue resolved: #2710  

- rdf4j-util split into several smaller 'common' modules to allow smaller footprint of individual dependencies and easier modular boundaries
- removed some obsolete code (OpenRDFException, OpenRDFUtil) and migrated some other classes to different packages (RDF4JException, IsolationLevel)
- added upgrade pointers to release notes

This is quite a large refactor, but it gives us greater flexibility, cleans up a few old things, and sets us up to make followups like #1494 (each package in a single module) easier to achieve. 

----


jeenbroekstra:
Looks like our use of multithreading for the maven build does not agree with the new modular structure. Verified locally that a single-threaded build succeeds. Possibly caused by my relying on transitive dependencies for including libraries like commons-annotations in some places (via rdf4j-model). I'll see if more explicit dependencies can fix this, otherwise may temporarily disable multithreaded builds in our Github actions.

jeenbroekstra:
- [ ] update BOM and rdf4j-client pom
- [ ] verify rdf4j-util is no longer referenced

hmottestad:
It's a bit big for me to review. 

Does this affect any of the ShaclSail code in any way?

Could you confirm that this builds with the following method?
1. mvn clean
1. Delete the RDF4J folder within your .m2 folder (so that you don't have any cached versions of RDF4J)
1. Disconnect from the internet
1. mvn install -DskipTests

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/http/HTTPStoreConnectionTest.java
```diff
@@ -13,8 +13,8 @@
 
 import java.io.StringReader;
 
-import org.eclipse.rdf4j.IsolationLevel;
-import org.eclipse.rdf4j.RDF4JException;
+import org.eclipse.rdf4j.common.exception.RDF4JException;
+import org.eclipse.rdf4j.common.transaction.IsolationLevel;
```
barthanssens, line 17: Good idea to move IsolationLevel and RDF4JException  to common

core/common/text/src/main/java/org/eclipse/rdf4j/common/text/StringUtil.java
```diff
@@ -40,7 +40,10 @@
 	 * @param news The String is the new content.
 	 * @param text The String in which the substitution is done.
 	 * @return The result String containing the substitutions; if no substitutions were made, the result is 'text'.
+	 * 
+	 * @deprecated use {@link String#replace(CharSequence, CharSequence) instead.
 	 */
+	@Deprecated
```
barthanssens, line 46: Perhaps the method could be removed altogether, instead of just being deprecated ?

jeenbroekstra: It probably could, but I didn't want to make that decision in this PR.

jeenbroekstra:
> It's a bit big for me to review.

Yeah, I'd be happy with just checking out the branc
> 
> Does this affect any of the ShaclSail code in any way?

A little. There's some refactoring with the exception classes.

> Could you confirm that this builds with the following method?
> 
>     1. mvn clean
> 
>     2. Delete the org/eclipse/rdf4j folder within your .m2 folder (so that you don't have any cached versions of RDF4J)
> 
>     3. Disconnect from the internet
> 
>     4. mvn install -DskipTests

I just ran that (except I used `mvn -o` rather than physically disconnecting), and it fails, but funnily enough not because of this refactor, but because it can't find the serql parser (obviously have a few dangling dependencies somewhere).

I'll fix those in a separate PR, then rebase this one and try again.

jeenbroekstra:
Ran the offline build test again with the dependency fixes merged in, all green.

jeenbroekstra:
Unless there are major objections I'll merge this tomorrow morning - I'd like to get this in as soon as possible the minimize the risk of merge conflicts. Of course, I'll be on hand to help out on any other PRs that get conflicts _because_ of this.

Issue #3216, opened 2021-07-21
SHACL - RDFS reasoner needs to be more performant on larger schemas


hmottestad:
The VerySimpleRdfsBackwardsChainingConnection supports RDFS backwards chaining for the ShaclSail to conform to rdfs:subClassOf support mandated by the spec.

For backwards chaining rdf:type requests we have two options:
1. backwards chain the request and fire multiple hasStatements to super
1. backwards chain the request and fire a getStatements to super and match the backwards chained result

We are currently doing option 1, but if the backwards chained result is large then it might make sense to use option 2 instead.

If the reasoner receives a request for hasStatement(ex:a rdf:type ex:Mercedes) then we backwards chain ex:Mercedes and get [ex:Cat, ex:LuxuryCar, .....]. Now we can query super.getStatements(ex:a rdf:type ?type) and see if any of the returned ?type match our list.

PR #3235, opened 2021-07-26
GH-3216 VerySimpleRdfsBackwardsChainingConnection handles larger inferred sets

hmottestad:
GitHub issue resolved: #3216 

 - some simplification to return earlier
 - check size of inferred set and if it is above 10 then we use getStatements to get the actual rdf:type(s) and check them against the inferred set

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/VerySimpleRdfsBackwardsChainingConnection.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/VerySimpleRdfsBackwardsChainingConnection.java
@@ -26,6 +26,8 @@ import org.eclipse.rdf4j.model.vocabulary.RDF;
 import org.eclipse.rdf4j.sail.SailConnection;
 import org.eclipse.rdf4j.sail.SailException;
 import org.eclipse.rdf4j.sail.helpers.SailConnectionWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Very simple RDFS backwardschaining connection that supports type inference on hasStatement and getStatement. It does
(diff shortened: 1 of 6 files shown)


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/VerySimpleRdfsBackwardsChainingConnection.java
```diff
@@ -47,17 +50,34 @@
...
-					.stream()
-					.map(type -> super.hasStatement(subj, pred, type, false, contexts))
-					.reduce((a, b) -> a || b)
-					.orElse(false);
+
+			Set<Resource> types = rdfsSubClassOfReasoner.backwardsChain((Resource) obj);
+			if (types.size() > 10) {
+				try (CloseableIteration<? extends Statement, SailException> statements = super.getStatements(subj,
+						RDF.TYPE, null, false, contexts)) {
+					return statements.stream()
```
jeenbroekstra, line 66: So if I understand correctly, this is based on the assumption that typically, the number of (explicit) type statements for a given subject is going to be fewer than 10, correct? 

That's a sensible enough rule of thumb, though of course it may fluctuate wildly between domains. As a quick performance fix this is ok with me, but I'm thinking for a future larger improvement: rather than putting a hardcoded limit in, wouldn't it make sense to tie this in with the store statistics? I know those statistics are currently unaccessible outside the store implementation itself, might a thing to fix in 4.0.

jeenbroekstra (approved):
LGTM. Just one comment with a possible further future improvement (that will need some refactoring).

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/VerySimpleRdfsBackwardsChainingConnection.java
hmottestad: This is bang on!

I thought 10 was a decent size, but I'm sure this isn't the case for all users. 

Access to statistics would be great! We could always generate out of statistics here, but I feel  it would just add unnecessary complexity.

hmottestad: Opened an issue: https://github.com/eclipse/rdf4j/issues/3249

Issue #3218, opened 2021-07-21
SHACL - QualifiedMinCount should be optimized for bulk validation


hmottestad:
Essentially detect if AllTargets is provided as the overriding node and pass it along to the qualified shape.

Issue #3217, opened 2021-07-21
SHACL - Automatically simplify some plan nodes


hmottestad:
Plan nodes like Union and Unique should provide a getInstance method so they can return the provided node if it makes sense.

An example is if Union is called with only one node, in this case it should return the node instead of wrapping it in a union.

PR #3236, opened 2021-07-26
GH-3217 Union and Unique plan nodes can return the provided node if they are redundant

hmottestad:
GitHub issue resolved: #GH-3217 

 - use getInstance methods instead of constructors
 - check input so that for 
     - Unique we can return the provided node if the is already unique
     - UnionNode we can return the provided node if the effective node count is 1
 - also (because it was hard to untangle from this code) GH-3218 optimize QualifiedMinCount to better handle bulk validation

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/PropertyShape.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/PropertyShape.java
@@ -187,12 +187,12 @@ public class PropertyShape extends Shape implements ConstraintComponent, Identif
 //
 //				PlanNode allTargetsPlan = getAllLocalTargetsPlan(connectionsGroup, negatePlan);
 //
-//				Unique invalid = new Unique(planNode);
+//				Unique invalid = Unique.getInstance(planNode);
 //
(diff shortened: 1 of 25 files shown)


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/PropertyShape.java
```diff
@@ -187,12 +187,12 @@ public PlanNode generateTransactionalValidationPlan(ConnectionsGroup connections
 //
 //				PlanNode allTargetsPlan = getAllLocalTargetsPlan(connectionsGroup, negatePlan);
 //
-//				Unique invalid = new Unique(planNode);
+//				Unique invalid = Unique.getInstance(planNode);
```
jeenbroekstra, line 190: Does this commented out code block need to remain?

jeenbroekstra (approved):
LGTM.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/PropertyShape.java
hmottestad: It's the code to support pushing negation. I'd like to just leave it there for when/if I get back to reimplementing negation pushing.

Issue #3231, opened 2021-07-26
SHACL - Skip validation if there are no shapes or validation is disabled


hmottestad:
Validation should be skipped as early as possible if there are no shapes or if validation is disabled.

PR #3241, opened 2021-07-26
GH-3231 skip validation early

hmottestad:
GitHub issue resolved: #3231 

 - skip validation early if validation is disabled or we there are no shapes

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -388,7 +388,12 @@ public class ShaclSail extends NotifyingSailWrapper {
 	}
 
 	private void forceRefreshShapes() {
-		// there is no shapes cache, so this is no-op
+		if (shapesRepo != null) {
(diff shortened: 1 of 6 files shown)


hmottestad:
@jeenbroekstra this is the PR that I mentioned in the email. The changes to the transaction isolation of the shapes repo in particular. I also think the test coverage for that isolation aspect is poor.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -959,6 +959,17 @@ public IsolationLevel getDefaultIsolationLevel() {
 		return super.getDefaultIsolationLevel();
 	}
 
+	public boolean hasShapes() {
```
jeenbroekstra, line 962: Adding a public method makes an API change, which doesn't really fit with a patch release. But given that you're only using it from the ShaclSailConnection class you could just make it package-private I think.

hmottestad: Thanks!

jeenbroekstra (approved):

Issue #3029, opened 2021-04-28
SHACL - Configurable Max transaction size before failing over to bulk validation


hmottestad:
The SHACL validation can either be executed as transactional validation or bulk validation. 

Transactional validation uses the NotifyingSail to track all changes and attempts to optimize the validation to retrieve as little data from the database as possible. 

Bulk validation does not track any changes and instead revalidates the entire database at the end of the transaction. 

The performance differences between these two approaches depends on the ratio between the size of the database and the transaction. It also depends on the complexity of the shapes and the change, but that's not the scope for this issue. 

For the time being it'll be easier to implement a configurable limit rather than a ratio. 

**Tasks**
 - configurable limit of statements tracked per transaction 
 - functionality to unregister from the notifying sail, cleanup tracked data, enable bulk validation
 - logging for when this gets triggered
 - benchmarks to make sure it works and to decide on a decent default limit

PR #3245, opened 2021-07-26
GH-3029 limit transaction size for transactional validation in ShaclSail

hmottestad:
GitHub issue resolved: #3029  

 - NotifyingSailConnectionBase allows listeners to be removed while the transaction is ongoing
 - ShaclSailConnection removes itself as a listener once a certain number of statements has been reached
 - Limit is configurable by users on a sail basis
     - Name: TransactionalValidationLimit

----

### DIFF
--- a/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/config/ShaclSailSchema.java
+++ b/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/config/ShaclSailSchema.java
@@ -62,6 +62,7 @@ public class ShaclSailSchema {
 
 	public final static IRI VALIDATION_RESULTS_LIMIT_TOTAL = create("validationResultsLimitTotal");
 	public final static IRI VALIDATION_RESULTS_LIMIT_PER_CONSTRAINT = create("validationResultsLimitPerConstraint");
+	public final static IRI TRANSACTIONAL_VALIDATION_LIMIT = create("transactionalValidationLimit");
 
 	private static IRI create(String localName) {
 		return iri(NAMESPACE, localName);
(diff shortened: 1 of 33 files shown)


core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/config/ShaclSailSchema.java
```diff
@@ -62,6 +62,7 @@
 
 	public final static IRI VALIDATION_RESULTS_LIMIT_TOTAL = create("validationResultsLimitTotal");
 	public final static IRI VALIDATION_RESULTS_LIMIT_PER_CONSTRAINT = create("validationResultsLimitPerConstraint");
+	public final static IRI TRANSACTIONAL_VALIDATION_LIMIT = create("transactionalValidationLimit");
```
hmottestad, line 65: I'm open to name suggestions!

Our SHACL engine will analyse the changes in the transaction during validation, but to do this it also needs to track all those changes. Large changes also require large amounts of RAM to keep those changes in memory. Furthermore for larger changes it may be faster to just ignore the changes and instead revalidate the entire database during commit(). This last aspect kinda depends on how large your underlying database is.

hmottestad:
A database with 100 000 triples will take less time to revalidate than one with 100 000 000 triples.  So in theory a transaction with 50 000 triples might be faster to validate transactionally on the biggest database while using revalidation would be faster on the smaller database. 

This is however heavily dependent on the types of constraints. Datatype constraints scale really well, but qualified shape constraints do not.

In the future we could consider having each constraint decide which validation approach will be the fastest based on the changes in the transaction. 

Right now though I just want to introduce an upper limit so users don't run out of memory while uploading data.

hmottestad:
TODO: This doesn't seem to work correctly if it's the first transaction. So need to make some tests and see what we should do!

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/config/ShaclSailSchema.java
jeenbroekstra: Name sounds ok as it is. The concept it embodies probably needs to be explained in more detail in the documentation though.

jeenbroekstra (commented):
Looks good to me - I'm thinking you may want to update the SHACL Sail documentation though before you merge this.

hmottestad:
TODO:
- [ ] Update documentation

PR #3255, opened 2021-07-28
GH-3131: fix FedX resource leaks in error / cancel cases

aschwarte10:
GitHub issue resolved: #3131 

This change adds robustness to FedX in special error and cancelling
cases, e.g. when the query is only partially consumed. Before we have
seen situations where there were resource leaks and blocked threads.

Summary of changes:

* ParallleTaskBase has now performTaskInternal to remember the reference
to the closable iteration. Upon cancel() the closable iteration is also
closed. Subclasses of ParallelTaskBase now override performTaskInternal
* Scheduler classes that make use of a phaser (e.g.
ControlledWorkerJoin) now explicitly force termination of the phaser
upon close. This potentially unblocks running threads
* Close dependent connection now also handles unexpected error cases
(e.g. interrupts)
* Generally more safe-guards are added and inner iterations are closed
as early as possible

Note: MediumConcurrencyTest now has a helper method for executing queries partially. If this is used, the canceling case can be simulated due to partial consumption of query results.

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
@@ -46,9 +46,8 @@ public abstract class ParallelExecutorBase<T> extends LookAheadIteration<T, Quer
 	/* Variables */
 	protected volatile Thread evaluationThread;
 	protected FedXQueueCursor<T> rightQueue = FedXQueueCursor.create(1024);
-	protected CloseableIteration<T, QueryEvaluationException> rightIter;
-	protected volatile boolean closed;
-	protected boolean finished = false;
+	protected volatile CloseableIteration<T, QueryEvaluationException> rightIter;
+	protected volatile boolean finished = false;
 
 	public ParallelExecutorBase(FederationEvalStrategy strategy, QueryInfo queryInfo) throws QueryEvaluationException {
 		this.strategy = strategy;
@@ -58,25 +57,32 @@ public abstract class ParallelExecutorBase<T> extends LookAheadIteration<T, Quer
 
 	@Override
 	public final void run() {
+
+		if (isClosed()) {
+			return;
+		}
+
 		evaluationThread = Thread.currentThread();
 
 		if (log.isTraceEnabled()) {
 			log.trace("Performing execution of " + getDisplayId() + ", thread: " + evaluationThread.getName());
 		}
-
 		try {
 			performExecution();
 			checkTimeout();
+
+			if (log.isTraceEnabled()) {
+				log.trace(getDisplayId() + " is finished.");
+			}
+			done();
+
 		} catch (Throwable t) {
 			toss(ExceptionUtil.toException(t));
 		} finally {
+			// signal DONE to queue: unblock polling
+			rightQueue.done();
 			finished = true;
 			evaluationThread = null;
-			rightQueue.done();
-		}
-
-		if (log.isTraceEnabled()) {
-			log.trace(getDisplayId() + " is finished.");
 		}
 	}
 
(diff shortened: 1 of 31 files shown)


aschwarte10:
Note: this PR is following up on the findings reported by @damyan-ognyanov and his proposed fixes. I have borrowed one improvement fix from https://github.com/eclipse/rdf4j/pull/3132 and investigated further. While testing errroneous cases I found a number of further issues (see this PR for the summary).

@damyan-ognyanov what is your proposal for giving attribution? Should I add your name as author in the commit message as well?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
```diff
@@ -58,25 +57,35 @@ public ParallelExecutorBase(FederationEvalStrategy strategy, QueryInfo queryInfo
 
 	@Override
 	public final void run() {
-		evaluationThread = Thread.currentThread();
 
-		if (log.isTraceEnabled()) {
-			log.trace("Performing execution of " + getDisplayId() + ", thread: " + evaluationThread.getName());
```
jeenbroekstra, line 64: Was this removed accidentally?

jeenbroekstra (approved):
There's a lot going on here and obviously some pretty hairy conccurrency stuff, it's hard to judge from a diff review what exactly is going on, but each of the individual changes make sense. 

There's a couple of places where I initially thought "why isn't he using a try-with-resources" - it took me a while to realize that you are handing off the resource to another thread, so you can't do that. Might be worth putting in a small comment here and there to clarify that. 

All in all looks good to me from a fairly quick review

aschwarte10:
@jeenbroekstra thanks for the quick review. The concurrency mechanisms in the engine are really heavy, and I hope that this change makes it a bit better.

I see that there are failing tests, going to investigate now.

@damyan-ognyanov as mentioned above, let's synch on how we want to compose the final PR and combine your changes with the other improvements

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/ConsumingIteration.java
```diff
@@ -50,13 +50,21 @@ public ConsumingIteration(CloseableIteration<BindingSet, QueryEvaluationExceptio
```
damyan-ognyanov, line 50: Do not really like the idea of fetching solutions from argument iteration within the constructor.  There are possible scenarious where, while iterating, the underlying iteration may block (dueto  a lack of resources/threads in the pools) and the exectution may stall. Better to either postpone it until first `hasNext()` or reintroduce the configurable parameter (from my PR) that we could use to control the `max` size or even disable the caching entirely, if we need to do so ...

damyan-ognyanov (changes requested):
@aschwarte10 good work, I do not have any preference on how this PR is composed/attributes as long as it works and ends within the upcomming minor release (if possible). the only thing that bothers me so far is the constructor of `ConsumingIteration()` where solutions from an external iteration are being fetched. see my comment above ...

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelExecutorBase.java
aschwarte10: have fixed this again

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/ConsumingIteration.java
aschwarte10: I have added your changes to make this configurable as a second commit.

Would not change it ad-hoc right now to move the consuming part out of the constructor, I cannot oversee the consequences of that. I remember that this consuming was explicitly required (in the construction part) to avoid too many open connections for small intermediate result sets, i.e. kind of a buffer for the standard case.

If you want to further experiment, please share the details of your evaluations

damyan-ognyanov (approved):

Issue #3059, opened 2021-05-20
Contribution of a Spring/RDF4J integration


fkleedorfer:
As [discussed](https://www.eclipse.org/lists/rdf4j-dev/msg01828.html) on the mailinglist, we have built a spring integration for RDF4J and are looking into contributing it, this issue is just for tracking the progress and having a place for community discussions in case there is anything to discuss.

# Features

* Integration with the spring transaction system (modeled after [rdf4j-spring](https://github.com/CarbonLDP/rdf4j-spring))
* Result caching layer
* Operation (query/update) stats exposed via JMX
* Rdf4JTemplate (modeled after springs JDBCTemplate) and builders for assembling query/update executions
* DAO base class
* CRUD DAO base class
* a few minor non spring-specific things that have been missing or we thought would be useful, such as some expressions for the sparql builder (BIND, IN/NOT_IN),  a PropertyPath builder, 

# State of the Code

The code currently depends on rdf4j 3.2.2 and spring 5.2.7. Some of our developments are probably obsolete by now (e.g. the BIND expression). So before the contribution can be merged, there is some more maintenance to do, and it's likely that the code turns out to contribute to existing modules in addition to be picked apart into contributions . A closer look at our use of the rdf4j API will be greatly appreciated - I suspect we aren't always using it optimally. Also, the spring code is tested in domain-specific integration tests that cannot really be moved, so we'll have to write new tests. Anyone who wants to chip in to help with these tasks is highly welcome!

# Current Usage Context

In our project, we run a spring-boot server that offers a graphql endpoint and fetches the data via rdf4j. There may be a little bit of supporting code we could extract to make this architecture easier for users of our contribution - the Instrumentation class for commiting/rolling back the transaction after qraphql processing (or setting the transaction readonly if there are no mutations).

# Status

A pull request will follow as soon as I have the final approval from our management.

fkleedorfer:
Status update: still working on getting all the necessary people to approve. It's complicated.

jeenbroekstra:
> Status update: still working on getting all the necessary people to approve. It's complicated.

No worries @fkleedorfer, I know it can be tricky to get all the noses pointing in the same direction. Let us know if there's anything we can do.

fkleedorfer:
There is progress: our project consortium has until July 16 to object to this plan. After that, we can go ahead.

fkleedorfer:
Success! There haven't been any objections from project partners. As soon as I get to it, I'll start a PR and move the code over.

[...]

Issue #2767, opened 2021-01-20
Support BIND in SparqlBuilder


pdolezal:
I know there is [a limitation](https://rdf4j.org/documentation/tutorials/sparqlbuilder/#known-limitations) of [BIND](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#bind) support. I wanted to ask about the status of this implementation and plans to complete this feature. We want to switch from Spanqit, but we miss this part.

hmottestad:
I am not aware of any concrete plans to further develop the SparqlBuilder. We had some students fix a couple of bugs a while ago, and there are a number of issues documenting further improvements that users have requested. 

The biggest issue we faced last time someone worked on the SparqlBuilder was a significant lack of test coverage. 

We would very much welcome your help in implementing BIND support! I can assist in reviewing and testing your code.

fkleedorfer:
This is fixed in `develop`. Can we close it or do we wait for it to go to `main`?

PR #3269, opened 2021-08-01
Contribute Rdf4j Spring Integration

fkleedorfer:
Addresses https://github.com/eclipse/rdf4j/issues/3059

Note that this is work in progress, a number of things need to be done (help is greatly appreciated!)
* [ ] review `org.eclipse.rdf4j.model` and `org.eclipse.rdf4.sparqlbuilder.*`; and possibly extract them to other rdf4j modules (may be done after the PR is merged)
* [ ] add unit tests for individual components
* [ ] add a reference project to illustrate usage (starting from the minimal example in `spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/package-info.java`)
    - [ ] using one DAO
    - [ ] using an in-memory repo initialized from a TTL file
    - [ ] using all the optional subsystems


fkleedorfer:
@jeenbroekstra not sure how you want to proceed. Adding the tests could take some time, maybe you want to start reviewing the rest of the code?

jeenbroekstra:
Thanks for putting this up! I will try and find time to look at it soon (it's mostly weekends for me at the moment though). 

Before you continue adding tests etc, can you make sure that you've signed the ECA and that your git commit author address is the same address that you used to sign? The verification step of the build is currently not finding your record. 

Have a look at the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md), in particular the points on how to sign the Eclipse ECA. Also, while you're fixing up your commits anyway, can you prefix each commit message with the issue number? Something like `GH-3059 Initial commit of rdf4j-spring`, etc.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
```diff
@@ -0,0 +1,30 @@
+package org.eclipse.rdf4j.model.impl;
```
jeenbroekstra, line 1: This and other classes will need the RDF4J copyright header added. It should be this:

```
```

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Bind.java
```diff
@@ -0,0 +1,24 @@
+package org.eclipse.rdf4j.sparqlbuilder.constraint;
+
+import org.eclipse.rdf4j.sparqlbuilder.core.Assignable;
+import org.eclipse.rdf4j.sparqlbuilder.core.Variable;
+import org.eclipse.rdf4j.sparqlbuilder.graphpattern.GraphPattern;
+import org.eclipse.rdf4j.sparqlbuilder.util.SparqlBuilderUtils;
+
+public class Bind implements GraphPattern {
```
jeenbroekstra, line 8: Ah, interesting, there is actually a separate issue for that: GH-2767. This should probably be moved to the sparqlbuilder module though.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/propertypath/InversePredicatePath.java
```diff
@@ -0,0 +1,16 @@
+package org.eclipse.rdf4j.sparqlbuilder.constraint.propertypath;
+
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Iri;
+
+public class InversePredicatePath implements PredicatePathOrInversePredicatePath {
```
jeenbroekstra, line 5: Ah, I didn't realize there was lot more SparqlBuilder stuff. We'll need some time to go over this and figure out which parts are new and which overlap with existing functionality.

Might perhaps be a good idea to take these out of this PR and put them up separately? It's definitely going to be useful but it's not really related to the spring component, so I wouldn't want it to hold that up.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/exception/Rdf4JSpringException.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.rdf4j.spring.dao.exception;
+
+public class Rdf4JSpringException extends RuntimeException {
```
jeenbroekstra, line 3: Can we make this extend RDF4JException (which in turn extends RuntimeException)?

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/exception/Rdf4JDaoException.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.rdf4j.spring.dao.exception;
+
+public class Rdf4JDaoException extends Rdf4JSpringException {
```
jeenbroekstra, line 3: Minor editorial: in the existing code we tend to spell RDF4J all capitals when used in class names.

jeenbroekstra (commented):
I did a very quick first look through things, looking very good I think. Couple of inline comments, some editorial stuff. I think the biggest thing might be what to do with the SparqlBuilder extensions. They look very useful, but it's not really related to the spring stuff and getting it all tested and integrated might take some time. I'd suggest taking it out of this PR and submitting it in a separate PR.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
fkleedorfer: ok

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/propertypath/InversePredicatePath.java
fkleedorfer: Ok, I'll make a separate PR with just the SparqlBuilder stuff

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/exception/Rdf4JSpringException.java
fkleedorfer: ok

jeenbroekstra:
One more thing I almost forgot, kinda important: I will likely need to pass this by the Eclipse IP team as well, as it's quite a large contribution. I will take care of that, but just be aware that at some point there may be some questions coming your way regarding code ownership etc.

fkleedorfer:
>  just be aware that at some point there may be some questions coming your way regarding code ownership etc.

Thanks for the heads-up. I have notified our legal people that this is coming when starting the whole process.

fkleedorfer:
> can you prefix each commit message with the issue number? Something like GH-3059 Initial commit of rdf4j-spring, etc.

I thought we'd squash everything in the process of merging, no? If not, are you ok with the commits made so far being amended?

jeenbroekstra:
> > can you prefix each commit message with the issue number? Something like [GH-3059](https://github.com/eclipse/rdf4j/issues/3059) Initial commit of rdf4j-spring, etc.
> 
> I thought we'd squash everything in the process of merging, no? If not, are you ok with the commits made so far being amended?

We don't squash anymore, no - we used to do that for a while but there are technical issues with the IP attribution on the squashed commit that is created that way. So we're back to regular merge-commits for now unfortunately.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/propertypath/InversePredicatePath.java
fkleedorfer: On second thought: there is quite some use of the `ExtendedVariable` class in the spring-related code, so at least in that regard, the spring code depends on the sparqlbuilder code. How do you suggest we handle that? 

The only point of the `ExtendedVariable` is to get public access to the variable name, maybe we can just modify `Variable` to provide that?

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
fkleedorfer: done

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/propertypath/InversePredicatePath.java
fkleedorfer: I am keeping the code around for the time being

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/exception/Rdf4JSpringException.java
fkleedorfer: done (also for all other exceptions)

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/exception/Rdf4JDaoException.java
fkleedorfer: done

fkleedorfer:
BTW, I don't know what is up with the Eclipse ECA. I signed the agreement using my committer email address, but now when I try to sign it again, just to be sure, I get `Error, An error has occured and your Eclipse Contributor Agreement was not submitted. (ECA-04)`. Maybe that has something to do with the workflow failing?

jeenbroekstra:
> BTW, I don't know what is up with the Eclipse ECA. I signed the agreement using my committer email address, but now when I try to sign it again, just to be sure, I get `Error, An error has occured and your Eclipse Contributor Agreement was not submitted. (ECA-04)`. Maybe that has something to do with the workflow failing?

Looks like there is a known issue with the ECA database at the moment - see https://bugs.eclipse.org/show_bug.cgi?566859 . Fwiw I can manually check the ECA database and the address on your commits is not currently on file, but if I read that bug report correctly that might be a sync issue. Let's keep an eye on that bugzilla ticket and once that's resolved give it another go.

jeenbroekstra:
I have submitted a zip of the current state of this PR as the initial contribution for IP review.

jeenbroekstra:
ECA check now succeeds!

fkleedorfer:
I think I am done with the first round of the tests now. 

A few notes:

I feel like the `RDF4JCRUDDao` class might need some refacoring still, it feels a bit weird to implement it: all those methods with default implementations that throw exceptions should probably just be removed so subclasses must implement them. However, I don't have the time to do that right now. 

The 'reference project' mentioned in the PR description is more like a set of integration tests, but they fulfill the purpose. 

A shortcoming of the current implementation is that `RDF4JTemplate` does not support ASK queries. 

I haven't done any work with quads using this framework - we may need to refactor for everything to work with quads smoothly.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf.iri;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Iri;
+
+/**
+ * @since 4.0.0
+ * @author Gabriel Pickl
+ */
+public class MultiIRI extends SimpleIRI implements Iri {
```
jeenbroekstra, line 22: I am actually not convinced this utility wrapper is a good idea. And it shouldn't be part of the core model, but part of the sparqlbuilder module (if at all). I'd suggest we remove this class from the PR as it's not really related to Spring.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Bind.java
jeenbroekstra: If we're keeping this as part of this PR (which I'm fine with) we should move it to the sparqlbuilder module I think.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
fkleedorfer: It was added because of the need to a) write vocabulary classes (like RDF.java) with minimal effort and to use these IRIs in the sparqlbuilder. It's possible that this is just an artifact created by our limited knowledge of the RDF4J API at the time.

I think removing it would be ok, but the issues we had are probably shared by others and we may want to think about how to ease the creation/use of vocabulary classes, especially with the sparqlbuilder. 

There are some related classes which you might want to review:
* `ExtendedPrefix.java` - If we remove `MultiIRI`, we should remove that, too
* `ExtendedVariable.java` - this one is very useful for retrieving values from query results - I would much like to keep it, or change `Variable` such that the alias can be publicly retrieved.
* `TypeMappingUtils.java` 
* `QueryResultUtils.java` - uses `ExtendedVariable`

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Bind.java
fkleedorfer: I'll gladly move the code. However, there is much more in the `constraint` package in this module, including the whole `propertypath` subpackage. I suggest we move all of that to the sparqlbuilder module, keeping the package structure as is. Ok?

fkleedorfer: I'll gladly move the code. However, there is much more in the `constraint` package in this module, including the whole `propertypath` subpackage. I suggest we move all of that to the sparqlbuilder module, keeping the package structure as is. Ok?

fkleedorfer (commented):
Move code to sparqlbuilder module? 
... See code comments in Bind.java

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
fkleedorfer: An example to illustrate the problem that `MultiIRI` solves:

`org.eclipse.rdf4j.sparqlbuilder.rdf.RdfSubject#isA` takes `RdfObject...` as its parameters. Now if I want to use my vocabulary there, I have to write eg `PAINTING_ID.isA(Rdf.iri(EX.Painting))`. With MultiIRI, this just works without having to remember to pass the IRI through `Rdf.iri`. 

An alternative (and maybe better than `MultiIRI` would be to overload such methods with one that takes `IRI` arguments and delegate them to the original method after conversion to an `Iri`, e.g.:
```
default TriplePattern isA(IRI... objects) {
		return isA(Arrays.stream(objects).map(Rdf::iri).collect(Collectors.toList()).toArray(new Iri[0]));
	}
```

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Bind.java
jeenbroekstra: Ah yes, I should have been clearer sorry, I meant this comment for the whole package, not just this particular class. Thanks!

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
jeenbroekstra: Oh you're quite right that there is a "disconnect" between the sparqlbuilder APIs and the model API, so I fully understand why you'd want something like this. In several other places in the SparqlBuilder we have in fact solved this by overloading, like you suggested. 

I'd just like to keep this question separate from this PR, that's all, and not introduce this new public class until we've thought it through. The curse of maintaining a public library using Semantic Versioning is that once we introduce a new class or interface to the API, we're stuck with it for a long time - so I like to be careful about what we introduce. 

If you do wish to keep it, we should at the very least annotate it using the `@Experimental` or `@InternalUseOnly` annotations - those are our "cop-out" annotations for warning people that we'll not honor semantic versioning requirements if we change our mind later.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/resultcache/ResultCachingTupleQuery.java
```diff
@@ -0,0 +1,109 @@
...
+		}
+		this.localResultCacheRef = new WeakReference<>(localResultCache);
+	}
+
+	@Override
+	public TupleQueryResult evaluate() throws QueryEvaluationException {
+		BindingSet currentBindings = getDelegate().getBindings();
+		// TODO: this might be pretty slow due to the toString() call. Is there a better way to get
+		// a hash for a query with minmal risk of collision ?
+		Integer cacheKey = currentBindings.hashCode() + getDelegate().toString().hashCode();
```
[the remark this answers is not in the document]
jeenbroekstra: @fkleedorfer I must not have been paying attention before, as I wasn't really aware this PR introduced query result caching in this fashion. Sorry about that, my personal circumstance at the moment make it hard for me to spend as much time on RDF4J as I'd like.

Before we dive into the cache key mechanics: can you elaborate a little on what problem this cache mechanism solves? What exactly are we caching (and why), and what is the scope/duration of that cache? How is this cache controlled (e.g. how do we take care it does not get out of sync with the underlying repository)? Is any of this specific to Spring, or should we consider making this a general, separate component?

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/model/impl/MultiIRI.java
fkleedorfer: I removed MultiIRI and ExtendedPrefix from the PR. Probably won't come up again.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/resultcache/ResultCachingTupleQuery.java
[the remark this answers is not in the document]
fkleedorfer: I tried to explain what it does in `org/eclipse/rdf4j/spring/resultcache/package-info.java`. The problem it solves is answering multiple identical queries issued by the application over time.  The cache is connection-specific and is cleared upon connection close or any writes. If global caching is enabled, a global cache is also kept that is cleared whenever the application (any connection) writes. 
The cache is configured via `org.eclipse.rdf4j.spring.resultcache.ResultCacheProperties`, which provides a bit more insight as to how to control it.

To follow what is happening, look at `org.eclipse.rdf4j.spring.resultcache.CachingRepositoryConnection.prepareTupleQuery()`.

It is not strongly spring-related, but to extract it to another package might uproot other parts of the rdf4j-spring module.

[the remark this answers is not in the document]
fkleedorfer: The items that are cached are `org.eclipse.rdf4j.spring.resultcache.CachedGraphQueryResult` and `org.eclipse.rdf4j.spring.resultcache.CachedTupleQueryResult`. The whole result, basically.

jeenbroekstra:
Alternatively, if you give me write access to your fork, I can do a push and take care of this myself as well.

Issue #3322, opened 2021-09-09
Massive performance degradation and loss of results in SPARQL queries between RDF4J 2.5.2 and 3.7.2


PravinPrameKumar:
### Current Behavior

We are currently running RDF4J version 2.5.2 but need to upgrade to the latest version for security reasons. We tried upgrading to 3.7.2 but we found that the performance of many of our SPARQL queries have massively degraded - actually, they don't even complete execution. These same queries complete in under 20 milliseconds in version 2.5.2.
We can roughly see that the problematic queries have UNIONs in them. For one of these queries that hangs forever, we found a way to rewrite it without the UNION keyword which caused it to work again. This was not a trivial exercise, so it is not feasible for us to replace the hundreds of queries we have in our system. 

I isolated and stripped down a query and RDF store data to repro this. I found that the performance of this query scales with the (decreasing) size of the RDF store. A full TTL export of our test RDF store is about 15 MB. The query hangs when run on the full store (I didn’t bother waiting after many minutes). When I cut down the store to 7 MB, the query took about 56 seconds. From 5 to 2 MB, the query went from 46 to 8 seconds. Note that the sizes I refer to here are the ttl export sizes. The memorystore.data file for the attached example RDF store comes to 509 KB.
 I also found that the problematic queries run on 3.7.2 don’t return any results at all (they should and do return results on 2.5.2).

The attached zip file contains the following files, which describe and can be used to reproduce this issue:
Screenshots show the query running in the workbench and in our internal tool.
Logs are prefixed by 252 for the test machine with RDF4J 2.5.2 and 372 for version 3.7.2.
Here are some extracts showing the time difference for the two versions:
372:
2021-09-08 12:07:35,424 10008 [repositories/NativeJStore] DEBUG c.g.z.f.c.CompressingHttpServletResponse - Committing response without compression
2021-09-08 12:07:42,949 17533 [repositories/NativeJStore] INFO  o.e.r.h.s.r.TupleQueryResultView - Request for query -2131556384 is finished

252:
2021-09-08 12:07:23,803 28861 [repositories/NativeJStore] DEBUG c.g.z.f.c.CompressingHttpServletResponse - Committing response without compression
2021-09-08 12:07:23,805 28863 [repositories/NativeJStore] INFO  o.e.r.h.s.r.TupleQueryResultView - Request for query 3392903 is finished

I include two sets of logs (example1 and 2), no functional difference, I just ran the test twice.

### Expected Behavior

[...]

PR #3326, opened 2021-09-11
GH-3322 optimize union clauses to inject vars when we can

jeenbroekstra:
GitHub issue resolved: #3322  

- add new optimizer that switches scope change off for unions when possible: only should not be allowed when BIND or VALUES clauses are inside one of the union arguments
- this optimization allows us to use interleaved joining when combining VALUES clauses with unions in the majority of practical cases, which is much more efficient.

----


jeenbroekstra:
Note to self: the internal class that detects bind/values clauses may need some adjustments so that it doesn't trigger on occurrences in subselects or nested unions.

Issue #2046, opened 2020-03-28
Migrate to Java 11 as minimal requirement for RDF4J


jeenbroekstra:
Currently, RDF4J is built and guaranteed compatible with Java 8.  

However, newer versions of Java offer significant advantages, both in terms of quicker runtime performance, as well as features that offer benefits for the project's development. I'm think of support for modules, in particular, as part of our effort to stabilize our public API (see #619).  

I'm opening this issue to solicit opinions early on the pros and cons to having our next major version bump the minimal version of Java straight to version 11 (going with 11 as it is the LTS release). 

Some notes: 

1. If we bump versions for the next major release (say 4.0.0), we will for the medium term continue to do patch releases for 3.x.y in parallel - which will of course remain compatible with Java 8. 
2. it is an open question whether continued maintenance for 3.x.y will also include further minor releases (that is, adding new functionality), or only patches/bug fixes. 

I'm looking for opinions on whether our users, and third party implementors, are ready to make this jump. My personal opinion is that bumping JVM versions is a lot more acceptable in enterprise these days than it was in the days of Java 7 and 8.

barthanssens:
JDK11 LTS for a 4.0 sounds reasonable to me. I'm beginning to see migrations from 8 to 11 LTS in not-so-fast-moving organizations (YMMV), and IIRC public / free JDK8 updates from Oracle will probably end this December.

hmottestad:
We would appreciate comments from the following who are mentioned on our website: https://rdf4j.org/about/

 - Ontotext GraphDB
 - Halyard
 - Stardog
 - Amazon Neptune
 - Systap Blazegraph
 - MarkLogic RDF4J API
 - Strabon
 - Openlink Virtuoso RDF4J Provider

If you know someone who represents one of the above, it would be great if you could invite them to this discussion.

barthanssens:
I've sent a tweet to Ontotext, Stardog and SemanticWebCompany

barthanssens:
Response from ontotext (https://twitter.com/vassil_momtchev/status/1244548895893524482?s=20)

"Ironically Java 8 will outlive 11 in the enterprises. (see https://adoptopenjdk.net/support.html). Sadly from a developer's perspective, I don't think it's a good idea to drop the JDK8 support yet."

vassilmomtchev:
The official policy in Ontotext is:

- Keep compliant with OpenJDK
- Support Java 8 and 11 + all next LTS versions (I see no new releases on the horizon)
[...]

Issue #3350, opened 2021-09-27
Compile JDK17


barthanssens:
### Problem description

Check if RDF4J compiles on JDK17. Probably not an issue since we already compile on JDK16.

### Preferred solution

Compiles without changes

### Are you interested in contributing a solution yourself?

Yes

hmottestad:
Also need to update maven-source-plugin to the newest version covered by that CQ.

hmottestad:
And update the base docker image to jdk11 since jdk8 won't read the code compiled with jdk17.

PR #3384, opened 2021-10-30
GH-3350 update pom and docker to support jdk17

hmottestad:
GitHub issue resolved: #2046 
GitHub issue related: #3350 

 - updated some build plugins that don't work on jdk17 (versions already covered by CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20318 )
 - updated base docker image to jre11 which we need because we are targeting 11 with our compiler (using the docker image is the only way I know to test that I didn't break the war file)

----

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -23,8 +23,10 @@ jobs:
           ${{ runner.os }}-jdk${{ matrix.jdk }}-maven-
     - name: Check formatting
       run: mvn -B formatter:validate impsort:check xml-format:xml-check
+    - name: Check maven install
+      run: mvn -B clean install -DskipTests
     - name: Test
-      run: mvn -B clean test -P-formatting -DskipITs -Dmaven.javadoc.skip=true
+      run: mvn -B test -P-formatting -DskipITs -Dmaven.javadoc.skip=true
     - name: Print test failures
       if: failure()
       run: ./scripts/printTestResults.sh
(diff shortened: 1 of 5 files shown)


hmottestad:
@JervenBolleman I couldn't add you as a reviewer for this PR, but thought you might like to take a look.

hmottestad:
Do we need a CQ for the docker base image update?

jeenbroekstra:
> Do we need a CQ for the docker base image update?

I don't think we did one for the original, so let's leave it for now.

jeenbroekstra:
> updated base docker image to jre11 which we need because we are targeting 11 with our compiler (using the docker image is the only way I know to test that I didn't break the war file)

What I typically do is build the war using `mvn -Passembly package` and then drop it in a locally running tomcat server. I could imagine direcly running a docker image being easier though.

Could you share (perhaps as a writeup in the developer docs) how you use the docker image to test? What I mean is literally just the commands you use to produce the docker image and then run it. It's probably straightforward enough but we have almost zero documentation around our docker images, and this might be a good place to start.

hmottestad:
I use the bash script called build.sh and then I run "docker-compose up".

hmottestad:
I think there are some readme files in the same directory. Probably the biggest improvement would be to move the docker stuff to its own directory to increase visibility.

barthanssens:
> Do we need a CQ for the docker base image update?

IIRC this wans't necessary, because docker builds are not 'official' Eclipse deliverables (though memory could serve me wrong)

barthanssens:
> I think there are some readme files in the same directory. Probably the biggest improvement would be to move the docker stuff to its own directory to increase visibility.

Back in the days I wrote some stuff on the dev mailinglist as well...
Which reminds me I'll have to push the images to hub.docker.com... (at one point we were looking into doing this automatically)

hmottestad:
Related: #3389

Issue #3394, opened 2021-11-03
rdf4j-dev Develop build complains about Javadoc on jenkins


JervenBolleman:
Many javadoc issues are real I can try and fix a lot of them by changing my eclipse settings to spot them.

jeenbroekstra:
This is duplicating #3372

Issue #3372, opened 2021-10-22
Deployment of develop branch to sonatype fails with javadoc error


jeenbroekstra:
The daily deploy of the develop branch has been failing for a while now (notifications were turned off which is why I didn't notice earlier). It has an issue with the javadoc in the annotations package:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.8.1:jar (attach-javadocs) on project rdf4j-common-annotation: MavenReportException: Error while creating archive: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in http://download.oracle.com/javase/7/docs/api/ are in the unnamed module.
[ERROR] 
[ERROR] Command line was: /opt/tools/java/openjdk/jdk-11/latest/bin/javadoc @options @packages
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/home/jenkins/agent/workspace/rdf4j-develop-deploy/core/common/annotation/target/apidocs' dir.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :rdf4j-common-annotation
```

This will need fixing before we can do a milestone build or a release.

jeenbroekstra:
There's several bugs in the javadoc tool included in jdk 11.0.2 that have since been fixed, and some of them seem like they are directly affecting us (see e.g. https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8212233). I've requested that the openjdk versions on our Jenkins instance get patched: https://bugs.eclipse.org/bugs/show_bug.cgi?id=576968

jeenbroekstra:
Unfortunately, switching to AdoptOpenJDK 11 (which is at patch version 11.0.11) did not resolve the issue, so it looks as if the problem is elsewhere.

jeenbroekstra:
[...]

PR #3397, opened 2021-11-03
GH-3394 Used eclipse javadoc settings to fix many javadoc issues

JervenBolleman:
Found many and bulk fix a lot them. `<tt>` changed to html5 `<var>` with a regex but this might have missed a few cases or don it incompletly

GitHub issue resolved: #3394 

Found lots of syntax errors and fixed them removed some empty "@return" tags.
Would love to know which settings jenkins uses to get the same warnings.
Added implNote to the javadoc arguments.

----


jeenbroekstra:
I was actually already looking into the jenkins issues as part of #3372 - though I was more looking at it from the perspective of the build configuration (it really shouldn't error like this). 

A lot of these changes to the javadoc you made look potentially useful. Since the diff is rather large: can you quickly highlight/point out what exact configuration changes you made (and where)? 

> changed to html5 with a regex 

Can you clarify what you mean by this? As far as I know the build was _already_ configured to use html5

pom.xml
```diff
@@ -236,7 +236,7 @@
 				<jdk>[11,)</jdk>
 			</activation>
 			<properties>
-				<javadoc.opts>-Xdoclint:none -html5</javadoc.opts>
+				<javadoc.opts>-Xdoclint:none -html5 -tag implNote:a:"Implementation Note"</javadoc.opts>
```
jeenbroekstra, line 239: This is essentially adding implNote as a custom tag. It's a workaround fix but it really _shouldn't_ be necessary, as far as I know.

jeenbroekstra: Not that I object to a workaround fix if that gets our deployment unstuck, mind you :)

jeenbroekstra (approved):
Thanks for picking this up @JervenBolleman - it looks good to me! I'll merge this ASAP and hopefuly that will unblock our nightly build deployments.

Issue #3387, opened 2021-11-02
Add User Documentation for rdf4j-spring module


fkleedorfer:
### Problem description

The module `rdf4j-spring` has been added to the repository recently. Javadoc is present, but better user documentation has been requested, leading a new user to building a simple spring-boot application that uses an RDF4J repository as its data backend.

### Preferred solution

Two aspects seem necessary
1. Build a small demo project 
2. Write some user documentation, pointing to the demo project when appropriate and using it for creating source code or config snippets.

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

The right location for the demo project is a bit unclear. It has been suggested to make it a submodule of the rdf4j-examples module. This is a straightforward choice for us working on the codebase, but for someone who actually wants to tinker with the demo project, it is cumbersome: they would have to check out the whole huge rdf4j project. Also, linking to the demo from somewhere else isn't super straightforward, and it's not even easy to find it in the checked out project, let alone build it quickly. If we could create a `rdf4j-spring-demo` project in the eclipse organization, users could clone just that, which would be a lot easier and faster to do. I am not sure, however, if we have that option.

fkleedorfer:
@jeenbroekstra I'll start work on the demo as an rdf4j-examples module, for now, but as I said, I'd prefer it to be a top-level project on github somewhere.

fkleedorfer:
Correction: It can't be a submodule of `rdf4j-examples` as that module is has `jar` packaging, not `pom` packaging. I'll add the demo as a toplevel module for now, until it finds its final home. 

Apparently, some rdf4j examples are just demos of java code with essentially the same project config, for which it's ok to have them all in the same module. On the other hand, at least `rdf4j-spring-demo` requires its own module, as the project config is part of the demo. I don't know if there are other use cases for demos with dedicated modules/projects, but at least for `rdf4j-spring-demo`, it may be necessary in the future to maintain different versions of the demo for different spring versions. 

Options:
1. Refactor rdf4j-examples such that it becomes a parent module and put all the existing examples in a submodule; add `rdf4j-examples-rdf4j-spring` (as it would have to be called according to the naming convention) as another submodule 
2. Add another rdf4j submodule (`rdf4j-demos` maybe?) as a parent for all demos/examples that need a dedicated module, put `rdf4j-demo-rdf4j-spring` there
3. Create a separate toplevel project on github, e.g. `eclipse/rdf4j-examples`, which has one submodule for each example
[...]

PR #3398, opened 2021-11-04
Add Demo Project for RDF4J-Spring

fkleedorfer:
Related Issue: #3387 

Add folder `spring-components/rdf4j-spring-demo` containing a maven project that uses `rdf4j-spring`.

The `rdf4j-spring-demo` project is explicitly **not a maven submodule** of `rdf4j-spring-components` so the pom is self-contained, which makes it clearer to users how to start their own `rdf4j-spring` project.

As discussed in #3387, I am not sure that this is the best place for this demo project, but I think it is the best place I can put it without changes to the rdf4j folder/project structure.

----

### DIFF
--- a/spring-components/rdf4j-spring-demo/pom.xml
+++ b/spring-components/rdf4j-spring-demo/pom.xml
@@ -1,59 +1,54 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
-    <groupId>org.eclipse.rdf4j</groupId>
-    <artifactId>rdf4j-spring-demo</artifactId>
-    <name>RDF4J: Demo rdf4j-spring project</name>
-    <version>4.0.0-SNAPSHOT</version>
-    <description>Demo of a spring-boot project using an RDF4J repo as its backend</description>
-    <properties>
-        <spring.boot.version>2.4.3</spring.boot.version>
-    </properties>
-
-    <!-- explicitly NOT submodule of maven module in parent folder so as to show what a self-contained, minimal pom looks like -->
-    <dependencies>
-        <dependency>
-            <groupId>org.eclipse.rdf4j</groupId>
-            <artifactId>rdf4j-spring</artifactId>
-            <version>4.0.0-SNAPSHOT</version>
-        </dependency>
-
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter</artifactId>
-        </dependency>
-
-    </dependencies>
-
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>org.springframework.boot</groupId>
-                <artifactId>spring-boot-dependencies</artifactId>
-                <version>${spring.boot.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-        </dependencies>
-    </dependencyManagement>
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.8.1</version>
-                <configuration>
-                    <fork>true</fork>
-                    <release>11</release>
-                    <encoding>utf8</encoding>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.springframework.boot</groupId>
-                <artifactId>spring-boot-maven-plugin</artifactId>
-                <version>${spring.boot.version}</version>
-            </plugin>
-        </plugins>
-    </build>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>org.eclipse.rdf4j</groupId>
+	<artifactId>rdf4j-spring-demo</artifactId>
+	<name>RDF4J: RDF4J-Spring Demo</name>
+	<version>4.0.0-SNAPSHOT</version>
+	<description>Demo of a spring-boot project using an RDF4J repo as its backend</description>
+	<properties>
+		<spring.boot.version>2.4.3</spring.boot.version>
+	</properties>
+	<!-- explicitly NOT submodule of maven module in parent folder so as to show what a self-contained, minimal pom looks like -->
+	<dependencies>
+		<dependency>
+			<groupId>org.eclipse.rdf4j</groupId>
+			<artifactId>rdf4j-spring</artifactId>
+			<version>4.0.0-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter</artifactId>
+		</dependency>
+	</dependencies>
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-dependencies</artifactId>
+				<version>${spring.boot.version}</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>3.8.1</version>
+				<configuration>
+					<fork>true</fork>
+					<release>11</release>
+					<encoding>utf8</encoding>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
+				<version>${spring.boot.version}</version>
+			</plugin>
+		</plugins>
+	</build>
 </project>
--- a/spring-components/rdf4j-spring-demo/pom.xml
+++ b/spring-components/rdf4j-spring-demo/pom.xml
@@ -30,6 +30,17 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<scope>test</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>org.junit.vintage</groupId>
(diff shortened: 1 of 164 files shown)


spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/ArtDemoCli.java
```diff
@@ -0,0 +1,57 @@
+package org.eclipse.rdf4j.spring.demo;
```
jeenbroekstra, line 1: Missing copyright header

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/ArtDemoConfig.java
```diff
@@ -0,0 +1,28 @@
+package org.eclipse.rdf4j.spring.demo;
```
jeenbroekstra, line 1: Missing header

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/dao/ArtistDao.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * ******************************************************************************
```
jeenbroekstra, line 2: This header looks odd - something went wrong with the comment annotation i think? Also noticed this in a couple of other files.

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/model/EX.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+
+/**
+ * @since 4.0.0
+ * @author Florian Kleedorfer
+ */
+public class EX {
+	private static final String base = "http://example.org/";
+	public static final IRI Artist = SimpleValueFactory.getInstance().createIRI(base, "Artist");
```
jeenbroekstra, line 22: Could we replace this with `Values.iri` instead? It's what we're hoping to promote as the recommended pattern these days.

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/ArtDemoCli.java
```diff
@@ -0,0 +1,57 @@
...
+import org.eclipse.rdf4j.spring.demo.model.Artist;
+import org.eclipse.rdf4j.spring.demo.model.Painting;
+import org.eclipse.rdf4j.spring.demo.service.ArtService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.CommandLineRunner;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class ArtDemoCli implements CommandLineRunner {
```
jeenbroekstra, line 16: Might be good to put some javadoc on all classes in the demo that briefly explains what it does.

jeenbroekstra (commented):
Nice work! Particularly enjoying that even after 20+ years, the default example for anything RDF-related is still artists and artworks :)

I'm fine with putting this in this spot for now. We should make an effort to refactor the `examples` module so that we can put this in as an independent submodule before release though.

Left a few inline comments, mostly just about editorial stuff, like copyright headers and javadoc.

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/model/EX.java
fkleedorfer: I'll look into it. No reservations whatsoever.

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/dao/ArtistDao.java
fkleedorfer: It seems this is what happens when you copy/paste the rdf4j header into intellij's license profile. The best I can do with intellij (very helpful especially with new classes) is this:

```
/*
 * *****************************************************************************
 * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
 * *****************************************************************************
 */
```

fkleedorfer: the whole `rdf4j-spring` module has this problem. Is it ok if I fix that in this PR?

fkleedorfer:
For the user documentation, I am not sure where to put it - I mean, is it a tutorial, or rather a programming guide, or a tool?. I am going with `site/content/documentation/programming/spring.md` for now.

spring-components/rdf4j-spring-demo/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+					<encoding>utf8</encoding>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
+				<version>${spring.boot.version}</version>
+			</plugin>
+		</plugins>
+	</build>
```
hmottestad, line 63: Do any of these spring-boot plugins or dependencies need to have a CQ logged? As in a works-with CQ. Or maybe it's enough to check clearly defined: https://www.eclipse.org/projects/handbook/#ip-clearlydefined

It's a bit unclear if we need to do anything or not. There is a question about CQ for content on a project website, which is exempt because it's "separate from project content". Dunno if this would be.

> Are project websites subject to the IP Due Diligence Process?
>
> Website content is separate from project content. Project teams are expected to respect licenses for all web site content, but are not required to submit website content (including third-party libraries) for review by the IP Team.

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/dao/ArtistDao.java
jeenbroekstra: I'd really prefer if we could stick to the single, prescribed format for the license header. @hmottestad I know you are an IntelliJ user - can you advise on how you've configured it to use the correctly formatted copyright header? 

@fkleedorfer  as for fixing it in this in PR, that's not a problem.

spring-components/rdf4j-spring-demo/pom.xml
jeenbroekstra: We use spring-boot 2.4.x in other parts of the project, so in that sense this should be fine. I am a little confused why this uses a different patch version though (the actual spring-components module uses 2.4.12).

jeenbroekstra:
> For the user documentation, I am not sure where to put it - I mean, is it a tutorial, or rather a programming guide, or a tool? I am going with site/content/documentation/programming/spring.md for now.

That's a good enough place for now. In the end it depends on the writing style (tutorial vs. reference) more than anything else. And we can always reshuffle this if needed.

spring-components/rdf4j-spring-demo/pom.xml
fkleedorfer: I'll fix that on Monday

fkleedorfer:
update: the user docs are in the works, I'll probably have a first version finished on Monday

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/dao/ArtistDao.java
fkleedorfer: oh, no I meant, I'll make them look the way they are everywhere else, the intellij bit was just meant fyi. It would be nice of course if the formatter could be made to hammer the copyright into the required form, but it's not a problem - it's just a regex replace. So I'll fix that in all the places I touched.

fkleedorfer: done

fkleedorfer:
I finally decided to make `rdf4j-spring-demo` a submodule of `rdf4j-spring-components` after all. It just makes everything harder not to. Case in point: already now, the spring-boot patch version was different from the one used elsewhere in the project. Also, formatting is so much easier if it's a submodule. So, there's that.

I also fixed the copyright headers in `rdf4j-spring`, and I added the `PredictableUUIDSource` class (and friends).

fkleedorfer:
... and the user documentation is finished, so from my point of view, this PR is good to go.

spring-components/rdf4j-spring-demo/pom.xml
fkleedorfer: ... I finally decided to change the demo to be a full child module of `rdf4j-spring-components`. It's just less of a hassle.

fkleedorfer:
Sorry, the unit test support needs more work...

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/ArtDemoCli.java
```diff
@@ -0,0 +1,90 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: This is not the correct copyright header I'm afraid. It should be this:

```
 * Copyright (c) 2021 Eclipse RDF4J contributors.
```

Note that there's no ", Aduna, and others." <- that is something that is _only_ used for our legacy code from the Sesame project, not for anything new.

spring-components/rdf4j-spring-demo/pom.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ ******************************************************************************
+  ~ Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 4: No real need to put the header in an xml file, to be honest.

jeenbroekstra (commented):
Looks really good! There's unfortunately still a bit of a problem with the copyright header used (see inline comment) but other than that this is good to go.

I've created a followup issue (GH-3428) to look into reorganizing the examples module and moving this demo code there, but for now this is great.

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/ArtDemoCli.java
fkleedorfer: Sorry for that! I just opened a random source file and got the header from there. Not a safe practice, it seems ;-)

spring-components/rdf4j-spring-demo/pom.xml
fkleedorfer: ok, removed (from properties file, too)

spring-components/rdf4j-spring-demo/src/main/java/org/eclipse/rdf4j/spring/demo/ArtDemoCli.java
fkleedorfer: fixed.

jeenbroekstra (approved):

Issue #3407, opened 2021-11-09
Sail level query optimisers


hmottestad:
### Problem description

At the moment it is hard to create query optimisers that are specific to a sail. Our two main sails have widely different performance characteristics and it would be useful to create optimisers that can leverage knowledge about these specifics.

The current option is to call `setOptimizerPipeline(...)` in the `EvaluationStrategyFactory`. The main issue with this approach is that the `StrictEvaluationStrategyFactory` uses this pipeline to override the default pipeline of the `StrictEvaluationStrategy`. You could of course instantiate a `StandardQueryOptimizerPipeline` and get the default optimisers from there. This brings us to the second big issue which is that the `StandardQueryOptimizerPipeline` needs a number of arguments, namely the current `EvaluationStrategy` the transactions `TripleSource` and the `EvaluationStatistics`. These are all pretty hard to get a hold of at the sail level.

### Preferred solution

Add functionality to the `EvaluationStrategyFactory` in order to inject new optimisers at runtime.

This functionality should consist of:
 - a functional interface (supplier) that is provided with `EvaluationStrategy strategy, TripleSource tripleSource, EvaluationStatistics evaluationStatistics` in order to instantiate a `QueryOptimizer`
 - methods on the `EvaluationStrategyFactory` to add instances of the above mentioned interface (with the ability to specify if the new `QueryOptimizer` should be inserted at the beginning or end of the current pipeline)
 - getters for the above
 - getter for the pipeline in `EvaluationStrategy`

### Are you interested in contributing a solution yourself?

Yes

PR #3409, opened 2021-11-09
GH-3407 sail level query optimizers

hmottestad:
GitHub issue resolved: #3407 

### Briefly describe the changes proposed in this PR:

Add functionality to the `EvaluationStrategyFactory` in order to inject new optimisers at runtime.

This functionality should consist of:
 - a functional interface (supplier) that is provided with `EvaluationStrategy strategy, TripleSource tripleSource, EvaluationStatistics evaluationStatistics` in order to instantiate a `QueryOptimizer`
 - methods on the `EvaluationStrategyFactory` to add instances of the above mentioned interface (with the ability to specify if the new `QueryOptimizer` should be inserted at the beginning or end of the current pipeline)
 - getters for the above
 - getter for the pipeline in `EvaluationStrategy`

Create a `Value` optimiser for the MemoryStore to swap out values in the query with the "interned" value from the `MemValueFactory`.

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryOptimizerFunctionalInterface.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryOptimizerFunctionalInterface.java
@@ -1,0 +1,18 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.query.algebra.evaluation;
+
+import org.eclipse.rdf4j.query.algebra.evaluation.impl.EvaluationStatistics;
+
+public interface QueryOptimizerFunctionalInterface {
+
+	QueryOptimizer getOptimizer(EvaluationStrategy strategy, TripleSource tripleSource,
+			EvaluationStatistics evaluationStatistics);
+
+}
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/AbstractEvaluationStrategyFactory.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/AbstractEvaluationStrategyFactory.java
@@ -7,9 +7,13 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.impl;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategyFactory;
+import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizerFunctionalInterface;
 import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizerPipeline;
 
 /**
@@ -25,6 +29,8 @@ public abstract class AbstractEvaluationStrategyFactory implements EvaluationStr
 	private boolean trackResultSize;
 
 	private QueryOptimizerPipeline pipeline;
+	private final List<QueryOptimizerFunctionalInterface> queryOptimizersPre = new ArrayList<>();
+	private final List<QueryOptimizerFunctionalInterface> queryOptimizersPost = new ArrayList<>();
 
 	@Override
 	public void setQuerySolutionCacheThreshold(long threshold) {
@@ -54,5 +60,23 @@ public abstract class AbstractEvaluationStrategyFactory implements EvaluationStr
 	@Override
 	public void setTrackResultSize(boolean trackResultSize) {
 		this.trackResultSize = trackResultSize;
+	}
+
+	@Override
+	public void addOptimizer(QueryOptimizerFunctionalInterface queryOptimizerFunctionalInterface,
+			boolean beforeOtherQueryOptimizers) {
+		if (beforeOtherQueryOptimizers) {
+			queryOptimizersPre.add(queryOptimizerFunctionalInterface);
+		} else {
+			queryOptimizersPost.add(queryOptimizerFunctionalInterface);
+		}
+	}
+
+	public List<QueryOptimizerFunctionalInterface> getQueryOptimizersPre() {
+		return Collections.unmodifiableList(queryOptimizersPre);
+	}
+
+	public List<QueryOptimizerFunctionalInterface> getQueryOptimizersPost() {
+		return Collections.unmodifiableList(queryOptimizersPost);
 	}
 }
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
@@ -15,11 +15,21 @@ import java.util.TimerTask;
 import org.eclipse.rdf4j.common.concurrent.locks.Lock;
 import org.eclipse.rdf4j.common.transaction.IsolationLevels;
 import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.query.Binding;
+import org.eclipse.rdf4j.query.BindingSet;
+import org.eclipse.rdf4j.query.Dataset;
+import org.eclipse.rdf4j.query.algebra.BindingSetAssignment;
+import org.eclipse.rdf4j.query.algebra.TupleExpr;
+import org.eclipse.rdf4j.query.algebra.ValueConstant;
+import org.eclipse.rdf4j.query.algebra.Var;
 import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategy;
 import org.eclipse.rdf4j.query.algebra.evaluation.EvaluationStrategyFactory;
+import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizer;
 import org.eclipse.rdf4j.query.algebra.evaluation.federation.FederatedServiceResolver;
 import org.eclipse.rdf4j.query.algebra.evaluation.federation.FederatedServiceResolverClient;
 import org.eclipse.rdf4j.query.algebra.evaluation.impl.StrictEvaluationStrategyFactory;
+import org.eclipse.rdf4j.query.algebra.helpers.AbstractQueryModelVisitor;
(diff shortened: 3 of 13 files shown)


hmottestad:
@jeenbroekstra you want to take a look?

Some things I'm uncertain about:
 - naming
 - how best to handle updating the Value in a Binding in a BindingSet
 - if I should extract the new optimizer from the MemoryStore into it's own file

jeenbroekstra:
Sounds good, I'll try and take a closer look later this week.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryOptimizerFunctionalInterface.java
```diff
@@ -0,0 +1,18 @@
...
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.query.algebra.evaluation;
+
+import org.eclipse.rdf4j.query.algebra.evaluation.impl.EvaluationStatistics;
+
+public interface QueryOptimizerFunctionalInterface {
```
jeenbroekstra, line 13: I don't understand the point of this functional interface to be honest. It does not seem to make the code easier or more compact.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/AbstractEvaluationStrategyFactory.java
```diff
@@ -55,4 +61,22 @@ public boolean isTrackResultSize() {
...
+		}
+	}
+
+	public List<QueryOptimizerFunctionalInterface> getQueryOptimizersPre() {
+		return Collections.unmodifiableList(queryOptimizersPre);
+	}
+
+	public List<QueryOptimizerFunctionalInterface> getQueryOptimizersPost() {
+		return Collections.unmodifiableList(queryOptimizersPost);
+	}
```
jeenbroekstra, line 81: Why compile two separate collections (that then later on get concatenated into a single pipeline) instead of just immediately adjusting the pipeline itself?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
```diff
@@ -201,6 +213,70 @@ public long getSyncDelay() {
 	public synchronized EvaluationStrategyFactory getEvaluationStrategyFactory() {
 		if (evalStratFactory == null) {
 			evalStratFactory = new StrictEvaluationStrategyFactory(getFederatedServiceResolver());
+			evalStratFactory.addOptimizer((strategy, tripleSource, evaluationStatistics) -> {
+				return new QueryOptimizer() {
```
jeenbroekstra, line 217: If `addOptimizer` accepts a `QueryOptimizer` (instead of a `QueryOptimizerFunctionalInterace`) this lambda becomes unnecessary.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemValueFactory.java
```diff
@@ -103,14 +103,14 @@ public MemValue getMemValue(Value value) {
 	 * See getMemValue() for description.
 	 */
 	public MemResource getMemResource(Resource resource) {
-		if (resource instanceof IRI) {
+		if (resource == null) {
+			return null;
+		} else if (resource.isIRI()) {
 			return getMemURI((IRI) resource);
-		} else if (resource instanceof BNode) {
+		} else if (resource.isBNode()) {
```
jeenbroekstra, line 110: Good catch.

jeenbroekstra (commented):
I like the idea of injecting store-specific optimizers - but I'm not sure I understand the mechanics of your approach.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryOptimizerFunctionalInterface.java
hmottestad: The reason is that none of the variables are available when creating a EvaluationStrategyFactory, so they need to be passed at a later point.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/AbstractEvaluationStrategyFactory.java
hmottestad: We don't have access to the pipeline yet. It doesn't get set until we instantiate an actual EvaluationStrategy. The pipeline in the EvaluationStrategyFactory can only be used if we don't want to use any of the standard optimizers.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/MemoryStore.java
hmottestad: Won't get access to any of the variables then. Eg. strategy, tripleSource and evaluationStatistics

hmottestad:
@jeenbroekstra maybe it's easier if I name it something referring to "inject", as in that we can inject optimizers into the existing pipeline.

JervenBolleman:
I feel this is not quite the right direction for the OptimizerPipeline. I love the idea of store related optimizers but I think this is not a responsibility of the `EvaluationStrategy`. An query should be optimized before it is given to the `EvaluationStrategy` to evaluate. There should basically be a `QueryOptimizerPipeline` per store that builds on the Standard one. Separating this would make it easier to customize the behaviour of the optimizer pipelines. Also less shared state in the `EvaluationStrategy` would be nice, while cheap building the optimizer pipleline would now become something that must be done for each query instead of once for each store/connection. An other option is to inject this in a  `EvaluationStrategyFactory` with a simple constructor call.

hmottestad:
How would I optimize the query before passing it to the evaluation factory?

hmottestad:
Also I think that we are currently building a new evaluation strategy and optimizer pipeline for each query.

JervenBolleman:
I would just have an optimizer pipeline object and encourage it to be called before the query/tuplexpr is passed in.
Something like this in `SailSourceConnection`
```java
// inside evaluateInternal
TripleSource tripleSource = new SailDatasetTripleSource(vf, rdfDataset);
EvaluationStrategy strategy = getEvaluationStrategy(dataset, tripleSource);
...
//		tupleExpr = strategy.optimize(tupleExpr, store.getEvaluationStatistics(), bindings);
// instead do
tupleExpr = new StandardQueryOptimizerPipeline().optimize(tupleExpr);
// or even better
tupleExpr = getOptimizerPipeline().optimize(tupleExpr);
```
then `getOptimizerPipeline()` could be something like

```java
protected QueryOptimizerPipeline getOptimizerPipeline(){
return new QueryOptimizerPipeline(){
@Override
	public Iterable<QueryOptimizer> getOptimizers() {
		return Arrays.asList(
				new BindingAssigner(),
				new BindingSetAssignmentInliner(),
				new ConstantOptimizer(strategy),
				new RegexAsStringFunctionOptimizer(tripleSource.getValueFactory()),
				new CompareOptimizer(),
				new ConjunctiveConstraintSplitter(),
				new DisjunctiveConstraintOptimizer(),
				new SameTermFilterOptimizer(),
				new UnionScopeChangeOptimizer(),
				new QueryModelNormalizer(),
				new QueryJoinOptimizer(evaluationStatistics),
				new IterativeEvaluationOptimizer(),
				new FilterOptimizer(),
				new OrderLimitOptimizer(),
				new ParentReferenceCleaner(),
                                new ReplaceValuesWithStoreNativeInstances(getValueFactory())); // your new optimizer step
	}
} 
}
```
or just a final variable of the connection

What do you think?

hmottestad:
It's currently the strategy that is responsible for optimizing the query. I still feel that it's a natural place to add the extension to support sail level optimizers.

JervenBolleman:
@hmottestad  okay, sounds good to me.

jeenbroekstra (approved):
I think I'm ok with this now - thanks for clarifying @hmottestad. I'd like to hear @JervenBolleman 's current opinion as well if possible.

JervenBolleman:
I am ok with this as is. @hmottestad would you mind testing out how this codes by updating the Fedx optimization steps with this approach. As there the strategy is not new per query but retrieved from a federation manager.

hmottestad:
> I am ok with this as is. @hmottestad would you mind testing out how this codes by updating the Fedx optimization steps with this approach. As there the strategy is not new per query but retrieved from a federation manager.

Fedex doesn't use the `EvaluationStrategyFactory` interface. So I'm not quite sure where to get started there.

JervenBolleman:
I really like the actual optimizer step by the way, that is super sweet. 

> > I am ok with this as is. @hmottestad would you mind testing out how this codes by updating the Fedx optimization steps with this approach. As there the strategy is not new per query but retrieved from a federation manager.
> 
> Fedex doesn't use the `EvaluationStrategyFactory` interface. So I'm not quite sure where to get started there.

I opened #3483 to discuss the FedX issue of it not using an `EvaluationStrategyFactory`. Maybe we can fix that and then this issue is cleaner to apply? what do you think? Somehow I feel that there is cleanup to be done in this section and then the code ends up nicer. Maybe if we make the `EvaluationStrategyFactory` less mutable I think the OptimizerPipelines might have all the required variables at initialization and we don't need the methods `getQueryOptimizersPre` etc.

Still that can all be done in a different pull request and there is no need to do that work in this one.

hmottestad:
The way the optimizer pipeline can be set from the `StrictEvaluationStrategyFactory` when instantiating a `StrictEvaluationStrategy` is quite odd. It first instantiates the strategy, then checks if it should override the pipeline. Instead of just passing in the optimizer pipeline.

Issue is that it doesn't have access to the variables that the `StandardQueryOptimizerPipeline` needs.

Would probably be a lot cleaner if the pipeline was stored in the `EvaluationStrategyFactory` and then passed into the `EvaluationStrategy`. 

Maybe a `QueryOptimizerPipelineFactory` that we set in the `EvaluationStrategyFactory`, which then generates the pipeline when we instantiate the `EvaluationStrategy`. The `QueryOptimizerPipelineFactory` could then cache the optimizers it wants to and reuse them, or create new ones if there are optimizers that are dependent on the transactions state.

Then we can extend a `StandardQueryOptimizerPipelineFactory` to add our own optimizers to a specific Sail.

jeenbroekstra (changes requested):
Noticing a few "wip" commits in the log so assuming for now that this needs some further attention before it's ready to be merged.

hmottestad:
Yep, I want to try the pipeline-factory approach.

hmottestad:

PR #3413, opened 2021-11-10
GH-3416 LMDB-based SAIL.

kenwenzel:
- based on NativeStore
  - uses similar encodings for triples and values
  - uses (varint encoded) long IDs everywhere
- uses LWJGL interfaces to LMDB
  - fast implementation that is available for multiple platforms  
- uses one db for indexes and one db for values

GitHub issue resolved: #3416

This change adds a full LMDB-based SAIL that stores values and triples in two LMDB databases.

----


kenwenzel:
```
#LMDB
Benchmark                                                       Mode  Cnt      Score      Error  Units
TransactionsPerSecondBenchmark.largerTransaction               thrpt    5      2.355 ±    0.128  ops/s
TransactionsPerSecondBenchmark.largerTransactionLevelNone      thrpt    5      2.849 ±    0.075  ops/s
TransactionsPerSecondBenchmark.mediumTransactionsLevelNone     thrpt    5   2502.306 ±  271.070  ops/s
TransactionsPerSecondBenchmark.transactions                    thrpt    5   9820.817 ±  819.341  ops/s
TransactionsPerSecondBenchmark.transactionsLevelNone           thrpt    5  12051.377 ± 1414.688  ops/s
TransactionsPerSecondBenchmark.veryLargerTransactionLevelNone  thrpt    5      0.023 ±    0.004  ops/s

Benchmark                                               Mode  Cnt     Score     Error  Units
QueryBenchmark.complexQuery                             avgt    5     0.283 ±   0.053  ms/op
QueryBenchmark.distinctPredicatesQuery                  avgt    5  3231.650 ± 157.282  ms/op
QueryBenchmark.groupByQuery                             avgt    5     0.121 ±   0.012  ms/op
QueryBenchmark.removeByQuery                            avgt    5     2.764 ±   0.061  ms/op
QueryBenchmark.removeByQueryReadCommitted               avgt    5     5.022 ±   0.347  ms/op
QueryBenchmark.simpleUpdateQueryIsolationNone           avgt    5     0.286 ±   0.014  ms/op
QueryBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt    5     0.280 ±   0.017  ms/op
```
```
# Native Store
Benchmark                                                       Mode  Cnt    Score     Error  Units
TransactionsPerSecondBenchmark.largerTransaction               thrpt    5    4.757 ±   1.746  ops/s
TransactionsPerSecondBenchmark.largerTransactionLevelNone      thrpt    5    6.601 ±   1.447  ops/s
TransactionsPerSecondBenchmark.mediumTransactionsLevelNone     thrpt    5  602.569 ±  48.934  ops/s
TransactionsPerSecondBenchmark.transactions                    thrpt    5  625.174 ± 122.350  ops/s
TransactionsPerSecondBenchmark.transactionsLevelNone           thrpt    5  570.538 ± 190.161  ops/s
TransactionsPerSecondBenchmark.veryLargerTransactionLevelNone  thrpt    5    0.056 ±   0.025  ops/s

Benchmark                                               Mode  Cnt      Score      Error  Units
QueryBenchmark.complexQuery                             avgt    5     78.985 ±   67.389  ms/op
QueryBenchmark.distinctPredicatesQuery                  avgt    5   2353.538 ±  564.378  ms/op
QueryBenchmark.groupByQuery                             avgt    5     37.243 ±   21.944  ms/op
QueryBenchmark.removeByQuery                            avgt    5   2547.820 ± 1349.694  ms/op
QueryBenchmark.removeByQueryReadCommitted               avgt    5  10934.396 ± 1283.774  ms/op
QueryBenchmark.simpleUpdateQueryIsolationNone           avgt    5  10116.993 ± 4609.040  ms/op
QueryBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt    5  11694.331 ± 2701.513  ms/op
```

hmottestad:
That query performance is pretty impressive!

kenwenzel:
> That query performance is pretty impressive!
> 
> Edit: Any chance it's too good to be true? It's miles ahead of the MemoryStore.

I'm also not sure if the numbers are correct. I've re-run the benchmarks several times with the same results.
To get more confidence I've also added corresponding tests in compliance/sparql. Those tests pass - at least on my computer.

BTW: I've used MDB_NOSYNC and MDB_NOMETASYNC for the benchmarks. Maybe most of the data is cached in (off-heap) memory.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ContextStore.java
```diff
@@ -0,0 +1,230 @@
...
+ * Each context is encoded in the file as a record, as follows:
+ *
+ * <pre>
+ *   byte 1 - 8      : the number of statements in the content, as a long.
+ *   byte 9          : boolean flag indicating the type of context identifier (1 = IRI, 0 = blank node)
+ *   byte 10 - 11    : the length of the encoded context identifier
+ *   byte 12 - A     : the UTF-8 encoded the encoded context identifer
+ * </pre>
+ *
+ * @author Jeen Broekstra
```
JervenBolleman, line 56: I think this Author field should be different.

JervenBolleman:
I think this is fantastic. I would love to have it with longs as identifiers. Making it usable for large datasets like UniProt or Wikidata.

I put some small issues in already from a brief first look. But to me it looks like good quality code.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ContextStore.java
kenwenzel: Some files like ContextStore.java are just (almost) verbatim copies from NativeStore. Therefore I didn't touch the author fields.
Should I change or extend the field even for small changes?

kenwenzel:
> I think this is fantastic. I would love to have it with longs as identifiers. Making it usable for large datasets like UniProt or Wikidata.
> 
> I put some small issues in already from a brief first look. But to me it looks like good quality code.

I also think that longs are the way to go. To keep storage space efficiency also [VarInts](https://en.wikipedia.org/wiki/Variable-length_quantity) can be used.

core/sail/lmdb/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+			<artifactId>lwjgl</artifactId>
+			<classifier>natives-macos</classifier>
+			<version>${lwjgl.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.lwjgl</groupId>
+			<artifactId>lwjgl</artifactId>
+			<classifier>natives-windows</classifier>
+			<version>${lwjgl.version}</version>
+		</dependency>
```
jeenbroekstra, line 56: We will need to to check the IP status of these and if necessary file CQs for them.

compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbComplexSPARQLQueryTest.java
```diff
@@ -0,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra, line 2: This copyright header needs to be changed - even if you copied it from an existing file, it's essentially a new contribution I'd say.

jeenbroekstra (commented):
This looks very impressvie @kenwenzel , thanks! I'll try and find time soon to do a more comprehensive review as soon as possible. 

I've noticed you have copy-pasted a lot of existing code into this new module to be able to reuse it (can tell by the use of old copyright headers and author tags). Two things: 

1. I'd prefer treating these as "new" (so with a new copyright header and author) even if they are derived from existing code.
2. I haven't looked in detail but do we need code duplication in this fashion, or is there some way we can organize this to make code reuse a little easier?

core/sail/lmdb/pom.xml
jeenbroekstra: We'll need to file CQs against this: they are not sufficiently documented in ClearlyDefined, and not previously registered for an IP check with Eclipse Foundation either. I'll start filing some CQs.

jeenbroekstra:
@kenwenzel can you please create an issue in our issue tracker that describes the proposed improvement, and link this PR to it?

hmottestad:
I'm on OS X with the new ARM based  M1 processor. Seems that support is coming in the next release (3.3.0).

https://github.com/LWJGL/lwjgl3/issues/601

I'm currently not able to run the tests or benchmarks. We could consider trying out the 3.3.0 snapshot version.

kenwenzel:
Short update from my side: I'm going to switch to varint encoded long ids and I'm planning to push this change within the next few days.
I'm also working on a zero-copy design for the record iterators on another branch.

kenwenzel:
> @kenwenzel can you please create an issue in our issue tracker that describes the proposed improvement, and link this PR to it? And if you have them, I'd appreciate some pointers to the source of the LMDB libraries that you're using, and any licensing information for them - I'll need that info to log them for IP review.

You mean besides #3416? So I should create an LMDB-specific issue with some explanation?

jeenbroekstra:
Ah, no, you can use that issue, but it was not clear to me that was what you were using. Can you link them up by mentioning the issue number in your commit messages and the PR description please?

kenwenzel:
> I'm on OS X with the new ARM based M1 processor. Seems that support is coming in the next release (3.3.0).
> 
> [LWJGL/lwjgl3#601](https://github.com/LWJGL/lwjgl3/issues/601)
> 
> I'm currently not able to run the tests or benchmarks. We could consider trying out the 3.3.0 snapshot version.

I've pushed a commit with an update to LWJGL 3.3.0.

kenwenzel:
A short update: I have some open issues with this.
- drop custom comparator and use natural sort order for indexes (callbacks from native to Java are slow with LWJGL)
- change group varint encoding to be compatible with natural sort order of the numbers (prerequisite for the above)
- use varints everywhere in value store

hmottestad:
I got it running on my laptop :) Thanks!

I took a look at the benchmarks and changed the complex query one to this:

```
	@Benchmark
	public long complexQuery() {

		try (SailRepositoryConnection connection = repository.getConnection()) {
			long count = connection
				.prepareTupleQuery(query4)
				.evaluate()
				.stream()
				.count();
			System.out.println(count);
			return count;
		}
	}
```

It print `0` which is wrong. You can change the one in the MemoryStore and see what it's supposed to be.

core/sail/lmdb/pom.xml
```diff
@@ -0,0 +1,173 @@
...
+			<artifactId>rdf4j-repository-manager</artifactId>
+			<version>${project.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-queryparser-serql</artifactId>
+			<version>${project.version}</version>
+			<scope>test</scope>
+		</dependency>
```
hmottestad, line 136: SERQL shouldn't be included, it's been removed.

kenwenzel:
> I got it running on my laptop :) Thanks!
> 
> I took a look at the benchmarks and changed the complex query one to this:
> 
> ```
> 	@Benchmark
> 	public long complexQuery() {
> 
> 		try (SailRepositoryConnection connection = repository.getConnection()) {
> 			long count = connection
> 				.prepareTupleQuery(query4)
> 				.evaluate()
> 				.stream()
> 				.count();
> 			System.out.println(count);
> 			return count;
> 		}
> 	}
> ```
> 
> It prints `0` which is wrong. You can change the one in the MemoryStore and see what it's supposed to be.

Thank you! I will dig into this.

jeenbroekstra:
Minor remark - you're using isse number 3415 in several commits, but the actual related issue is #3416.

kenwenzel:
> Minor remark - you're using isse number 3415 in several commits, but the actual related issue is #3416.

Oops, I'll rewrite the commits when it is finally working.

core/sail/lmdb/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+			<artifactId>lwjgl-lmdb</artifactId>
+			<classifier>natives-macos-arm64</classifier>
+			<version>${lwjgl.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.lwjgl</groupId>
+			<artifactId>lwjgl-lmdb</artifactId>
+			<classifier>natives-macos</classifier>
+			<version>${lwjgl.version}</version>
+		</dependency>
```
jeenbroekstra, line 38: Can we make the various natives extensions optional dependencies perhaps? It seems a bit wasteful to force downloading all of them when any particular deployment will only need one. Also thinking about the size of our fat jars :)

kenwenzel:
> That query performance is pretty impressive!
> 
> Edit: Any chance it's too good to be true? It's miles ahead of the MemoryStore.

@hmottestad I hope that I've fixed the problems now.
I have to investigate why complexQuery is much slower with LMDB than with NativeStore.

core/sail/lmdb/pom.xml
kenwenzel: Yes, I think this is possible. How are optional dependencies handled in unit tests?

jeenbroekstra: That should not make a difference. An optional dependency is still available for us. It just means that any project that depends on our module does not automatically include them, transitively.

core/sail/lmdb/pom.xml
jeenbroekstra: I've gotten CQs for the two main libraries (lwjgl and lwjgl-lmdb) approved. If we include the native extensions as optional runtimes only, we won't need to file additional CQs for those (especially since they're just different compilations of essentially the same code).

jeenbroekstra:
@kenwenzel this is up to you, but if you are comfortable with some of us pairing up to get this polished, can you give @eclipse/technology-rdf4j-committers access to your forked repo so we can push to your branch?

kenwenzel:
> @kenwenzel this is up to you, but if you are comfortable with some of us pairing up to get this polished, can you give @eclipse/technology-rdf4j-committers access to your forked repo so we can push to your branch?

@jeenbroekstra  Sure, how can I do this? I can't find "eclipse/technology-rdf4j-committers" in the members list.

kenwenzel:
Besides polishing the following points are left:
- use VarInts for namespace IDs in value store and reduce max key size before hashing is used (currently 512 bytes which is the max key size of LMDB)
- invent a perfomant GC algorithm for values (Which was the initial motivation for this PR)

hmottestad:
I've taken a look at the performance of `ComplexQuery` and found out that LMDB and the MemoryStore choose ever so slightly different query plans. This seems to make a massive difference for some reason, although it's not quite obvious why.

The main difference is that the MemoryStore does a join without any common variables as the initial (bottom) join, while LMDB has moved this join up a few nodes in the plan. I've added a visualisation of the two plans below.

In order to force the same plan to be used I disabled the query optimizer and ran both benchmarks again. The results are incredibly close 1030ms for LMDB and 930ms for the MemoryStore. This is very promising @kenwenzel. I would love to contribute some more benchmarks to your branch if you'd be ok with that.

### MemoryStore
graphviz-memory


### LMDB
graphviz-lmdb


If you want to produce these visualisations yourself you can take a look at the query explanation documentation here: https://rdf4j.org/documentation/programming/repository/#explaining-queries

kenwenzel:
> I've taken a look at the performance of `ComplexQuery` and found out that LMDB and the MemoryStore choose ever so slightly different query plans. This seems to make a massive difference for some reason, although it's not quite obvious why.
> 
> The main difference is that the MemoryStore does a join without any common variables as the initial (bottom) join, while LMDB has moved this join up a few nodes in the plan. I've added a visualisation of the two plans below.
> 
> In order to force the same plan to be used I disabled the query optimizer and ran both benchmarks again. The results are incredibly close 1030ms for LMDB and 930ms for the MemoryStore. This is very promising @kenwenzel. I would love to contribute some more benchmarks to your branch if you'd be ok with that.

Thank you for the visualizations :-) What do you mean by ComplexQuery?

hmottestad:
> @hmottestad Thank you for the visualizations :-) What do you mean by ComplexQuery? I thought that I've fixed the issue.
> 
> BTW, cost estimates are currently done by iterating the first 1000 values, then moving to the last value and then extrapolating based on the difference in subject, predicate, object or context ids.

You've fixed the issue @kenwenzel. I was just running the benchmarks locally and your LMDB store is faster than the MemoryStore. After investigating I found out that the query plans are slightly different and I currently digging into it a bit trying to figure out what would make it faster for the MemoryStore. Even with caching I would still expect an in-memory database than a disk based one.

jeenbroekstra:
> > @kenwenzel this is up to you, but if you are comfortable with some of us pairing up to get this polished, can you give @eclipse/technology-rdf4j-committers access to your forked repo so we can push to your branch?
> 
> @jeenbroekstra Sure, how can I do this? I can't find "eclipse/technology-rdf4j-committers" in the members list.

Argh. I assumed too much on behalf of github, thinking that it would be possible to add the team as a whole as collaborators on your fork. But I've just given it a try and it looks like that's not possible and can only give access to individual collaborators. 

Something simpler we can try first: there should be a checkbox "Allow edits by maintainers" visible for you on your pull request somewhere. If you enable that option it should be possible for me and other committers to add commits to your branch.

jeenbroekstra:
Never mind, I just gave it a quick try with some small changes to the pom.xml and it already worked! I've pushed an update to your branch, don't forget to pull before you continue working on it.

kenwenzel:
> Never mind, I just gave it a quick try with some small changes to the pom.xml and it already worked! I've pushed an update to your branch, don't forget to pull before you continue working on it.

How should we move forward? Should I first rewrite the commit messages and potentially squash my commits before you and @hmottestad base your work on them?

hmottestad:
Would be nice if you did that before I do any work. Could you also rebase onto develop so we get the latest changes from there too?

If you need to force push in the future then a heads up would be nice, but it's your branch after all!

FYI: I'm not about to do anything on your branch just yet.

jeenbroekstra:
I'll stay off your branch for now as well.

kenwenzel:
@jeenbroekstra The SPARQL compliance tests for LMDB fail now due to the optional dependencies to LWJGL native libraries.
Should I add those dependencies to the compliance tests or just remove the compliance tests for LMDB?

jeenbroekstra:
@kenwenzel Ah, yes, adding them as runtime dependence in the compliance test module should fix that

kenwenzel:
```
# LMDB
Benchmark                                                       Mode  Cnt      Score      Error  Units
TransactionsPerSecondBenchmark.largerTransaction               thrpt    5     31.845 ±    0.861  ops/s
TransactionsPerSecondBenchmark.largerTransactionLevelNone      thrpt    5     44.121 ±    2.161  ops/s
TransactionsPerSecondBenchmark.mediumTransactionsLevelNone     thrpt    5  19986.561 ±  774.509  ops/s
TransactionsPerSecondBenchmark.transactions                    thrpt    5  21565.831 ±  815.203  ops/s
TransactionsPerSecondBenchmark.transactionsLevelNone           thrpt    5  38754.785 ± 1597.893  ops/s
TransactionsPerSecondBenchmark.veryLargerTransactionLevelNone  thrpt    5      0.268 ±    0.016  ops/s

Benchmark                                               Mode  Cnt     Score     Error  Units
QueryBenchmark.complexQuery                             avgt    5    33.494 ±   2.694  ms/op
QueryBenchmark.distinctPredicatesQuery                  avgt    5  1231.341 ± 120.133  ms/op
QueryBenchmark.groupByQuery                             avgt    5    18.464 ±   1.735  ms/op
QueryBenchmark.removeByQuery                            avgt    5   231.470 ±  14.651  ms/op
QueryBenchmark.removeByQueryReadCommitted               avgt    5   683.210 ±  25.924  ms/op
QueryBenchmark.simpleUpdateQueryIsolationNone           avgt    5   514.449 ±  20.417  ms/op
QueryBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt    5  1095.385 ± 216.371  ms/op
```
Findings:
- insertion speed also improved for non-hashed values (benchmarks only generate very short literals)
- degraded query perfomance because of 2 hops for lookups of hashed values (id -> hash and hash+id -> value)
- TODO: rework storing of hashed values again

kenwenzel:
This is now feature complete in comparison to the native store. Two open points are:

1. add a perfomant GC algorithm for values (Which was the initial motivation for this PR)
2. maybe drop the complex logic for transaction isolation and solely rely on LMDB's transactions (problem: only one writer at a time is allowed)

Point 1 is really necessary while point 2 would be nice to have.

How should we proceed? Do you want to merge this PR first and then we add GC and maybe other features?

hmottestad:
Previously we've marked features as experimental or for internal use only. That way we can merge early while still allowing for large changes later on.

JervenBolleman:
As the work is stand alone as a new sail, I agree with @hmottestad to merge it with an experimental tag.

jeenbroekstra (approved):

jeenbroekstra:
Great work @kenwenzel , thanks!

PR #3445, opened 2021-11-15
GH-1815: Escape String literals in SparqlBuilder

### LINKED ISSUE #1815: String literals are not properly escaped
The approach used for escaping string literals doesn't work for all cases. For example, a string literal won't be properly escaped if its value ends with a single quote (i. e. `value'`). The values like `val'''ue` will also fail. Though the second case seems to be unlikely, I do have some literals ending with a single quote.

The following code fails with `MalformedQueryException`:

```
ConstructQuery query = Queries.CONSTRUCT().where(GraphPatterns.tp(SparqlBuilder.var("s"), RDFS.LABEL, Rdf.literalOf("value'")));
ParsedGraphQuery parsedGraphQuery = QueryParserUtil.parseGraphQuery(QueryLanguage.SPARQL, query.getQueryString(), null);
```

Does it make sense to use backslashes according to [SPARQL 1.1 Query Language – Escape sequences in strings](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#grammarEscapes). I.e. `value'` is escaped as `value\'`, `val"ue` becomes `val\"ue` etc.?

fkleedorfer:
GitHub issue resolved: #1815 

Handles escape sequences, such as `\n`, in string literals as prescribed by the spec
https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#grammarEscapes

Note: codepoint escape sequences are not treated in any special way. Users who want unusual characters in their literals must find a way to get them into the Java String, e.g. by encoding them in Java source code as, e.g. `\u2021`, in which case they will be properly represented in the literal's String representation.
(see https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#codepointEscape )

Fixes #1815 

----

### DIFF
--- a/core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/RdfLiteral.java
+++ b/core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/rdf/RdfLiteral.java
@@ -8,9 +8,9 @@
 
 package org.eclipse.rdf4j.sparqlbuilder.rdf;
 
-import org.eclipse.rdf4j.sparqlbuilder.util.SparqlBuilderUtils;
-
 import java.util.Optional;
+
+import org.eclipse.rdf4j.sparqlbuilder.util.SparqlBuilderUtils;
 
 /**
  * Denotes an RDF literal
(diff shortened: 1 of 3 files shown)


fkleedorfer:
@RoKish you might want to check if this fits your needs

fkleedorfer:
One possible problem with this is that it should be done in the reverse sense when reporting results from the repository. Otherwise, repeated reading a string from the repository and writing it back will lead to more and more escaped backslashes, e.g.:

```
user input: "\tHello, world!\n"
write converted: "\\tHello, world!\\n"
read: "\\tHello, world!\\n" -> written back as user input
write converted: "\\\\tHello, world!\\\\n"
...
```
I suspect that escape sequences in string literals are translated in query results, but we should make sure, and I haven't managed to find the relevant code yet.

jeenbroekstra (approved):
Nice work :+1:

Issue #3446, opened 2021-11-15
Use prefixed IRIs automatically for declared prefixes


fkleedorfer:
### Problem description

The SparqlBuilder allows for `PREFIX` declarations in queries, but creating prefixed IRIs is cumbersome:
(from the docs:)
```
query.prefix(foaf)
    .select(name)
    .where(
         x.has(foaf.iri("name"), name));
```
We have to use `foaf.iri("name")` to generate `foaf:name`:
```
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
SELECT ?name
WHERE { ?x foaf:name ?name . }
```
However, it is common practice to have a class per namespace that contains all required IRIs as constants, in this case `org.eclipse.rdf4j.model.vocabulary.FOAF.NAME`:
```
query.prefix(foaf)
    .select(name)
    .where(
         x.has(FOAF.NAME, name));
```
 When using those constants, their absolute IRI is used each time, making queries less readable:
```
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
SELECT ?name
WHERE { ?x <http://xmlns.com/foaf/0.1/name> ?name . }
```

### Preferred solution

If a prefix is declared for a query, all uses of the namespace should be shortened with the prefix label, ie. using `foaf.iri("name")` should yield exactly the same result as using `FOAF.NAME` in the above example.

### Are you interested in contributing a solution yourself?

Yes

PR #3451, opened 2021-11-15
GH-3446: Use prefixed IRIs automatically for declared prefixes

fkleedorfer:
GitHub issue resolved: #3446 

The SparqlBuilder now replaces all occurrences of a registered namespace with its label. Some convenience methods added.

----


fkleedorfer:
Working on some corner cases:
* do not replace prefixes in strings
* do not replace prefixes if contiuation is not a legal localname

fkleedorfer:
Done

core/sparqlbuilder/pom.xml
```diff
@@ -15,5 +15,10 @@
 			<artifactId>rdf4j-model</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.rdf4j</groupId>
+			<artifactId>rdf4j-rio-turtle</artifactId>
```
jeenbroekstra, line 20: I would prefer if we didn't need this, especially since it's just for the use of a single utility class. Let's have a think about maybe moving or copying the code to a better place.

fkleedorfer: Sure. Somewhere in core?

jeenbroekstra: Yep. `rdf4j-model` would be a good place. It already has a class `URIUtil` which has several closely related utility methods - I wonder if we could stick this in there somewhere. Bonus: it's already a dependency for the sparqlbuilder so you won't have to add another module.

fkleedorfer: Ok I'll do it

fkleedorfer: ... but we have to copy the method, right? Otherwise it's a breaking change. What do you prefer?

fkleedorfer: Hmm the method depends on a couple of other methods in the class. Copy them all?

fkleedorfer: done

jeenbroekstra (approved):
LGTM!

Issue #3457, opened 2021-11-16
Improve SparqlBuilder's integration of property paths


fkleedorfer:
### Problem description

Currently, special knowledge is required to build a triple pattern with a property path:
```
return Queries.SELECT(x)
                .where(
                                x.isA(EX.Person)
                                .andHas(Expressions
                                             .path(EX.parent)
                                             .then(EX.parent)
                                             .then(FOAF.name)
                                             .build(),
                                grandparentName)
                .getQueryString();
```

Without knowing that `Expressions.path()` is the way to get a `PropertyPathBuilder`, and that its `build()` method will make something that fits here, one can't use it. Also, `Expressions` is the wrong place for the property-path methods, as a property path is not an expression according to the standard - so nobody would go looking for it there.

Consequence: adoption of this feature is almost impossible

### Preferred solution

Offer the property path builder to the user as one way of creating a triple. Thus, users can stumble upon it when using their IDE's code completion. For example:

```
return Queries.SELECT(x)
                .where(
                                x.isA(EX.Person)
                                .andHas(p -> p.
                                             .pred(EX.parent)
                                             .then(EX.parent)
                                             .then(FOAF.name),
                                grandparentName)
                .getQueryString();
```
Here, there is an  `andHas(Consumer<EmptyPropertyPathBuilder, RdfObject)` method that handles management of the builder. The only thing the client has to do is to configure the path as needed. This method is suggested by IDE autocompletion and should be rather self-explanatory.

Also, move all property path related methods away from `Expressions` as it is just the wrong place.

### Are you interested in contributing a solution yourself?

Yes

PR #3459, opened 2021-11-16
GH-3457: improve property path integration in sparqlbuilder

fkleedorfer:
GitHub issue resolved: #3457 

* Methods are added to the classes that have factory methods for triple patterns (and similar objects), which allow for providing a property path. 
* all property path related methods are removed from `Expressions` as they are not expressions
* Tests were added
* The existing PropertyPaths class was marked `@Deprecated` 

----


jeenbroekstra (commented):
Looks good to me. Can you clarifiy if this this is a breaking change? That is: if I have code that created a property using the old method, will that still work?

fkleedorfer:
None of my changes have been breaking so far with respect to the main branch. I was going to ask, though, if breaking the sparqlbuilder's interface is an option. If so, we could make it auch that all those Iri, RdfPredicate, etc classes are only used internally, and at the surface, the RDF4J model classes are used. Also, We might get rid of the heavy delegating that's happening in some places. Food for thought - but not in this PR

fkleedorfer:
Noticed that now, we can add the previously missing tests from section 9 in the sparql spec. Did that with the last commit

jeenbroekstra (approved):

Issue #3463, opened 2021-11-18
Upgrade to junit 5.8.1


hmottestad:
This version removes some INFO logging.

hmottestad:
CG: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23781

PR #3464, opened 2021-11-18
GH-3463 upgrade junit to 5.8.1

hmottestad:
GitHub issue resolved: #3463  

----

### DIFF
--- a/spring-components/pom.xml
+++ b/spring-components/pom.xml
@@ -21,6 +21,13 @@
 	<dependencyManagement>
 		<dependencies>
 			<dependency>
+				<groupId>org.junit</groupId>
+				<artifactId>junit-bom</artifactId>
+				<version>${junit.version}</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+			<dependency>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-dependencies</artifactId>
 				<version>${spring.boot.version}</version>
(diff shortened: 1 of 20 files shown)


hmottestad:
A few tests are using random data, which means that if they fail then there is no way to reproduce that exact run locally. Hopefully this will be better with fixed seeds.

testsuites/repository/src/main/java/org/eclipse/rdf4j/repository/TupleQueryResultTest.java
```diff
@@ -232,8 +233,8 @@ public void testNotClosingResult() {
 				TupleQuery tupleQuery = repCon.prepareTupleQuery(QueryLanguage.SPARQL, queryString);
 
 				// see if open results hangs test
-				try (TupleQueryResult result = tupleQuery.evaluate()) {
-				}
+				// DO NOT CLOSE THIS
+				tupleQuery.evaluate();
```
jeenbroekstra, line 237: Good find.

spring-components/pom.xml
```diff
@@ -20,6 +20,13 @@
 	<description>Components to use with Spring</description>
 	<dependencyManagement>
 		<dependencies>
+			<dependency>
+				<groupId>org.junit</groupId>
+				<artifactId>junit-bom</artifactId>
+				<version>${junit.version}</version>
+				<type>pom</type>
```
jeenbroekstra, line 27: Does this need to be included here if it's already in the root pom?

hmottestad: The spring one overrides it.

Issue #3477, opened 2021-11-21
Platform-dependent test fails on windows


fkleedorfer:
### Current Behavior

`org.eclipse.rdf4j.sail.memory.QueryPlanRetrievalTest` uses `\n` for line separation, but the classes being tested use the platform-specific separator.

On windows, the test fails.

### Expected Behavior

The test should not fail

### Steps To Reproduce

build on windows.

### Version

4.0.0.SNAPSHOT

### Are you interested in contributing a solution yourself?

Yes

PR #3478, opened 2021-11-21
GH-3477: Fix platform-dependent test

fkleedorfer:
Also adapts ExplanationImpl to use the platform-specific line separator.

GitHub issue resolved: #3477 

A test uses `\n` line endings but the classes being tested use the platform-specific line separator. 

----

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/explanation/ExplanationImpl.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/explanation/ExplanationImpl.java
@@ -60,6 +60,7 @@ public class ExplanationImpl implements Explanation {
 
 	@Override
 	public String toDot() {
-		return "digraph Explanation {\n" + genericPlanNode.toDot() + "\n}\n";
+		String sep = System.lineSeparator();
+		return "digraph Explanation {" + sep + genericPlanNode.toDot() + sep + "}" + sep;
 	}
 }
--- a/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
+++ b/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/QueryPlanRetrievalTest.java
@@ -35,7 +35,9 @@ import org.junit.Test;
 
 public class QueryPlanRetrievalTest {
 
-	public static final String MAIN_QUERY = String.join("\n", "",
+	private static final String LINE_SEP = System.lineSeparator();
+
+	public static final String MAIN_QUERY = String.join(LINE_SEP, "",
 			"{",
 			"    {",
 			"        OPTIONAL {",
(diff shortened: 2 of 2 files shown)


core/query/src/main/java/org/eclipse/rdf4j/query/explanation/ExplanationImpl.java
```diff
@@ -60,6 +60,7 @@ public String toString() {
 
 	@Override
 	public String toDot() {
-		return "digraph Explanation {\n" + genericPlanNode.toDot() + "\n}\n";
+		String sep = System.lineSeparator();
+		return "digraph Explanation {" + sep + genericPlanNode.toDot() + sep + "}" + sep;
```
jeenbroekstra, line 64: To catch these kinds of things more consistently, would it be worth looking at a Github action that regularly run a build on a Windows environment?

jeenbroekstra (approved):

core/query/src/main/java/org/eclipse/rdf4j/query/explanation/ExplanationImpl.java
hmottestad: I was looking into setting the system default line separator through maven. But gave up. Should be possible. 

Running tests on windows would probably be worth it, if it's free. 

Btw. The tests you've fixed here, are they actually failing on windows or are you just doing a search replace for \n?

fkleedorfer: They failed.

fkleedorfer: As for setting the separator in maven, I also looked for that and gave up. It also doesn't strike me as very robust to do that. 

I am actually surprised that rdf4j even bothers for the system line separator. In many cases it just generates strings with \n - and I think that is totally fine. In some cases - the "clean" code - the resulting strings are platform dependent. Instead of fixing the tests, we could make sure that the system line separator is never ever used. Unless you are opening the file in MS word or Notepad, you would probably never notice  and it's more portable for scripting as nobody I have ever heard of writes scripts that take the system line separator into account.

hmottestad: I agree with you @fkleedorfer about using only `\n` instead of the system line separator. What do you think @jeenbroekstra ?

jeenbroekstra: I am not entirely sure I follow. Are you suggesting that all RDF4J writers should just output unix line endings regardless of the platform they run on? My kneejerk is that that sounds like a change that makes it easier for us as devs at the cost of user functionality. But perhaps I misunderstand. Let's discuss further on forum or mailinglist (or open a ticket with a proposal).

jeenbroekstra (approved):

Issue #3476, opened 2021-11-21
Modules not named pedantically according to spec


fkleedorfer:
### Current Behavior

The rdf4j module names, i.e. the strings in the pom's `<name>` elements, are not all following the project's guidelines.

The guideline is that each module name should start with `RDF4J: ` (note the colon and the space). About a quarter of them do not.
Normally, I would not care ~~too much~~ at all, but unfortunately, Intellij sorts them by name in the maven view:
![grafik](https://user-images.githubusercontent.com/551624/142757932-76e4310a-ac11-4fae-b764-00fcb7165573.png)

So the modules are shown in essentially two lists: one with colons, one without. 

It's a nuisance.

### Expected Behavior

All names should have the same prefix, namely `RDF4J: `

### Version

4.0.0.SNAPSHOT

### Are you interested in contributing a solution yourself?

Yes

PR #3481, opened 2021-11-21
GH-3476: Pedantically rename maven modules according to guidelines

fkleedorfer:
Renames all module names such that they start with
'RDF4J: '

GitHub issue resolved: #3476

maven modules are renamed so they all start with the same prefix

----

### DIFF
--- a/assembly-descriptors/pom.xml
+++ b/assembly-descriptors/pom.xml
@@ -7,6 +7,6 @@
 		<version>4.0.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>rdf4j-assembly-descriptors</artifactId>
-	<name>RDF4J Assembly Descriptors</name>
+	<name>RDF4J: Assembly Descriptors</name>
 	<description/>
 </project>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -8,7 +8,7 @@
 	</parent>
 	<artifactId>rdf4j-bom</artifactId>
 	<packaging>pom</packaging>
-	<name>RDF4J BOM</name>
+	<name>RDF4J: BOM</name>
 	<description>RDF4J Bill of Materials (BOM)</description>
 	<dependencyManagement>
 		<dependencies>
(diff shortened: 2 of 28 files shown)


hmottestad:
Does this affect how the modules are used or how they get published? Would it break anything for existing users of these modules?

fkleedorfer:
I don't think so, but I might be wrong. Should anyone rely on the module names never ever changing? 

Hopefully not related: It seems one test fails: It is one in module `rdf4j-sail-extensible-store`, which is affected in this PR - but only of a change of the maven module name. I cannot reproduce this test failure locally, and I would rather suspect the test is somehow instable.

hmottestad:
That one started failing for me too. I thought it was because of the use of randomness, but I fixed that. So I guess it must be something else. I'll look into it.

fkleedorfer:
Still cannot reproduce. FWIW, its in ` org.eclipse.rdf4j.repository.TupleQueryResultTest.testStreaming()` (run by subclass `org.eclipse.rdf4j.sail.extensiblestore.compliance.ExtensibleStoreTupleQueryResultTest`. ) 

The error happens in one of those loops that run 1000x, I guess whatever the authors were testing for, they've found it ;-)

```
 Error:  Tests run: 5, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.158 s <<< FAILURE! - in org.eclipse.rdf4j.sail.extensiblestore.compliance.ExtensibleStoreTupleQueryResultTest
Error:  org.eclipse.rdf4j.sail.extensiblestore.compliance.ExtensibleStoreTupleQueryResultTest.testStreaming  Time elapsed: 0.346 s  <<< ERROR!
java.lang.ArrayIndexOutOfBoundsException: Index 56 out of bounds for length 32
```

hmottestad:
https://github.com/eclipse/rdf4j/issues/3484

It's my code that's failing, but the test isn't mine.

jeenbroekstra (approved):
Nice cleanup, thanks @fkleedorfer . Fwiw this shouldn't affect anything other than what maven outputs on the console when running builds, as far as I know (and perhaps how our modules are described in Maven Central) - but none of that should have any influence on any part of the build - these names are for human consumption only.

Issue #3458, opened 2021-11-16
Remove ExtendedVariable from SparqlBuilder


fkleedorfer:
### Problem description

Recently, the class `ExtendedVariable (extends Variable)` was added to the SparqlBuilder. Its sole purpose is to make the variable name publicly accessible via `public String getVarName()`.  In my opinion, there is no harm done by exposing the name of the variable. 

Context: The `ExtendedVariable` is used in `rdf4j-spring` to pre-bind values in a query and extract values from a `BindingSet`. Thus, it provides a way to bridge from the SparqlBuilder to the query evaluation framework as it can be used with both subsystems. 

### Preferred solution

It would make things easier if we just added the `public String getVarName()` method to the `Variable` class.

### Are you interested in contributing a solution yourself?

Yes

PR #3482, opened 2021-11-21
GH-3458: Remove ExtendedVariable class

fkleedorfer:
* all occurrences replaced with Sparqlbuilder's Variable class
* changed Variable class:
    - public constructor
    - getter for field varName
    - changed field name from alias to varName (closer to spec)

GitHub issue resolved: #3458 

`ExtendedVariable` was introduced by the rdf4j-spring code to access the constructor and `varName` property. Making these things accessible to the outside directly in the `Variable` class seems cleaner.

----


fkleedorfer:
In this PR, the Constructor of the `Variable` class is made public. Maybe that's a bit much. Alternatively we could enforce creation of variables through `SparqlBuilder.var()`. However, for the `rdf4j-spring` package, we need the accessor for the `varName` field and some way of instantiating variables.

jeenbroekstra (approved):

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/core/Variable.java
```diff
@@ -21,15 +21,19 @@
  */
 public class Variable implements Projectable, RdfSubject, RdfPredicate, RdfObject, Operand, Orderable, Groupable,
 		GraphName, Assignable {
-	private String alias;
+	private String varName;
 
-	Variable(String varName) {
-		this.alias = varName;
+	public Variable(String varName) {
```
jeenbroekstra, line 26: I have no _very_ strong feelings on this, but if you can leave it package-private without having to jump through too many hoops, then I think I'd prefer sticking with that.

jeenbroekstra (commented):
LGTM. As soon as we decide on the constructor visibility one way or the other this is good to go I think.

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/core/Variable.java
fkleedorfer: Ok, I'll make it package-private (thanks for making me look that up btw :-)) and we'll go through `SparqlBuilder.var()` in `rdf4j-spring`.

fkleedorfer: done

Issue #3496, opened 2021-12-01
Reduce runtime of test NativeStoreConsistencyIT


kenwenzel:
### Problem description

The test NativeStoreConsistencyIT uses RepositoryUtil.difference which is very slow. On my machine the tests takes over 90s to finish.

### Preferred solution

The test could simply use methods on LinkedHashModel to compute the differences between two sets of statements since mapping of blank nodes is not required. On my machine the runtime of the test could be reduced to < 10s.

PR #3497, opened 2021-12-01
GH-3496 Replace RepositoryUtil.difference by methods of LinkedHashModel.

kenwenzel:
GitHub issue resolved: #3496 

Speed up test NativeStoreConsistencyIT by using other methods for computing the difference.

----


hmottestad:
Looks like the test only passes when the two models are equal. So it _might_ be faster to do .equals and only calculate the difference if it's not equal.

kenwenzel:
> Looks like the test only passes when the two models are equal. So it _might_ be faster to do .equals and only calculate the difference if it's not equal.

Yes, good point. Do you like to add the change oder should I update my commit?

hmottestad:
LGTM

Issue #3504, opened 2021-12-10
Upgrade log4j-core to 2.15 or newer


barthanssens:
### Current Behavior

Elasticsearch compliance uses log4j-core 2.13.2, which has a security vulnerability.
Upgrade this dependency just to be sure

### Expected Behavior

Upgrade to 2.15.0

### Version

3.7.4

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

See https://nvd.nist.gov/vuln/detail/CVE-2021-44228

barthanssens:
May require a CQ, since this version hasn't been validated (yet) by ClearlyDefined.io

barthanssens:
Also a (sub)dependency of solr4j and elasticsearch-transport

jeenbroekstra:
> May require a CQ, since this version hasn't been validated (yet) by ClearlyDefined.io

Can  you run it through the [IP license tool](https://www.eclipse.org/projects/handbook/#ip-license-tool) to see what it says?

jeenbroekstra:
Actually see latest post on eclipse-committers (https://www.eclipse.org/lists/eclipse.org-committers/msg01333.html):
 
>  A serious vulnerability that affects versions of Apache Log4J between 2.0.0 and 2.14.1. Please take a few minutes to ensure that your project is not affected. Note that there are numerous vulnerabilities reported against versions < 2.0 as well (which are no longer "supported") and that it's probably just a good idea to update to the latest and greatest. Note that Log4j version 2.15 has been vetted through our IP Due Diligence process so you're good to go to incorporate that version into your project.

So no need to log a CQ I think

barthanssens:
See also discussion https://github.com/eclipse/rdf4j/discussions/3503

jeenbroekstra:
There appears to be a followup to the CVE which states that the original fix in 2.15.0 was incomplete. A new release, 2.16.0 has been done, and we should bump to that version.

In addition, following up from discussion #3503, we should fix the pom in the solr sail: it has an optional dependency on solr-core, which currently pulls in log4j-core as a transitive dependency. This doesn't affect our own release distribution since it's an optional dependency, but we should still add log4j-core to the list of exclusions in the pom.

barthanssens:
Hmz, excluding the log4j-artifacts from sail/solr/pom.xml was already part of the previous commit for 3.7.x, should it be excluded from other poms as well ?

barthanssens:
[...]

PR #3505, opened 2021-12-10
GH-3504: security upgrade of log4j-core

barthanssens:

GitHub issue resolved: #3504 

- upgrade dependency to 2.15 to fix vulnerability 

----

### DIFF
--- a/compliance/elasticsearch/pom.xml
+++ b/compliance/elasticsearch/pom.xml
@@ -102,7 +102,7 @@
 		<dependency>
 			<groupId>org.apache.logging.log4j</groupId>
 			<artifactId>log4j-core</artifactId>
-			<version>2.13.2</version>
+			<version>2.15.0</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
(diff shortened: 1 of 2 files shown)


jeenbroekstra (approved):
LGTM. @barthanssens you want to run a patch release straightaway?

jeenbroekstra (commented):
I've just noticed that this _only_ bumps the version of log4j-core, not of log4j-api, log4j-slf4j-impl, or log4j-to-slf4j (which are referenced transitively in various modules, including the solr, elasticsearch, and lucene sails). Should we make sure we bump all of those to get things in sync? 

EDIT: actually just read the CVE and it looks like it's specifically log4j-core that's affected. So let's get this merged and released first and look at the other deps later.

barthanssens:
The API jar might actually be ok (see last comment on https://issues.apache.org/jira/browse/LOG4J2-3201), but to be sure (and to make things a bit cleaner, it should indeed be better to bump everything log4j-related...)

jeenbroekstra:
Just taking a closer look at the parts of RDF4J affected - it doesn't look like this directly affects anything we distribute, they're only test dependencies correct? If so, perhaps an immediate patch release is unnecessary.

barthanssens:
Indeed, looks like it's not part of jar/wars we distribute, so probably not needed right away...

PR #3540, opened 2021-12-21
GH-3539 use a cleaner and shutdown hook instead of finalize()

JervenBolleman:
GitHub issue resolved: #3539 

Briefly describe the changes proposed in this PR:

Use a [Cleaner](https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html) and [shutdown hook](https://docs.oracle.com/javase/8/docs/technotes/guides/lang/hook-design.html) to make sure unused 
resources are removed cleanly and efficiently without spooky behaviour. 

----
PR Author Checklist (see the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md) for more details):

 - [x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [ ] I've added tests for the changes I made
 - [x] I've applied [code formatting](https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)
 - [x] I've [squashed](https://rdf4j.org/documentation/developer/squashing) my commits where necessary 
 - [x] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change

PR #3542, opened 2021-12-21
nativerdf use cleaner to close overflowmemory backing store when not used.

JervenBolleman:
GitHub issue resolved: #3541

Similar to #3540 but now for native store, doesn't use a shutdown hook so might be better/cleaner.

@kenwenzel I now see where the code came from. Still would like your opinion though :)

This can not be backported to the 3 series as requires java11
----

### DIFF
--- a/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowToDiskTest.java
+++ b/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowToDiskTest.java
@@ -17,11 +17,13 @@ import java.nio.file.Files;
 import org.eclipse.rdf4j.model.Model;
 import org.eclipse.rdf4j.sail.base.SailStore;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 
 public class MemoryOverflowToDiskTest {
 
 	@Test
-	public void testCleaner() throws IOException, InterruptedException {
+	@Timeout(5)
+	public void testCleanerRemovesTempDirWhenMemoryOverflowModelGetsGCed() throws IOException, InterruptedException {
 		File file = Files.createTempDirectory("model").toFile();
 
 		Model model = createModel(file);
@@ -32,7 +34,6 @@ public class MemoryOverflowToDiskTest {
 		while (file.exists()) {
 			System.gc();
 			Thread.sleep(10);
-			System.out.println();
 		}
 
 		assertFalse(file.exists());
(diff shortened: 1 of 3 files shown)


kenwenzel:
@JervenBolleman This looks good. Unfortunately I haven't found a way to explicitly close the MemoryOverflowModel.

hmottestad:
Could you make a test for this?

In the ShaclSail I've done this: 

```
	@Test
	public void testThatGarbadgeCollectionWillShutdownTheThreadPool()
			throws InterruptedException, NoSuchFieldException, IllegalAccessException {

		ExecutorService[] executorServices = startShaclSailAndTask();

		assertFalse(executorServices[0].isShutdown());

		for (int i = 0; i < 100; i++) {
			System.gc();
			if (executorServices[0].isShutdown()) {
				return;
			}
			System.out.println(i);
			Thread.sleep(100);
		}

		fail("Executor service should have been shutdown due to GC");
	}
```

Where I'm testing that the executorServices eventually gets shutdown.

jeenbroekstra (approved):
Nice! I wasn't aware of the new Cleaner functionality, looks good. Do we have a regression test in place for this?

JervenBolleman:
I tried making a test for this. However, I think the concept as it was originally implemented might not be worth it. The cleaning will work but on my machine testing this takes hours because performance is down the drain in the cases where this starts to apply.

hmottestad:
Isn't there some sort of minimum value for triggering the overflow model? Maybe it can be adjusted to always trigger the overflow model? Or you can extract your code into a method and make the method package private, that way you can call it from a test in the same package without having to make the method public.

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/MemoryOverflowToDiskTest.java
```diff
@@ -0,0 +1,51 @@
...
+		model = null;
+
+		System.gc();
+		while (file.exists()) {
+			System.gc();
+			Thread.sleep(10);
+			System.out.println();
+		}
+
+		assertFalse(file.exists());
```
jeenbroekstra, line 38: Doesn't this run a risk of looping endlessly? Not sure we have a global test timeout configured here.

hmottestad: True. I'll fix it.

hmottestad: Fixed it with this: https://mkyong.com/junit5/junit-5-timeouts-examples/

jeenbroekstra (approved):

JervenBolleman:
@hmottestad thanks for improving the test case :) should I hit the merge button?

Issue #3558, opened 2021-12-30
Improve lock managers by using StampedLock and LongAdder


hmottestad:
### Problem description

Related to: #3435

Overhead due to read locks is significant during SHACL validation on a MemoryStore.

### Preferred solution

Use StampedLock and LongAdder in the effected lock managers. Ref: http://concurrencyfreaks.blogspot.com/2013/09/combining-stampedlock-and-longadder-to.html

### Are you interested in contributing a solution yourself?

Yes

hmottestad:
<img width="1203" alt="Screenshot 2021-12-30 at 08 01 21" src="https://user-images.githubusercontent.com/797185/147729029-1a325757-3963-4cc1-a138-6df32d6b285a.png">

Slowest thread. There are multiple calls to `getReadLock()` and in total for all those calls (on the slowest thread) we are spending 14% of our time even though there are no contending writes.

hmottestad:
**3x improvement to concurrent queries against the MemoryStore.**

PR #3564, opened 2021-12-31
GH-3558 improve lock managers

hmottestad:
Resolves GH-3558

- Reimplement the two basic LockManagers with StampedLock
- Reimplement read/write locks using a combination of StampedLock and LongAdder from: http://concurrencyfreaks.blogspot.com/2013/09/combining-stampedlock-and-longadder-to.html?m=1

----


hmottestad:
### Before

```

```

### After

```

```

core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/ReadPrefReadWriteLockManager.java
```diff
@@ -69,12 +70,7 @@ public Lock tryReadLock() {
 	 * released.
 	 */
 	@Override
-	public synchronized Lock getReadLock() throws InterruptedException {
-		// Wait for the writer to finish
-		while (isWriterActive()) {
-			waitForActiveWriter();
-		}
-
+	public Lock getReadLock() throws InterruptedException {
```
jeenbroekstra, line 73: I'm not sure you can get rid of synchronization here. The `createReadLock` method itself is not thread-safe and its implementation has a possible race condition with checking the state of the write lock. Callers of `createReadLock` should take care of synchronization I think...

core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/AbstractReadWriteLockManager.java
```diff
@@ -86,26 +98,121 @@ protected void waitForActiveWriter() throws InterruptedException {
...
 	protected void waitForActiveReaders() throws InterruptedException {
-		activeReaders.waitForActiveLocks();
+		while (isReaderActive()) {
+			Thread.onSpinWait();
+		}
 	}
 
 	/**
 	 * Creates a new Lock for reading and increments counter for active readers. The lock is tracked if lock tracking is
 	 * enabled. This method is not thread safe itself, the calling method is expected to handle synchronization issues.
```
jeenbroekstra, line 108: Is this still true given that you're now relying on coordination between two LongAdders for synchronization?

hmottestad: Haven't updated all the documentation yet.

core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/ReadPrefReadWriteLockManager.java
hmottestad: Previous code spins while waiting for active writers to complete, that is now handled by the underlying code instead.

core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/AbstractReadWriteLockManager.java
hmottestad: Also I haven't incorporated any of the lock tracking back into the read-write lock code. 

Got a bit stuck right now because I discovered I need to write an entire suite of tests for verifying what features the previous code supported when it comes to tracking locks, automatically releasing GCed locks and detecting deadlocks.

hmottestad:
## Before

```
Benchmark                             Mode  Cnt    Score    Error  Units
ParallelQueryBenchmark.mixedWorkload  avgt    3  398.387 ± 89.162  ms/op
```

## After

```
Benchmark                             Mode  Cnt    Score    Error  Units
ParallelQueryBenchmark.mixedWorkload  avgt    3  151.281 ± 21.076  ms/op
```

hmottestad:
@jeenbroekstra So I've still got some tests to figure out. Essentially just an issue with differences in performance of my machine vs the CI server. Might move those tests into a benchmark or something instead. 

Performance in general though is much improved with this new way of locking. I've still got some work to do on updating all the java docs and comments, but it would be great if you could take another look at this and maybe give it a try on your own machine if you have a chance.

hmottestad:
## Before

```

```

## After

```

```

hmottestad:
3x improvement to concurrent queries against the MemoryStore (on my machine).

Issue #3572, opened 2022-01-05
Concurrent WeakObjectRegistry


hmottestad:
### Problem description

Performance of the MemValueFactory is poor when there are multiple threads performing read operations because the WeakObjectRegistry does not support any concurrency.

### Are you interested in contributing a solution yourself?

Yes

hmottestad:
Could not get the following to work correctly: https://docs.spring.io/spring-framework/docs/4.0.0.M1_to_4.2.0.M2/Spring%20Framework%204.2.0.M2/org/springframework/util/ConcurrentReferenceHashMap.html

hmottestad:
Related to #3435

PR #3573, opened 2022-01-05
GH-3572 Concurrent WeakObjectRegistry

hmottestad:
GitHub issue resolved: #3572 

 - use an array of WeakHashMap objects to allow multithreaded access for both read and writes
     - we mod the hash of the key to index into the array
     - initial size of a WeakHashMap is 16 and we create `Runtime.getRuntime().availableProcessors() * 2` maps
         - this gives some overhead for MemoryStores with very very few values
         - as the maps fill up there should be less and less overhead since we are storing the same number of objects in total
 - use StampedLocks to allow multiple readers and single writer per WeakHashMap
     - this allows multiple writers if we hit different WeakHashMaps
 - remove external synchronization
 - use lambdas to support "getOrAdd" operations against the WeakObjectRegistry
 - deprecate three methods that are not thread-safe and create new methods that are thread-safe to achieve the same goal

----


hmottestad:
### Before
```
Benchmark                                      Mode  Cnt    Score     Error  Units
MemValueFactoryConcurrentBenchmark.onlyReads   avgt    5  491.734 ± 110.388  ms/op
MemValueFactoryConcurrentBenchmark.onlyWrites  avgt    5  707.663 ±  79.292  ms/op
MemValueFactoryConcurrentBenchmark.readHeavy   avgt    5  217.618 ±   7.088  ms/op
```

### After
```
```

hmottestad:
### Before
```
Benchmark                                                     Mode  Cnt      Score   Error  Units
QueryBenchmark.complexQuery                                   avgt          49.555          ms/op
QueryBenchmark.different_datasets_with_similar_distributions  avgt           3.250          ms/op
QueryBenchmark.groupByQuery                                   avgt           2.082          ms/op
QueryBenchmark.long_chain                                     avgt        1023.599          ms/op
QueryBenchmark.lots_of_optional                               avgt         259.100          ms/op
QueryBenchmark.minus                                          avgt       13122.140          ms/op
QueryBenchmark.nested_optionals                               avgt         259.591          ms/op
QueryBenchmark.pathExpressionQuery1                           avgt          22.957          ms/op
QueryBenchmark.pathExpressionQuery2                           avgt           1.815          ms/op
QueryBenchmark.query_distinct_predicates                      avgt         172.088          ms/op
QueryBenchmark.simple_filter_not                              avgt          10.976          ms/op

```

### After

```
```

hmottestad:
### Before
```
```

### After
```
Benchmark                    (isolationLevel)  Mode  Cnt    Score    Error  Units
NotifyingSailBenchmark.load              NONE  avgt    5  157.878 ± 22.213  ms/op
NotifyingSailBenchmark.load          SNAPSHOT  avgt    5  198.236 ± 22.420  ms/op
NotifyingSailBenchmark.load      SERIALIZABLE  avgt    5  392.916 ± 25.133  ms/op
```

hmottestad:
So this is adding a small overhead when loading data into a MemoryStore. I'll investigate.

hmottestad:
### Before
```
Benchmark                                      Mode  Cnt    Score    Error  Units
MemValueFactoryConcurrentBenchmark.onlyReads   avgt    5  450.121 ± 78.835  ms/op
MemValueFactoryConcurrentBenchmark.onlyWrites  avgt    5  749.964 ± 62.857  ms/op
MemValueFactoryConcurrentBenchmark.readHeavy   avgt    5  217.716 ± 11.002  ms/op
```

### After
```
Benchmark                                      Mode  Cnt    Score    Error  Units
MemValueFactoryConcurrentBenchmark.onlyReads   avgt    5   75.562 ±  1.790  ms/op
MemValueFactoryConcurrentBenchmark.onlyWrites  avgt    5  416.135 ± 16.399  ms/op
MemValueFactoryConcurrentBenchmark.readHeavy   avgt    5   83.085 ±  0.777  ms/op
```

hmottestad:
The last benchmarks are specifically designed to show the performance improvements with this PR. That benchmark has three benchmarks that simulate various access patterns against the MemValyeFactory using multiple threads. The other benchmarks I've posted are just to show that there is no significant performance degradation for queries or when adding data.

Due to other bottlenecks in the system there is little performance improvements elsewhere due to this patch. One of those bottlenecks is locking, which I'm trying to address in the following PR https://github.com/eclipse/rdf4j/pull/3564

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleParser.java
```diff
@@ -278,12 +278,11 @@ protected void parsePrefixID() throws IOException, RDFParseException, RDFHandler
...
 
 		// Read the namespace URI
-		IRI namespace = parseURI();
+		String namespaceStr = parseURI().toString();
 
-		// Store and report this namespace mapping
 		String prefixStr = prefixID.toString();
-		String namespaceStr = namespace.toString();
 
+		// Store and report this namespace mapping
```
hmottestad, line 285: This is just to clean up some code that i stumbled upon while fixing a bug else where.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/CloseableIterator.java
```diff
@@ -0,0 +1,17 @@
...
+ *  are made available under the terms of the Eclipse Distribution License v1.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.sail.memory.model;
+
+import java.util.Iterator;
+
+public interface CloseableIterator<E> extends Iterator<E>, AutoCloseable {
```
jeenbroekstra, line 13: There is a `AbstractCloseableIterator` in `org.eclipse.rdf4j.common.iterator`. It's currently not `AutoCloseable`, just `Closeable`, but that should be simple to fix. Can you reuse that - or alternative put this interface in that package and make `AbstractCloseableIterator` extend it?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemValueFactory.java
```diff
@@ -228,13 +259,10 @@ public MemResource getOrCreateMemResource(Resource resource) {
...
-		MemIRI memURI = getMemURI(uri);
-
-		if (memURI == null) {
-			// Namespace strings are relatively large objects and are shared
-			// between uris
+	public MemIRI getOrCreateMemURI(IRI uri) {
+		return iriRegistry.getOrAdd(uri, () -> {
 			String namespace = uri.getNamespace();
+			assert namespace != null;
 			String sharedNamespace = namespaceRegistry.get(namespace);
```
jeenbroekstra, line 266: Should this also be handled using a `getOrAdd` instead of the `if-else` construction currently used?

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/BaseConcurrentBenchmark.java
```diff
@@ -0,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Eclipse RDF4J contributors.
```
jeenbroekstra, line 2: You need to update the year in your header template :)

jeenbroekstra (commented):
Great stuff @hmottestad - glad you're looking into the concurrency handing in general. This all looks good to me mostly, a few remarks inline.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/CloseableIterator.java
hmottestad: I moved the interface and renamed it. It's not possible to integrate it with the AbstractCloseableIterator because that one specifies that close should throw an IOException.

hmottestad:
```
Benchmark                                      Mode  Cnt    Score   Error  Units
MemValueFactoryConcurrentBenchmark.onlyReads   avgt   10   77.343 ± 0.272  ms/op
MemValueFactoryConcurrentBenchmark.onlyWrites  avgt   10  437.942 ± 3.844  ms/op
MemValueFactoryConcurrentBenchmark.readHeavy   avgt   10   84.729 ± 0.071  ms/op
```

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/CloseableIterator.java
hmottestad: I’m wondering if maybe I can just remove the entire interface now that I’m using a class instead of a lambda.

Issue #3574, opened 2022-01-06
Implement GC algorithm for values in LMDB store


kenwenzel:
### Problem description

This is a follow up issue of #3416.

While the LMDB store is feature complete in comparison to memory and native store it still misses the functionality to evict unused values.

### Preferred solution

- track subject, predicate, object and context IDs when statements are removed from the triple indexes
- before finally committing a transaction:
  - test tracked IDs against the triple indexes and remove IDs which are still in use
  - add remaining IDs to some kind of "recycle" list within the value store and remove the associated data 

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

Implement the GC algorithm on a second thread to impact update performance as few as possible.
While this may be the better option it is harder to implement and I would go with a synchronous implementation for the first PoC.

PR #3577, opened 2022-01-07
GH-3574 GC algorithm for LMDB store values

kenwenzel:
GitHub issue resolved: #3574

Implement GC algorithm for values in LMDB store.
The algorithm will work synchronously to the transactions and will do the following work:
- track subject, predicate, object and context IDs when statements are removed from the triple indexes
- before finally committing a transaction:
  - test tracked IDs against the triple indexes and remove IDs which are still in use
  - add remaining IDs to some kind of "recycle" list within the value store and remove the associated data

----


core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
```diff
@@ -23,14 +23,15 @@
...
 import org.lwjgl.system.MemoryStack;
-import org.lwjgl.system.MemoryUtil;
 import org.lwjgl.util.lmdb.MDBVal;
 
 /**
  * A record iterator that wraps a native LMDB iterator.
  *
  */
-public class LmdbRecordIterator implements RecordIterator {
+class LmdbRecordIterator implements RecordIterator {
```
jeenbroekstra, line 32: Nice, thanks.

jeenbroekstra (commented):
This all looks good to me. The only thing I can think of is perhaps have some explicit test coverage for the cleanup itself?

kenwenzel:
> This all looks good to me. The only thing I can think of is perhaps have some explicit test coverage for the cleanup itself?

Yes, I will add a specific test when the implementation is finished.

jeenbroekstra:
@kenwenzel this PR has been sitting in draft state for a while now. Any chance you can take a look at getting it finalized?

kenwenzel:
@jeenbroekstra Thank you for the reminder.
My challenge is to decide what to do with the unused values:
1. directly delete them including their IDs
2. add them to a to-be-deleted list which is processed in a certain time interval (like every X seconds)

Option 1 is simple but prevents the direct reuse of values which are deleted and inserted again in a short period of time.
Option 2 enables the reuse of values but is a bit more complex to implement because it requires another LMDB table for tracking the deleted IDs.

kenwenzel:
Option 2 is (almost) implemented.
TODOs:
- execute deletion every X seconds
- use persistent hash set to keep track of unused IDs during transaction

abrokenjester:
@kenwenzel we are planning to do a 4.3.0 release in a few weeks time. Do you think you will have time to revisit this PR and push it over the lline?

kenwenzel:
> @kenwenzel we are planning to do a 4.3.0 release in a few weeks time. Do you think you will have time to revisit this PR and push it over the lline?

Yes, I'll try my best.
@abrokenjester Do you have any suggestion regarding the persistent hash set?

abrokenjester:
> > @kenwenzel we are planning to do a 4.3.0 release in a few weeks time. Do you think you will have time to revisit this PR and push it over the lline?
> 
> Yes, I'll try my best. @abrokenjester Do you have any suggestion regarding the persistent hash set?

I haven't looked in detail at what you require here, but would a MapDB HashMap fit the bill? We use the MapDB library for persistent collections elsewhere in the project (e.g. in the `GroupIterator`, from memory, to keep track of resource values during iteration for SPARQL aggregation purposes).

abrokenjester:
Regarding the persistent hash thing: potentially, the CollectionFactory work that @JervenBolleman has been pushing might be relevant here.

kenwenzel:
> > > Regarding the persistent hash thing: potentially, the CollectionFactory work that @JervenBolleman has been pushing might be relevant here.
> > 
> > 
> > Yes, this would be really helpful. Is this already merged?
> 
> Yes the collection factory API was merged for 4.3.0.

Cool :-) Do you have an example for creating a hash set for long values that is overflowed to disk if it has grown to a certain size?

JervenBolleman:
@kenwenzel  not optimal but the following will work.

```java
CollectionFactory cf = new MapDbCollectionFactory();
Set<Long> ids = cf.createSet<Long>();
```

It will do what you want, and is is not as bad in performance as you might think at first glance.

kenwenzel:
> @kenwenzel not optimal but the following will work.
> 
> ```java
> CollectionFactory cf = new MapDbCollectionFactory();
> Set<Long> ids = cf.createSet<Long>();
> ```
> 
> It will do what you want, and is is not as bad in performance as you might think at first glance.

Why does the set switch to a disk-based backend already with 16 elements?
`private static final int SWITCH_TO_DISK_BASED_SET_AT_SIZE = 16;`

I thought of keeping at least 1 million values in memory.

Best regards,
Ken

kenwenzel:
> @kenwenzel not optimal but the following will work.
> 
> ```java
> CollectionFactory cf = new MapDbCollectionFactory();
> Set<Long> ids = cf.createSet<Long>();
> ```
> 
> It will do what you want, and is is not as bad in performance as you might think at first glance.

@JervenBolleman Maybe I am doing something wrong but I get the following benchmark results:

```
HashSet (in-memory):

Benchmark                                  Mode  Cnt     Score     Error  Units
QueryBenchmark.removeByQuery               avgt    5   791.903 ± 146.214  ms/op
QueryBenchmark.removeByQueryReadCommitted  avgt    5  1244.643 ± 177.180  ms/op

PersistentSet (LMDB-based):

Benchmark                                  Mode  Cnt     Score     Error  Units
QueryBenchmark.removeByQuery               avgt    5   969.686 ± 142.148  ms/op
QueryBenchmark.removeByQueryReadCommitted  avgt    5  1563.613 ± 262.295  ms/op

MapDbCollectionFactory(100).createSet():

Benchmark                                  Mode  Cnt      Score       Error  Units
QueryBenchmark.removeByQuery               avgt    5  46626.813 ± 12705.357  ms/op
QueryBenchmark.removeByQueryReadCommitted  avgt    5  45645.531 ±  9026.693  ms/op
```

`PersistentSet` is a quick and dirty implementation of an LMDB-based set.

JervenBolleman:
@kenwenzel do you really need to have mapdb do a commit  and fsync to disk every 100 entries with mapdb?

kenwenzel:
> @kenwenzel do you really need to have mapdb do a commit and fsync to disk every 100 entries with mapdb?

Ah OK. I was not aware that this number is used for fsyncing to disk. I try with a larger number.

JervenBolleman:
The main thing about using a collection factory is that we can change the set implementation later. For example as this is a set of longs we can think about using Roaring64Bitmaps instead.

I also imagine adding a createSet method that allows for type specific optimizations e.g. something in the following spirit.

```java
public <V> Set<V> createSet(<Class<V> valueType) {
    if (Long.class.equals(valueType)){
       //return optimized set for longs
    } else if (Integer.class.equals(valueType)){
       //return optimized set for ints
    } else {
      return new HashSet<V>();
   }
}
```

kenwenzel:
@abrokenjester @hmottestad This PR is ready to review.

hmottestad:
I've not had time to look at this yet. I'll try to get to it soon.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
```diff
@@ -134,6 +141,13 @@ class AddQuadOperation implements Operation {
...
 		@Override
 		public void execute() throws IOException {
+			if (!unusedIds.isEmpty()) {
+				// these ids are used again
+				unusedIds.remove(s);
+				unusedIds.remove(p);
+				unusedIds.remove(o);
+				unusedIds.remove(c);
+			}
 			boolean wasNew = tripleStore.storeTriple(s, p, o, c, explicit);
```
hmottestad, line 151: I'm wondering about how concurrency is handled. Is it possible that a concurrent operation can be adding/removing items from `unusedIds`? Could it be possible that `unusedIds` is empty on line 144, but due to a concurrent operation s, p, o or c could be added to `unusedIds` and removed from the value store right before the triple is stored on line 151?

kenwenzel: As LmdbStore is based on SnapshotSailStore there shouldn't be any concurrency issues. Every write access is executed only by one thread at any point in time.

hmottestad:
> @JervenBolleman I've run some updated benchmarks. What do you think?
> 
> ```
> new MapDbCollectionFactory(Integer.MAX_VALUE).createSet():
> 
> Benchmark                     Mode  Cnt      Score      Error  Units
> QueryBenchmark.removeByQuery  avgt    5  10689.992 ± 6913.359  ms/op
> 
> 
> PersistentSet (LMDB-based) - commit and SYNC after 100 updates:
> 
> Benchmark                     Mode  Cnt     Score      Error  Units
> QueryBenchmark.removeByQuery  avgt    5  5959.434 ± 1061.529  ms/op
> 
> 
> PersistentSet (LMDB-based) - commit  after 100 updates but DON'T SYNC:
> 
> Benchmark                     Mode  Cnt     Score     Error  Units
> QueryBenchmark.removeByQuery  avgt    5  1159.452 ± 287.824  ms/op
> 
> 
> PersistentSet (LMDB-based) - commit only for iterator and DON'T SYNC:
> 
> Benchmark                     Mode  Cnt     Score     Error  Units
> QueryBenchmark.removeByQuery  avgt    5  1038.176 ± 214.871  ms/op
> ```

I'm wondering about what the use of commit and fsync are. I have a couple of questions:
 1. Does unusedIds get repopulated from disk at startup?
 2. Does it matter if some IDs are missing from unusedIds after a crash?
 3. Will another thread be able to read stale data from unusedIds if commit or fsync are not used, or is this guaranteed by LMDB or maybe the OS level filesystem cache?
 4. Does LMDB guarantee that the backing file(s) will never get corrupted if commit or fsync is omitted? (for instance by using an append only log so that a corrupted indexes can be rebuilt to a known good state, even though data could be lost due to the lack of fsync)

kenwenzel:
> > @JervenBolleman I've run some updated benchmarks. What do you think?
> > ```
> > new MapDbCollectionFactory(Integer.MAX_VALUE).createSet():
> > 
> > Benchmark                     Mode  Cnt      Score      Error  Units
> > QueryBenchmark.removeByQuery  avgt    5  10689.992 ± 6913.359  ms/op
> > 
> > 
> > PersistentSet (LMDB-based) - commit and SYNC after 100 updates:
> > 
> > Benchmark                     Mode  Cnt     Score      Error  Units
> > QueryBenchmark.removeByQuery  avgt    5  5959.434 ± 1061.529  ms/op
> > 
> > 
> > PersistentSet (LMDB-based) - commit  after 100 updates but DON'T SYNC:
> > 
> > Benchmark                     Mode  Cnt     Score     Error  Units
> > QueryBenchmark.removeByQuery  avgt    5  1159.452 ± 287.824  ms/op
> > 
> > 
> > PersistentSet (LMDB-based) - commit only for iterator and DON'T SYNC:
> > 
> > Benchmark                     Mode  Cnt     Score     Error  Units
> > QueryBenchmark.removeByQuery  avgt    5  1038.176 ± 214.871  ms/op
> > ```
> 
> I'm wondering about what the use of commit and fsync are. I have a couple of questions:
> 
> 1. Does unusedIds get repopulated from disk at startup?
> 2. Does it matter if some IDs are missing from unusedIds after a crash?
> 3. Will another thread be able to read stale data from unusedIds if commit or fsync are not used, or is this guaranteed by LMDB or maybe the OS level filesystem cache?
> 4. Does LMDB guarantee that the backing file(s) will never get corrupted if commit or fsync is omitted? (for instance by using an append only log so that a corrupted indexes can be rebuilt to a known good state, even though data could be lost due to the lack of fsync)

1. No, it is just a temporary set used within ONE transaction.
2. They don't get GCed. But as unusedIds is only temporary this should not matter here at all.
3. The set is only accessed by the writing thread. I am not sure if PersistentSet is fully thread-safe at all as a LMDB write transaction would be handed over between multiple threads. (This seems to be allowed but I am not sure.)
4. Please see "MDB_NOSYNC" at http://www.lmdb.tech/doc/group__mdb.html#ga32a193c6bf4d7d5c5d579e71f22e9340
   LMDB claims to be very safe in this regard. But I also read that someone has managed to corrupt the underlying LMDB of the Monero blockchain...
   Maybe it would be good to use some threshold here and sync from time to time to disk as it is done by the MapDB-backed set. The same thing applies to the LMDB-based triple and values stores themselves.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
hmottestad: It's a bit hard to see. This code is called asynchronously from startTransaction(), but also from addStatement() (LmdbSailStore line 673).

hmottestad:
> > > @JervenBolleman I've run some updated benchmarks. What do you think?
> > > ```
> > > new MapDbCollectionFactory(Integer.MAX_VALUE).createSet():
> > > 
> > > Benchmark                     Mode  Cnt      Score      Error  Units
> > > QueryBenchmark.removeByQuery  avgt    5  10689.992 ± 6913.359  ms/op
> > > 
> > > 
> > > PersistentSet (LMDB-based) - commit and SYNC after 100 updates:
> > > 
> > > Benchmark                     Mode  Cnt     Score      Error  Units
> > > QueryBenchmark.removeByQuery  avgt    5  5959.434 ± 1061.529  ms/op
> > > 
> > > 
> > > PersistentSet (LMDB-based) - commit  after 100 updates but DON'T SYNC:
> > > 
> > > Benchmark                     Mode  Cnt     Score     Error  Units
> > > QueryBenchmark.removeByQuery  avgt    5  1159.452 ± 287.824  ms/op
> > > 
> > > 
> > > PersistentSet (LMDB-based) - commit only for iterator and DON'T SYNC:
> > > 
> > > Benchmark                     Mode  Cnt     Score     Error  Units
> > > QueryBenchmark.removeByQuery  avgt    5  1038.176 ± 214.871  ms/op
> > > ```
> > 
> > 
> > I'm wondering about what the use of commit and fsync are. I have a couple of questions:
> > 
> > 1. Does unusedIds get repopulated from disk at startup?
> > 2. Does it matter if some IDs are missing from unusedIds after a crash?
> > 3. Will another thread be able to read stale data from unusedIds if commit or fsync are not used, or is this guaranteed by LMDB or maybe the OS level filesystem cache?
> > 4. Does LMDB guarantee that the backing file(s) will never get corrupted if commit or fsync is omitted? (for instance by using an append only log so that a corrupted indexes can be rebuilt to a known good state, even though data could be lost due to the lack of fsync)
> 
> 1. No, it is just a temporary set used within ONE transaction.

Isn't it a shared variable for the entire LmdbSailStore, and there is only 1 LmdbSailStore (`backingStore` line 73 in LmdbStore)?

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
kenwenzel: Yes. This depends on the mode of the LMDB store. If multi-threading is active then it is executed via a separate worker thread for the triple store, else it is directly executed synchronously. Both modes are never mixed.

kenwenzel:
> > > > @JervenBolleman I've run some updated benchmarks. What do you think?
> > > > ```
> > > > new MapDbCollectionFactory(Integer.MAX_VALUE).createSet():
> > > > 
> > > > Benchmark                     Mode  Cnt      Score      Error  Units
> > > > QueryBenchmark.removeByQuery  avgt    5  10689.992 ± 6913.359  ms/op
> > > > 
> > > > 
> > > > PersistentSet (LMDB-based) - commit and SYNC after 100 updates:
> > > > 
> > > > Benchmark                     Mode  Cnt     Score      Error  Units
> > > > QueryBenchmark.removeByQuery  avgt    5  5959.434 ± 1061.529  ms/op
> > > > 
> > > > 
> > > > PersistentSet (LMDB-based) - commit  after 100 updates but DON'T SYNC:
> > > > 
> > > > Benchmark                     Mode  Cnt     Score     Error  Units
> > > > QueryBenchmark.removeByQuery  avgt    5  1159.452 ± 287.824  ms/op
> > > > 
> > > > 
> > > > PersistentSet (LMDB-based) - commit only for iterator and DON'T SYNC:
> > > > 
> > > > Benchmark                     Mode  Cnt     Score     Error  Units
> > > > QueryBenchmark.removeByQuery  avgt    5  1038.176 ± 214.871  ms/op
> > > > ```
> > > 
> > > 
> > > I'm wondering about what the use of commit and fsync are. I have a couple of questions:
> > > 
> > > 1. Does unusedIds get repopulated from disk at startup?
> > > 2. Does it matter if some IDs are missing from unusedIds after a crash?
> > > 3. Will another thread be able to read stale data from unusedIds if commit or fsync are not used, or is this guaranteed by LMDB or maybe the OS level filesystem cache?
> > > 4. Does LMDB guarantee that the backing file(s) will never get corrupted if commit or fsync is omitted? (for instance by using an append only log so that a corrupted indexes can be rebuilt to a known good state, even though data could be lost due to the lack of fsync)
> > 
> > 
> > 
> > 1. No, it is just a temporary set used within ONE transaction.
> 
> Isn't it a shared variable for the entire LmdbSailStore, and there is only 1 LmdbSailStore (`backingStore` line 73 in LmdbStore)?

Yes, but SnapshotSailStore makes sure that only one writer is active at any time. The transactions are completely serialized.

kenwenzel:
@hmottestad It would also be possible to not permanently share the set in a member variable but it might have a performance impact. AFAIK, I should be right and there is only one active writer at a time when using the SnapshotSailStore. Or am I wrong?

hmottestad:
I don't know. I thought I would add some thread tracking code to your branch just to see if I can find any issues.

kenwenzel:
> I don't know. I thought I would add some thread tracking code to your branch just to see if I can find any issues.

Maybe this is not necessary. Both, NativeStore and LmdbStore rely in the fact that there is only a single writer thread for the backing store at a particular point in time. Concurrent writes would break the whole write logic of the value and triple stores.

hmottestad:
I think it's the fork and sink design that is responsible for making sure that there is only a single writer that is using the main store. 

I did some more testing and couldn't get it to break. 

So this PR is good to merge!

hmottestad (approved):

kenwenzel:
> I think it's the fork and sink design that is responsible for making sure that there is only a single writer that is using the main store.
> 
> I did some more testing and couldn't get it to break.
> 
> So this PR is good to merge!

Great :-) Thank you for investigating this in detail.

Issue #3585, opened 2022-01-08
Improve sorting performance


hmottestad:
### Problem description

 - improve overall performance
 - reduce memory footprint

### Are you interested in contributing a solution yourself?

Yes

PR #3586, opened 2022-01-08
GH-3585 improve sorting performance

hmottestad:
GitHub issue resolved: #3585 

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
@@ -144,12 +144,12 @@ public interface Literal extends Value {
 
 	/**
 	 * Retrieves the {@link TemporalAccessor temporal accessor} value of this literal.
-	 * 
+	 *
 	 * <p>
 	 * A temporal accessor representation can be given for literals whose label conforms to the syntax of the following
 	 * <a href="https://www.w3.org/TR/xmlschema11-2">XML Schema 1.1</a> date/time datatypes:
 	 * </p>
-	 * 
+	 *
 	 * <ul>
 	 *
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#dateTime">xsd:dateTime</a>,</li>
@@ -161,24 +161,24 @@ public interface Literal extends Value {
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#gMonthDay">xsd:gMonthDay</a>,</li>
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#gDay">xsd:gDay</a>,</li>
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#gMonth">xsd:gMonth</a>.</li>
-	 * 
+	 *
 	 * </ul>
-	 * 
+	 *
 	 * <p>
 	 * Temporal accessor representations may be converted to specific {@link java.time} values like
 	 * {@link OffsetDateTime} using target static factory methods, for instance
 	 * {@code OffsetDateTime.from(literal.temporalAccessorValue())}.
 	 * </p>
-	 * 
+	 *
 	 * <p>
 	 * Note however that {@link java.time} doesn't include dedicated classes for some legal XML Schema date/time values,
 	 * like offset dates (for instance, {@code 2020-11-16+01:00}) and {@code xsd:gDay} (for instance, {@code ---16}).
 	 * </p>
-	 * 
+	 *
 	 * @return the temporal accessor value of this literal
 	 *
 	 * @throws DateTimeException if this literal cannot be represented by a {@link TemporalAccessor} value
-	 * 
+	 *
 	 * @since 3.5.0
 	 * @author Alessandro Bollini
 	 *
@@ -192,7 +192,7 @@ public interface Literal extends Value {
 	 *          Datatypes Second Edition</a>: it is not included among temporal datatypes automatically assigned by
 	 *          {@link ValueFactory#createLiteral(TemporalAmount)} in order to provide better interoperability with the
 	 *          latter version of the standard.
-	 * 
+	 *
 	 * @implSpec The default method implementation throws an {@link UnsupportedOperationException} and is only supplied
 	 *           as a stop-gap measure for backward compatibility: concrete classes implementing this interface are
 	 *           expected to override it.
@@ -209,7 +209,7 @@ public interface Literal extends Value {
 	 * <a href="https://www.w3.org/TR/xmlschema-2/">XML Schema 2</a>
 	 * <a href="https://www.w3.org/TR/xmlschema-2/#duration">xsd:duration</a> datatype.
 	 * </p>
-	 * 
+	 *
 	 * <p>
 	 * The adoption of the <a href="https://www.w3.org/TR/xmlschema-2/">XML Schema 2</a> definition is a known deviation
 	 * from the <a href="http://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal">RDF 1.1</a> standard;
@@ -225,7 +225,7 @@ public interface Literal extends Value {
 	 * Note however that {@link java.time} doesn't include dedicated classes for legal XML Schema duration values
 	 * including both date and time components (for instance, {@code P1YT23H}).
 	 * </p>
-	 * 
+	 *
 	 * @return the temporal amount value of this literal
 	 *
 	 * @throws DateTimeException if this literal cannot be represented by a {@link TemporalAmount} value
@@ -290,4 +290,20 @@ public interface Literal extends Value {
 	@Override
 	int hashCode();
 
+	/**
+	 * Checks whether the supplied value is a "plain literal". A "plain literal" is a literal with no datatype and
+	 * optionally a language tag.
+	 *
+	 * @see <a href="http://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#dfn-plain-literal">RDF Literal
+	 *      Documentation</a>
+	 */
+	boolean isPlainLiteral();
+
+	/**
+	 * Checks whether the supplied value is a "simple literal". A "simple literal" is a literal with no language tag nor
+	 * datatype.
+	 *
+	 * @see <a href="http://www.w3.org/TR/sparql11-query/#simple_literal">SPARQL Simple Literal Documentation</a>
+	 */
+	boolean isSimpleLiteral();
 }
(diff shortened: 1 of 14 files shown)


hmottestad:
### Before
```
OOMBenchmark.sortAllObjectsWithSparql:    Count:  2
OOMBenchmark.sortDirectly:                Count:  3
```

### After
```
OOMBenchmark.sortAllObjectsWithSparql:    Count: 11
OOMBenchmark.sortDirectly:                Count: 74
```

core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
```diff
@@ -290,4 +290,20 @@ default TemporalAmount temporalAmountValue() throws DateTimeException {
...
+	 */
+	boolean isPlainLiteral();
+
+	/**
+	 * Checks whether the supplied value is a "simple literal". A "simple literal" is a literal with no language tag nor
+	 * datatype.
+	 *
+	 * @see <a href="http://www.w3.org/TR/sparql11-query/#simple_literal">SPARQL Simple Literal Documentation</a>
+	 */
+	boolean isSimpleLiteral();
```
hmottestad, line 308: These two methods improve performance because there are many points where we can hard code the values or compute the values more efficiently for new instances.

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
```diff
@@ -158,6 +158,38 @@ public static boolean isDecimalDatatype(IRI datatype) {
...
+			return true;
+		} else if (XSD.UNSIGNED_LONG == datatype) {
+			return true;
+		} else if (XSD.UNSIGNED_INT == datatype) {
+			return true;
+		} else if (XSD.UNSIGNED_SHORT == datatype) {
+			return true;
+		} else if (XSD.UNSIGNED_BYTE == datatype) {
+			return true;
+		}
```
hmottestad, line 191: This is faster than looking up in a hashmap because we don't need to compute the hash first. This is also much easier for the JVM to optimize.

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
```diff
@@ -1215,7 +1271,7 @@ private static String normalizeFPNumber(String value, String minMantissa, String
 			}
 			sb.append(mantissa.charAt(firstDigitIdx));
 			sb.append('.');
-			sb.append(mantissa.substring(firstDigitIdx + 1, dotIdx));
+			sb.append(mantissa, firstDigitIdx + 1, dotIdx);
 			sb.append(mantissa.substring(dotIdx + 1));
```
hmottestad, line 1275: should probably optimize the other substring call too

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -49,12 +49,17 @@
 	 * The literal's language tag.
 	 */
 	private String language;
+	transient private Optional<String> optionalLanguage = Optional.empty();
```
hmottestad, line 52: Because `getLanguage()` returns an optional and we often have to check for a language during sorting this ends up creating a lot of garbage through new Optional objects. Precomputing it is much faster.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -138,12 +158,16 @@ protected void setLanguage(String language) {
...
 	@Override
 	public Optional<String> getLanguage() {
-		return Optional.ofNullable(language);
+		if (optionalLanguage == null) {
+			optionalLanguage = Optional.ofNullable(language);
+		}
+		return optionalLanguage;
 	}
 
 	protected void setDatatype(IRI datatype) {
```
hmottestad, line 173: Having protected setters for everything is a real pain in the backside. It means we can't assume that the object is immutable which is why I've had to implement a "lateInit" boolean.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
```diff
@@ -66,6 +71,9 @@
 	 *--------------*/
 
 	protected SimpleLiteral() {
+		plain = false;
+		simple = false;
+		lateInit = true;
 	}
```
hmottestad, line 77: I wish we could remove this and make all the variables in this object final. Maybe we can deprecate this class and create a new one with everything being final to use for our own code.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
```diff
@@ -74,7 +74,8 @@ public static String getLabel(Optional v, String fallback) {
 		if (l instanceof SimpleLiteral) {
 			return ((SimpleLiteral) l).getXsdDatatype();
 		}
-		return Optional.empty();
+		// we return null to signify that the literal does not support the new XSD Datatype enum
+		return null;
```
hmottestad, line 78: Using null to signal lack of support for XSD datatype.

I was thinking about maybe using an interface for the XSD datatype enum, but then we won't be able to use `==`. The only real solution is to put the XSD class somewhere so we can depend on it from both the Literal class and the code that does sorting.

hmottestad: Out of scope for this PR.

hmottestad: https://github.com/eclipse/rdf4j/issues/3587

hmottestad:
Todo:
 [ ] document changes to the interface in the release notes

hmottestad:
Benchmark that uses the no-op Epsilon GC to measure how many times an operation can be run before running out of memory. 

### Before
```
OOMBenchmark.sortAllObjectsWithSparql:    Count:  2
OOMBenchmark.sortDirectly:                Count:  3
```

### After
```
OOMBenchmark.sortAllObjectsWithSparql:    Count: 11
OOMBenchmark.sortDirectly:                Count: 74
```

core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
jeenbroekstra: Plain literal as a concept doesn't really exist in RDF 1.1 anymore - a literal is either simple (no language tag), or if it has a language tag, it also has a datatype (and is therefore not plain). I wonder if we can't get rid of the logic surrounding that check completely. 

I'll try and find more time later this week to look into this PR in more depth. I think the idea looks good but I haven't fully grasped the consequences yet for compatibility, or for other store implementors. I also am a little concerned that having the logic about which literals plain/simple implemented in different places is going to cause inconsistencies (if I remember correctly this used to be handled by a utility class).

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
jeenbroekstra: I am not really following why you are letting a method that has an optional as it return type return `null`. That seems error-prone to me.

hmottestad: The point is for the method to return an empty optional if the datatype isn't an XSD datatype. But it also used to return an empty optional if the support for XSD datatypes was missing. So an empty optional would signal both "this is not an XSD datatype" and "I don't know if it's an XSD datatype".

core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
hmottestad: Yeah. It used to be a utility class, which was really slow since it needed check for a language tag and try to match the datatype each time it was called, which is typically log(n) times for each literal when sorting.

jeenbroekstra: Oh yeah not saying this isn't a better approach (it is), just that I'd like to make sure we migrate properly and consistently.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleLiteral.java
jeenbroekstra: There's `GenericLiteral` in the model-api package that might be a good spot for this?

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
jeenbroekstra: Ah I see, thanks. That makes sense then. We should reflect this distinction in the javadoc though.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
hmottestad: We could do a default implementation, but I would rather not. I don't really feel that these two methods need to be user facing, so I was thinking about making an ExtendedLiteral interface for them (and maybe other methods we want down the line). Then check for that interface with instanceof, and if not use the utility.

Though if you think we could drop one of the methods then maybe it's not too bad.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Literals.java
hmottestad: Yeah. Definitely. Though I have the issue mentioned above where I want to add support for core datatypes as enums directly in the Literal interface and force other implementers to support it too. I that would really simplify our handling of datatypes pretty much everywhere. 

Something like `Optional<CoreDatatype> getCoreDatatype()` and have CoreDatatype within the model-api module.

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
kenwenzel: How do you ensure that the passed datatype is exactly identical (and not only equal) to any of the constants?

hmottestad: There is a fallback after this sequence of if/elseif that uses the HashSet. So this is faster if you get a match without being significantly slower if you don't get a match.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
hmottestad: If we merge https://github.com/eclipse/rdf4j/pull/3596 then these interface changes may not be needed. Enum based datatypes will make the helper utils much faster and probably fast enough.

jeenbroekstra (changes requested):
@hmottestad Given that GH-3619 addresses the same issue and is more recent, I assume this PR can be discarded?

hmottestad:
Yeah. It can, but this branch is a bit faster for the "direct" benchmark which is a bit odd. Would be nice to know the difference, so I'll keep it around for just a bit longer.

Issue #3589, opened 2022-01-10
Cache toString in MemIRI


hmottestad:
### Problem description

toString() method in MemIRI mints a new string every time.

### Preferred solution

Cache it using a soft reference.

PR #3592, opened 2022-01-10
GH-3589 cache toString() in MemIRI 

hmottestad:
GitHub issue resolved: #3589 

 - cache the results of toString() using a soft reference to reduce the number of times we need to recompute it
    - typically the .toString() method is called when retrieving the MemIRI from MemValueFactory with an AbstractIRI because the .equals() method there uses the full string representation instead of the namespace and local name like MemIRI does.

----


jeenbroekstra (approved):
Good find - LGTM!

Issue #3594, opened 2022-01-10
NumericLiteral extends SimpleLiteral but should extend AbstractLiteral instead.


JervenBolleman:
### Current Behavior

As the NumericLiteral extends SimpleLiteral we allocate a number of strings and do number to string conversions even when this is not required. This is a flame graph of a custom store "select (count(?s) as ?c) WHERE {?s ?p ?o}" where quite a bit of time is spend in calling an `Long.toString` whose output is never used.

[output.pdf](https://github.com/eclipse/rdf4j/files/7839402/output.pdf)
.

### Expected Behavior

NumericLiteral should only give a string when asked for getLabel() or stringValue().

### Version

4.0.0-develop

### Are you interested in contributing a solution yourself?

Yes

PR #3595, opened 2022-01-10
GH-3594 Only do toString on the Number value when getLabel is called

JervenBolleman:
Signed-off-by: Jerven Bolleman <jerven.bolleman@sib.swiss>

GitHub issue resolved: #3594 

For a NumericLiteral store only the Number value and datatype.

----

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
@@ -145,15 +145,14 @@ public class NumericLiteral extends AbstractLiteral {
 
 	@Override
 	public boolean equals(Object o) {
-		if (o instanceof NumericLiteral)
-			return equals((NumericLiteral) o);
-		else
+		if (this == o) {
+			return true;
+		}
+		if (o instanceof NumericLiteral) {
+			return datatype == ((NumericLiteral) o).datatype && number.equals(((NumericLiteral) o).number);
+		} else {
 			return super.equals(o);
+		}
 	}
 
-	private boolean equals(NumericLiteral o) {
-		if (this == o)
-			return true;
-		return number.equals(o.number) && datatype.equals(o.datatype);
-	}
 }
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
@@ -7,38 +7,47 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.model.impl;
 
+import java.util.Objects;
+import java.util.Optional;
+
 import org.eclipse.rdf4j.model.IRI;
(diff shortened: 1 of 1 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
```diff
@@ -7,32 +7,41 @@
...
+		assert Objects.nonNull(number);
+		assert Objects.nonNull(datatype);
+		this.datatype = datatype;
 		this.number = number;
 	}
 
 	protected NumericLiteral(Number number, XSD.Datatype datatype) {
-		super(XMLDatatypeUtil.toString(number), datatype);
-		this.number = number;
+		this(number, datatype.getIri());
```
hmottestad, line 44: This nullifies the sorting performance improvements that depend on the XSD.Datatype enum. If we end up merging https://github.com/eclipse/rdf4j/pull/3596 then it won't matter, but if not it would be worthwhile to implement support for XSD.Datatype here in this class.

hmottestad (changes requested):
Can you hold on from merging this until we've had a discussion about my proposal for a new datatype enum: https://github.com/eclipse/rdf4j/pull/3596

core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
JervenBolleman: I would like to merge the XSD.Datatype work into this when it is ready.

hmottestad:
@JervenBolleman I've merged the CoreDatatype PR. If you rebase your branch I can double check it for you before you merge it.

JervenBolleman:
@hmottestad this has been rebased with the coredatype work used.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
```diff
@@ -113,4 +122,37 @@ public float floatValue() {
...
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o instanceof NumericLiteral) {
+			return datatype == ((NumericLiteral) o).datatype && number.equals(((NumericLiteral) o).number);
+		} else {
+			return super.equals(o);
+		}
+	}
```
hmottestad, line 156: I cleaned up the equals method a bit. Since datatype is an enum it is faster to check than the number value, that's why I switched them around.

hmottestad (approved):
I've considered if this would have an impact for our users (and those who extend RDF4J) wrt. that I want to keep the changes between M3 and the final 4.0.0 release as small as possible. I think it should be fine.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/NumericLiteral.java
JervenBolleman: Good for me.

JervenBolleman:
I will wait for the checks to run and then merge and close the issue.

Issue #3587, opened 2022-01-09
Datatype enum in the RDF4J model-api module


hmottestad:
### Problem description

We can't add general support for XSD datatype enums to the Literal interface because of how the modules are constructed.

The XSD datatypes are a fundamental part of RDF, unlike the other vocabularies. 

The `rdf:langString` should also be included. Essentially any datatype that is mentioned explicitly in the RDF or SPARQL spec.

We should have a general datatype enum class with all datatypes that we have to support for things like sorting. Would be nice to make it extensible, but I'm not sure if that would really work.

Being able to rely on a datatype enum within all our code will both simplify our code and improve performance.

With 4.0.0 we have a chance to add this as a required feature, otherwise we would have to have a fallback to allow implementers to not implement this method.

### Preferred solution

`Optional<CoreDatatype> getCoreDatatype()` on the Literal interface. 

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

_No response_we

Issue #2643, opened 2020-11-09
implement LINDT UCUM units of measure


VladimirAlexiev:
I think that LINDT UCUM units of measure of @maximelefrancois86 are very useful, see https://github.com/w3c/sparql-12/issues/129 for a description and discussion.

Is there interest to port LINDT from a Jena fork (https://github.com/OpenSensingCity/jena-ucum) to rdf4j?

A useful discussion is in https://lists.apache.org/thread.html/77cebe345679553c1df0a97d3c7ce29ad5555d2231522557420a8ae0@%3Cdev.jena.apache.org%3E . Excerpts from the mlist thread:

- Maxime Lefrancois:
  - our implementation uses the recent JSR 385, Units of Measurement API 2.0
  - uses https://github.com/unitsofmeasurement/uom-systems/tree/master/ucum-java8 (a new Jena dependency)
  - touches jena-core and jena-arq; may need integration in other modules (jena-tdb, ... )
  - Compare Jena 3.6.0 and Jena 3.6.0-ucum
https://github.com/apache/jena/compare/master...OpenSensingCity:jena-3.6.0-ucum
  - UCUM is very comprehensive and has different implementations in different programming languages. It would be possible to implement UCUM datatypes in other RDF-SPARQL engines.
  - not possible to implement as External module (extension): I had to add a new value space VSPACE_QUANTITY (and NodeValueQuantity for javax.measure.Quantity) to overload the SPARQL operators '<>=' and arithmetic functions '+-*/'.
  - see VSPACE_QUANTITY: https://github.com/OpenSensingCity/jena-ucum/blob/jena-3.6.0-ucum/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java#L566
  - see eg multiply: https://github.com/OpenSensingCity/jena-ucum/blob/jena-3.6.0-ucum/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/NodeValueOps.java#L283
- Andy Seaborne (ajs6f):
  - would this contribution have any effect (or should it) on jena-spatial? Would it be either necessary or if not, appropriate to integrate there?
  - What is the relation to QUDT? http://www.qudt.org. ANSWER: UCUM codes are in QUDT 2.1, see comments by dr-shorthair in https://github.com/w3c/sparql-12/issues/129
  - Comparator<NodeValue>? (Thinking of sorting.) Comparing by implicit value can get into 
[...]

PR #3596, opened 2022-01-10
GH-3587 CoreDatatype enum for literals

hmottestad:
GitHub issue resolved: #3587 

 - create an enum for all the core datatypes without depending on any vocabularies (like XSD.Datatype does)
     - also include rdf:langString
     - includes all the methods from XSD.Datatype
     - also has a bunch of other helper methods and a cache implementation to simplify implementation for us
 - add method to Literal interface to return an Optional<CoreDatatype>
     - this method is not allowed to return null and if it returns an empty optional then that is because the datatype is note one of the XSD datatypes or rdf:langString
 - implemented the new method in our code
 
The next step would be to take advantage of the new datatype enum throughout our code base to simplify and speed the places where we check the datatype against either rdf:langString or one of the XSD datatypes.

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
@@ -18,6 +18,8 @@ import java.util.Optional;
 
 import javax.xml.datatype.XMLGregorianCalendar;
 
+import org.eclipse.rdf4j.model.base.CoreDatatype;
+
 /**
  * An RDF-1.1 literal consisting of a label (the lexical value), a datatype, and optionally a language tag.
  *
@@ -144,12 +146,12 @@ public interface Literal extends Value {
 
 	/**
 	 * Retrieves the {@link TemporalAccessor temporal accessor} value of this literal.
-	 * 
+	 *
 	 * <p>
 	 * A temporal accessor representation can be given for literals whose label conforms to the syntax of the following
 	 * <a href="https://www.w3.org/TR/xmlschema11-2">XML Schema 1.1</a> date/time datatypes:
 	 * </p>
-	 * 
+	 *
 	 * <ul>
 	 *
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#dateTime">xsd:dateTime</a>,</li>
@@ -161,24 +163,24 @@ public interface Literal extends Value {
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#gMonthDay">xsd:gMonthDay</a>,</li>
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#gDay">xsd:gDay</a>,</li>
 	 * <li><a href="https://www.w3.org/TR/xmlschema11-2/#gMonth">xsd:gMonth</a>.</li>
-	 * 
+	 *
 	 * </ul>
-	 * 
+	 *
 	 * <p>
 	 * Temporal accessor representations may be converted to specific {@link java.time} values like
 	 * {@link OffsetDateTime} using target static factory methods, for instance
 	 * {@code OffsetDateTime.from(literal.temporalAccessorValue())}.
 	 * </p>
-	 * 
+	 *
 	 * <p>
 	 * Note however that {@link java.time} doesn't include dedicated classes for some legal XML Schema date/time values,
 	 * like offset dates (for instance, {@code 2020-11-16+01:00}) and {@code xsd:gDay} (for instance, {@code ---16}).
 	 * </p>
-	 * 
+	 *
 	 * @return the temporal accessor value of this literal
 	 *
 	 * @throws DateTimeException if this literal cannot be represented by a {@link TemporalAccessor} value
-	 * 
+	 *
 	 * @since 3.5.0
 	 * @author Alessandro Bollini
 	 *
@@ -192,7 +194,7 @@ public interface Literal extends Value {
 	 *          Datatypes Second Edition</a>: it is not included among temporal datatypes automatically assigned by
 	 *          {@link ValueFactory#createLiteral(TemporalAmount)} in order to provide better interoperability with the
 	 *          latter version of the standard.
-	 * 
+	 *
 	 * @implSpec The default method implementation throws an {@link UnsupportedOperationException} and is only supplied
 	 *           as a stop-gap measure for backward compatibility: concrete classes implementing this interface are
 	 *           expected to override it.
@@ -209,7 +211,7 @@ public interface Literal extends Value {
 	 * <a href="https://www.w3.org/TR/xmlschema-2/">XML Schema 2</a>
 	 * <a href="https://www.w3.org/TR/xmlschema-2/#duration">xsd:duration</a> datatype.
 	 * </p>
-	 * 
+	 *
 	 * <p>
 	 * The adoption of the <a href="https://www.w3.org/TR/xmlschema-2/">XML Schema 2</a> definition is a known deviation
 	 * from the <a href="http://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal">RDF 1.1</a> standard;
@@ -225,7 +227,7 @@ public interface Literal extends Value {
 	 * Note however that {@link java.time} doesn't include dedicated classes for legal XML Schema duration values
 	 * including both date and time components (for instance, {@code P1YT23H}).
 	 * </p>
-	 * 
+	 *
 	 * @return the temporal amount value of this literal
 	 *
 	 * @throws DateTimeException if this literal cannot be represented by a {@link TemporalAmount} value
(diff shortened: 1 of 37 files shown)


hmottestad:
@jeenbroekstra @JervenBolleman @aschwarte10 @barthanssens @rdstn What do you guys think about my approach to having an enum for all the core literal datatypes? I really believe that this will allow us to simplify and optimize a lot of our code that depends on checking datatypes.

JervenBolleman:
@hmottestad I think this reads nice and clean :) and looks like an improvement. Is it worth adding the GeoSPARQL datatype as well?

barthanssens (commented):
Nice improvement indeed

core/model/src/main/java/org/eclipse/rdf4j/model/datatypes/XMLDatatypeUtil.java
```diff
@@ -289,6 +290,78 @@ public static boolean isValidValue(String value, IRI datatype) {
...
+		} else if (datatype == CoreDatatype.XSD_LANGUAGE) {
+			result = Literals.isValidLanguageTag(value);
+		}
+
+		return result;
+	}
+
 	/**
 	 * Verifies if the supplied lexical value is a valid decimal or not.
 	 *
```
barthanssens, line 367: Just wondering if a switch wouldn't be better/more readable.
I suppose it's just a matter of taste (assuming the JVM optimizes the code anyway)

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/CoreDatatype.java
```diff
@@ -0,0 +1,282 @@
...
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Optional;
+
+import javax.xml.datatype.Duration;
+import javax.xml.datatype.XMLGregorianCalendar;
+
+import org.eclipse.rdf4j.model.IRI;
+
+public enum CoreDatatype {
```
jeenbroekstra, line 20: Can we make `CoreDatatype` implement a basic interface and then use that interface consistently? That way, we can later on add additional enum classes with other datatype vocabularies, if we want.

Also, perhaps we should call this enum "BuiltInDatatype" as they correspond to the list of XML Schema built-in datatypes defined here: https://www.w3.org/TR/xmlschema11-2/#built-in-datatypes  (plus the one extra odd case of langString).

hmottestad:
> @hmottestad I think this reads nice and clean :) and looks like an improvement. Is it worth adding the GeoSPARQL datatype as well?

I hadn't thought about those. I can see that geo:wktLiteral is used a fair bit. Any other that we should include?

jeenbroekstra (commented):
I really like this approach @hmottestad,  good stuff.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/CoreDatatype.java
hmottestad: I had initially used an interface and made CoreDatatypes implement that interface. The reason I dropped that approach was that I got worried that if we use `==` then we would get into trouble if someone implements their own XSD datatypes and expects it to function with our code.

I'm open for other names, BuiltInDatatype would work well too. We aren't limited to XSD datatypes though, any datatype that we use directly in our code would make sense to add.

jeenbroekstra: Fair enough. l'm fine with sticking with CoreDatatype (and no interface) then, at least for this iteration.

hmottestad: https://github.com/eclipse/rdf4j/issues/2643

An interface would definitely make it easier to support units of measure.

jeenbroekstra: If you want to go for an interface after all: we can stick a warning on the interface documentation to indicate restrictions in third parties extending it. Or even choose to mark it `@InternalUseOnly` for now (most normal usage shouldn't _directly_ use it, anyway). If we do go that route, I'd vote for calling this enum `BuiltInDatatype` (and document that it reflects the XML Schema built-ins), and perhaps use `CoreDatatype` for the more general interface.

hmottestad:
@jeenbroekstra I have created an interface with three inner classes for the three datatype namespaces. What do you think?

core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/XSD.java
```diff
@@ -186,58 +189,58 @@
...
-		NEGATIVE_INTEGER(XSD.NEGATIVE_INTEGER, false, false, true, true, true, false, false),
-		NON_NEGATIVE_INTEGER(XSD.NON_NEGATIVE_INTEGER, false, false, true, true, true, false, false),
-		POSITIVE_INTEGER(XSD.POSITIVE_INTEGER, false, false, true, true, true, false, false),
-		UNSIGNED_LONG(XSD.UNSIGNED_LONG, false, false, true, true, true, false, false),
-		UNSIGNED_INT(XSD.UNSIGNED_INT, false, false, true, true, true, false, false),
-		UNSIGNED_SHORT(XSD.UNSIGNED_SHORT, false, false, true, true, true, false, false),
-		UNSIGNED_BYTE(XSD.UNSIGNED_BYTE, false, false, true, true, true, false, false),
-		YEARMONTHDURATION(XSD.YEARMONTHDURATION, false, true, false, true, false, false, false);
+		DURATION(create("duration"), true, true, false, false, false, false, false),
+		DATETIME(create("dateTime"), true, false, false, false, false, false, true),
```
jeenbroekstra, line 198: I'm not sure I follow - it looks like you are now creating IRI constants for the XSD datatypes both here _and_ in the inner enum in `CoreDatatype`. Shouldn't one reuse the other?

hmottestad: Good point. The XSD vocabulary class should probably use the IRI constants from the CoreDatatype.XSD enums.

hmottestad:
I want to test out dropping the use of optional. Instead of retuning null though I thought we should return a fixed enum to indicate that the datatype is not a core datatype. 

This would be for performance. If it's not really faster then I think we should stick to optional.

hmottestad:
I've switched away from using Optional. If there is no matching CoreDatatype then we return CoreDatatype.NONE. This way we don't have any null pointer issues.

```
// we can compare directly
yourLiteral.getCoreDatatype() == CoreDatatype.XSD.STRING;

// for convenience we can "cast" directly
CoreDatatype.XSD xsdDatatype = yourLiteral.getCoreDatatype().asXSDDatatypeOrNull();

// or just check
boolean isXsd = yourLiteral.getCoreDatatype().isXSDDatatype();
```

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/CoreDatatype.java
```diff
@@ -0,0 +1,340 @@
...
+		return null;
+	}
+
+	default RDF asRDFDatatypeOrNull() {
+		return null;
+	}
+
+	default GEO asGEODatatypeOrNull() {
+		return null;
+	}
```
jeenbroekstra, line 52: Can we remove the `OrNull` bit from the method names here and return an Optional instead? We could then do:

```java
CoreDatatype.XSD xsdDatatype = yourLiteral.getCoreDatatype().asXSDDatatype().orElse(null);
```

which IMHO is a bit cleaner.

core/model-api/src/main/java/org/eclipse/rdf4j/model/Literal.java
```diff
@@ -280,7 +280,7 @@ default TemporalAmount temporalAmountValue() throws DateTimeException {
 	 *           match any of the core datatypes. Instead of minting a new Optional object every time this method is
 	 *           called it's recommended to use {@link CoreDatatype#asOptional()}.
 	 */
-	<T extends CoreDatatype> Optional<T> getCoreDatatype();
+	CoreDatatype getCoreDatatype();
```
jeenbroekstra, line 283: Happy with that - will need a javadoc tweak (but you've already got that on your TODO I saw 👍 )

jeenbroekstra (commented):
Nice! Just one more suggestion but overall this looks good to me.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/CoreDatatype.java
hmottestad: I kinda wanted to stay away from using Optional. Performance wasn't great and probably won't be until GraalVM becomes the norm. I would rather return null in this case, but I can change the method name and add some javadocs.

hmottestad: Let me double check the performance though. I have a branch where I've rewritten the sorting code to use CoreDatatype, so I can check there if it's actually any slower.

hmottestad: Performance was actually perfectly fine.

Issue #3634, opened 2022-01-24
LMDB value store does not cache own values


kenwenzel:
### Current Behavior

If an ID for a value is looked up from persistent storage and the value was created by LMDB store's value factory then the corresponding value -> ID mapping is not correctly cached.

### Expected Behavior

The value -> ID mapping should be stored in ValueStore.valueIDCache after looking up the ID from persistent storage.

### Version

4.0.0-M2

### Are you interested in contributing a solution yourself?

Yes

PR #3635, opened 2022-01-24
GH-3634 Fix caching of LMDB value store

kenwenzel:
GitHub issue resolved: #3634 

Also cache IDs of own values after lookup from persistent storage.

----

### DIFF
--- a/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
+++ b/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
@@ -575,6 +575,8 @@ class ValueStore extends AbstractValueFactory {
 				if (isOwnValue) {
 					// Store id in value for fast access in any consecutive calls
 					((LmdbValue) value).setInternalID(id, revision);
+					// Store id in cache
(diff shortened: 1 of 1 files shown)


jeenbroekstra (approved):

fkleedorfer:
I got curious about LMDB because of this implementation, and it seems client-side caching is not required because lookups are so fast (. If you don't mind me asking: Is the value cache really needed? It could actually slow things down or become a bottleneck due to locking...

For example, https://github.com/lmdbjava/lmdbjava states as one important advantage of LMDB: "Freedom from application-side data caching (memory-mapped files are more efficient)".

kenwenzel:
@fkleedorfer This is an interesting question and not easy to answer.
LMDB is fast if the relevant pages can be kept in memory which is the case when your db's size is not very larger than the available RAM size.
Furthermore, when searching for a value the RDF4J LMDB store first needs to convert it to a byte array and then uses this byte array as key for the lookup.
This conversion involves memory allocations and multiple operations for translating the data. Due to the B+-Tree the lookup within LMDB also has the complexity of log(n).
I also think that locking may not be an issue because the caches are using ConcurrentHashMap that relies on multiple locks (> 16) for different hash segments.

Nevertheless, it may be worth to test the performance of the LMDB store without caching to finally answer your question.

Issue #3646, opened 2022-01-27
SailSourceBranch does not clear changes if flushing to sink fails


kenwenzel:
### Current Behavior

If changes are flushed to a SailSink and one change fails to apply the rest of the changes are kept in the queue.
If at a later point the sail repository is shutdown then it tries to flush these stale changes again.

### Expected Behavior

The changes should be cleared if an error occurs while flushing the changes to a sink.

### Version

4.0.0-M2

### Are you interested in contributing a solution yourself?

Yes

PR #3647, opened 2022-01-27
GH-3646 Clear changes if SailException occurs.

kenwenzel:
GitHub issue resolved: #3646 

Correctly clear changes if an exception occurs while flushing to sink.

----

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
@@ -267,6 +267,11 @@ class SailSourceBranch implements SailSource {
 					prepared = null;
 				}
 			}
+		} catch (SailException e) {
+			// clear changes if flush fails
+			changes.clear();
+			prepared = null;
+			throw e;
 		} finally {
 			semaphore.unlock();
 		}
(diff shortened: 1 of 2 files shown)


core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
```diff
@@ -267,6 +267,10 @@ public void flush() throws SailException {
 					prepared = null;
 				}
 			}
+		} catch (SailException e) {
+			// clear changes if flush fails
+			changes.clear();
```
jeenbroekstra, line 272: Can we add a regression test of some sort that demonstrates the situation this fixes? I realize there isn't a lot of unit test coverage for this code, but perhaps it would fit in as part of the OptmisticIsolationTest suite? 

Also: should you do the same kind of thing for the `prepared` iterms (presunably inside the `prepare()` method)?

kenwenzel: Yes, for sure. I also think this could somehow be integrated into the OptimisticIsolationTest suite. I will take a look into this.

kenwenzel:
@jeenbroekstra I also reset the prepared attribute now and I've also added a basic test that fails if the changes are flushed again on shutdown of the sail store.

jeenbroekstra (approved):
Good stuff, thanks @kenwenzel

Issue #2874, opened 2021-02-24
Concurrent reads from SailConnection


hmottestad:
The ShaclSail uses concurrent reads from a single connection to be able to validate shapes in parallel. We should investigate if this could be an issue. 

Java collections generally state that even though they don't support concurrency, they generally allow concurrent reads. So it makes sense for our connections to support concurrent reads and long as writes+reads are synchronized. 

One place this could be an issue is the observed statements in the Changeset. 

Essentially we need to look for statefulness in the code that would be connected to read operations.

jeenbroekstra:
This is not something we can fix generally, or even want to. We have no full control over what third party implementations do, which means that you cannot rely on this behavior in your ShaclSail, even _if_ you manage to fix it in the specific nativestore/memorystore sail implementations.

hmottestad:
We could make it required behavior starting with 4.0.0. 

It's important for the performance of the ShaclSail. And it would be useful behavior for supporting parallel query evaluation in the future. HashJoin could benefit from that.

jeenbroekstra:
> We could make it required behavior starting with 4.0.0.

Not something this fundamental to the architecture. Not without serious polling of sail/repository implementors and getting their approval. At the _very_ least you will have to solicit approval from vendors like Ontotext, Stardog, Blazegraph and others. 

> It's important for the performance of the ShaclSail.

That's unfortunate, but unless you can get buy-in from third parties on the idea, you'll just have to find another solution that fits within the semantics of the framework. 

I also still don't buy that it is necessary, by the way. If you want to read things in parallel threads, just open two connections.

hmottestad:
Opening multiple connections is a bit problematic since they would all need to contain the same changes as the current transaction. 

And parallel validation is particularly useful for bulk validation. Currently I'm seeing a 3x speedup on my 4 core MacBook.

hmottestad:
Btw. Updated the title to reflect that this is only needed for SailConnections and not RepositoryConnections.

jeenbroekstra:
> Opening multiple connections is a bit problematic since they would all need to contain the same changes as the current transaction.

...and I assume that you cannot rely on using something like READ_UNCOMMITTED isolation because that is only a minimal guarantee on what _might_ happen, not a guarantee that you definitelly _will_ see all changes before they're committed. Correct?

> Btw. Updated the title to reflect that this is only needed for SailConnections and not RepositoryConnections.

[...]

PR #3655, opened 2022-02-01
GH-2874 interface to signal a SailConnection that supports concurrent reads

hmottestad:
GitHub issue resolved: #2874 

 - an interface with a single method to signal support
 - implemented in the MemoryStore and NativeStore, wrapper connections and the ShaclSail
 - no tests planned for this at the moment

----

### DIFF
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ThreadSafetyAware.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ThreadSafetyAware.java
@@ -1,0 +1,24 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Distribution License v1.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.sail.features;
+
+/**
+ * An interface used to signal thread safety features of a sail or its related classes.
+ */
+public interface ThreadSafetyAware {
+
+	/**
+	 * A class may support concurrent reads from multiple threads against the same object. This ability may change based
+	 * on an object's current state.
+	 *
+	 * @return true if this object supports concurrent reads
+	 */
+	boolean supportsConcurrentReads();
+
+}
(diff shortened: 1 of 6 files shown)


core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ConcurrentReadsSailConnection.java
```diff
@@ -0,0 +1,26 @@
...
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.sail.features;
+
+/**
+ * And interface used to signal that the SailConnection supports concurrent reads on the same connection.
+ */
+public interface ConcurrentReadsSailConnection {
```
jeenbroekstra, line 14: I'd name this more generally, e.g. `ThreadSafetyAware`. That gives us (some) license in future to add more related signal methods, if needed.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ConcurrentReadsSailConnection.java
```diff
@@ -0,0 +1,26 @@
...
+
+	/**
+	 * All sails supports concurrent reads as long as each reader uses their own connection. Some connections also
+	 * support concurrent reads on the same connection.
+	 *
+	 * @return true if this connection supports concurrent reads
+	 */
+	default boolean supportsConcurrentReads() {
+		return true;
```
jeenbroekstra, line 23: Should the default be (conservatively) false?

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ConcurrentReadsSailConnection.java
hmottestad: Sure.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ConcurrentReadsSailConnection.java
hmottestad: If a class doesn't support concurrent reads then it shouldn't implement this class. Wrapper classes that implement this interface should call the delegate if the delegate implements this interface.

That being said I've noticed that most of the implementations at the moment are overriding this to provide a conditional return (eg. the memory store connection that supports concurrent reads up to, but not including SERIALIZABLE).

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/features/ThreadSafetyAware.java
```diff
@@ -0,0 +1,24 @@
...
+
+	/**
+	 * A class may support concurrent reads from multiple threads against the same object. This ability may change based
+	 * on an object's current state.
+	 *
+	 * @return true if this object supports concurrent reads
+	 */
+	boolean supportsConcurrentReads();
+
+}
```
hmottestad, line 24: @jeenbroekstra I've renamed the class as you suggested. Do you think my javadocs make sense?

Issue #3623, opened 2022-01-20
Remote transactions set PRESERVE_BNODE_IDS to true and break the expected behaviour of blank node uniqueness across imported files


avataar:
### Current Behavior

Given a remote repository, a transaction and importing two different files that share a blank node identifier, for example:

```java
try (RepositoryConnection connection = repository.getConnection()) {
    connection.begin();
    connection.add(new StringReader("<urn:a> <urn:b> _:c ."), RDFFormat.TURTLE);
    connection.add(new StringReader("<urn:a> <urn:d> _:c ."), RDFFormat.TURTLE);
    connection.commit();
}
```

Both instances of `_:c` will be imported exactly as `_:c`.

### Expected Behavior

Blank nodes should be scoped within a given serialization and thus the two instances should resolve to two unique blank nodes. This is actually how RDF4J works with local transactions.

The issue was introduced in ancient Sesame times here: https://bitbucket.org/openrdf/sesame/commits/2246487cdf0966eb46451498e443d471a7291b62, and later the code was migrated to the class Transaction. I couldn't find any explanation as to why it was needed.

As a side note: using `connection.getParserConfig().set(BasicParserSettings.PRESERVE_BNODE_IDS, false)` on a remote connection doesn't seem to do anything. It should either work as expected (affecting the settings) or throw an exception that setting parser settings on a remote connection isn't possible.

### Version

3.7.4

### Are you interested in contributing a solution yourself?

Yes

jeenbroekstra:
> As a side note: using connection.getParserConfig().set(BasicParserSettings.PRESERVE_BNODE_IDS, false) on a remote connection doesn't seem to do anything. It should either work as expected (affecting the settings) or throw an exception that setting parser settings on a remote connection isn't possible.

Setting parser config on a remote connection  does affect actual parser settings, but it's not the parser that's being used at the server end you're configuring. The client-side also has a parser (used for processing server responses), and _that_ is what you configure. There's currently no way to control the server's parser remotely via the API, though there's a feature request to allow that kind of thing (#446).

jeenbroekstra:
> The issue was introduced in ancient Sesame times here: https://bitbucket.org/openrdf/sesame/commits/2246487cdf0966eb46451498e443d471a7291b62, and later the code was migrated to the class Transaction. I couldn't find any explanation as to why it was needed.

[...]

PR #3662, opened 2022-02-03
GH-3623 Use unique bnode ids in remote transactions

DenitsaStoianova:
Remove setting of PRESERVE_BNODE_IDS parameter to true when a remote transaction is created. So when data is added to transaction the parser will use the default value of the parameter which is false and will create new identifiers to ensure uniqueness of blank nodes across multiple files.
If preserveNodeId parameter of HttpServletRequest is set to true, PRESERVE_BNODE_IDS will be reconfigured and bnode ids will be preserved.

GitHub issue resolved: #3623  

----

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
@@ -417,7 +417,6 @@ class Transaction implements AutoCloseable {
 		Future<RepositoryConnection> result = submit(() -> {
 			RepositoryConnection conn = rep.getConnection();
 			ParserConfig config = conn.getParserConfig();
-			config.set(BasicParserSettings.PRESERVE_BNODE_IDS, true);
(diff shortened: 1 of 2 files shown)


jeenbroekstra (commented):
It took me a while to puzzle my way through this, at first I thought this might be too simple a fix and would cause unwanted behavior in other use cases. But I've noticed/remembered that bnode preservation is something that is actually communicated from the client to server via a parameter, and that parameter should govern behavior - not this hardcoded setting here. So: good spot, and I think this fix makes sense.

Could you perhaps take a look at adding a regression test?

DenitsaStoianova:
Added test which checks whether by default blank nodes are added with unique identifiers to transactions or their identifiers are preserved by specifying a parameter for server instructions when parsing request data.

jeenbroekstra:
Thanks @DenitsaStoianova, that looks great! There's a small code formatting issue it seems - can you run `mvn process-resources` to fix and then commit / push?

jeenbroekstra (approved):
Lgtm, thanks!

Issue #3665, opened 2022-02-04
Update rdf4j-spring


fkleedorfer:
### Problem description

In the project that contributed `rdf4j-spring`, we still haven't been able to switch to the 4.0.0 milestone, and we've had some minor changes to our rdf4j-spring code that we'd like to contribute - also with the goal to switch to the rdf4j implementation soon.

The changes are:

* handle the case of a  lost WeakReference in the ClearableAwareUpdate gracefully
* minor convenience changes
* provide helpful log output if a commit fails in a ShaclSail 
* enforce read-only behaviour in transactions explicitly marked read-only

### Preferred solution

I'll make a PR with all these changes if that's ok.

### Are you interested in contributing a solution yourself?

Yes

PR #3666, opened 2022-02-04
GH-3665: rdf4j spring update

fkleedorfer:
GitHub issue resolved: #3665

In the project that contributed `rdf4j-spring`, we still haven't been able to switch to the 4.0.0 milestone, and we've had some minor changes to our rdf4j-spring code that we'd like to contribute - also with the goal to switch to the rdf4j implementation soon.

The changes are:

* handle the case of a  lost WeakReference in the ClearableAwareUpdate gracefully
* minor convenience changes
* provide helpful log output if a commit fails in a ShaclSail 
* enforce read-only behaviour in transactions explicitly marked read-only

----


spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/support/ShaclSailValidationReportHelper.java
```diff
@@ -0,0 +1,41 @@
...
+				break;
+			}
+		}
+		if (thowable != null && thowable instanceof ShaclSailValidationException) {
+			Model model = ((ShaclSailValidationException) thowable).validationReportAsModel();
+			StringWriter w = new StringWriter();
+			Rio.write(model, System.err, RDFFormat.TURTLE);
+		}
+	}
+}
```
hmottestad, line 41: Could you try to check for ValidationException instead?

I could also recommend pretty printing and blank node inlining.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/support/DataInserter.java
```diff
@@ -41,7 +42,18 @@ public void insertData(Resource dataFile) {
...
+					Set.of(
+							RDFFormat.TURTLE,
+							RDFFormat.TRIG,
+							RDFFormat.NTRIPLES,
+							RDFFormat.NQUADS))
+					.orElseThrow(
+							() -> new IllegalArgumentException(
+									"Failed to determine file format of input file "
+											+ dataFile));
+			con.add(dataFile.getInputStream(), "", fmt);
```
jeenbroekstra, line 56: I'd change this to 

```java
RDFFormat fmt = Rio.getParserFormatForFilename(dataFile.getName()).orElseThrow(....)
```
Advantage is that this uses all registered parsers to determine a suitable format, instead of a hard-coded list.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/support/ShaclSailValidationReportHelper.java
jeenbroekstra: More generally: should this helper code live here? It does not look like something specific to spring, at first glance. Though the same is true for SHACL's `ValidationReport` class itself - perhaps this is a good opportunity to find a good spot for both these things in one of the common-packages?

jeenbroekstra (commented):
Looks sensible to me insofar as I can judge :)  couple of inline remarks

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/support/ShaclSailValidationReportHelper.java
fkleedorfer: > More generally: should this helper code live here? It does not look like something specific to spring, at first glance. Though the same is true for SHACL's `ValidationReport` class itself - perhaps this is a good opportunity to find a good spot for both these things in one of the common-packages?

If not here, then probably the `rdf4j-shacl` module, right? You won't need this class unless you are (or might be) using shacl. In that case, I'd put it in `src/main/java/org/eclipse/rdf4j/sail/shacl`. Ok?

> Could you try to check for ValidationException instead?

yes. good point.
 
> I could also recommend pretty printing and blank node inlining.

Good points, too. I would have liked these features when working on our shacl shapes. Following blank node ids around is tedious ;-)

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/support/DataInserter.java
fkleedorfer: ah, yes. Thanks.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/support/ShaclSailValidationReportHelper.java
fkleedorfer: addressed these in the latest commits to the PR

fkleedorfer:
(Ready to merge on my end)

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailValidationReportHelper.java
```diff
@@ -0,0 +1,49 @@
...
+		for (thowable = t; thowable != null; thowable = thowable.getCause()) {
+			if (thowable instanceof ValidationException) {
+				break;
+			}
+		}
+		if (thowable != null && thowable instanceof ValidationException) {
+			Model model = ((ValidationException) thowable).validationReportAsModel();
+			Rio.write(model, System.err, RDFFormat.TURTLE, WRITER_CONFIG);
+		}
+	}
```
hmottestad, line 48: Could we pass in the OutputStream instead of hard coding it, especially since the method name says 'log' but there isn't a logger.

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailValidationReportHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.rdf4j.rio.WriterConfig;
+import org.eclipse.rdf4j.rio.helpers.BasicWriterSettings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @since 4.0.0
+ * @author Florian Kleedorfer
+ */
+public class ShaclSailValidationReportHelper {
```
hmottestad, line 28: Could you mark the class as `@InternalUseOnly` so we can make changes in the future without having to first deprecate things?

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailValidationReportHelper.java
fkleedorfer: I don't like that System.err is hard-coded here either. I actually would want the output to be end up in the logger, but using the logger directly would require to pass the loglevel (or hardcode that). I'll make the change you suggest and add a method that just returns the output as a string, so it's easier to log it if that's what you want

core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailValidationReportHelper.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * *****************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ * *****************************************************************************
+ */
```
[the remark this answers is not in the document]
fkleedorfer: sorry about that.

jeenbroekstra (approved):

Issue #3672, opened 2022-02-10
UUID in StrictEvaluationStrategy is costly to create


hmottestad:
The StrictEvaluationStrategy registers each instance with the EvaluationStrategies cache. The key is a UUID, which is a good key because it is generally considered to be unique both internally to one machine and externally to other machines. In order to instantiate a UUID the JVM uses secure random, which is a very limited resource per physical machine. Since each query evaluation instantiates a new StrictEvaluationStrategy the UUID generation becomes a bottleneck for parallel queries.

PR #3673, opened 2022-02-10
GH-3672 

hmottestad:
GitHub issue resolved: # 

----


hmottestad:
Pity that this passes. That means we don't have any tests that cover the registration of evaluation factories :(

jeenbroekstra:
If I remember correctly this cache was added to aid on-disk overflow using MapDB of query results/iterations. This uses Java object serialization but we can't serialize the strategy object itself, so its use as a field on serializable objects was declared transient and retrieved from the cache upon deserialization - specifically in the OrderComparator. See https://bitbucket.org/openrdf/sesame/commits/8efdaed81ec6f969c15e28f029d7364e6236f2a5 .

As far as I can tell that use case is still there. I'll try and look at this in more detail over the weekend but unless I missed something this PR will break MapDB on-disk overflow of iterations in query evaluation. Clearly we haven't got this test-covered properly :(

hmottestad:
I've figured this out. The pre-compilation that @JervenBolleman introduced means that we don't need to have a reference to the evaluation strategy anymore. So this is safe to merge. 

Another question I ran into was why the OrderComparator needs to be serializable at all? I can't see why we would need to serializable what is essential the evaluation plan for a query. I understand that we use serializable to spill to disk, but I don't understand why we want to spill OrderComparator to disk. The  order iteration (or something like that) spills its BindingSets to disk, presumably to not run out of memory when sorting a large amount of data.

PR #3701, opened 2022-02-27
GH-3671 isomorphism bug

jeenbroekstra:
GitHub issue resolved: #3671 

- added regression test that demonstrates problem
- increasing sha length to 256 bits fixes the issue

----

### DIFF
--- a/compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicTest.java
+++ b/compliance/model/src/test/java/org/eclipse/rdf4j/model/util/IsomorphicTest.java
@@ -85,61 +85,56 @@ public class IsomorphicTest {
 		internallyIsomorphic_2 = getModel("internallyIsomorphic.ttl");
 		manyProperties_2 = getModel("manyProperties.ttl");
 		manyProperties2_2 = getModel("manyProperties2.ttl");
-
 	}
 
 	@Test
 	public void empty() {
-
 		isomorphic(empty, empty_2);
-
 	}
 
 	@Test
 	public void blankNodes() {
-
 		isomorphic(blankNodes, blankNodes_2);
-
 	}
 
 	@Test
 	public void shacl() {
-
 		isomorphic(shacl, shacl_2);
-
 	}
 
 	@Test
 	public void longChain() {
-
 		isomorphic(longChain, longChain_2);
-
 	}
 
 	@Test
 	public void sparqlTestCase() {
-
 		isomorphic(sparqlTestCase, sparqlTestCase_2);
+	}
 
+	@Test
+	public void testGH3671Case() {
+		// The test tends to succeed the first time around but then fails on a subsequent pass
+		for (int i = 0; i < 10; i++) {
+			// we need to re-parse the model in every parse for the failure to occur
+			Model m1 = getModel("GH-3671-case.ttl");
+			Model m2 = getModel("GH-3671-case.ttl");
+			isomorphic(m1, m2);
+		}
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
I'm speculating that the problem has to do with how we combine multiple hashes to calculate unique values, and that there can happen a collission more frequently than anticipated when using a lower size for the hashes.

I've ran a few benchmarks to see if going to a stronger hash would mean a significant performance decrease: 

Using murmur3_128:

```
Benchmark                                          Mode  Cnt  Score   Error  Units
IsomorphicBenchmark.shacl                          avgt   10  3.408 ± 0.212  ms/op
IsomorphicBenchmark.shaclValidationReport          avgt   10  9.749 ± 0.387  ms/op
IsomorphicBenchmark.shaclValidationReport_changed  avgt   10  9.994 ± 0.514  ms/op
```

Using sha256:

```
Benchmark                                          Mode  Cnt   Score   Error  Units
IsomorphicBenchmark.shacl                          avgt   10   3.718 ± 0.168  ms/op
IsomorphicBenchmark.shaclValidationReport          avgt   10  10.669 ± 0.091  ms/op
IsomorphicBenchmark.shaclValidationReport_changed  avgt   10  10.717 ± 0.324  ms/op
```

It is (naturally) slightly slower, but not significantly so.

hmottestad:
Is there no way to detect collisions and maybe resolve them or fail over to the legacy isomorphic algorithm?

jeenbroekstra:
> Is there no way to detect collisions and maybe resolve them or fail over to the legacy isomorphic algorithm?

No.

hmottestad:
> > Is there no way to detect collisions and maybe resolve them or fail over to the legacy isomorphic algorithm?
> 
> No. Or rather: perhaps there is, but it would need significant investment of research and dev time, which I currently don't have. In the mean time, we need to put reliability over performance and merge this as a short term fix.

Just wondering if maybe there is some underlying issue that is causing collisions to be more frequent?

jeenbroekstra:
> > > Is there no way to detect collisions and maybe resolve them or fail over to the legacy isomorphic algorithm?
> > 
> > 
> > No. Or rather: perhaps there is, but it would need significant investment of research and dev time, which I currently don't have. In the mean time, we need to put reliability over performance and merge this as a short term fix.
> 
> Just wondering if maybe there is some underlying issue that is causing collisions to be more frequent?

Oh, yes, most likely. I suspect that it has to do with how we calculate combined hashes. I _think_ we're following the letter of the algorithm as outlined in the technical paper there, but perhaps I've missed something. Certainly, in theory 128 bits should be enough to not cause collisions in graphs of this size (or any practical size, really). Yet clearly we do get a collision of sorts, and clearly using a larger hash size mitigates the issue. Perhaps there's an overflow in the hash combination algorithm that makes it collapse to a smaller value space? The trouble is that I can't quite put my finger on it, and I've not been able to reduce the example - as soon as I take even a few triples out the isomorphism consistently succeeds.

hmottestad:
You're good to merge this as is. 

Could you file a follow up issue though to look into the collisions? Maybe something along the lines of hashing with both murmur and sha to try to find out exactly where the collision is occurring.

hmottestad (approved):

jeenbroekstra:
> You're good to merge this as is.
> 
> Could you file a follow up issue though to look into the collisions? Maybe something along the lines of hashing with both murmur and sha to try to find out exactly where the collision is occurring.

Yep good point, will do.

Issue #3676, opened 2022-02-13
url-encode.xsl can stack overflow on long strings in workbench


nguyenm100:
### Current Behavior

url-encode.xsl is called by table.xsl to format the results by workbench but if the string is too long, it can cause a stack overflow and fail to display any result.

url-encode uses recursion to walk the supplied string: https://github.com/eclipse/rdf4j/blob/main/tools/workbench/src/main/webapp/transformations/url-encode.xsl#L64

### Expected Behavior

display correct results

### Steps To Reproduce

execute a query that will return a long string (say 2k chars).

### Version

3.7.4

nguyenm100:
hi @jeenbroekstra , I submitted a patch for this issue but it doesn't include any unit-tests as I didn't see anything relating to transformation/xslt testing in the existing codebase that I could repurpose.  was wondering if you could advise on how best to test this change.  I did test on an external site: https://xslttest.appspot.com/ and of course against our dataset.

jeenbroekstra:
One thing to note is that this appears to be specific to Chrome-based browsers - I could not reproduce the issue at first (using Firefox), only when I switched to Brave did I see the problem.

PR #3735, opened 2022-03-14
A patch for url-encode.xsl potential to stackoverflow

nguyenm100:
GitHub issue resolved: #3676

Use a divide and conquer approach to iterate through the xslt string in order to prevent a stackoverflow for large strings.  see: https://stackoverflow.com/a/9081077/5237756

----


jeenbroekstra:
Thanks for submitting this patch @nguyenm100 ! I'll need to find some time to build the workbench and run a test with this - can't think of an easier way to test this unfortunately.

jeenbroekstra (approved):
I've managed to reproduce this locally (using a Chrome-based browser, the problem does not occur in Firefox it seems) and also verified that the patch appears to fix the problem. Thanks again @nguyenm100!

Issue #3737, opened 2022-03-15
Maven artifacts aren't fully Java 8 compatible.


pulquero:
### Current Behavior

Trying to use the workbench webapp on Java 8 gives:
`
java.lang.NoSuchMethodError: java.nio.ByteBuffer.rewind()Ljava/nio/ByteBuffer;
	org.eclipse.rdf4j.sail.nativerdf.datastore.HashFile.<init>(HashFile.java:125)
`

### Expected Behavior

No error to occur.

### Steps To Reproduce

OpenJDK 1.8.0_322 Temurin
Use tomcat 9 to run RDF4J workbench.

### Version

3.7.5

### Anything else?

Issue seems to be due to not using the new release argument for the maven-compiler-plugin.

See https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release and https://github.com/eclipse/jetty.project/issues/3244 for a similar issue.

jeenbroekstra:
We should check the Jenkins config. I vaguely remember setting the JVM used for release deployment to 11 for the 4.0 milestones, we may have forgotten to switch it back Java 8 for the patch release.

jeenbroekstra:
I'll take a closer look at adding that release argument, but for now I have created separate Jenkins jobs specifically for deploying releases using Java 8:

- https://ci.eclipse.org/rdf4j/view/release%20deployment/job/rdf4j-java8-deploy-release-ossrh/
- https://ci.eclipse.org/rdf4j/view/release%20deployment/job/rdf4j-java8-deploy-release-sdk/

jeenbroekstra:
We went with setting the compiler flags (using profiles for different JDK versions) in the end, so those separate Jenkins jobs are no longer necessary.

pulquero:
Will there be a 3.7.6 (re-)release of 3.7.5? 3.7.5 is un-usuable for me, and it contains a fix I need for my testing?

PR #3743, opened 2022-03-20
GH-3737 use release flag to compile for Java 8

jeenbroekstra:
GitHub issue resolved: #3737  

Replace the use of of the `source` and `target` compiler arguments with the (new since Java 9) `release` argument. This will ensure that when we do 3.7.x releases, the produced artifacts will be compatible with Java 8 (even if the artifacts are compiled using a newer JVM).

A potential downside: this flag will cause an error if you are _actually_ using Java 8 to compile the code, as the `release` argument was only introduced in Java 9. However, given that we are all running more modern Javas on our machines, and that our Jenkins deployment jobs also have been configured to use Java 11, I think  this may be the simplest way to fix the issue.

See https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -239,6 +239,49 @@
 			</build>
 		</profile>
 		<profile>
+			<id>java8-compiler-flags</id>
+			<activation>
+				<jdk>1.8</jdk>
+			</activation>
+			<build>
+				<pluginManagement>
+					<plugins>
+						<plugin>
+							<groupId>org.apache.maven.plugins</groupId>
+							<artifactId>maven-compiler-plugin</artifactId>
+							<configuration>
+								<fork>true</fork>
+								<source>1.8</source>
+								<target>1.8</target>
+								<encoding>utf8</encoding>
+							</configuration>
+						</plugin>
+					</plugins>
+				</pluginManagement>
+			</build>
+		</profile>
+		<profile>
+			<id>java9-compiler-flags</id>
+			<activation>
+				<jdk>[9,)</jdk>
+			</activation>
+			<build>
+				<pluginManagement>
+					<plugins>
+						<plugin>
+							<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
I overlooked that this will break our github actions.... We would need to tweak this using a profile to get it working. 

Alternatively, we just scratch this, and instead make sure that we use the Java 8 Jenkins jobs I created when doing 3.x releases (see comment on issue ticket).

hmottestad:
I've been thinking a bit about this one. If we switch to java 9+ for GitHub CI we would be able to use the release flag to target java 8, but when running tests we would still be running them in a newer JVM. 

The release flag is definitely the way to go, since it also checks that all the java apis that are called form the code actually exist in the given version. 

I think we can set the release flag from the properties section, and then presumably from the command line or environment as well. Maybe we can tweak the GitHub CI to unset the release flag and instead set the source and target version instead?

jeenbroekstra:
> I've been thinking a bit about this one. If we switch to java 9+ for GitHub CI we would be able to use the release flag to target java 8, but when running tests we would still be running them in a newer JVM.
> 
> The release flag is definitely the way to go, since it also checks that all the java apis that are called form the code actually exist in the given version.
> 
> I think we can set the release flag from the properties section, and then presumably from the command line or environment as well. Maybe we can tweak the GitHub CI to unset the release flag and instead set the source and target version instead?

Possibly. I'm just on the fence if the effort is really worth it, given that we are close to getting 4.0 released. We're unlikely to do many more Java 8 releases.

jeenbroekstra:
I can possibly do something using maven profiles to handle this. Trying it out now.

jeenbroekstra:
This did the trick - happy with it. We'll need to be a little careful in the main->develop sync after this is merged to make sure develop config doesn't end up in a weird state.

Issue #3753, opened 2022-03-27
Closed Changeset still used


hmottestad:
The Changeset has a close method and is actively closed by the supporting code. A closed changeset may still be used.

Issue #3754, opened 2022-03-27
Changeset should be cleared on close to help GC


hmottestad:
Clearing a Changeset on close will make it easier to garbage collect.

PR #3755, opened 2022-03-27
GH-3754 clear changeset on close

hmottestad:
GitHub issue resolved: #3753 #3754 

----

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
@@ -103,13 +103,25 @@ abstract class Changeset implements SailSink, ModelFactory {
 	 */
 	private boolean statementCleared;
 
+	private boolean closed;
+
 	@Override
 	public void close() throws SailException {
-		// no-op
+		closed = true;
+		refbacks = null;
+		prepend = null;
+		observed = null;
+		approved = null;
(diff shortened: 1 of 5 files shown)


core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
```diff
@@ -348,7 +377,28 @@ protected void setChangeset(Changeset from) {
 		this.statementCleared = from.statementCleared;
 	}
 
+	public static Changeset simpleClone(Changeset from) {
```
jeenbroekstra, line 380: Couldn't this be an instance method as easily? And why `simpleClone` rather than `clone`? Either way, might be good to stick a small javadoc on this to explain its purpose.

jeenbroekstra (commented):
Thanks for separating this out  @hmottestad . Mostly looks fine to me, just one remark.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
hmottestad: Maybe it would be better to call it a shallow clone because it doesn't clone any of the data structures and instead just copies of the the references (pointers). 

I'm not sure why I went with a static method, I'll try to change is to an instance method and see what happens.

Issue #3765, opened 2022-04-02
Uploading RDF-star via Workbench does not work


robertbbb:
### Current Behavior

Dear all,

I am not able to upload Turtle-star or Trig-star to a repository via the Add screen in the workbench. When using those formats I get the exception 

org.eclipse.rdf4j.repository.RepositoryException: value parameter should be a URI, BNode or Literal

### Expected Behavior

When uploading Turtle-star data via the Add screen the data should be uploaded. Instead, it corrupts the repository.

### Steps To Reproduce

1. I use versions RDF4J 3.7.6 on Tomcat 9.0.62, JRE 1.8.0_321
2. I try to upload the following via the Add screen, with Turtle-star format set

```
@prefix : <http://example.org/>.
<<:a :b :c>> :d :e.
```

3. I get back an Internal Server Error with message: org.eclipse.rdf4j.repository.RepositoryException: value parameter should be a URI, BNode or Literal

4.Afterwards the whole repository gets corrupted, nothing else can be loaded in it.

### Version

3.7.6

### Are you interested in contributing a solution yourself?

No

jeenbroekstra:
I can reproduce the error. It turns out the data is actually added to the store, and it also accepts additional data being added - so the repository is not actually corrupted. However, the Workbench UI keeps showing 500 errors.

jeenbroekstra:
Cause is a bug in the memory store: although the store itself supports RDF-star, its on-disk persistence mechanism does not.

jeenbroekstra:
Thanks for reporting this @robertbbb. Working on a fix, will be available in the next patch release (3.7.7) when that becomes available.

PR #3767, opened 2022-04-03
GH-3765 add support for rdf-star triple persistence in memorystore

jeenbroekstra:
GitHub issue resolved: #3765  

- added support for reading/writing RDF-star triples in memory store custom persistence format
- extended rdf-star tests to explicitly run with a persistence-enabled memory store

----

### DIFF
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/FileIO.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/FileIO.java
@@ -32,9 +32,11 @@ import org.eclipse.rdf4j.model.Literal;
 import org.eclipse.rdf4j.model.Namespace;
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.Triple;
 import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.model.ValueFactory;
 import org.eclipse.rdf4j.model.util.Literals;
+import org.eclipse.rdf4j.rio.helpers.RDFStarUtil;
 import org.eclipse.rdf4j.sail.SailException;
 import org.eclipse.rdf4j.sail.base.SailDataset;
 import org.eclipse.rdf4j.sail.base.SailSink;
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Verified locally that with these fixes in place, Turtle-star upload via the Workbench to a memory store with disk sync now works as expected.

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/MemoryRDFStarSupportTest.java
```diff
@@ -7,19 +7,25 @@
...
  */
 public class MemoryRDFStarSupportTest extends RDFStarSupportTest {
 
+	@TempDir
+	File tempDir;
+
 	@Override
 	protected Repository createRepository() {
-		return new SailRepository(new MemoryStore());
+		return new SailRepository(new MemoryStore(tempDir));
```
hmottestad, line 28: Would it make sense to duplicate this so we have one class with persistence and another without?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/FileIO.java
```diff
@@ -306,6 +314,9 @@ private Value readValue(DataInputStream dataIn) throws IOException, ClassCastExc
 			String label = readString(dataIn);
 			IRI datatype = (IRI) readValue(dataIn);
 			return vf.createLiteral(label, datatype);
+		} else if (valueTypeMarker == RDFSTAR_TRIPLE_MARKER) {
+			IRI rdfStarEncodedTriple = (IRI) readValue(dataIn);
+			return RDFStarUtil.fromRDFEncodedValue(rdfStarEncodedTriple);
```
hmottestad, line 319: Do any of the tests check that this code works by adding data, shutting down the repo, then creating a new one with the same directory and check if all the data is still there?

core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/MemoryRDFStarSupportTest.java
jeenbroekstra: We can do that easily enough, but I'm not sure if it adds anything: the tests without persistence test exactly the same thing, only without persistence. If the test _with_ persistence succeeds, the one without will, also.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/FileIO.java
jeenbroekstra: I _think_ we cover that in some of the integration tests, but I'm not 100% sure to be honest. I was also a bit surprised the FileIO itself has no direct unit test coverage. Might be worth putting a followup card in for that.

jeenbroekstra: Added a card to look into this further: https://github.com/eclipse/rdf4j/issues/3771

Issue #3782, opened 2022-04-08
Regression/Change in 4.0 M3: query parser fails with "variable 'o1' in projection not present in GROUP BY" (ORDER BY clause)


aschwarte10:
### Current Behavior

We have applied the 4.0 M3 milestone in our application, and one of our query parsing/rendering unit tests is failing (while it was working with 3.7.x and 4.0-M2)

It looks like some regression (or at least a change) was introduced in the recent milestone

I have reduced the query example to the minimum where it occurs at parsing time

```
SELECT DISTINCT ?s (?o AS ?o1) 
WHERE {
	?s ?p ?o 
} GROUP BY ?s ?o 
ORDER BY ?s ?o1 
```

Fails with: `variable 'o1' in projection not present in GROUP BY`

### Expected Behavior

As this is a change compared to previous versions, we need to have a good understanding of "correctness w.r.t the SPARQL spec". I do not have sufficient knowledge whether ORDER BY is executed on the final projected result or on an intermediate result.

If the change / regression is not intended and correct w.r.t the spec, the expectation is that the query parser behaves as in previous versions of RDF4J

### Steps To Reproduce

Can be reproduced with the following unit test snippet

```
String query = "SELECT DISTINCT ?s (?o AS ?o1) \n"
                + "WHERE {\n"
                + "    ?s ?p ?o \n"
                + "} GROUP BY ?s ?o \n"
                + "ORDER BY ?s ?o1 ";
QueryParserUtil.parseOperation(QueryLanguage.SPARQL, query, null);
```

which fails with

```
org.eclipse.rdf4j.query.MalformedQueryException: variable 'o1' in projection not present in GROUP BY.
	at org.eclipse.rdf4j.query.parser.sparql.SPARQLParser.buildQueryModel(SPARQLParser.java:220)
	at org.eclipse.rdf4j.query.parser.sparql.SPARQLParser.parseQuery(SPARQLParser.java:178)
	at org.eclipse.rdf4j.query.parser.QueryParserUtil.parseOperation(QueryParserUtil.java:49)
	...
Caused by: org.eclipse.rdf4j.query.parser.sparql.ast.VisitorException: variable 'o1' in projection not present in GROUP BY.
	at org.eclipse.rdf4j.query.parser.sparql.TupleExprBuilder.visit(TupleExprBuilder.java:659)
	at org.eclipse.rdf4j.query.parser.sparql.TupleExprBuilder.visit(TupleExprBuilder.java:235)
[...]

PR #3788, opened 2022-04-09
GH-3782 clean up handling of GROUP BY projection element verification

abrokenjester:
GitHub issue resolved: #3782  

When a query has a GROUP BY clause, the projection may only contain:

 1. aggregate expressions
 2. constants
 3. simple expressions consisting of just a variable (if and only if that var is in the GROUP BY)

code changes: 

- check presence of all non-simple expression types in projeciton when using GROUP BY
- clearer errors are given
- added test cases

----

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
@@ -17,7 +17,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.LongAdder;
 
 import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
 import org.eclipse.rdf4j.model.IRI;
@@ -636,14 +635,16 @@ public class TupleExprBuilder extends AbstractASTVisitor {
 		if (group != null) {
 			for (ProjectionElem elem : projElemList.getElements()) {
 				if (!elem.hasAggregateOperatorInExpression()) {
-					// non-aggregate projection elem is only allowed to be a simple var or constant (see
+					// non-aggregate projection elem is only allowed to be a constant or a simple expression (see
 					// https://www.w3.org/TR/sparql11-query/#aggregateRestrictions)
 					ExtensionElem extElem = elem.getSourceExpression();
 					if (extElem != null) {
 						ValueExpr expr = extElem.getExpr();
-						throw new VisitorException(
-								"non-aggregate expression '" + expr
-										+ "' not allowed in projection when using GROUP BY.");
+						if (!(expr instanceof ValueConstant)) {
+							throw new VisitorException(
+									"non-aggregate expression '" + expr
+											+ "' not allowed in projection when using GROUP BY.");
+						}
 
 					} else {
 						Set<String> groupNames = group.getBindingNames();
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
It appears there are some W3C manifest test cases that contradict my current reading of the spec. Needs a bit more investigation.

aschwarte10 (commented):
From what I can tell the fix looks good. It addresses the finding and corresponds to the spec interpretation summarized in the issue (i.e. renamings are considered simple expressions).

I have tested this piece of code on our codebase.

The specific unit test is still failing, but with the agreed interpretation of the spec this is also expected. Note that for the below query (kind of artificial, but still covers some aspects) RDF4J 3.7.x and RDF4j 4.0 M2 were accepting it

```
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
PREFIX dc: <http://purl.org/dc/elements/1.1/>
SELECT DISTINCT 
	?s (?o AS ?o1) 
	(<http://www.w3.org/2001/XMLSchema#boolean>(?o) AS ?o2) 
	(COUNT(DISTINCT ?p) AS ?count) 
	(rdf:type AS ?typeProperty) 
WHERE {
	{ 	
		?s ?p ?o . 
		?s rdf:type foaf:Person . 
		BIND ('42' AS ?knowsname ) . 
		?o rdf:type dc:Document . } 
		MINUS 
		{
			?s rdf:type rdfs:Resource . 
			OPTIONAL { ?s ?p ?o . } 
		} 
		FILTER (NOT EXISTS { ?s rdfs:label ?o }) 
	} GROUP BY ?s ?o 
		HAVING (?count=5) 
		ORDER BY ?s ?o1 
		LIMIT 10 OFFSET 5 
		VALUES (?s) { (<http:example.org/dummy>) }
```

With this branch we are seeing (as-expected):

```
non-aggregate expression 'FunctionCall (http://www.w3.org/2001/XMLSchema#boolean)
   Var (name=o)
' not allowed in projection when using GROUP BY
```


Removing the functional expression from the projection of our example query, let's the parsing succeed

@jeenbroekstra maybe the above query is also of interest for you. We seem to use it in the context of query rendering (where we also do parsing)

I am good to merge. Maybe a second pair of eyes for confirmation

jeenbroekstra:
@aschwarte10 thanks for the feedback, and that additional query looks interesting. If you have the means, could you rewrite it to this and try it out:

```sparql
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
PREFIX dc: <http://purl.org/dc/elements/1.1/>
SELECT DISTINCT 
	?s (?o AS ?o1) 
	?o2
	(COUNT(DISTINCT ?p) AS ?count) 
	(rdf:type AS ?typeProperty) 
WHERE {
	{ 	
		?s ?p ?o . 
		?s rdf:type foaf:Person . 
		BIND ('42' AS ?knowsname ) . 
		?o rdf:type dc:Document . } 
		MINUS 
		{
			?s rdf:type rdfs:Resource . 
			OPTIONAL { ?s ?p ?o . } 
		} 
		FILTER (NOT EXISTS { ?s rdfs:label ?o }) 
	} GROUP BY ?s ?o (<http://www.w3.org/2001/XMLSchema#boolean>(?o) AS ?o2) 
		HAVING (?count=5) 
		ORDER BY ?s ?o1 
		LIMIT 10 OFFSET 5 
		VALUES (?s) { (<http:example.org/dummy>) }
```

In other words: move the boolean cast to _inside_ the GROUP BY, and project the resulting alias (o2).
That should be syntactically fine, but if you can try I'd like to verify that it also gives the expected result when evaluated.

Issue #1755, opened 2019-12-18
Move format-specific Rio settings to parser/writer package


barthanssens:
Currently the format-specific (RDF/XML, JSON-LD... ) are all part of the org.eclipse.rdf4j.rio.helpers package, which means that format-specific (extra) dependencies are sometimes required in the pom.xml.

A cleaner separation would be better, having only abstract / format-agnostic settings in rio.helpers and moving format-specific setting to the parser/writer package (e.g. org.eclipse.rdf4j.rio.jsonld)

jeenbroekstra:
If this is something we want to pick up for 5.0, we should make it a priority to put deprecation notices in for the stuff we intend to move, in the next 4.x release.

barthanssens:
Working on a branch to deprecate the current org.eclipse.rdf4j.rio.helpers; classes in 4.3.0 (except the "Basic" settings and perhaps the generic - non-LD - JSON settings) and create class(es) per parser package

barthanssens:
A benefit is that removing the deprecated classes in e.g. 5.0 will also remove the jsonld-java dependency from the helper package

barthanssens:
Removing the classes in helper package to be scheduled in 5.0 ?

PR #3791, opened 2022-04-12
GH-1755 move parser-specific settings out of generic helper package

jeenbroekstra:
GitHub issue resolved: #1755  

- parser/writer settings for binary, turtle, json-ld, rdf/sjon, and trix have been moved into their parser-specific packages/modules
- settings for n-triples and xml-based parsers/writers have been left as they are generically reused in several places

@barthanssens I'd like your thoughts on whether we really get any benefits of this change. I'm personally not 100% convinced it's the right thing to do.

----

### DIFF
--- a/core/rio/api/src/test/java/org/eclipse/rdf4j/rio/RDFWriterTest.java
+++ b/core/rio/api/src/test/java/org/eclipse/rdf4j/rio/RDFWriterTest.java
@@ -63,7 +63,6 @@ import org.eclipse.rdf4j.model.vocabulary.SP;
 import org.eclipse.rdf4j.model.vocabulary.SPIN;
 import org.eclipse.rdf4j.rio.helpers.BasicParserSettings;
 import org.eclipse.rdf4j.rio.helpers.BasicWriterSettings;
-import org.eclipse.rdf4j.rio.helpers.JSONLDMode;
 import org.eclipse.rdf4j.rio.helpers.StatementCollector;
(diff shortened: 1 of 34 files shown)


jeenbroekstra:
@hmottestad @JervenBolleman @aschwarte10 any of you have any opinions on this refactor? _If_ we want to do this, we should do it before 4.0.

hmottestad:
I would really prefer to keep the merges to develop to bug fixes only since we are so close to releasing 4.0.0.

I haven't looked through your code though, so I don't know how big of a change this is. Could it be targeted at 4.1.0? Or maybe deprecate some stuff now so that it's easier to remove in 5.0.0?

I've got a lot of breaking changes in my performance branch, so I'm fairly eager to work towards 5.0.0 sooner rather than later.

jeenbroekstra:
It's package reshuffling, so kind of a pain to do in a minor release (you'd have to introduce deprecated duplicates of the moved classes in the old location, which is annoying when using code completion), but not impossible. But my main thing is really that I'm not convinced this refactor really gains us anything. The basic idea (have all parser-specific stuff in that parser's package/module instead of in the generic rio api module) is sound enough but it ignores the fact that a lot of this is still shared between multiple parser implementations.

And you're right: it's rather close to release. Perhaps it's better to just park this for now, then do a more comprehensive reorg (including some of the shared stuff) later, for 5.0.

Issue #3807, opened 2022-04-19
Code cleanup


hmottestad:
Code cleanup:

 - remove unused imports
 - remove extra `;`
 - disable lock tracking and global debug property after test is done
 - use junit 5 more
 - use `<>` where possible
 - make private variables final if possible
 - cleanup superfluous static/public/final
 - use List.of(...) where possible (it's faster and cleaner than Arrays.asList(...))
 - remove implemented interface that is already inherited
 - close unclosed resources
 - redundant initialization
 - use try-with-resource

PR #3808, opened 2022-04-19
GH-3807 code cleanup

hmottestad:
GitHub issue resolved: #3807 

See: #3807

----


hmottestad:
No need to merge this before 4.0.0.

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/util/WorkbenchRequest.java
```diff
@@ -116,7 +116,7 @@ public int getInt(String name) throws BadRequestException {
 
 	@Override
 	public String getParameter(String name) {
-		String result = null;
+		String result;
```
jeenbroekstra, line 119: Interesting that this can now be done - in older versions of Java this would result in a "value may have not have been initialized" compile error. Which version of Java was this changed in? I assume the behavior is now that it implicitly initializes local vars to `null`?

jeenbroekstra (approved):
Haven't gone through every file but from what I've seen it all looks sensible to me.

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/util/WorkbenchRequest.java
hmottestad: Don't know how new of a feature this is. The compiler just checks that there is no code path where an unitialized variable can be read.

Useful for things like:

```
int res;
if(something()){
    res = 1;
}else{
    res = 7;
}


```

Issue #3811, opened 2022-04-19
In the QueryEvaluationContext the now value is not thread safe.


JervenBolleman:
### Current Behavior

The now value is not volatile nor thread safely initialized.

### Expected Behavior

That it is thread safe

### Version

4.0.0-develop

### Are you interested in contributing a solution yourself?

Yes

Issue #3798, opened 2022-04-13
Binary operators should be precompiled (building on GH-2741)


JervenBolleman:
### Problem description

This allows for passing through optimizations in the evaluation of queries.

### Are you interested in contributing a solution yourself?

Yes

PR #3812, opened 2022-04-19
GH-3811 ensure thread safety of the shared now value.

JervenBolleman:
GitHub issue resolved: #3811 

Ensure the now value is instantiated to the same value but lazily at a low cost.

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
@@ -7,6 +7,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.impl;
 
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
 import java.util.Date;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
@@ -14,6 +16,7 @@ import java.util.function.Predicate;
 
 import org.eclipse.rdf4j.model.Literal;
 import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.ValueFactory;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.query.Binding;
 import org.eclipse.rdf4j.query.BindingSet;
@@ -31,12 +34,24 @@ import org.eclipse.rdf4j.query.algebra.evaluation.QueryBindingSet;
 public interface QueryEvaluationContext {
 
 	public class Minimal implements QueryEvaluationContext {
-		// The now time in the unix epoch or zero
-		private final long nowInMillis;
-		// The now time as a literal, lazy set if not yet known.
-		private Literal nowInLiteral;
+		// Doubly checked locking is simpler but with Project Loom coming
+		// synchronized on this is less interesting.
+		private static final VarHandle HANDLE_TO_THE_NOW_LITERAL;
+		static {
+			try {
+				HANDLE_TO_THE_NOW_LITERAL = MethodHandles
+						.privateLookupIn(QueryEvaluationContext.Minimal.class, MethodHandles.lookup())
+						.findVarHandle(QueryEvaluationContext.Minimal.class, "nowInLiteral", Literal.class);
 
+			} catch (NoSuchFieldException | IllegalAccessException e) {
+				throw new UnsupportedOperationException("The nowInLiteral field is missing");
+			}
+		}
+		// The now time as a literal, lazy set if not yet known.
+		// No need for volatile as we use the varhandle for this.
+		private Literal nowInLiteral;
 		private final Dataset dataset;
+		private final ValueFactory vf;
 
 		/**
 		 * Set the shared now value to an prexisting object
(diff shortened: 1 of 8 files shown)


hmottestad:
Why does it need to be thread safe?

JervenBolleman:
Currently it doesn't. I just thought you wanted it to be threadsafe because in the Performance branch you used a synchronized block around it. I think it is a nice idea because it would be hard to other wise guarantee this once we do introduce a parallel query evaluation strategy.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
```diff
@@ -33,26 +33,31 @@ public interface QueryEvaluationContext {
...
+			this.nowL = now;
+			// This is valid because the now field will never be read.
+			this.now = 0L;
 			this.dataset = dataset;
 		}
 
 		public Minimal(Dataset dataset) {
+			this.now = System.currentTimeMillis();
 			this.dataset = dataset;
 		}
```
jeenbroekstra, line 45: Normally I would suggest that the second constructor should just invoke the first one:

```java
public Minimal(Dataset dataset) {
    this(System.currentTimeMillis(), dataset);
}
```

But what's going on on line 38 confuses me. Why are you resetting the `now` value to 0 (and why are you doing it in the first constructor but not in the second one)? And if the `now` field is "never read", why even have one?

hmottestad:
Double checked locking might be simpler to understand and reason about. 

Or you can try out VarHandle. Allows you to do atomic compare and set. It's a bit overkill maybe, but super cool 🤤

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
JervenBolleman: @jeenbroekstra  I added comments about what is going on.

JervenBolleman:
@hmottestad if going for atomic's a simple AtomicReference will do, no need for varhandles.

hmottestad:
I'm playing a bit with VarHandles here: https://github.com/eclipse/rdf4j/blob/1bb3e1efb967a8ea5a1d96d4153b56c0a218fe46/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/SequentialRecordCache.java

Essentially it allows a write-once value to be declared without volatile, but still do atomic compare and set as well as volatile reads. 

In my case I'm lazily initializing  a variable. If the variable is not null, then we can safely use. If it is null then it could either be uninitialized or a stale read. So then I do a volatile read to be sure it's not stale, and then if it's still null I initialize the variable using compare and set.

I don't think I'll keep this code. It's a bit overkill and I don't really need to lazily initialize the variable. I might pick it up again somewhere else where locking or synchronization adds a lot of overhead. 

Can I propose using volatile and double checked locking for your code? System.currentTimeMillis() is platform dependent and there used to be some issues with poor performance in Linux. https://www.mo4tech.com/cant-believe-it-system-currenttimemillis-has-a-performance-problem.html

JervenBolleman:
On review I did now implement it with a varhandle. In retrospect I think this is the nicest and best solution with the lowest runtime impact.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
JervenBolleman: Going with the simpler varhandle approach.

hmottestad:
Could you make some tests to cover this code? There is a query optimizer that rewrites calls to now() with a constant value, so I don't actually think we have any tests that cover this code path. You could try to just return `null` to see if there are any tests that fail.

I'm also going to start pushing out the 4.0.0 release, but since this is a minor bug fix we can target it towards 4.0.1.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
```diff
@@ -31,28 +34,64 @@
...
 		@Override
 		public Literal getNow() {
 			// creating a new date is expensive because it uses the XMLGregorianCalendar implementation which is very
-			// complex
-			if (now == null) {
-				now = SimpleValueFactory.getInstance().createLiteral(new Date());
+			// complex.
+			if (nowInLiteral == null) {
+				nowInLiteral = (Literal) HANDLE_TO_THE_NOW_LITERAL.compareAndExchange(null,
+						vf.createLiteral(new Date()));
```
hmottestad, line 92: Screenshot 2022-04-22 at 09 46 17


Also the javadocs for `compareAndExchange` say that the expected value is returned, if I understand them correctly: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/invoke/VarHandle.html#compareAndExchange(java.lang.Object...)

JervenBolleman:
@hmottestad thank you for the review as the code was quite broken :( I now added a number of tests and the code is now exercised correctly on this branch. (It would have executed on GH-3798-more-functions-precompiled)

Issue #3806, opened 2022-04-18
Excessive file manipulation in SequentialRecordCache used in o.e.r.sail.nativerdf.TripleStore slows down bulk statement removal leading to transaction timeout exception


cadmiumkitty:
### Current Behavior

Executing `connection.clear(<context>)` (and `CLEAR SILENT GRAPH <graph name>`) outside the transaction (also inside the transaction, but I don't have the right profiler snapshot to show it) results in a significant amount of time spent in o.e.r.common.io.NioFile.size() and o.e.r.common.io.NioFile.read() reading the records back from the SequentialRecordCache, even though the size or content of the cache are not changing.

### Expected Behavior

Faster `connection.clear(<context>)` (and `CLEAR SILENT GRAPH <graph name>`). It should be possible to tell SequentialRecordCache that writing to the cache is complete, so the size of the file can be cached, and records pre-read in bulk (perhaps a configuration option for number of records read ahead)

### Steps To Reproduce

1. Run Rdf4j server 3.7.7

2. Create a remote repository with RemoteRepositoryManager using the configuratoin snippet below:

```
  rep:repositoryImpl [
      rep:repositoryType "openrdf:SailRepository" ;
      sr:sailImpl [
          sail:sailType "openrdf:NativeStore" ;
          sail:connectionTimeOut "600000" ;
          ns:tripleIndexes "spoc,opsc,ospc" ;
          ns:forceSync false ;
          ns:valueCacheSize "10000000" ;
          ns:valueIDCacheSize "10000000" ;
          ns:namespaceCacheSize "10000";
          ns:namespaceIDCacheSize "10000" 
        ]
    ].
```

3. Import about 5K statements (in my case very simple SKOS taxonomy with about 1K concepts).

4. Run the following code in the client:

```
RepositoryConnection connection = ...
connection.setIsolationLevel(IsolationLevels.SNAPSHOT);
connection.begin();
connection.clear(...);
... some additional logic here...
connection.commit();
```

5. Observe operation taking about 10-20 seconds (running on a pretty old XPS-15 with Intel(R) Core(TM) i7-4712HQ CPU @ 2.30GHz and HDD with 4G RAM allocated to Rdf4j server)

6. Observe the profiler results that show that a lot of time is spent in TripleStore.removeTriples() working with SequentialRecordCache first writing deleted triples, then reading them back one by one always checking the file size.

[...]

PR #3813, opened 2022-04-19
GH-3806 improve performance of clear() in NativeStore

hmottestad:
GitHub issue resolved: #3806 

----

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/SequentialRecordCache.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/SequentialRecordCache.java
@@ -9,6 +9,8 @@ package org.eclipse.rdf4j.sail.nativerdf;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
 import java.nio.ByteBuffer;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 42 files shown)


hmottestad:
Currently just updated benchmarks

hmottestad:
Screenshot 2022-04-19 at 15 47 43

hmottestad:
### After
```
Benchmark                                                    Mode  Cnt     Score   Error  Units
QueryWriteBenchmark.clear                                    avgt       1680.806          ms/op
QueryWriteBenchmark.clearByQuery                             avgt       1667.793          ms/op
QueryWriteBenchmark.clearByQueryReadCommitted                avgt       1674.576          ms/op
QueryWriteBenchmark.clearContext                             avgt       1597.303          ms/op
QueryWriteBenchmark.clearContextReadCommitted                avgt       1674.378          ms/op
QueryWriteBenchmark.clearReadCommitted                       avgt       1682.097          ms/op
QueryWriteBenchmark.removeByQuery                            avgt        233.723          ms/op
QueryWriteBenchmark.removeByQueryReadCommitted               avgt        408.298          ms/op
QueryWriteBenchmark.simpleUpdateQueryIsolationNone           avgt        291.329          ms/op
QueryWriteBenchmark.simpleUpdateQueryIsolationReadCommitted  avgt        445.185          ms/op
```

hmottestad:
Big question is if the memory based cache will affect integrity in case of something like a power outage. I don't think it will, but it's hard to be sure.

hmottestad:
With 1GB as the limit for the benchmark we are always staying in memory. I want to post the results without the in-memory cache, since the various fixes for aligning to 4KB and doing read ahead also make a big difference.

kenwenzel:
> Big question is if the memory based cache will affect integrity in case of something like a power outage. I don't think it will, but it's hard to be sure.

The problem might be that a partially applied transaction cannot be rolled back when the record cache (some kind of WAL) is missing.

hmottestad:
I can't see that it ever gets read back in again. 

The WAL seems to based on editing the B-tree and adding a flag to denote that a statement was removed or added. And then during commit the status file is updated to "committing" and each statement that is flagged for removal is removed and those flagged as added statements just have their flag cleared. 

For recovery the transaction status file is read and if it's in  the commit phase then it just resumes iterating over the statements and doing what the flag says. If it's not in the commit phase then it assumes that no action has been taken based on the flags and essentially just rests them. 

The recovery mechanism just calls the commit and rollback methods. The updatedTriplesCache is used to track which statements have been updated during the transaction, so that commit and rollback only need to iterate over affected statements instead of all the statements. When the recovery mechanism invokes the commit and rollback methods there isn't a updatedTriplesCache so it instead needs to iterate over all the statements in the indexes. 

It looks like the updatedTriplesCache is kept in a specific file, but I can't see how it is used during recovery and it makes me wonder if it too could be  kept in memory.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/SequentialRecordCache.java
```diff
@@ -55,36 +89,78 @@ public SequentialRecordCache(File cacheDir, int recordSize) throws IOException {
 	public SequentialRecordCache(File cacheDir, int recordSize, long maxRecords) throws IOException {
 		super(maxRecords);
 		this.recordSize = recordSize;
+		this.cacheDir = cacheDir;
+	}
 
-		File cacheFile = File.createTempFile("txncache", ".dat", cacheDir);
-		nioFile = new NioFile(cacheFile);
+	private void init() throws IOException {
```
hmottestad, line 95: Don't really need to use lazy initialization, but it was fun to play with VarHandle a bit. Going to revert back I think and remove the VarHandle code.

hmottestad:
```

private Map<Integer, Long> removeTriples(RecordIterator iter) throws IOException {

		byte[] data = iter.next();
		if (data == null) {
			// no triples to remove
			return Collections.emptyMap();
		}

		final HashMap<Integer, Long> perContextCounts = new HashMap<>();

		// Store the values that need to be removed in a tmp file and then
		// iterate over this file to set the REMOVED flag
		RecordCache removedTriplesCache = new InMemRecordCache();
		try {
			try (iter) {
				while (data != null) {
					if ((data[FLAG_IDX] & REMOVED_FLAG) == 0) {
						data[FLAG_IDX] |= REMOVED_FLAG;
						removedTriplesCache.storeRecord(data);
						int context = ByteArrayUtil.getInt(data, CONTEXT_IDX);
						perContextCounts.merge(context, 1L, Long::sum);
					}
					data = iter.next();

					if (shouldOverflowToDisk(removedTriplesCache)) {
						logger.debug("Overflowing RecordCache to disk due to low free mem.");
						assert removedTriplesCache instanceof InMemRecordCache;
						InMemRecordCache old = (InMemRecordCache) removedTriplesCache;
						removedTriplesCache = new SequentialRecordCache(dir, RECORD_LENGTH);
						removedTriplesCache.storeRecords(old);
						old.clear();
					}
				}
			}

			updatedTriplesCache.storeRecords(removedTriplesCache);

			// Set the REMOVED flag by overwriting the affected records
			for (TripleIndex index : indexes) {
				BTree btree = index.getBTree();

				try (RecordIterator recIter = removedTriplesCache.getRecords()) {
					while ((data = recIter.next()) != null) {
						btree.insert(data);
					}
				}
			}
		} finally {
			removedTriplesCache.discard();
		}

		return perContextCounts;
	}

```

hmottestad:
This is the code that handles marking a statement for removal. 

It's got three parts. 

The first part collects records (bytes) for each statement from the btree and stores them in a cache. 

Then it writes that cache into the transaction cache. (Which I'm unsure if really requires persistence). 

Then it loops through the first cache and writes to the btree. But it doesn't actually delete anything, just writes the record back with a modified flag to indicate that the statement should be deleted. 

Later on commit() is called and then the transaction cache is used to perform the actual deletes.

kenwenzel:
In its current implementation the transaction cache may also be incomplete. So it should not matter If you keep it in memory or not. Hence also with the current implementation it may happen that a transaction can not be completely rolled back if commit() fails and the cache is incomplete.
BTW, at this point the rollback is only required in case of an IO error as the base implementation of the SAIL already handles transactions.
Therefore it may even be possible to directly modify the b-tree without using the flags as intermediate step. But I'm unsure if this is even less safe than the current approach.

hmottestad:
Flags also allow to perform a rollback from isolation level NONE.

hmottestad:
TODO: 

- [ ] Rebase and cleanup commits

jeenbroekstra (approved):
Other than cleaning up the commits (which you already noted yourself) this all looks fine to me. Nice job!

Issue #3826, opened 2022-04-23
Update custom function tutorial to cover use of TripleSource


jeenbroekstra:
The current tutorial for custom functions (https://rdf4j.org/documentation/tutorials/custom-sparql-functions/ ) is a little out of date: it does not show the use of the `TripleSource` input argument to use the current store state.

PR #3827, opened 2022-04-24
GH-3826 update tutorial, add default impl for deprecated method

jeenbroekstra:
GitHub issue resolved: #3826  

- ~added default implementation of deprecated `evaluate` method (to allow implementations to avoid overriding it if they already implement the new variant with a `TripleSource` argument)~
- slightly updated the tutorial and example code

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/Function.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/Function.java
@@ -35,12 +35,11 @@ public interface Function {
 	 * @param args         the function input arguments.
 	 * @return the function result value.
 	 * @throws ValueExprEvaluationException
-	 * @deprecated since 3.3.0. Use {@link #evaluate(TripleSource, Value...)} instead.
+	 * @deprecated since 3.3.0. Use {@link #evaluate(TripleSource, Value...)} instead. A reference to a ValueFactory can
+	 *             be retrieved using {@link TripleSource#getValueFactory()} if needed.
 	 */
 	@Deprecated
-	default Value evaluate(ValueFactory valueFactory, Value... args) throws ValueExprEvaluationException {
-		throw new UnsupportedOperationException("use evaluate(TripleSource, Value...) instead");
-	}
+	Value evaluate(ValueFactory valueFactory, Value... args) throws ValueExprEvaluationException;
 
 	/**
 	 * Evaluate the function over the supplied input arguments.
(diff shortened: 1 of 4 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/Function.java
```diff
@@ -32,11 +32,12 @@ public interface Function {
...
 	 * @throws ValueExprEvaluationException
-	 * @deprecated since 3.3.0. Use {@link #evaluate(TripleSource, Value...)} instead. A reference to a ValueFactory can
-	 *             be retrieved using {@link TripleSource#getValueFactory()} if needed.
+	 * @deprecated since 3.3.0. Use {@link #evaluate(TripleSource, Value...)} instead.
 	 */
 	@Deprecated
-	Value evaluate(ValueFactory valueFactory, Value... args) throws ValueExprEvaluationException;
+	default Value evaluate(ValueFactory valueFactory, Value... args) throws ValueExprEvaluationException {
+		throw new UnsupportedOperationException("use evaluate(TripleSource, Value...) instead");
+	}
```
hmottestad, line 40: Should this maybe be targeted at 4.1.0?

jeenbroekstra: I'm a bit on the fence on that. I originally only wanted to change the docs and example, but I discovered this while doing so. What this change allows is a cleaner implementation where you don't _have_ to implement the deprecated method anymore (but still can if you want, so it's backward compatible). 

I think it's defensible to do this as a bug fix and put it in 4.0.1. Especially since we aim for a quick 4.0.1 release anyway, hopefully. The only thing I don't know for sure is if this is binary compatible.

This reminds me we should reenable the API compatibility checks for the main branch, now that 4.0 is out the door.

jeenbroekstra: Heh. Just ran a check on this, and it's not binary compatible. Which according to the japicmp plugin would make this a breaking change, needing a major version.

jeenbroekstra:
Grmbl. Shot myself in the foot didn't I, re-enabling api compatibility checks before sneaking this in :grin:

hmottestad:
@jeenbroekstra would you like to split out the documentation changes and merge those?

jeenbroekstra:
> @jeenbroekstra would you like to split out the documentation changes and merge those?

Yeah good idea. Just really struggling to find any time to work on this at the moment... Hopefully I can sit down at it over the next weekend.

Issue #3805, opened 2022-04-18
NullPointerException in LMDB TxnRef when using RepositoryConnection and iterating through TupleQueryResult in a transaction


cadmiumkitty:
### Current Behavior

Remote client using RepositoryConnection iterating over a TupleQueryResult in a transaction against LmdbStore repository fails with the transaction being rolled back due to NullPointerException exception in TxnRef (and server subsequently closing the connection).

### Expected Behavior

Client code using transactions should be executed without exceptions when using LmdbStore on the server.

### Steps To Reproduce

1. Run Rdf4j server 4.0.0-M3

2. Create a repository using RemoteRepositoryManager with the following snippet (I have not tested without the LuceneSail, but suspect the behaviour is the same):

```
<https://example.com/repository> a rep:Repository ;
  rdfs:label "Some label" ;
  rep:repositoryImpl [
      rep:repositoryType "openrdf:SailRepository" ;
      sr:sailImpl [
          sail:delegate [
              sail:sailType "rdf4j:LmdbStore" 
          ];
          sail:sailType "openrdf:LuceneSail" ;
          sl:indexDir "index/"
      ]
  ].

```

3. Execute the following code on the client:

```
TupleQuery select = ...

connection.setIsolationLevel(IsolationLevels.READ_UNCOMMITTED);
connection.begin();

try (TupleQueryResult result = select.evaluate()) {
    if (result.hasNext()) {
        ... some other logic ....
        connection.commit();
    }
}
```

4. Observe the client failing with the transaction being rolled back due to the connection being closed by the server.

5. Observe the following exception on the server (probably because there are several threads involved in processing requests and TxnRef is maintaining a HashMap of states using Thread.currentThread() as key):

```
[INFO ] 2022-04-17 14:59:45,406 [repositories/xxx] POST transaction start
[INFO ] 2022-04-17 14:59:45,408 [repositories/xxx] transaction started
[INFO ] 2022-04-17 14:59:45,418 [repositories/xxx] PUT txn query request
[INFO ] 2022-04-17 14:59:45,431 [repositories/xxx] PUT txn query request finished
[ERROR] 2022-04-17 14:59:45,434 [repositories/xxx] Query evaluation error
org.eclipse.rdf4j.query.QueryEvaluationException: java.lang.NullPointerException
	at org.eclipse.rdf4j.sail.base.SailDatasetTripleSource$Eval.convert(SailDatasetTripleSource.java:67)
[...]

PR #3839, opened 2022-04-27
GH-3805 Fix transaction handling in LMDB store.

kenwenzel:
GitHub issue resolved: #3805 

This change allows to freely pass LMDB store iterations between threads.

----


hmottestad:
Going to take a little bit of time to review this. You are introducing a new locking system that I need to try to understand.

hmottestad:
@kenwenzel Have you considered using one of the existing lock managers? There are fast implementations of read-preferred and write-preferred locks, as well as a stamped lock. The benefit of these is that you can enable various levels of tracking, logging and automatic unlocking if the lock gets garbage collected.

kenwenzel:
@hmottestad The lock is a StampedLock and it existed already in the last version. Let me know if the other locks are faster. In this case we could simply switch to another implementation.

hmottestad:
https://github.com/eclipse/rdf4j/tree/main/core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks

hmottestad:
The ReadPrefReadWriteLockManager is faster than a plain stamped lock, but there is a small overhead because of the lock object. It uses a combination of a StampedLock and two LongAdders to achieve faster read locks than just using the StampedLock. I also intend to switch it over to using a boolean and VarHandle instead.

The biggest reason to use the lock managers is that they make debugging locking bugs easier because they can be configured to detect abandoned locks and also possible deadlocks. 

We use these for transaction management and iterators that are user facing. So that if a user forgets to close the iterator for a query result then the lock is automatically released when it becomes phantom reachable.

kenwenzel:
The current implementation uses only one lock instance for the whole triple store. The lock itself is acquired while fetching data within the iterators next() method and is released immediately afterwards. Therefore it cannot happen that the user forgets to free the lock. I am unsure if it is really helpful to use ReadPrefReadWriteLockManager in this case as the locks are very short-lived and also guaranteed to be freed.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -0,0 +1,253 @@
...
+import org.lwjgl.PointerBuffer;
+import org.lwjgl.system.MemoryStack;
+
+/**
+ * Manager for LMDB transactions.
+ */
+class TxnManager {
+
+	private final Mode mode;
+	private final Set<Txn> active = new HashSet<>();
```
hmottestad, line 37: Looks like you are using a HashSet for objects that don't implement hashCode() or equals().

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -0,0 +1,253 @@
...
+		RESET,
+		ABORT,
+		NONE
+	}
+
+	class Txn implements Closeable, AutoCloseable {
+
+		long txn;
+		List<Long> staleTxns;
+		private volatile long version;
```
hmottestad, line 164: Doesn't need to be volatile since it is only written to from code holding a write lock and read from code holding a read/write lock.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -0,0 +1,253 @@
...
+	enum Mode {
+		RESET,
+		ABORT,
+		NONE
+	}
+
+	class Txn implements Closeable, AutoCloseable {
+
+		long txn;
+		List<Long> staleTxns;
```
hmottestad, line 163: Can these be private?

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -0,0 +1,253 @@
...
+/**
+ * Manager for LMDB transactions.
+ */
+class TxnManager {
+
+	private final Mode mode;
+	private final Set<Txn> active = new HashSet<>();
+	private final long[] pool;
+	private final StampedLock lock = new StampedLock();
+	private long env;
```
hmottestad, line 40: Can this be final?

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TripleStore.java
```diff
@@ -779,13 +785,13 @@ void endTransaction(boolean commit) throws IOException {
 						E(mdb_txn_commit(writeTxn));
 					} finally {
 						recordCache = null;
-						readTxnRef.activate();
-						readTxnRef.lock().unlockWrite(stamp);
+						txnManager.activate();
+						txnManager.lock().unlockWrite(stamp);
```
hmottestad, line 789: txnManager.activate() could potentially throw an exception so that the write lock is never unlocked.

hmottestad:
In general you are referencing the stamped lock directly and not storing it in a variabel, except for here: [/LmdbRecordIterator.java#L117](https://github.com/eclipse/rdf4j/blob/7a14aed5bd751a851bc18bd9c239f7b3d50ca50e/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java#L117)

hmottestad:
And here [LmdbRecordIterator.java#L100](https://github.com/eclipse/rdf4j/blob/7a14aed5bd751a851bc18bd9c239f7b3d50ca50e/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java#L100) there is some code between the read lock and the try/finally section which could potentially throw and exception so that the read lock doesn't get released.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: Yes, I'm relying on the default implementations. The Txn objects are only equal if they are identical. 
I think that this is a common practice. Would you propose something else?

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: `version` is written in `setActive()` and read in `version()` that themselves are not synchronized. Doesn't this imply the use of volatile here?

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: Yes, for sure ;-)

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TripleStore.java
kenwenzel: I'll wrap it in a try {...} finally {...} block.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
hmottestad: IdentityHashMap is typically faster and that also signals to other developers that you are intentionally relying on reference equality.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
hmottestad: You only call those methods while inside a read or write-locked section. The StampedLock guarantees happens-before relationships.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: OK, I'll use an IdentityHashMap<Txn, Boolean>

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: Yes, you are right. I just thought that there is also the possibility to call `version()` outside of a locked section.
Do you think it is faster if `volatile` is removed?

hmottestad: Might be ever so slightly faster. One issue with using volatile here is that you at some point call `version++` which is a non-atomic operation on a volatile variable. So you are already assuming that there needs to be some kind of external locking to prevent race conditions.

kenwenzel: I've removed the volatile modifier.

kenwenzel:
@hmottestad @jeenbroekstra Should I rebase this against main?

jeenbroekstra:
> @hmottestad @jeenbroekstra Should I rebase this against main?

Yeah for a bug fix that's a good idea.

hmottestad:
I've cherry picked your commit into this PR here to merge into main: https://github.com/eclipse/rdf4j/pull/3883

hmottestad:
Merged with https://github.com/eclipse/rdf4j/pull/3883

Issue #3847, opened 2022-04-29
Quicker "quick" maven profile


hmottestad:
Would be nice to also skip the formatting and japicmp, so it essentially becomes a quick way to just build the project.

PR #3848, opened 2022-04-29
GH-3847 skip japicmp and formatters when using the quick profile in maven

hmottestad:
GitHub issue resolved: #3847 

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -204,6 +204,10 @@
 			<properties>
 				<skipTests>true</skipTests>
 				<skipITs>true</skipITs>
+				<formatter.skip>true</formatter.skip>
+				<impsort.skip>true</impsort.skip>
+				<xml-format.skip>true</xml-format.skip>
+				<japicmp.skip>true</japicmp.skip>
 			</properties>
 		</profile>
 		<profile>


pom.xml
```diff
@@ -204,6 +204,10 @@
 			<properties>
 				<skipTests>true</skipTests>
 				<skipITs>true</skipITs>
+				<formatter.skip>true</formatter.skip>
+				<impsort.skip>true</impsort.skip>
+				<xml-format.skip>true</xml-format.skip>
+				<japicmp.skip>true</japicmp.skip>
 			</properties>
 		</profile>
```
hmottestad, line 212: @jeenbroekstra What is your opinion on skipping the formatters and japicmp in the quick profile? 

I want to speed up the build a bit on my local machine when I need to use `mvn clean install` for sharing my local RDF4J changes with another maven project. I usually also try to remember to use the parallel build (-T 1C), which makes things even faster. Some of the formatters don't support parallel builds (dunno if it's just a lack of explicit support or if something is actually broken, I just got a bunch of warnings). 

For doing a full build there is still `mvn clean install -DskipTests` that does all the formatting and also stops with an error if the api changes are breaking. Since `-DskipTests` does pretty much what the `quick` profile does today I was thinking that maybe there is room to differentiatet it by also skipping formatting and japicmp.

jeenbroekstra: Makes total sense to me, it's more or less the whole point of that profile, I just never got around to updating it. Good call!

jeenbroekstra (approved):

Issue #3860, opened 2022-05-04
High churn cause MemoryStore to run out of memory


hmottestad:
### Current Behavior

The cleanup thread releases its write lock very often, so with high churn we run out of memory before the cleanup thread is able to complete.

### Expected Behavior

Cleanup thread should detect low memory and stop releasing the write lock until it has finished cleaning up.

### Version

4.0.0

### Are you interested in contributing a solution yourself?

Yes

PR #3861, opened 2022-05-04
GH-3860 handle high churn of memory store

hmottestad:
GitHub issue resolved: #3860 

 - detect when there is less than 64MB of free RAM (with Xmx>=256MG) and stop releasing the write lock while running the snapshot cleanup in the MemoryStore
 - the snapshot cleanup process already tracks what it has already cleaned to improve performance, but the HashSets used for tracking are now clearer if we have little memory

----


hmottestad:
I didn't make a test for this, since it's tricky to test. Instead I have a benchmark that fails without this fix: https://github.com/eclipse/rdf4j/blob/develop/core/sail/memory/src/test/java/org/eclipse/rdf4j/sail/memory/benchmark/ParallelMixedReadWriteBenchmark.java

jeenbroekstra (approved):

PR #3863, opened 2022-05-05
wip

hmottestad:
GitHub issue resolved: # 

----


hmottestad:
## Before
```
```

## After
```
Benchmark                                                     Mode  Cnt      Score     Error  Units
QueryBenchmark.complexQuery                                   avgt    5      3.154 ±   0.083  ms/op
QueryBenchmark.different_datasets_with_similar_distributions  avgt    5      1.647 ±   0.082  ms/op
QueryBenchmark.groupByQuery                                   avgt    5      1.108 ±   0.037  ms/op
QueryBenchmark.long_chain                                     avgt    5    537.601 ±  10.472  ms/op
QueryBenchmark.lots_of_optional                               avgt    5    134.055 ±   1.422  ms/op
QueryBenchmark.minus                                          avgt    5  13941.495 ± 307.702  ms/op
QueryBenchmark.nested_optionals                               avgt    5    167.978 ±   2.089  ms/op
QueryBenchmark.pathExpressionQuery1                           avgt    5     12.738 ±   0.672  ms/op
QueryBenchmark.pathExpressionQuery2                           avgt    5      1.198 ±   0.005  ms/op
QueryBenchmark.query_distinct_predicates                      avgt    5     76.526 ±   0.508  ms/op
QueryBenchmark.simple_filter_not                              avgt    5      6.469 ±   0.088  ms/op
```

JervenBolleman:
@hmottestad I read the code and the changes make sense to me.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/GenericSimpleStatement.java
```diff
@@ -0,0 +1,110 @@
...
+import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.base.AbstractStatement;
+
+/**
+ * A simple default implementation of the {@link Statement} interface for statements that don't have an associated
+ * context. For statements that do have an associated context, {@link ContextStatement} can be used.
+ *
+ * @see SimpleValueFactory
+ */
+public class GenericSimpleStatement<R extends Resource, I extends IRI, V extends Value> extends AbstractStatement {
```
jeenbroekstra, line 25: Why do we need a new class `GenericSimpleStatement` and `GenericContextStatement`? What is their relation to the already-existing `SimpleStatement` and `ContextStatement`?

hmottestad: The use java generics, which gives a decent performance boost by allowing the JVM to skip type checking and casting.

hmottestad: They are otherwise copies of SimpleStatement and ContextStatement.

hmottestad: I could make SimpleStatement and ContextStatement extend these two new classes, that would mean that we are pretty much moving all the code from those classes to the two new generic variants.

jeenbroekstra: What I don't fully understand is why these classes are in the model module, if the performance improvements are specifically targeted at the memory store. I'm sorry, I'm a bit short on time to look more closely, so don't take this as dismissing the idea, I'm just trying to get my head around it, with limited time.

hmottestad: It's because `MemStatement` extends `ContextStatement`. And alternative solution would be to have `MemStatement` implement `Statement` directly. The reason the `MemStatement` needs to be able to specify what classes are used for s, p, o and c is because the nature of the MemoryStore allows us to use "==" to check if two MemIRI (and other Mem****) are the same without ever needing to use `.equals(...)`. Performance wise there is a decent benefit in being able to specify that MemStatement uses MemResource, MemIRI and MemValue. 

Adding GenericSimpleStatement and GenericContextStatement to the model module makes sense because this is where SimpleStatement and ContextStatement reside. It might make even more sense if I made them extend each other and deprecated the old ones.

e.g.

```
@Deprecated
public class SimpleStatement extends GenericSimpleStatement<Resource, IRI, Value> {
...
}

@Deprecated
public class ContextStatement extends GenericContextStatement<Resource, IRI, Value, Resource> {
...
}
```

We could also just skip having both a GenericSimpleStatement and a GenericContextStatement. Instead we could just have GenericStatement which implements the full Statement interface. 

I am hoping that the generic versions can be used in other sails too. And I am also hoping to introduce more generics in general, particularly at the Sail level since that code isn't user facing.

jeenbroekstra: Fwiw I never really liked the separation of SimpleStatement and ContextStatement in the first place - IMO it's a leftover from when "contexts" / "quads" / "named graphs" were still an experimental thing.

hmottestad: I'll go with GenericStatement then.

hmottestad: And I can mark it internal use only, for the time being.

Issue #3867, opened 2022-05-06
Improve performance for generic code for flushing changes


hmottestad:
In AbstractSailConnection the method flushPendingUpdates is synchronized, causing a lot of contention. There may also be worth checking if there is any unnecessary flushing.

hmottestad:
<img width="1349" alt="Screenshot 2022-05-06 at 09 42 49" src="https://user-images.githubusercontent.com/797185/167088998-b33c61af-b3ca-46a1-99a2-cbe9488534ac.png">

hmottestad:
## After
![Screenshot 2022-05-06 at 11 02 27](https://user-images.githubusercontent.com/797185/167102392-cbd86680-9e43-4b6a-b85e-a7c1c135708c.jpg)

PR #3868, opened 2022-05-06
GH-3867 improve generic flushing performance

hmottestad:
GitHub issue resolved: #3867 

 - improve change tracking to reduce unneeded flushing
 - reduce use of synchronization

----

### DIFF
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractNotifyingSailConnection.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractNotifyingSailConnection.java
@@ -33,37 +33,28 @@ public abstract class AbstractNotifyingSailConnection extends AbstractSailConnec
 
 	@Override
 	public void addConnectionListener(SailConnectionListener listener) {
-		synchronized (listeners) {
-			listeners.add(listener);
-		}
+		listeners.add(listener);
 	}
 
 	@Override
 	public void removeConnectionListener(SailConnectionListener listener) {
-		synchronized (listeners) {
-			listeners.remove(listener);
-		}
+		listeners.remove(listener);
 	}
 
 	protected boolean hasConnectionListeners() {
-		synchronized (listeners) {
-			return !listeners.isEmpty();
-		}
+		return !listeners.isEmpty();
 	}
 
 	protected void notifyStatementAdded(Statement st) {
-		synchronized (listeners) {
-			for (SailConnectionListener listener : listeners) {
-				listener.statementAdded(st);
-			}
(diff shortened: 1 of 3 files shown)


hmottestad:
# Results

When running `ComplexLargeBenchmark#noPreloadingNonEmptyParallel()` we see that `flushPendingChanges()` is a fairly large portion of evaluating a query (TupleExpr). After applying the improvements from this branch the time spent in `flushPendingChanges()` becomes so small that it only registers as a small sliver in the flame graph and I had to add a zoomed in view to show that it's stil there.

## Before
167088998-b33c61af-b3ca-46a1-99a2-cbe9488534ac

## After
Screenshot 2022-05-06 at 11 02 27

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractNotifyingSailConnection.java
```diff
@@ -33,37 +33,28 @@ public AbstractNotifyingSailConnection(AbstractSail sailBase) {
 
 	@Override
 	public void addConnectionListener(SailConnectionListener listener) {
-		synchronized (listeners) {
-			listeners.add(listener);
-		}
+		listeners.add(listener);
```
hmottestad, line 36: Note that we are already using a thread-safe collection: `private final List<SailConnectionListener> listeners = new CopyOnWriteArrayList<>();`

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -712,17 +714,32 @@ public boolean addInferredStatement(Resource subj, IRI pred, Value obj, Resource
...
+					if (!hasStatement(dataset, subj, pred, obj, contextsToCheck)) {
+						notifyStatementAdded(vf.createStatement(subj, pred, obj, ctx));
+						sink.approve(subj, pred, obj, ctx);
+					}
+				} else {
+					sink.approve(subj, pred, obj, ctx);
 				}
-				sink.approve(subj, pred, obj, ctx);
 			}
 		}
```
hmottestad, line 744: `hasStatement(s,p,o,c)` is a vararg, which means that the JVM needs to create an array for time it is called with one or more arguments. The typical case for us though is that `contexts` is either an empty array or an array with a single element.

hmottestad: Of all the tests in the MemoryStore, there is only one test that calls down to this code with more than one context.

Issue #3871, opened 2022-05-10
SHACL Validation fails with IndexOutOfBoundsException


maxstolze:
### Current Behavior

Our integration tests use SHACL validation. We are migrating to RDF4J-4.0 and we are seeing problems with the validation: the `ValidationQuery.getValueVariable()` method throws an IndexOutOfBoundsException. As the logic of the class is rather contrived and our SHACL constraints are not trivial, we don't currently know which shape triggers the problem (but it's a MinCountConstraint, we know that much).

However, setting the `valueIndex` in `[ValidationQuery:66](https://github.com/eclipse/rdf4j/blob/16d19d987a0adbe392580635cb1abc2505697226/core/sail/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ast/ValidationQuery.java#L66)` seems to be the culprit, it sets the integer counting the variables to `variables.size()`, not `variables.size() - 1` as is used in the other related cases. Stepping through the code until the Exception is thrown reveals that the Exception happens the first time the `valueIndex` is initialized in line 66 (and not in line 63 or 70) and subsequently the `getValueVariable()` method is called.

### Expected Behavior

The SHACL validation should not produce an IndexOutOfBoundsException

### Steps To Reproduce

With Java 14 on Windows 10, try to evaluate these shapes:
```
@prefix mms:   <https://mms.researchstudio.at/mms#> .
@prefix sh:    <http://www.w3.org/ns/shacl#> .
@prefix dtype: <http://www.linkedmodel.org/schema/dtype#> .
mms:EnumerationTypeShape
        a              sh:NodeShape ;
        sh:targetClass mms:EnumerationType ;
        sh:property    [ sh:path     dtype:hasMember ;
                         sh:minCount 1 ;
                         sh:node     mms:EnumerationMemberShape ] .

    mms:EnumerationMemberShape
        a           sh:NodeShape ;
        sh:property [ sh:path     [ sh:inversePath dtype:hasMember ] ;
                      sh:minCount 1 ;
                      sh:maxCount 1 ;
                      sh:class    mms:EnumerationType ] .
```

### Version

4.0

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

We'll make a PR that includes a test case that reveals the bug and then we'll try to fix it. Just changing the initialization of `valueIndex` does not suffice (but you get a different exception)

PR #3872, opened 2022-05-10
GH-3871: Add revealing test case and fix one-off error

maxstolze:
GitHub issue resolved: #3871

* adds a test case in `core/sail/shacl/src/test/resources/test-cases/complex/mms/` that reveals the bug
* fixes the one-off error in `ValidationQuery`, the test still fails with another error - hence draft PR. 
* Help is welcome!
----


maxstolze:
After solving the `ArrayIndexOutOfBoundsException` issue, the code currently fails in `ValidationQuery.union()`, where the two ValidationQueries are checked in a number of assert statements. To see which queries are joined, we print them to stdout, here's the output:
```
union of a: 
?target_0000000000 a <https://mms.researchstudio.at/mms#EnumerationType> .
?target_0000000000 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .

FILTER(NOT EXISTS{?value <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .})
union of b: 
?target_0000000000 a <https://mms.researchstudio.at/mms#EnumerationType> .
?target_0000000000 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
?value_0 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
?value_1 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
FILTER(?value_0 != ?value_1)

union of a: 
{
?target_0000000000 a <https://mms.researchstudio.at/mms#EnumerationType> .
?target_0000000000 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .

FILTER(NOT EXISTS{?value <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .})
} UNION {
?target_0000000000 a <https://mms.researchstudio.at/mms#EnumerationType> .
?target_0000000000 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
?value_0 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
?value_1 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
FILTER(?value_0 != ?value_1)

}
union of b: 
?target_0000000000 a <https://mms.researchstudio.at/mms#EnumerationType> .
?target_0000000000 <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
?value <http://www.linkedmodel.org/schema/dtype#hasMember> ?target_0000000001 .
 FILTER(!(EXISTS{?value a <https://mms.researchstudio.at/mms#EnumerationType>}))
2022-05-10 17:24:47,556 [ForkJoinPool-1-worker-5] DEBUG org.eclipse.rdf4j.sail.shacl.ShaclSailConnection - Cleanup

java.lang.AssertionError
	at org.eclipse.rdf4j.sail.shacl.ast.ValidationQuery.union(ValidationQuery.java:84)
	at java.base/java.util.stream.ReduceOps$2ReducingSink.accept(ReduceOps.java:123)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
  [...]
```

The problem seems caused by one of the value variables being called `value` whereas the other is `value_1`

@hmottestad I think we need your help here

maxstolze:
Can't find the cause by our own and will wait now for @hmottestad or someone else who knows, why this happens here.

hmottestad:
I've found the issue and I've tested a fix. Thanks for creating a test case. 

I would like to use your test case in my code, so if you could take a look at the contributor guide and make sure you have signed the Eclipse Contributor Agreement. That way I can merge your branch in with my fix. 

https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md#legal-stuff

In short the issue is that some property shape constraints produce values and others don't. Min/max count don't, but class does. So it's correct that the index for the value is out of bounds. The problem here is that the use of sh:node stops the ShaclSail from splitting the shape into multiple shapes and then runs into the bug where the union code doesn't correctly check if the constraint produces a value or not.

hmottestad:
And as a temporary fix you can disable SPARQL based validation, and revert to using the regular bulk or transactional validation for everything. This will make some validations slower. 

You can set the property `org.eclipse.rdf4j.sail.shacl.sparqlValidation` to the `false`. Both as strings. See https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/System.html#setProperty(java.lang.String,java.lang.String)

hmottestad:
I've posted a PR here: https://github.com/eclipse/rdf4j/pull/3877

maxstolze:
Okay, thank you! I've signed the Eclipse Contributor Agreement and you can merge the test case as well.

hmottestad:
Merged into [GH-3871-shacl-index-out-of-bounds](https://github.com/eclipse/rdf4j/tree/GH-3871-shacl-index-out-of-bounds)

Issue #1090, opened 2018-10-02
FileFormat constructor allows empty set of mimetypes


HansBrende:
In the `FileFormat` class, I noticed that, although the constructor uses an `assert !mimeTypes.isEmpty() : "mimeTypes must not be empty";`, no exception is thrown in the case of an empty list of mimetypes, unless the developer has enabled assertions.

This could cause problems with the following methods:

```java
getDefaultMIMEType()
```
and
```java
hasDefaultMimeType(String)
```

will both throw an `IndexOutOfBoundsException` if an empty list of mimetypes had been supplied to the constructor.

Would not "fail-fast" behavior be more appropriate here? I could see how using `assert` might be more appropriate if all instances of the `FileFormat` class were created by the rdf4j library itself (e.g., for a package-private constructor), but given that it's a public constructor in a public class, it seems to me that using `assert` is not necessarily sufficient for argument validation.

**Note:** the 2 other `assert` statements requiring the collections to be non-null do not pose the same problem, because the null-check is handled down the line by `List.addAll(collectionThatIsNotNull)`.

HansBrende:
One possible alternative to the "fail-fast" approach would be to allow an empty set of mimetypes, but then return `null` for `getDefaultMIMEType()`. OR return "application/octet-stream". OR return `charset == null ? "application/octet-stream" : "text/plain"`. Seems like the best idea would be to simply fail on bad input though.

barthanssens:
Hmz, what about throwing an IllegalArgumentException ?

PR #3875, opened 2022-05-11
GH-1090: added IllegalArgumentException when mimeTypes isEmpty in FileFormat.java

Sanchit-Trivedi:
GitHub issue resolved: #1090 

Instead of silently failing with the assert check previously used, an empty mimeType list now throws an IllegalArgumentException 

`assert !mimeTypes.isEmpty() : "mimeTypes must not be empty"`

----

### DIFF
--- a/core/common/io/src/main/java/org/eclipse/rdf4j/common/lang/FileFormat.java
+++ b/core/common/io/src/main/java/org/eclipse/rdf4j/common/lang/FileFormat.java
@@ -97,8 +97,11 @@ public class FileFormat {
 	public FileFormat(String name, Collection<String> mimeTypes, Charset charset, Collection<String> fileExtensions) {
 		assert name != null : "name must not be null";
 		assert mimeTypes != null : "mimeTypes must not be null";
-		assert !mimeTypes.isEmpty() : "mimeTypes must not be empty";
 		assert fileExtensions != null : "fileExtensions must not be null";
+
+		if (mimeTypes.isEmpty()) {
+			throw new IllegalArgumentException("mimeTypes must not be empty");
+		}
 
 		this.name = name;
 		this.mimeTypes.addAll(mimeTypes);
--- a/core/common/io/src/test/java/org/eclipse/rdf4j/common/lang/FileFormatTest.java
+++ b/core/common/io/src/test/java/org/eclipse/rdf4j/common/lang/FileFormatTest.java
@@ -1,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) $2022 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.common.lang;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class FileFormatTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void testEmptyMimeType() {
+        thrown.expect(IllegalArgumentException.class);
+        new FileFormat("PLAIN TEXT", new ArrayList<String>(), null, Arrays.asList("txt"));
+    }
+}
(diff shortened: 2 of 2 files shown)


core/common/io/src/test/java/org/eclipse/rdf4j/common/lang/FileFormatTest.java
```diff
@@ -0,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) $2022 Eclipse RDF4J contributors.
```
jeenbroekstra, line 2: There's a `$` in front of the year that shouldn't be there.

core/common/io/src/test/java/org/eclipse/rdf4j/common/lang/FileFormatTest.java
```diff
@@ -0,0 +1,27 @@
...
+public class FileFormatTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void testEmptyMimeType() {
+        thrown.expect(IllegalArgumentException.class);
+        new FileFormat("PLAIN TEXT", new ArrayList<String>(), null, Arrays.asList("txt"));
+    }
```
jeenbroekstra, line 26: We've switched over to JUnit 5 a while back and are slowly migrating our existing tests over to it. Do you want to rewrite this as a JUnit 5 test?

core/common/io/src/main/java/org/eclipse/rdf4j/common/lang/FileFormat.java
```diff
@@ -97,9 +97,12 @@ public FileFormat(String name, String mimeType, Charset charset, Collection<Stri
 	public FileFormat(String name, Collection<String> mimeTypes, Charset charset, Collection<String> fileExtensions) {
 		assert name != null : "name must not be null";
 		assert mimeTypes != null : "mimeTypes must not be null";
-		assert !mimeTypes.isEmpty() : "mimeTypes must not be empty";
 		assert fileExtensions != null : "fileExtensions must not be null";
 
+		if (mimeTypes.isEmpty()) {
```
jeenbroekstra, line 102: I know the bug report was only about the mime types, but perhaps we should do this for the other arguments as well - what do you think? For the single object inputs we could just use a `Objects.requireNonNull` and for the other collection parameter (file extensions) we could do a similar check to what's you've added here.

jeenbroekstra (commented):
Thanks for working on this @Sanchit-Trivedi !

core/common/io/src/test/java/org/eclipse/rdf4j/common/lang/FileFormatTest.java
Sanchit-Trivedi: Sure I can do that. Does [this](https://pastebin.com/DxmTfit8) look good to you?

core/common/io/src/main/java/org/eclipse/rdf4j/common/lang/FileFormat.java
Sanchit-Trivedi: Quoting from #1090 HansBrende's original issue 

> Note: the 2 other assert statements requiring the collections to be non-null do not pose the same problem, because the null-check is handled down the line by List.addAll(collectionThatIsNotNull).

Therefore I did not go ahead with adding checks for the same. 

However, I do think we can add `Objects.requireNonNull` for the single inputs instead of differing it to  `List.addAll()` and a similar check for the file extensions parameter. Let me know if you would like me to go ahead and implement these.

jeenbroekstra: Ah good catch. Actually let's keep things simple for now, and leave as-is. Thanks :)

jeenbroekstra:
Also looks as if the Java formatting checks don't like something in your PR. You may want to run a `mvn process-resources` on your branch before committing/pushing.

Sanchit-Trivedi:
Fixed the formatting issues and added the upgraded tests @jeenbroekstra.

jeenbroekstra (approved):
Nice work :)

jeenbroekstra:
@Sanchit-Trivedi I just noticed this is targeted against the `develop` branch. As this is a bug fix: could you rebase it against the main branch instead?

Sanchit-Trivedi:
@jeenbroekstra 

I did run the mvn process-resources and got a [Successful](https://user-images.githubusercontent.com/46599622/168413919-bd1a84c3-9d18-4e09-a472-7de31b6bd7d2.png) message after running the command. Still, the Github actions fail.

Is there something else I need to do before running this command?

Also will this require rebasing ? I created the fork from the main branch only but just targeted the PR to the develop branch

jeenbroekstra:
At a guess: did you format and then make some other changes? It seems to complaining about your test code this time.

And no need to rebase if you started off main, we can just retarget this PR.

hmottestad:

Issue #3681, opened 2022-02-15
RDF lists created using RDFCollections.asRDF() are not properly serialized using TurtleWriter


jetztgradnet:
### Current Behavior

When serializing an RDF list created using `RDFCollections.asRDF()` with `TurtleWriter` the first segment of the list is not properly inlined as it is not recognized as a well-formed list.

The result looks as this:
```
@prefix ex: <http://example.com/ns#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .

ex:Cities ex:list [ a rdf:List;
      rdf:first ex:NewYork;
      rdf:rest (ex:Rio ex:Tokyo)
    ] .
```
Please notice the first part with explicit `rdf:first` and `rdf:rest` statements instead of the inlining of the list, which is done for the last two segments of the list.

This is caused by `org.eclipse.rdf4j.rio.turtle.TurtleWriter.isWellFormedCollection()` not recognizing the `rdf:type rdf:List` type statement for the first part of the list segment which is generated in `RDFCollections.asRDF()` (or more specifically in `org.eclipse.rdf4j.model.util.RDFCollections.consumeCollection(Iterable<?>, Resource, Consumer<Statement>, ValueFactory, Resource...)`).
A simple solution would be to adjust `org.eclipse.rdf4j.rio.turtle.TurtleWriter.isWellFormedCollection()` to recognize and accept that statement when checking for additional statements within the list.

### Expected Behavior

The proper visualization of the list would instead look like this:
```
@prefix ex: <http://example.com/ns#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .

ex:Cities ex:list (ex:NewYork ex:Rio ex:Tokyo) .
```

### Steps To Reproduce

This test case reproduces the behavior:

```
package com.metaphacts.services.ontologies;

import static org.junit.Assert.assertEquals;

import java.io.StringWriter;
import java.util.List;

import org.eclipse.rdf4j.model.BNode;
import org.eclipse.rdf4j.model.IRI;
import org.eclipse.rdf4j.model.Model;
import org.eclipse.rdf4j.model.impl.TreeModel;
import org.eclipse.rdf4j.model.util.RDFCollections;
import org.eclipse.rdf4j.model.util.Values;
import org.eclipse.rdf4j.model.vocabulary.RDF;
import org.eclipse.rdf4j.rio.RDFFormat;
[...]

PR #3878, opened 2022-05-13
GH-3681 RDF collections formatting issue in Turtle

The-Nightwing:
GitHub issue resolved: #3681

Description
The Turtle parser (Rio.write) does not add "a rdf:List" statements while parsing, so I also didn't add rdf:List while writing it out using Statements object, that's why we commented out the line which was trying to write out rdf:List to our data.
```
public static void consumeCollection(Iterable<?> values, Resource head, Consumer<Statement> consumer,
			ValueFactory vf,
			Resource... contexts) {
		Objects.requireNonNull(values, "input collection may not be null");
		Objects.requireNonNull(consumer, "consumer may not be null");
		Objects.requireNonNull(vf, "injected value factory may not be null");

		Resource current = head != null ? head : vf.createBNode();

		//Statements.consume(vf, current, RDF.TYPE, RDF.LIST, consumer, contexts); 
               //--just comment out this above line in order to get the desired output which is list items are properly inlined.

		Iterator<?> iter = values.iterator();
		while (iter.hasNext()) {
			Object o = iter.next();
			Value v = o instanceof Value ? (Value) o : Literals.createLiteralOrFail(vf, o);
			Statements.consume(vf, current, RDF.FIRST, v, consumer, contexts);
			if (iter.hasNext()) {
				Resource next = vf.createBNode();
				Statements.consume(vf, current, RDF.REST, next, consumer, contexts);
				current = next;
			} else {
				Statements.consume(vf, current, RDF.REST, RDF.NIL, consumer, contexts);
			}
		}
	}
```

----
The desired output in the issue 3681 can be achieved by making this change in consumeCollection.

### DIFF
--- a/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
+++ b/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
@@ -40,7 +40,12 @@ public class ElasticsearchSailExample {
 	 * @param args
 	 */
 	public static void main(String[] args) throws Exception {
+		long startTime = System.nanoTime();
+        // QueryResultFunction();
 		createSimple();
+		long endTime = System.nanoTime();
+        long timeElapsed = endTime - startTime;
+        System.out.println("Execution time in milliseconds for All the Queries: " + timeElapsed / 1000000);
 	}
 
 	/**
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleIRI.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleIRI.java
@@ -8,7 +8,7 @@
 package org.eclipse.rdf4j.model.impl;
 
 import java.util.Objects;
-
+import org.eclipse.rdf4j.model.util.Values;
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.base.AbstractIRI;
 import org.eclipse.rdf4j.model.util.URIUtil;
@@ -100,6 +100,15 @@ public class SimpleIRI extends AbstractIRI {
 		}
 
 		return iriString.substring(localNameIdx);
+	}
+
+	//conversion function to decode percent-encoded chars, Issue: 1291
+	@Override
+	public IRI percentageDecode(IRI url){
+		String url_ = url.toString();
+		url_ = url_.replaceAll("%", "");
+		url = Values.iri(url_);
+        return url;
 	}
 
 }
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFCollections.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFCollections.java
@@ -208,7 +208,7 @@ public class RDFCollections {
 
 		Resource current = head != null ? head : vf.createBNode();
 
-		Statements.consume(vf, current, RDF.TYPE, RDF.LIST, consumer, contexts);
+		//Statements.consume(vf, current, RDF.TYPE, RDF.LIST, consumer, contexts);
 
 		Iterator<?> iter = values.iterator();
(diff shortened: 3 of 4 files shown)


jeenbroekstra:
Could you please digitally sign the Eclipse Contributor Agreement (ECA)? See the [Contributor Guidelines](https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md) for more details on how to sign. Unfortunately, we cannot accept your contribution until these formalities are out of the way. It shouldn't take more than 5-10 minutes to sort out, let me know if you need a hand with anything.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleIRI.java
```diff
@@ -102,4 +102,13 @@ public String getLocalName() {
...
 	}
 
+	//conversion function to decode percent-encoded chars, Issue: 1291
+	@Override
+	public IRI percentageDecode(IRI url){
+		String url_ = url.toString();
+		url_ = url_.replaceAll("%", "");
+		url = Values.iri(url_);
+        return url;
+	}
```
jeenbroekstra, line 112: This looks like it may have accidentally ended up in this PR - I don't think this has anything to do with #3681?

core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFCollections.java
```diff
@@ -208,7 +208,7 @@ public static void consumeCollection(Iterable<?> values, Resource head, Consumer
 
 		Resource current = head != null ? head : vf.createBNode();
 
-		Statements.consume(vf, current, RDF.TYPE, RDF.LIST, consumer, contexts);
+		//Statements.consume(vf, current, RDF.TYPE, RDF.LIST, consumer, contexts);
```
jeenbroekstra, line 211: IMO this hides the issue rather than fixing it. I don't have a problem with taking this line out (it's not _required_ to have this type statement for a collection), but I think we want a more comprehensive fix, as well.

I think what you want to look at is the `TurtleWriter` class. The handling of writing lists is in the `handleStatementInternal` method (line 299-321). In particular, take a look at line 307 - this is where it decides if it will process the statement as part of a list:

```java
				if ((pred.equals(RDF.FIRST) || pred.equals(RDF.REST)) && isWellFormedCollection(subj)) {
					// we only use list shorthand syntax if the collection is considered well-formed
					handleList(st, canShortenObjectBNode);
```

My guess is the if-condition is overlooking the fact that it can also be a `rdf:type rdf:List` statement that should be handled as a list. As for _how_ it should handle it: my gut feel is that it should simply skip this statement.

compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
```diff
@@ -40,7 +40,12 @@ public class ElasticsearchSailExample {
 	 * @param args
 	 */
 	public static void main(String[] args) throws Exception {
+		long startTime = System.nanoTime();
+        // QueryResultFunction();
 		createSimple();
+		long endTime = System.nanoTime();
+        long timeElapsed = endTime - startTime;
+        System.out.println("Execution time in milliseconds for All the Queries: " + timeElapsed / 1000000);
```
jeenbroekstra, line 48: This as well - doesn't look like it's related to the collection issue?

jeenbroekstra (changes requested):

The-Nightwing:
I have submitted the request.

jeenbroekstra:
> I have submitted the request.

Can you double-check that you've done this correctly? All you need to do is sign up for an Eclipse account, then log in and navigate to the ECA form, which you need to tick and submit. As said in the contributor guidelines, it's important that you use the same e-mail address for all of this that you use for your commits.

hmottestad:
Screenshot 2022-05-15 at 07 12 17

The-Nightwing:
Okay let me check again.

Issue #2524, opened 2020-09-23
QueryStringUtil.getTupleQueryString doesnt handle inline binding expressions


pulquero:
The handling of binding substitution fails for queries of the form "select (func(?x) as ?y)". For a binding x=foo the output is "select (func('foo' as ?x) as ?y)" when it should be "select (func('foo') as ?y)". The current code only works for "select ?x ?y ?z" type queries. Regex needs to be modified to ignore patterns that are preceeded by 'as'.

hmottestad:
I think I understand what this is about, but could you make a small example for us so we can make a test. That would help us a lot!

pulquero:
select (count(?o) as ?c) {?s ?p ?o}  with binding o='foo'

pulquero:
Example fix:
```
.
.
.		for (String name : bindings.getBindingNames()) {
			String replacement = QueryStringUtil.valueToString(bindings.getValue(name));
			if (replacement != null) {
				String pattern = "[\\?\\$]" + name + "(?!\\s?\\))(?=\\W)";
				select = select.replaceAll(pattern, "(" + Matcher.quoteReplacement(replacement) + " as ?" + name + ")");
				pattern = "[\\?\\$]" + name + "(?=\\s?\\)\\W)";
				select = select.replaceAll(pattern, Matcher.quoteReplacement(replacement));
				pattern = "[\\?\\$]" + name + "(?=\\W)";
				where = where.replaceAll(pattern, Matcher.quoteReplacement(replacement));
			}
		}
.
.
.
```

jeenbroekstra:
Related issues: #2212 and #520

hmottestad:
This should be a decent first issue for hacktoberfest. Just write a couple of tests and try out the fix from @pulquero . It's a good patch for the time being.

hmottestad:
Tried to assign @Sanchit-Trivedi

PR #3879, opened 2022-05-13
GH-2524 : Added regex for handling inline binding expressions

Sanchit-Trivedi:
GitHub issue resolved: #2524

The handling of binding substitution fails for queries of the form select (func(?x) as ?y). For a binding x=foo the output is select (func('foo' as ?x) as ?y) when it should be select (func('foo') as ?y). The current code only works for select ?x ?y ?z type queries. Regex has been modified to ignore patterns that are preceded by 'as'.

----


jeenbroekstra:
Given that this is a bug fix: could you rebase against the main branch, instead of develop?

jeenbroekstra:
I think you need to run formatting here again as well - looks like it's not happy with the formatting of the javadoc in this case.

hmottestad:
Could you add some tests too?

hmottestad:

Issue #327, opened 2016-09-06
Include timing stats in workbench query results


knoan:
(Migrated from https://openrdf.atlassian.net/browse/SES-2110)

It would be useful to include some basic execution timing stats (e.g. elapsed server-side wall-clock time) in the query results page of the workbench.

jeenbroekstra:
This is marked as a "good first issue" but I've just realized this is probably not straightforward - especially if we want it to report server-side elapsed time. The response document is in one of several standardized formats, and AFAIK none of these formats have an option to carry this kind of metadata. So getting the server time back to the frontend would have to be done by some custom extension, possibly a response header.

It might be easier to go with client-side time at first.

nguyenm100:
I could have sworn I saw query time on the result set of workbench in an older (3.x?) but I no longer see it.  Did I imagine it?  If not, agree this would be great to have.

PR #3880, opened 2022-05-14
GH-327 Include timing stats in workbench query results

The-Nightwing:
GitHub issue resolved: #327

Made changes in the ElasticSearchSailExample.java, it had multiple queries running in only one function, so to print runtime, i just calculated runtime for that only function which ran multiple queries together. 

Please point out if the way is wrong.

----

### DIFF
--- a/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
+++ b/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
@@ -40,7 +40,11 @@ public class ElasticsearchSailExample {
 	 * @param args
 	 */
 	public static void main(String[] args) throws Exception {
+		long startTime = System.nanoTime();
 		createSimple();
+		long endTime = System.nanoTime();
+        long timeElapsed = endTime - startTime;
+        System.out.println("Execution time in milliseconds for All the Queries: " + timeElapsed / 1000000);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleIRI.java
```diff
@@ -102,4 +102,13 @@ public String getLocalName() {
...
 	}
 
+	//conversion function to decode percent-encoded chars, Issue: 1291
+	@Override
+	public IRI percentageDecode(IRI url){
+		String url_ = url.toString();
+		url_ = url_.replaceAll("%", "");
+		url = Values.iri(url_);
+        return url;
+	}
```
jeenbroekstra, line 112: This should not be part of this pull request.

compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
```diff
@@ -40,7 +40,11 @@ public class ElasticsearchSailExample {
 	 * @param args
 	 */
 	public static void main(String[] args) throws Exception {
+		long startTime = System.nanoTime();
 		createSimple();
+		long endTime = System.nanoTime();
+        long timeElapsed = endTime - startTime;
+        System.out.println("Execution time in milliseconds for All the Queries: " + timeElapsed / 1000000);
```
jeenbroekstra, line 47: I don't follow how this addresses the issue? The feature request is about seeing query evaluation time in the RDF4J Workbench tool - a web application. This calculates execution time in a bit of example code, instead.

jeenbroekstra (commented):
Thanks for submitting these PRs. Can I suggest that perhaps you can best concentrate on just one issue for now, and make sure we get that done properly? Then move on to the next. There's a bit of a setup problem with the pull rquests you've created sofar (they all contain code that is not relevant to the issue the PR is for), and also I think you may want to have a bit of a discussion with us, so we can come up with a good solution for each issue together.

Issue #1291, opened 2019-02-11
encoded vs un-encoded IRIs


VladimirAlexiev:
Maybe I misunderstand the specs, but it seems to me I should be able to compare encoded vs un-uncoded IRIs as equal, but I cannot. (Tested in GraphDB 8.7.2, RDF4J 2.3.2)

This returns false (tried `sameTerm` and `=`):
```sparql
select * {
  bind (
    sameTerm(<https://en.wikipedia.org/wiki/University_of_Tromsø>,
     <https://en.wikipedia.org/wiki/University_of_Troms%C3%B8>)
        as ?x)
  }
```

This returns different strings:
```sparql
select * {
    bind(str(<https://en.wikipedia.org/wiki/University_of_Tromsø>) as ?s1)
    bind(str(<https://en.wikipedia.org/wiki/University_of_Troms%C3%B8>) as ?s3)
  }
```

What the standards say:
- https://www.w3.org/TR/sparql11-query/#func-str "returns the codepoint representation of an IRI". 
  - Doesn't say that anything needs to be encoded. Since it talks of IRIs, I think unicode chars should be preserved. And the example shows that "@", at least , is preserved.
- https://www.w3.org/TR/rdf-concepts/#section-Graph-URIref "Because of the risk of confusion between RDF URI references that would be equivalent if derefenced, the use of %-escaped characters in RDF URI references is strongly discouraged. See also the URI equivalence issue of the Technical Architecture Group"

See https://phabricator.wikimedia.org/T210738 for a similar issue raised against Wikidata/Blazegraph.

barthanssens:
some additional reading: https://tools.ietf.org/html/rfc3987#section-5, https://tools.ietf.org/html/rfc3987#section-5.3.1

VladimirAlexiev:
"Any kind of IRI comparison REQUIRES that all escapings or encodings in the protocol or format that carries an IRI are resolved."

barthanssens:
Hmz, does it also implies that IRI should be unescaped/decoded when created (ValueFactory) / stored ?
Otherwise it seems like a costly operation...

VladimirAlexiev:
I don't know the structure of the code, but it seems most natural to store them in decoded form.
cc @jeenbroekstra

jeenbroekstra:
Currently away so I can't comment in too much detail, but two points for consideration:

1. there is a difference between codepoint escape sequences (`\uHEX`), which are processed, and %-encodings (which are not, as a rule).
[...]

PR #3881, opened 2022-05-14
GH-1291 Add convenience functions for comparing encoded vs un-encoded IRIs

The-Nightwing:
GitHub issue resolved: #1291

An abstract function has been made in IRI.java, and its implementation is given in SimpleIRI.java.

### DIFF
--- a/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
+++ b/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
@@ -40,7 +40,11 @@ public class ElasticsearchSailExample {
 	 * @param args
 	 */
 	public static void main(String[] args) throws Exception {
+		long startTime = System.nanoTime();
 		createSimple();
+		long endTime = System.nanoTime();
+        long timeElapsed = endTime - startTime;
+        System.out.println("Execution time in milliseconds for All the Queries: " + timeElapsed / 1000000);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/examples/ElasticsearchSailExample.java
```diff
@@ -40,7 +40,11 @@ public class ElasticsearchSailExample {
 	 * @param args
 	 */
 	public static void main(String[] args) throws Exception {
+		long startTime = System.nanoTime();
 		createSimple();
+		long endTime = System.nanoTime();
+        long timeElapsed = endTime - startTime;
+        System.out.println("Execution time in milliseconds for All the Queries: " + timeElapsed / 1000000);
```
jeenbroekstra, line 47: This shouldn't be part of this pull request - I think you need to make sure you are creating new branches _before_ picking up another issue.

core/model/src/main/java/org/eclipse/rdf4j/model/impl/SimpleIRI.java
```diff
@@ -102,4 +102,13 @@ public String getLocalName() {
...
 	}
 
+	//conversion function to decode percent-encoded chars, Issue: 1291
+	@Override
+	public IRI percentageDecode(IRI url){
+		String url_ = url.toString();
+		url_ = url_.replaceAll("%", "");
+		url = Values.iri(url_);
+        return url;
+	}
```
jeenbroekstra, line 112: Couple of things: first of all, I don't think `IRI` itself (or the `SimpleIRI` implementation) is really the best place for this. What we're looking for in issue #1291 is probably more a SPARQL function, or perhaps a custom option in the Rio parsers to use percent-decoding.

Second: this function as you currently have it doesn't percent-decode. All it does is remove any %-chars from the IRI - but that's not decoding it. 

I would suggest that we continue discussion on the issue ticket itself, to talk a bit more about the best solution, if you're still interested in working on this.

PR #3896, opened 2022-05-15
4.0.1-release-notes

hmottestad:
Javadocs, release-notes and news item for 4.0.1


jeenbroekstra (approved):
LGTM. Btw don't forget creating the release tag on github (right hand column). It's something I often overlook myself as well, and it seems a bit redundant, but if we don't do it it just shows an out-of-date version number there (there seems to be no way to remove that section).

Issue #3902, opened 2022-05-16
Fix MemEvaluationStatistics regression introduced during rewrite


hmottestad:
MemEvaluationStatistics was rewritten in #3862 but it wasn't a pure refactor and an important case was omitted.

Setting this a an internal task since this regression hasn't been released to any users.

PR #3903, opened 2022-05-16
GH-3902 fix MemEvaluationStatistics regression

hmottestad:
GitHub issue resolved: #3902 

Original code first looked up all four S,P,O,C in the memory value factory before finding out what the smallest statement list size was. I rewrote it to look up one value at a time instead of all four at once, that way we could return earlier if the statement list size was 0. 

I forgot to take account of that the memory value factory returns null when the value is unknown, which entails that there can't be any statements with the value. Fixed that now.

----


jeenbroekstra (approved):
Good catch

jeenbroekstra:
Looks like japicmp is giving you grief. Does not look related to your changes though. Hopefully a rebase will fix it?

Issue #3907, opened 2022-05-17
Deprecate Iteration interface in favour of only using CloseableIteration


hmottestad:
`CloseableIteration` extends `Iteration` and adds a close method through `AutoCloseable`. Because of how the JVM deals with interfaces there is an increased need to itable stubs when calling methods in a class that implements `CloseableIteration` from variable declared as `Iteration`.

hmottestad:
@jeenbroekstra 

I would like to deprecate the `Iteration` interface because of the performance improvements I'm able to get by only using `CloseableIteration`. I don't know exactly why it makes such a big difference, but it's got something to do with how interface calls are made in Java compared to regular virtual calls. There is a slow path where the JVM creates a stub code that gets optimized later on. In my profiler this shows up as `itable stub`. I think that the reason this affects us so much is that we have a lot of short lived iterations, for instance during join operations where we create a new iteration for the right side for every element on the left side.

Here is some background info I found: 
 - https://wiki.openjdk.java.net/display/HotSpot/InterfaceCalls
 - https://shipilev.net/blog/2015/black-magic-method-dispatch/

I have a PR open here: https://github.com/eclipse/rdf4j/pull/3909

I'm not sure how to do a clean break from using the `Iteration` interface. Currently I'm leaning towards reimplementing all the classes that use `Iteration` in new packages. An example would be for the package `org.eclipse.rdf4j.common.iteration` I would create a new package `org.eclipse.rdf4j.common.iteration2` with a copy of all the classes but swapping out the use of `Iteration` with `CloseableIteration`. Would definitely come up with a better name instead of `iteration2` :P

jeenbroekstra:
I have no objection to marking it deprecated. 

As for replacing/renaming: I have long thought that maybe we no longer really need the iteration classes - their main reason for existing was that we needed an iterator that can throw a checked exception. Since RDF4J Exceptions are now all unchecked, perhaps we really no longer need Iteration and its subtypes, and we can in fact just work with Iterator implementations. This performance issue you found might be a good excuse to revisit that idea.

hmottestad:
We still want to have AutoCloseable, which I'm not able to find as a combination with Iterator or Iterable. I did find https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/BaseStream.html which might be a good choice if it means that we can take advantage of the stream optimizations in GraalVM.

hmottestad:
[...]

PR #3909, opened 2022-05-17
GH-3907 deprecate Iteration

hmottestad:
GitHub issue resolved: #3907 

- deprecate the Iteration interface

----


core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIterationSpliterator.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Consumer;
+
+/**
+ * A {@link Spliterator} implementation that wraps a {@link CloseableIteration}.
+ */
+public class CloseableIterationSpliterator<T, E extends Exception> extends Spliterators.AbstractSpliterator<T> {
```
jeenbroekstra, line 19: Can we make this package-private? I think we only really need it to support stream conversion, which happens internally.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/DelayedIteration.java
```diff
@@ -16,6 +16,7 @@
  * or where a created iteration consumes scarce resources like JDBC-connections or memory. Subclasses must implement the
  * <var>createIteration</var> method, which is called once when the iteration is first needed.
  */
+@Deprecated(since = "4.1.0", forRemoval = true)
 public abstract class DelayedIteration<E, X extends Exception> extends AbstractCloseableIteration<E, X> {
```
jeenbroekstra, line 20: You've marked this (and several other abstract iteration classes) as marked for removal. They're in active use though. What are you aiming to replace them by?

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
```diff
@@ -23,9 +27,32 @@
...
+				.stream(new CloseableIterationSpliterator<>(this), false)
+				.onClose(() -> {
+					try {
+						close();
+					} catch (RuntimeException e) {
+						throw e;
+					} catch (Exception e) {
+						throw new RuntimeException(e);
+					}
+				});
```
jeenbroekstra, line 53: I think you can replace this by using the new method you added to `Iterations`, instead.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
```diff
@@ -23,9 +27,32 @@
  *   // process the exception that can be thrown while processing.
  * }
  * </pre>
+ *
+ * @implNote In the future this interface will stop extending {@link Iteration} and instead declare the same interface
+ *           methods directly. The interface will also stop requiring implementations to automatically close when
+ *           exhausted, instead making this an optional feature and requiring the user to always call close.
```
jeenbroekstra, line 33: I think it's worth also marking this interface as `@Deprecated` (though not necessarily for removal), and start having a think about replacing the whole concept of Iterations with something else, for example something like (just spitballing here):

```java
public interface ResultIterator<T> extends Iterator<T>, AutoCloseable 
```

Not saying we _do_ want to go that way, but making this deprecated now gives us options in 5.0.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/DelayedIteration.java
hmottestad: In 5.0.0 I plan on doing a search replace for `Iteration` with `CloseableIteration` and then removing the deprecation annotations.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
hmottestad: Sounds good.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/DelayedIteration.java
hmottestad: Maybe it's simpler if I just use `@Deprecated(since = "4.1.0")` instead.

hmottestad: There. What do you think @jeenbroekstra ?

Issue #3696, opened 2022-02-24
BIND seems over zealous 


peterjohnlawrence:
### Current Behavior

I have a particular query in which a variable that is later used in a BIND expression, is used in two basic graph patterns. However, the query only seems to execute one of these BGPs before it applies the BIND. 
The test query is 
```
 { 
  ?bind   rdfs:label ?b1 .
  ?bind   a def:Unit  . 
  FILTER (?b1 = 'Unit2') . 
  BIND(?bind   as ?n0)
  ?n0 def:hasProductBatteryLimit ?n1 
 }
```

This incorrectly returns a value of :Unit2 for the variables ?bind and ?n0

### Expected Behavior

I would expect the query to satisfy the two BGPs 

```
  ?bind   rdfs:label ?b1 .
  ?bind   a def:Unit  . 
```

and then apply the filter.

### Steps To Reproduce

Using the two attached  *.TTL files in any repository (memory seems to exhibit the issue)

This query should return no records:

```
PREFIX def: <http://inova8.com/calc2graph/def/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
SELECT  *
 { 
  ?bind   rdfs:label ?b1 
  ?bind   a def:Unit  .
  FILTER (?b1 = 'Unit2') . 
  BIND(?bind   as ?n0)
  ?n0 def:hasProductBatteryLimit ?n1 
 }
```

However, it returns with the variable ?bind set to  :Unit2, which is incorrect . 
:Unit2 does satisfy the BGP { ?bind   a def:Unit  .} but not the  { ?bind   rdfs:label ?b1  . FILTER (?b1 = 'Unit2') . } pattern.

Rearranging the query, or including {...}, does not seem to change the behaviour.

[Calc2Graph.zip](https://github.com/eclipse/rdf4j/files/8134980/Calc2Graph.zip)

### Version

3.7.4

### Are you interested in contributing a solution yourself?

Perhaps?

jeenbroekstra:
what does the (optimized) query plan look like when you try to `explain()` it?

peterjohnlawrence:
Projection (resultSizeActual=0, totalTimeActual=0.647ms, selfTimeActual=0.003ms)
╠══ProjectionElemList
║     ProjectionElem "bind"
║     ProjectionElem "b1"
║     ProjectionElem "n0"
║     ProjectionElem "n1"
[...]

PR #3925, opened 2022-05-22
GH-3696 fix QueryJoinOptimizer to handle BIND arguments wrapped in a filter condition

jeenbroekstra:
GitHub issue resolved: #3696  

- fix in QueryJoinOptimizer to ensure it handles BIND arguments wrapped in e.g. a filter condition as priority arguments as well
- added compliance test
- small improvement in the optimizer pipeline: push filters down _before_ join optimization

----


jeenbroekstra:
This test case should probably be moved into the SPARQL compliance tests. Related to this: #3926

hmottestad:
Btw. I think that the reason the build doesn't fail is that the class name doesn't end in "test".

jeenbroekstra:
I think the core of the problem is in the `QueryJoinOptimizer`. The (optimized) query plan for the failing query is this:

```
Projection
╠══ProjectionElemList
║     ProjectionElem "bind"
║     ProjectionElem "b1"
║     ProjectionElem "n0"
║     ProjectionElem "n1"
╚══Join (JoinIterator)
   ├──StatementPattern (costEstimate=1, resultSizeEstimate=1)
   │     Var (name=n0)
   │     Var (name=_const_1bf65419_uri, value=http://example.org/has, anonymous)
   │     Var (name=n1)
   └──Extension
      ╠══Join (JoinIterator)
      ║  ├──Filter (costEstimate=4, resultSizeEstimate=4)
      ║  │  ╠══Compare (=)
      ║  │  ║     Var (name=b1)
      ║  │  ║     ValueConstant (value="Unit2")
      ║  │  ╚══StatementPattern (costEstimate=1, resultSizeEstimate=2)
      ║  │        Var (name=bind)
      ║  │        Var (name=_const_9285ccfc_uri, value=http://www.w3.org/2000/01/rdf-schema#label, anonymous)
      ║  │        Var (name=b1)
      ║  └──StatementPattern (costEstimate=1, resultSizeEstimate=2)
      ║        Var (name=bind)
      ║        Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
      ║        Var (name=_const_62cbbc05_uri, value=http://example.org/Unit, anonymous)
      ╚══ExtensionElem (n0)
            Var (name=bind)
```

The Join optimer pulls the `(n0, has, n1)` statement pattern forward (becaue it cheaper than the full Extension block), but in doing so it messes with  the variable scope: there is a hard requirement that the target variable of an extension can not have been used before.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StandardQueryOptimizerPipeline.java
```diff
@@ -56,9 +56,9 @@ public Iterable<QueryOptimizer> getOptimizers() {
 				new UnionScopeChangeOptimizer(),
 				new QueryModelNormalizer(),
 				new ProjectionRemovalOptimizer(), // Make sure this is after the UnionScopeChangeOptimizer
+				new FilterOptimizer(),
 				new QueryJoinOptimizer(evaluationStatistics),
 				new IterativeEvaluationOptimizer(),
-				new FilterOptimizer(),
```
jeenbroekstra, line 61: This isn't _necessary_ for the fix, but pushing filter conditions down before doing join optimization reduces the number of tree walks the join optimizer potentially has to do.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryJoinOptimizer.java
```diff
@@ -236,11 +235,11 @@ protected <M extends Map<Var, Integer>> M getVarFreqMap(List<Var> varList, M var
...
 
-		protected List<Extension> getExtensions(List<TupleExpr> expressions) {
-			List<Extension> extensions = new ArrayList<>();
+		protected List<TupleExpr> getExtensions(List<TupleExpr> expressions) {
+			List<TupleExpr> extensions = new ArrayList<>();
 			for (TupleExpr expr : expressions) {
-				if (expr instanceof Extension) {
-					extensions.add((Extension) expr);
+				if (TupleExprs.containsExtension(expr)) {
+					extensions.add(expr);
```
jeenbroekstra, line 242: This is the core of the fix. The problem was that the join argument is a filter condition which in turn  has the Extension as an argument. This is a bit obfuscated because later on the FilterOptimizer removes the filter condition altogether. The fix ensures that we also look deeper down in the algebra tree, so that any wrapped Extensions are also treated as priority arguments.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/helpers/TupleExprs.java
```diff
@@ -60,6 +62,33 @@ public static boolean containsSubquery(TupleExpr t) {
...
+					|| (n instanceof Projection && ((Projection) n).isSubquery())
+					|| (n instanceof Service)) {
+				return false;
+			} else {
+				queue.addAll(getChildren(n));
+			}
+		}
+		return false;
+	}
+
```
hmottestad, line 91: Creating an ArrayDeque is fairly expensive. I've optimized a similar method (on the develop branch). Could you rewrite this method to match that approach?

```
public static boolean containsSubquery(TupleExpr t) {

		TupleExpr n = t;
		Deque<TupleExpr> queue = null;

		do {
			if (n instanceof Projection && ((Projection) n).isSubquery()) {
				return true;
			} else if (n instanceof Join) {
				// projections already inside a Join need not be
				// taken into account
				return false;
			} else {
				List<TupleExpr> children = getChildren(n);
				if (!children.isEmpty()) {
					if (queue == null) {
						queue = new ArrayDeque<>();
					}
					queue.addAll(children);
				}
			}

			n = queue != null ? queue.poll() : null;

		} while (n != null);

		return false;
	}
```

hmottestad:
Looks good.

jeenbroekstra:
Ok what the hell? 

This has started failing again after a rebase, and when trying to debug it appears it doesn't even pass through the QueryJoinOptimizer anymore.

jeenbroekstra:
@hmottestad why is there a duplicate QueryJoinOptimizer class?

jeenbroekstra:
@hmottestad it looks as if the optimizer classes were refactored into a new package (`org.eclipse.rdf4j.query.algebra.evaluation.optimizer`) but unfortunately the old classes were left in place as-is (instead of making them deprecated extensions). I'll do some cleanup.

jeenbroekstra:
Argh. My cleanup of the refactored optimizer code has created a binary incompatibility. 

@hmottestad where did we land on the discussion about compatibility for minor releases? I vaguely remember that we discussed dropping binary compatibility for minor releases, only guaranteeing source compatibility. But I don't know if we made a final decision on this.

hmottestad:
I ended up duplicating a lot of the optimizers because I needed to change the super class (extends). I should have deprecated the original classes. I presume I forgot to update the tests to point to the new classes?

I'll open an issue for checking if I've deprecated all the classes and updated the tests.

Issue #2081, opened 2020-04-09
Organize and rename ComplexSPARQLQueryTest testsuite


jeenbroekstra:
The `ComplexSPARQLQueryTest` is a test suite for every kind of SPARQL query test we need in terms of regression testing. However, it is currently organized as a single class with a host of individual test methods that are inconsistently named and not grouped or ordered in any meaningful manner, making maintenance hard.

We should re-organize the class in a suite comprised of a number of sub-suites (similarly to how this is done, for example, with the `org.eclipse.rdf4j.repository.OptimisticIsolationTest` suite), each suite covering a particular aspect of SPARQL (e.g. "aggregates", "property paths", etc). 

We should also consider renaming the class as these are not necessarily "complex" SPARQL queries - they are just test cases that are not covered by the official W3C manifest tests.

PR #3926, opened 2022-05-23
GH-2081 reorg sparql custom compliance tests as JUnit 4 Test suite

jeenbroekstra:
GitHub issue resolved: #2081  

- reorganized ComplexSPARQLQueryTest class into a suite of several test classes
- using JUnit 4 Suite setup as I could not get things working with the JUnit 5 `@Suite` annotation
- rough separation into related test cases, can be refined/extended further as necessary

----


hmottestad:
I was wondering about doing some rewrites of those tests but was wondering if they are not published on maven? In which case others might be using them.

jeenbroekstra:
> I was thinking about doing some rewrites of those tests but was wondering if they are not published on maven? In which case others might be using them.

That's a good point - I may leave the old set of tests in (deprecated) purely for backward compatibility. They won't run if we don't have concrete implementations on our end, but other parties can continue using them.

JervenBolleman:
Maybe ask around if anyone does? people can always continue to use the old test jars.

jeenbroekstra:
> Maybe ask around if anyone does? people can always continue to use the old test jars.

Unfortunately I've found in the past that "asking around" is not a very reliable system. People don't pay attention to that kind of stuff until they bump versions and suddenly stuff breaks.

In any case leaving the old tests in places in deprecated form is not a big chore, it's a single Java class after all :)

JervenBolleman:
> > Maybe ask around if anyone does? people can always continue to use the old test jars.
> 
> Unfortunately I've found in the past that "asking around" is not a very reliable system. People don't pay attention to that kind of stuff until they bump versions and suddenly stuff breaks.
> 
> In any case leaving the old tests in places in deprecated form is not a big chore, it's a single Java class after all :)

Ok, maybe add a `@Deprecated(forRemoval=true)` so we can delete it at version 5 ?

jeenbroekstra:
Updated this, restoring the old suite as a deprecated class, and improved the javadoc a bit here and there.

testsuites/sparql/pom.xml
```diff
@@ -96,6 +96,11 @@
 			<artifactId>rdf4j-repository-contextaware</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>rdf4j-repository-manager</artifactId>
+			<version>${project.version}</version>
+		</dependency>
```
jeenbroekstra, line 103: Ugh - I should take this out again, I put this in to try something but went with a different approach instead.

hmottestad:
Could it be useful to mark the new classes as experimental or internal only so that we can keep improving them and maybe move them to junit 5?

jeenbroekstra:
> Could it be useful to mark the new classes as experimental or internal only so that we can keep improving them and maybe move them to junit 5?

Good point, will do.

Issue #3954, opened 2022-06-03
Regression when parsing queries with math operations on aggregates without GROUP BY


hmottestad:
### Current Behavior

The following query throws an exception during parsing since 4.0.0. It works fine with 3.7.7.

```
SELECT (COUNT (DISTINCT *) AS ?count) (?count / ?count AS ?result) {
    ?s a ?o .
}
```

```
Exception in thread "main" org.eclipse.rdf4j.query.MalformedQueryException: non-aggregate expression 'MathExpr (/)
   Var (name=count)
   Var (name=count)
' not allowed in projection when using GROUP BY.
	at org.eclipse.rdf4j.query.parser.sparql.SPARQLParser.buildQueryModel(SPARQLParser.java:220)
	at org.eclipse.rdf4j.query.parser.sparql.SPARQLParser.parseQuery(SPARQLParser.java:178)
	at org.eclipse.rdf4j.query.parser.QueryParserUtil.parseQuery(QueryParserUtil.java:93)
	at org.eclipse.rdf4j.query.parser.QueryParserUtil.parseTupleQuery(QueryParserUtil.java:108)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.prepareTupleQuery(SailRepositoryConnection.java:288)
	at org.eclipse.rdf4j.repository.sail.SailRepositoryConnection.prepareTupleQuery(SailRepositoryConnection.java:59)
	at org.eclipse.rdf4j.repository.base.AbstractRepositoryConnection.prepareTupleQuery(AbstractRepositoryConnection.java:143)
	at org.eclipse.rdf4j.repository.RepositoryConnection.prepareTupleQuery(RepositoryConnection.java:212)
	at org.eclipse.rdf4j.sail.memory.Temp.main(Temp.java:37)
Caused by: org.eclipse.rdf4j.query.parser.sparql.ast.VisitorException: non-aggregate expression 'MathExpr (/)
   Var (name=count)
   Var (name=count)
' not allowed in projection when using GROUP BY.
	at org.eclipse.rdf4j.query.parser.sparql.TupleExprBuilder.visit(TupleExprBuilder.java:644)
	at org.eclipse.rdf4j.query.parser.sparql.TupleExprBuilder.visit(TupleExprBuilder.java:234)
	at org.eclipse.rdf4j.query.parser.sparql.ast.ASTSelect.jjtAccept(ASTSelect.java:32)
[...]

PR #3955, opened 2022-06-05
GH-3954 fix for query regression with group by

hmottestad:
GitHub issue resolved: #3954 

 - use recursion to check if a projected variable is effectively a constant, an aggregation or declared in the group by expression

----


hmottestad:
@tokovach this should fix the issue you reported

hmottestad:
@jeenbroekstra I'll wait for you to chime in on whether or not this is according to the SPARQL 1.1 specification. The queries pass the SPARQL Validator, but I'm not sure if we that can be relied on for conformance. I feel that this at least follows the intent of the specification.

I also considered the possibility for infinite recursion. Made some tests, but ended up removing them since it crashed in some unrelated code. So I don't think it's worth the effort.

jeenbroekstra:
Thanks @hmottestad I'll take another look tomorrow.

jeenbroekstra (approved):
This looks fine to me. Nice solution @hmottestad . From a cursory read these kinds of examples are indeed legal SPARQL, for example in section 16.1.2 (SELECT expressions):

> [...] The rules of assignment in SELECT expression are the same as for assignment in BIND. The expression combines variable bindings already in the query solution, **or defined earlier in the SELECT clause**, to produce a binding in the query solution.

(emphasis mine)

So yeah, this seems legit.

Issue #3590, opened 2022-01-10
Reduce synchronization for tracking iterations in the AbstractSailConnection


hmottestad:
### Problem description

There is a decent amount of synchronization connected to tracking iterations in the AbstractSailConnection. 

### Preferred solution

Alternatives:
 1. Use ConcurrentHashMap.
 2. Use Cleaner and two LongAdders to track the number of active vs. closed iterations.

The first options is simpler to implement, but the second will be considerably more performant.

### Are you interested in contributing a solution yourself?

Yes

PR #3956, opened 2022-06-05
GH-3590 Use Cleaner instead of Map for closing unreachable iterations

hmottestad:
GitHub issue resolved: #3590 

Instead of tracking all iterations in a map we instead tell the JVM to automatically close them if the users has forgotten. We register the iterations with the [Java Cleaner](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html) (introduced in Java 9 as a better alternative to finalize) together with a Runnable that closes the iteration. The Cleaner executes the runnable when the iteration has become phantom reachable (e.g. eligible for GC). We use our existing ConcurrentCleaner which wraps an array of Cleaner instances to spread the load when objects are registered from multiple threads.

I've retained the tracking functionality for debugging purposes and it can be enabled the same way as we would previously have enabled tracking with stack traces.

A change for users will be that a connection will now `throw new SailException("Connection closed before all iterations were closed.")` if there are unclosed iterations. Before throwing this exception we call `System.gc()` multiple times to hopefully trigger the closing of any "lost" iterations. 

----

### DIFF
--- a/testsuites/repository/src/main/java/org/eclipse/rdf4j/testsuite/repository/TupleQueryResultTest.java
+++ b/testsuites/repository/src/main/java/org/eclipse/rdf4j/testsuite/repository/TupleQueryResultTest.java
@@ -8,9 +8,6 @@
 package org.eclipse.rdf4j.testsuite.repository;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


hmottestad:
## Before

```
```

## After

```
Benchmark                                           Mode  Cnt     Score    Error  Units
ComplexLargeBenchmark.noPreloadingNonEmptyParallel  avgt   10  2595.491 ± 30.023  ms/op
```

testsuites/repository/src/main/java/org/eclipse/rdf4j/testsuite/repository/TupleQueryResultTest.java
```diff
@@ -35,28 +31,17 @@
...
-	@BeforeClass
-	public static void setUpClass() throws Exception {
-		System.setProperty("org.eclipse.rdf4j.repository.debug", "true");
-	}
-
-	@AfterClass
-	public static void afterClass() throws Exception {
-		System.setProperty("org.eclipse.rdf4j.repository.debug", "false");
-	}
-
```
hmottestad, line 59: Rewrote this test to use Junit 5. I know that it's a public class in our Test Suite so I'm not sure if this would be backwards compatible or not. If not then we can create a new package and copy my rewrite over to that package and deprecate the old class.

Issue #3992, opened 2022-06-17
Define predefined prefixes for SPARQL requests


ate47:
### Problem description

Same as [Is it possible to define default prefixes with connection.prepareQuery(String)?](https://github.com/eclipse/rdf4j/discussions/3980)

Add the possibility to define default SPARQL prefixes without having to declare them in the query?

Like I want to be able to run the query

```sparql
SELECT * {ex:aaa ?p ?o}
```

without having to specify `PREFIX ex: <http://example.org/>`.

### Preferred solution

Add in the [org.eclipse.rdf4j.query.parser.sparql.SPARQLParser](https://github.com/eclipse/rdf4j/blob/main/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java) class a Map<String, IRI> to define default prefixes and use them if not defined in a new query.

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

For now I was concatenating `PREFIX ex: <http://example.org/> ` to my query before running it, but  if someone try to define `ex:` it fails.

PR #3993, opened 2022-06-17
GH-3992 add sparql parser default prefixes

ate47:
GitHub issue resolved: #3992

This PR add a constructor to the SPARQLParser class and a new method in the SPARQLParserFactory to add custom default prefixes to the parser.

```java
// SPARQLParser
SPARQLParser(Set<Namespace> customDefaultPrefixes)

// SPARQLParserFactory
QueryParser getParser(Set<Namespace> customDefaultPrefixes);
```

Example

```java
// create a set with the ex namespace 
Set<Namespace> customDefaultPrefixes = new HashSet<>();
customDefaultPrefixes.add(Values.namespace("ex", "http://example.org/"));

// create a parser with the default prefixes
SPARQLParser parser = new SPARQLParser(customDefaultPrefixes);

// we can use the default prefix without having to specify it
parser.parseQuery("SELECT ?s {?s ex:aaa ex:ooo}", null);

// we can redefine the prefix in the query if we want another one (this is not an error)
parser.parseQuery("PREFIX ex: <http://example2.org/> SELECT ?s {?s ex:aaa ex:ooo}", null);


// we cannot redefine a prefix in the same request (this is still an excepted error)
parser.parseQuery("PREFIX ex: <http://example.org/> PREFIX ex: <http://example2.org/> SELECT ?s {?s ex:aaa ex:ooo}", null);
```

----

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/PrefixDeclProcessor.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/PrefixDeclProcessor.java
@@ -50,11 +50,13 @@ public class PrefixDeclProcessor {
 	 * query, verifies that prefixes are not redefined and replaces any {@link ASTQName} nodes in the query with
 	 * equivalent {@link ASTIRI} nodes.
 	 *
-	 * @param qc The query that needs to be processed.
+	 * @param qc                    The query that needs to be processed.
+	 * @param customDefaultPrefixes Custom prefixes to add, will override SPARQL default prefixes. can't be null.
 	 * @return A map containing the prefixes that are declared in the query (key) and the namespace they map to (value).
 	 * @throws MalformedQueryException If the query contains redefined prefixes or qnames that use undefined prefixes.
 	 */
-	public static Map<String, String> process(ASTOperationContainer qc) throws MalformedQueryException {
+	public static Map<String, String> process(ASTOperationContainer qc, Map<String, String> customDefaultPrefixes)
+			throws MalformedQueryException {
 		List<ASTPrefixDecl> prefixDeclList = qc.getPrefixDeclList();
 
 		// Build a prefix --> IRI map
@@ -71,8 +73,16 @@ public class PrefixDeclProcessor {
 			prefixMap.put(prefix, iri);
 		}
 
+		int preDefaultPrefixes = 0;
+
+		// insert the default prefixes if presents
+		for (Entry<String, String> defaultPrefix : customDefaultPrefixes.entrySet()) {
+			preDefaultPrefixes += insertDefaultPrefix(prefixMap, defaultPrefix.getKey(), defaultPrefix.getValue());
+		}
+
 		// insert some default prefixes (if not explicitly defined in the query)
-		final int defaultPrefixesAdded = insertDefaultPrefix(prefixMap, "rdf", RDF.NAMESPACE)
+		final int defaultPrefixesAdded = preDefaultPrefixes
+				+ insertDefaultPrefix(prefixMap, "rdf", RDF.NAMESPACE)
 				+ insertDefaultPrefix(prefixMap, "rdfs", RDFS.NAMESPACE)
 				+ insertDefaultPrefix(prefixMap, "rdf4j", RDF4J.NAMESPACE)
 				+ insertDefaultPrefix(prefixMap, "sesame", SESAME.NAMESPACE)
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java
@@ -15,9 +15,12 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
+import org.eclipse.rdf4j.model.Namespace;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.util.Namespaces;
 import org.eclipse.rdf4j.query.Dataset;
 import org.eclipse.rdf4j.query.IncompatibleOperationException;
 import org.eclipse.rdf4j.query.MalformedQueryException;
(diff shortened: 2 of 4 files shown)


barthanssens:
Hi,

thanks for the PR, just wondering if e.g. a `Set<Namespace>` might be more elegant.
That would also allow for (re)using the predefined namespaces from org.eclipse.rdf4j.model.vocabulary....

ate47:
True! I wasn't aware of the existence of this package, I've replaced the map to a set parameter

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/PrefixDeclProcessor.java
```diff
@@ -55,6 +55,21 @@ public class PrefixDeclProcessor {
 	 * @throws MalformedQueryException If the query contains redefined prefixes or qnames that use undefined prefixes.
 	 */
 	public static Map<String, String> process(ASTOperationContainer qc) throws MalformedQueryException {
+		return process(qc, null);
```
EmptyStar, line 58:
```suggestion
		return process(qc, Collections.emptyMap());
```
Passing an empty map here would make the overloaded method simpler.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/PrefixDeclProcessor.java
```diff
@@ -71,8 +86,18 @@ public static Map<String, String> process(ASTOperationContainer qc) throws Malfo
 			prefixMap.put(prefix, iri);
 		}
 
+		int preDefaultPrefixes = 0;
+
+		// insert the default prefixes if presents
+		if (customDefaultPrefixes != null && !customDefaultPrefixes.isEmpty()) {
```
EmptyStar, line 92: This `if` statement could be removed if the argument could be made non-nullable in favor of an empty map, then the following `for` loop would simple do nothing if the map were empty.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/PrefixDeclProcessor.java
```diff
@@ -71,8 +86,18 @@ public static Map<String, String> process(ASTOperationContainer qc) throws Malfo
...
+			for (Entry<String, String> defaultPrefix : customDefaultPrefixes.entrySet()) {
+				preDefaultPrefixes += insertDefaultPrefix(prefixMap, defaultPrefix.getKey(), defaultPrefix.getValue());
+			}
+		}
+
 		// insert some default prefixes (if not explicitly defined in the query)
-		final int defaultPrefixesAdded = insertDefaultPrefix(prefixMap, "rdf", RDF.NAMESPACE)
+		final int defaultPrefixesAdded = preDefaultPrefixes
```
EmptyStar, line 99: A non-null map at this point could replace `preDefaultPrefixes` with `customDefaultPrefixes.size()` rather than requiring a separate integer to track the number of custom prefixes.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java
```diff
@@ -57,6 +60,29 @@
...
+	/**
+	 * Create a new SPARQLParser.
+	 * 
+	 * @param customPrefixes the default namespaces to apply to this parser.
+	 * @throws java.lang.NullPointerException if defaultPrefixes is null.
+	 */
+	public SPARQLParser(Set<Namespace> customPrefixes) {
+		Objects.requireNonNull(customPrefixes, "defaultPrefixes can't be null!");
+		if (customPrefixes.isEmpty()) {
+			this.customPrefixes = null;
```
EmptyStar, line 74:
```suggestion
			this.customPrefixes = Collections.emptyMap();
```
An empty map would work well here.

EmptyStar (changes requested):
I'd like to make a case for using [Collections.emptyMap()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Collections.html#emptyMap()) in favor of passing around `null` for the custom prefix map. Your code would be simpler and probably safer if you could always assume the prefix map to exist. I pointed out a few bits of code that could be made more elegant with an empty map.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/PrefixDeclProcessor.java
ate47: the `insertDefaultPrefix(...)` method is returning 1 only if the prefix was added in the `prefixMap`, so we can't use `customDefaultPrefixes.size()` because maybe a custom default prefix was defined by the user in the query.

ate47:
> Thanks for the enhancement ! Perhaps the only thing left to do would be squashing the commits (see also https://rdf4j.org/documentation/developer/squashing/ on how to do this, and https://rdf4j.org/documentation/developer/merge-strategy/ why this is a probably a good idea)

Done, I was thinking it was squashed enough.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java
```diff
@@ -57,6 +59,27 @@
...
+			this.customPrefixes = Namespaces.asMap(customPrefixes);
+		}
+	}
+
+	/**
+	 * Create a new SPARQLParser without any default prefixes.
+	 */
+	public SPARQLParser() {
+		this(null);
+	}
```
hmottestad, line 82: I feel that since we have two constructors then we shouldn't allow null. Could you add a null check with `Objects.requireNonNull` in the first constructor and pass in an empty set from the second constructor?

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParserFactory.java
```diff
@@ -35,4 +38,15 @@ public QueryLanguage getQueryLanguage() {
...
+	/**
+	 * @param customDefaultPrefixes the default prefixes
+	 * @return a parser with predefined custom default prefixes.
+	 */
+	public QueryParser getParser(Set<Namespace> customDefaultPrefixes) {
+		if (customDefaultPrefixes == null || customDefaultPrefixes.isEmpty()) {
+			return singleton;
+		}
+		return new SPARQLParser(customDefaultPrefixes);
+	}
```
hmottestad, line 51: Could you add a similar null check here too?

core/queryparser/sparql/src/test/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParserTest.java
```diff
@@ -517,6 +532,222 @@ public void testGroupByProjectionHandling_renameVariableWithAggregation() {
...
+		defaultPrefixes.add(new SimpleNamespace("ex", "http://example.org/"));
+		SPARQLParser parser = new SPARQLParser(defaultPrefixes);
+
+		parser.parseQuery(queryTuple, null);
+	}
+
+	@Test
+	public void testDefaultPrefixesOverride() {
+		final String namespaceEx1 = "http://example1.org/";
+		final String namespaceEx2 = "http://example2.org/";
```
hmottestad, line 549: Thank you for adding tests!

hmottestad (changes requested):
I've added some comments. If you agree with my comments then ping me with @hmottestad when you're done and I'll merge this PR so it will be included in the upcoming 4.1.0 release.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParser.java
ate47: done

Issue #3998, opened 2022-06-19
Improve snapshot monitoring in MemoryStore


hmottestad:
Use ConcurrentHashMap with LongAdder instead of synchronization.

PR #3999, opened 2022-06-19
GH-3998 snapshot monitoring performance

hmottestad:
GitHub issue resolved: #3998 

----

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/UnionSailDataset.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/UnionSailDataset.java
@@ -1,17 +1,15 @@
 /*******************************************************************************
- * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * Copyright (c) 2022 Eclipse RDF4J contributors.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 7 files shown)


hmottestad:
## Before
```
```

## After
```
Benchmark                                               Mode  Cnt  Score   Error  Units
ConcurrentQueryBenchmark.getNamespaces                  avgt    5  1.353 ± 0.068  ms/op
ConcurrentQueryBenchmark.getNamespacesSharedConnection  avgt    5  3.831 ± 0.084  ms/op
ConcurrentQueryBenchmark.hasStatement                   avgt    5  1.718 ± 0.211  ms/op
ConcurrentQueryBenchmark.hasStatementSharedConnection   avgt    5  2.708 ± 0.119  ms/op
```

hmottestad:
Need to rework this a bit. Removing the LongAdder from the HashMap when it becomes zero is problematic because it may just temporarily be zero. If I understand things correctly then once the snapshot version increments there can be no new uses of older snapshot versions. If the snapshot version is 10, and then gets incremented to 11 we can safely assume that once the counter for version 10 reaches zero that snapshot is no longer in use.

hmottestad:
I think I'm just going to go ahead and merge this since it fixes some code that I merged in a previous PR, and not any existing code.

Issue #4010, opened 2022-06-22
LuceneSail reindex method throws Exception, but a IOException in the javadoc


ate47:
### Current Behavior

The [LuceneSail#reindex()](https://github.com/eclipse/rdf4j/blob/main/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java#L581) method is throwing an Exception, but the Javadoc is saying an IOException.

### Expected Behavior

throwing a IOException or writing Exception in the Javadoc

### Version

4.1.0-SNAPSHOT

### Are you interested in contributing a solution yourself?

Yes

PR #4011, opened 2022-06-22
GH-4010 replace Exception to SailException in LuceneSail reindex method

ate47:
GitHub issue resolved: #4010 

Instead of having Exception in the LuceneSail's reindex, I've used the SailException, I used the SailException and not the IOException like the Javadoc suggested to keep the consistency with the other sail methods.

----

 - [x] I've added tests for the changes I made (useless, it's only refactoring)

### DIFF
--- a/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
+++ b/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
@@ -576,78 +576,82 @@ public class LuceneSail extends NotifyingSailWrapper {
 	 * Starts a reindexation process of the whole sail. Basically, this will delete and add all data again, a
 	 * long-lasting process.
 	 *
-	 * @throws IOException
+	 * @throws SailException If the Sail could not be reindex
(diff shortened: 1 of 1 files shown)


core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
```diff
@@ -576,78 +576,82 @@ public void setFederatedServiceResolver(FederatedServiceResolver resolver) {
...
+	 * @throws SailException If the Sail could not be reindex
 	 */
-	public void reindex() throws Exception {
-		// clear
-		logger.info("Reindexing sail: clearing...");
-		luceneIndex.clear();
-		logger.info("Reindexing sail: adding...");
-
-		luceneIndex.begin();
+	public void reindex() throws SailException {
```
hmottestad, line 581: @jeenbroekstra I was wondering if we consider this a backwards compatible change or not? Japicmp doesn't seem to be complaining. If the change is ok with you, then I propose that we can remove the throws declaration all together since SailException is a RuntimeException.

Issue #4037, opened 2022-07-04
Error evaluating TupleQuery since 4.1.0-M1


maxstolze:
### Current Behavior

Error evaluating TupleQuery:

```
org.eclipse.rdf4j.spring.dao.exception.RDF4JSpringException: Error evaluating TupleQuery:
QueryRoot
   Distinct
      Projection
         ProjectionElemList
            ProjectionElem "id"
            ProjectionElem "f_copyOf"
            ProjectionElem "f_standardId"
            ProjectionElem "f_name"
            ProjectionElem "dateCreated"
            ProjectionElem "dateModified"
            ProjectionElem "f_private"
            ProjectionElem "f_description"
            ProjectionElem "creator"
            ProjectionElem "editor"
            ProjectionElem "f_featureKind"
            ProjectionElem "class"
            ProjectionElem "quantityKind"
            ProjectionElem "quantityKindLabel"
            ProjectionElem "unit"
            ProjectionElem "unitLabel"
            ProjectionElem "allowMultiple"
            ProjectionElem "optionId"
            ProjectionElem "optionValue"
            ProjectionElem "optionDescription"
            ProjectionElem "path"
         Extension
            Filter
               Not
                  And
                     Bound
                        Var (name=organizationId)
                     Bound
                        Var (name=projectId)
               Join
                  Join
                     LeftJoin
                        LeftJoin
                           LeftJoin
                              LeftJoin
                                 Join
                                    Join
                                       Join
                                          Join
                                             Join
                                                Join
                                                   Join
                                                      StatementPattern
                                                         Var (name=id)
                                                         Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
                                                         Var (name=_const_78de6dd_uri, value=https://mms.researchstudio.at/mms#Feature, anonymous)
                                                      StatementPattern
                                                         Var (name=id)
                                                         Var (name=_const_9763fc31_uri, value=http://schema.org/name, anonymous)
                                                         Var (name=f_name)
                                                   StatementPattern
                                                      Var (name=id)
                                                      Var (name=_const_cbf3f98a_uri, value=https://mms.researchstudio.at/mms#private, anonymous)
                                                      Var (name=f_private)
                                                StatementPattern
                                                   Var (name=id)
                                                   Var (name=_const_4af93ef4_uri, value=http://schema.org/dateCreated, anonymous)
                                                   Var (name=dateCreated)
[...]

PR #4039, opened 2022-07-04
GH-4037 fix for removing from VarNameCollector

hmottestad:
GitHub issue resolved: #4037 

 - optimized the code
 - create a new HashSet if we need to modify the set produced by the VarNameCollector

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/QueryModelNormalizerOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/QueryModelNormalizerOptimizer.java
@@ -231,13 +231,40 @@ public class QueryModelNormalizerOptimizer extends AbstractSimpleQueryModelVisit
 			leftJoin.getCondition().visit(optionalVarCollector);
 		}
 
-		Set<String> problemVars = optionalVarCollector.getVarNames();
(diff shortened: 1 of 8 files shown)


hmottestad:
Still needs tests.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/QueryModelNormalizerOptimizer.java
```diff
@@ -231,13 +231,40 @@ private boolean isWellDesigned(LeftJoin leftJoin) {
...
+					if (problemVars.contains(leftBindingName)) {
+						HashSet<String> ret = new HashSet<>(problemVars);
+						ret.removeAll(leftBindingNames);
+						return ret;
+					}
+				}
+			}
+		}
+		return problemVars;
+	}
```
hmottestad, line 265: This is the fix for the reported bug. I've also taken the liberty to check other places where we are assuming that a returned set is modifiable (which I feel is a bad assumption).

This code is covered by the tests in the SPARQL compliance module.

hmottestad:
@maxstolze does this fix work for you? I wasn't able to reproduce the issue with the query and data you provided, but the stack trace makes me confident that this PR should fix the issue.

maxstolze:
As far as I could test it, it fixes the issue. Thank you!

jeenbroekstra (approved):
Looks good to me!

Issue #4047, opened 2022-07-07
Simplify iterations that are frequently recreated to reduce itable stubbing


hmottestad:
Simplify iterations that are frequently recreated to reduce itable stubbing.

PR #4048, opened 2022-07-07
GH-4047 simplify iteratios to reduce JVM generated itable stubs

hmottestad:
GitHub issue resolved: #4047 

----

### DIFF
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
@@ -154,7 +154,7 @@ public class Var extends AbstractQueryModelNode implements ValueExpr {
 		}
 		Var var = (Var) o;
 
-if (cachedHashCode != 0 && var.cachedHashCode != 0 && cachedHashCode != var.cachedHashCode) {
+		if (cachedHashCode != 0 && var.cachedHashCode != 0 && cachedHashCode != var.cachedHashCode) {
 			return false;
 		}
 
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
@@ -153,11 +153,13 @@ public class Var extends AbstractQueryModelNode implements ValueExpr {
 			return false;
 		}
 		Var var = (Var) o;
-		if (cachedHashCode != 0 && var.cachedHashCode != 0 && cachedHashCode != var.cachedHashCode) {
+
+if (cachedHashCode != 0 && var.cachedHashCode != 0 && cachedHashCode != var.cachedHashCode) {
 			return false;
 		}
 
-		return anonymous == var.anonymous && Objects.equals(name, var.name) && Objects.equals(value, var.value);
+		return anonymous == var.anonymous && !(name == null && var.name != null || value == null && var.value != null)
+				&& Objects.equals(name, var.name) && Objects.equals(value, var.value);
 	}
 
 	@Override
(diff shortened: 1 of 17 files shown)


hmottestad:
## Before
```
Benchmark                                    Mode  Cnt   Score   Error  Units
ParallelQueryBenchmark.mixedQueriesAndReads  avgt    5  28.726 ± 0.358  ms/op
```

## After
```
Benchmark                                    Mode  Cnt   Score   Error  Units
ParallelQueryBenchmark.mixedQueriesAndReads  avgt    5  22.821 ± 0.469  ms/op
```

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
```diff
@@ -153,7 +153,9 @@ public boolean equals(Object o) {
 			return false;
 		}
 		Var var = (Var) o;
-		return anonymous == var.anonymous && Objects.equals(name, var.name) && Objects.equals(value, var.value);
+
+		return anonymous == var.anonymous && !(name == null && var.name != null || value == null && var.value != null)
+				&& Objects.equals(name, var.name) && Objects.equals(value, var.value);
```
jeenbroekstra, line 158: What makes this better? It reads to me as if it is just more complex. Doesnt `Objects.equals` take `null` into account?

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementIterator.java
```diff
@@ -156,25 +159,32 @@ public static boolean cacheAwareHasStatement(MemStatementList smallestList, MemR
...
-				// First check if we match the specified SPO, then check the context, then finally check the
-				// explicit/inferred and snapshot.
-				// Checking explicit/inferred and snapshot requires reading a volatile field, which is fairly slow and
-				// the
-				// reason we check this last.
-				matchingStatements++;
-				return statement;
+			// First check if we match the specified SPO, then check the context, then finally check the
+			// explicit/inferred and snapshot. Checking explicit/inferred and snapshot requires reading a volatile
+			// field, which is fairly slow and the reason we check this last.
```
jeenbroekstra, line 172: Nice

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementIterator.java
```diff
@@ -8,17 +8,17 @@
...
-import org.eclipse.rdf4j.common.iteration.LookAheadIteration;
 import org.eclipse.rdf4j.sail.SailException;
 
 /**
  * A StatementIterator that can iterate over a list of Statement objects. This iterator compares Resource and Literal
  * objects using the '==' operator, which is possible thanks to the extensive sharing of these objects in the
  * MemoryStore.
  */
-public class MemStatementIterator<X extends Exception> extends LookAheadIteration<MemStatement, X> {
+public class MemStatementIterator implements CloseableIteration<MemStatement, SailException> {
```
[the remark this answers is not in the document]
jeenbroekstra: If this is the case, do you see any legit use cases for `LookAheadIteration` as an (abstract) class, still? If not, should we mark it deprecated?

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/Var.java
hmottestad: In the case that two Var objects are identical with the exception of one having a Value and the other not, then we can skip the equals check for the name.

core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementIterator.java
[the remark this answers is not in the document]
hmottestad: I think that we should sacrifice DRY for performance when it comes to hot code such as the iteration code. Itable stubbing might be less of an issue when we move away from having both the Iteration and CloseableIteration interfaces. Stubbing also only happens on the first method invocation of an object, so long lived iterations will not really be affected much.

If possible I would prefer trying to create some static methods in a helper class for common use cases such as a LookAheadIteration, instead of using inheritance. I also want to play with keeping inheritance but removing polymorphic methods, essentially marking all iteration classes as final and adding abstract methods for the cases where we need to override a method.

hmottestad:
@jeenbroekstra I have deprecated most all of the iterations now.

Issue #4065, opened 2022-07-17
public API methods incorrectly marked deprecated


jeenbroekstra:
Several public API methods, for example in the `QueryResult` interface as well as `QueryResults` utility methods are marked deprecated incorrectly. Looks like this happened because of the deprecation of a lot of code related to Iterations. 

This leads to deprecation warnings in user code that really shouldn't happen. I suggest we do a bit of a cleanup, overriding such methods where necessary.

jeenbroekstra:
@hmottestad just making you aware this is an issue that we probably want to fix before releasing 4.1.0.

PR #4067, opened 2022-07-18
GH-4065 public api deprecation fixes

jeenbroekstra:
GitHub issue resolved: #4065  

- redefine method previously inherited from now deprecated classes to avoid deprecation warnings in places we don't want them. 

----

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResult.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/QueryResult.java
@@ -9,6 +9,7 @@ package org.eclipse.rdf4j.query;
 
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+import java.util.stream.Stream;
 
 import org.eclipse.rdf4j.common.iteration.CloseableIteration;
 import org.eclipse.rdf4j.common.iterator.CloseableIterationIterator;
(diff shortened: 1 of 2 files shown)


core/query/src/main/java/org/eclipse/rdf4j/query/QueryResult.java
```diff
@@ -18,11 +20,41 @@
...
 	}
 
+	/**
+	 * Returns {@code true} if the query result has more elements. (In other words, returns {@code true} if
+	 * {@link #next} would return an element rather than throwing a {@link NoSuchElementException}.)
+	 *
+	 * @return {@code true} if the iteration has more elements.
+	 * @throws QueryEvaluationException if an error occurs while executing the query.
+	 */
+	boolean hasNext() throws QueryEvaluationException;
```
hmottestad, line 38: Since QueryEvaluationException is a runtime exception would it be allowed to drop the `throws` declaration? Or would japicmp complain?

jeenbroekstra: I guess we could, but I left it in to make it clear that this is an exception type that could happen. Though perhaps having that only in the javadoc is enough.

jeenbroekstra: Actually I think (haven't tested) that we still need this because it overrides a generic `hasNext() throws X` method in the superclass.

hmottestad: That would make sense.

hmottestad:
This looks good. Thanks for realizing that the deprecation cascaded up to user level. If the `throws` declaration is required then feel free to merge this PR :)

Issue #4068, opened 2022-07-18
Unclosed iterations in the federation module


hmottestad:
Fix unclosed iterations in the federation module.

PR #4078, opened 2022-07-25
GH-4068 fix unclosed iterations

hmottestad:
GitHub issue resolved: #4068 

 - update tests to make the unclosed iterations visible
 - add missing close() and handleClose() methods
 - close iterations in case of exceptions

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
@@ -61,11 +61,11 @@ import org.slf4j.LoggerFactory;
  * Prior to evaluation various optimizations are performed, see
  * {@link org.eclipse.rdf4j.federated.optimizer.FedXOptimizer} for further details.
  * <p>
- *
+ * <p>
  * Since 4.0 FedX supports write operations using the supplied {@link WriteStrategy}, e.g. by writing to a designated
  * federation member. Note: the {@link WriteStrategy} is initialized lazily upon first access to a write operation, see
  * {@link #getWriteStrategyInternal()}.
- *
+ * <p>
  * Implementation notes: - not all methods are implemented as of now
  *
  * @author Andreas Schwarte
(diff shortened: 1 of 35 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/FederatedDescribeIteration.java
```diff
@@ -96,4 +96,13 @@ protected BindingSet convert(BindingSet sourceObject) throws QueryEvaluationExce
 			}
 		};
 	}
+
+	@Override
+	protected void handleClose() throws QueryEvaluationException {
+		try {
+			super.handleClose();
+		} finally {
+			queryInfo.close();
```
hmottestad, line 105: Not sure if the queryInfo object should be closed here or not.

tools/federation/src/test/java/org/eclipse/rdf4j/federated/FedXBaseTest.java
```diff
@@ -83,14 +81,18 @@ public static void initLogging() throws Exception {
...
 
 		String queryString = readQueryString(queryFile);
 
 		Query query = queryManager().prepareQuery(queryString);
 
 		if (query instanceof TupleQuery) {
+			// Some query results will automatically close themselves when they are exhausted. To properly test that
+			// query results are closed correctly we need to evaluate the query without retrieving any elements.
+			((TupleQuery) query).evaluate().close();
+
```
hmottestad, line 95: This is one of the two main changes to the tests that make the unclosed iterations visible.

tools/federation/src/test/java/org/eclipse/rdf4j/federated/FedXBaseTest.java
```diff
@@ -99,6 +101,10 @@ protected void execute(RepositoryConnection conn, String queryFile, String expec
 			}
 
 		} else if (query instanceof GraphQuery) {
+			// Some query results will automatically close themselves when they are exhausted. To properly test that
+			// query results are closed correctly we need to evaluate the query without retrieving any elements.
+			((GraphQuery) query).evaluate().close();
+
```
hmottestad, line 107: This is one of the two main changes to the tests that make the unclosed iterations visible.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/GraphToBindingSetConversionIteration.java
```diff
@@ -68,4 +68,8 @@ protected BindingSet convert(Statement st) {
 		return result;
 	}
 
+	@Override
+	protected void handleClose() throws QueryEvaluationException {
+		graph.close();
+	}
```
hmottestad, line 74: The graph object was not being closed, which is what was causing my original issues.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConnection.java
```diff
@@ -165,17 +162,26 @@ protected CloseableIteration<? extends BindingSet, QueryEvaluationException> eva
...
+				if (originalQuery instanceof PassThroughTupleExpr && res instanceof EmptyIteration) {
+					((PassThroughTupleExpr) originalQuery).setPassedThrough(true);
+				}
+				res = new StopRemainingExecutionsOnCloseIteration(res, queryInfo);
+				return res;
+			} catch (Throwable t) {
+				if (res != null) {
+					res.close();
+				}
+				throw t;
```
hmottestad, line 182: This is my approach to making sure that we handle any possible exceptions. In this particular case there might not be any possible exceptions, but I've previously come across constructors with null checks that would potentially fail without closing the iteration. The StopRemainingExecutionsOnCloseIteration constructor doesn't contain any null checks, but it might in the future.

aschwarte10:
Thanks for looking into the non-closed iterations. 

I did not yet have time to look into the PR in detail (will only be able to do this by Wednesday). From high level skimming the changes make sense. Regarding whether the queryInfo needs to be closed as well, I would need to double check - can't remember this from the back of my head.

May I ask in which scenario you discovered the unclosed connections? Or was it by going through the code.

Thanks for also maintaining and cleaning the code - I guess this has not been touched for a while :)

hmottestad:
> May I ask in which scenario you discovered the unclosed connections? Or was it by going through the code.

The AbstractSailConnection used to track all iterations and close them automatically when the connection was closed. I found this to be a bit of a resource hog, so instead I'm using a Cleaner to close them during GC. Since this doesn't guarantee that iterations are actually closed in a timely manner I've chosen to implement a counter of active iterations and then throw an exception if there are any open iterations when the connection is closed. I then ran into issues when integrating RDF4J 4.1.0-M1 with GraphDB where a test was complaining about my new exception.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/QueryResultIteration.java
```diff
@@ -76,8 +76,15 @@ public void remove() throws QueryEvaluationException {
...
 	protected void handleClose() throws QueryEvaluationException {
-		inner.close();
-		abortQuery();
+		try {
+			inner.close();
+		} finally {
+			try {
+				abortQuery();
+			} finally {
+				queryInfo.close();
```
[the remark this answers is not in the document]
aschwarte10: Calling `queryInfo.close()` here does not harm, but also is not required. `org.eclipse.rdf4j.federated.QueryManager.abortQuery(QueryInfo)` (which is called from the `abortQuery()` invocation) takes care for closing it as well

Note that queryInfo.close() should be safe to be called multiple times

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/FederatedDescribeIteration.java
aschwarte10: I think here it is good to have. But even if not: I think there is a very much outer place for error handling of all query types which makes sure to abort in case of any errors

aschwarte10 (commented):
I have scanned the PR twice and did not see any obvious issues. I think this is a good way ahead and I am good to go once this is ready and all tests are passing.

Would it be possible to have another 4.1 Milestone build after this one is merged? Error handling in federation is quite difficult and even subtle changes may cause issues at runtime. With a milestone build we would be able to test this alongside our development and have even more confidence.

Thanks @hmottestad for taking the efforts 👍

hmottestad:
> I have scanned the PR twice and did not see any obvious issues. I think this is a good way ahead and I am good to go once this is ready and all tests are passing.
> 
> Would it be possible to have another 4.1 Milestone build after this one is merged? Error handling in federation is quite difficult and even subtle changes may cause issues at runtime. With a milestone build we would be able to test this alongside our development and have even more confidence.
> 
> Thanks @hmottestad for taking the efforts 👍

I'm hoping to not have to create another milestone build. It'll still be at least another week before 4.1 is out the door, maybe you can start testing the snapshot build? If you find any critical issues we can push the release date.

Issue #4079, opened 2022-07-25
Use Export/Version on package-info to describe the OSGi meta-data


laeubi:
Thanks for submitting this @stbischof. Could you please also log an issue in our tracker that describes this enhancement, and link it to your PR? Also, given that this is an enhancement rather than a bug fix, if would be good if you could rebase this against the develop branch.

_Originally posted by @jeenbroekstra in https://github.com/eclipse/rdf4j/issues/3736#issuecomment-1073126062_

jeenbroekstra:
@laeubi can you give some background on what this improvement actually does? I don't know a lot about a OSGi - how does this improvement benefit developers who use OSGi?

stbischof:
@laeubi maybe you should copy the MANIFEST.MF file that is generated here. There you will see that we mainly generate manifest-entrys that describe whick packages are exported (for externat use). without this enty an osgi framework will not be able to load the jar and/or gives other classes that import rdf4j access to not exported classes..

laeubi:
I have recently seen https://github.com/eclipse/rdf4j/commit/eb239114cc33fab91baa67007ae06317ab9416cf (revert incompatible changes to interface) The version annotations, together with BND-Baseline plugin can be used to enforce [Semantic Versioning 2.0.0](https://semver.org), so maybe this would be interesting for RDF4j even though it does not use OSGi directly (yet)?

jeenbroekstra:
> I have recently seen [eb23911](https://github.com/eclipse/rdf4j/commit/eb239114cc33fab91baa67007ae06317ab9416cf) (revert incompatible changes to interface) The version annotations, together with BND-Baseline plugin can be used to enforce [Semantic Versioning 2.0.0](https://semver.org), so maybe this would be interesting for RDF4j even though it does not use OSGi directly (yet)?

We actually use the japicmp maven plugin to verify semver compliance in our code base. 

I wouldn't be against using OSGi for this (and other) purposes, but nobody has ever been able to satisfactorily explain to me what OSGi is _for_, and what benefits adopting it would give us. Then again the problem has always been that there's really no one with OSGi expertise on our core team, so if you're interested in being our resident expert, I'd love to hear more :)

PR #3736, opened 2022-03-15
Add better OSGi support on each module/jar

stbischof:
* use bnd to generate manifest
* define Exported packages using @Export in package-info.java

Signed-off-by: Stefan Bischof <stbischof@bipolis.org>

----
PR Author Checklist (see the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md) for more details):

 - [x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [ ] I've added tests for the changes I made
 - [ ] I've applied [code formatting](https://github.com/eclipse/rdf4j/blob/main/.github/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)
 - [x] I've [squashed](https://rdf4j.org/documentation/developer/squashing) my commits where necessary 
 - [ ] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change

PR #4080, opened 2022-07-25
Add better OSGi support on each module/jar

laeubi:
GitHub issue resolved: #4079

This is a rebase on devlop of https://github.com/eclipse/rdf4j/pull/3736

FYI @stbischof
----

### DIFF
--- a/core/common/io/src/main/java/org/eclipse/rdf4j/common/io/package-info.java
+++ b/core/common/io/src/main/java/org/eclipse/rdf4j/common/io/package-info.java
@@ -1,0 +1,8 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors. All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Distribution License v1.0 which accompanies this distribution, and is
+ * available at http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+
+@org.osgi.annotation.bundle.Export
+package org.eclipse.rdf4j.common.io;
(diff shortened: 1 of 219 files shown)


core/common/io/src/main/java/org/eclipse/rdf4j/common/io/package-info.java
```diff
@@ -0,0 +1,8 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors. All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Distribution License v1.0 which accompanies this distribution, and is
+ * available at http://www.eclipse.org/org/documents/edl-v10.php.
```
jeenbroekstra, line 4: It looks like the copyright header was accidentally malformed in several places - it's missing newlines.

It should be like this:

```
 * Copyright (c) 2022 Eclipse RDF4J contributors.
```

However, given that these are `package-info.java` files, I would be fine with just removing these copyright headers. @hmottestad I remember something being said about copyright headers in the last Eclipse review, were package-info.java files included in that comment?

jeenbroekstra (commented):
Thanks for picking this up again @laeubi. I can't really judge the usefulness of this (not knowing a lot about OSGi) but apart from a minor editorial issue withs some copyright headers this looks fine to me!

I note there are test failures, but it seems unlikely that your changes would have caused those. I'll try giving it a bump.

core/common/io/src/main/java/org/eclipse/rdf4j/common/io/package-info.java
hmottestad: That comment was about some test files. I think we should stick with copyright headers in all *.java files.

stbischof: It is very useful for us because we could now use rdf4j in our platform. The Platform is OSGi based and so we couldn't use RDF4J without the proper Manifest-information. Also if you want to use it in eclipse plugins you would need it because this is also osgi based.

laeubi:
I further dived into this and noticed that there is already a 
```
<dependency>
    <artifactId>rdf4j-runtime-osgi</artifactId>
    <version>4.1.0</version>
</dependency>
```
artifact, that currently bundles everything into one big OSGi-jar. This is not optimal, but given the many RDF4J modules and its different packing (jar/war) might be a way to start instead of adding OSGi infos to each artifact individually.

Issue #4091, opened 2022-07-30
Release review


hmottestad:
Todo from the review
- [x] CONTRIBUTING file
- [x] Missing copyright headers
- [x] Eclipse foundation terms of use
- [x] SPDX

PR #4093, opened 2022-07-30
Create CONTRIBUTING.md

hmottestad:
Relates to https://github.com/eclipse/rdf4j/issues/4091


hmottestad:
We were given the option of adding this file or updating our readme file with the terms of use. See https://github.com/eclipse/rdf4j/pull/4094

PR #4105, opened 2022-08-02
Remove local maven config file and add .mvn folder to gitignore

hmottestad:
Remove the local maven config file and add the .mvn folder to gitignore.

----


jeenbroekstra:
I'm on mobile so can't check, but I seem to remember we added this for a specific reason:  the way we refer to the project root fails in Jenkins or in GitHub when this is not defined, I think. I'll try and look up the details in the commit history.

jeenbroekstra:
So we introduced this file as part of the PR that introduced automated code formatting checks (see https://github.com/eclipse/rdf4j/pull/1340 ) . At the time this was done because Jenkins was having trouble with the relative path reference we use to link to the code formatting file. I can't remember if adding this was actually the fix, or if this is merely something I used to figure out the version of maven used in Jenkins.

In any case: if you want to remove it, I'd recommend running a test build on Jenkins to make sure the code formatter part of the build doesn't break there.

Issue #4107, opened 2022-08-04
Unclosed iteration in fedx


hmottestad:
There are still unclosed iterations in fedx.

PR #4108, opened 2022-08-04
GH-4107 fix unclosed iteration fedx

hmottestad:
GitHub issue resolved: #4107 

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelTask.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelTask.java
@@ -44,4 +44,11 @@ public interface ParallelTask<T> {
 	 * Optional implementation to cancel this task on a best effort basis
 	 */
 	void cancel();
+
+	/**
+	 * Optional implementation to close this task on a best effort basis.
+	 */
+	default void close() {
+
+	}
 }
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/union/ControlledWorkerUnion.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/union/ControlledWorkerUnion.java
@@ -15,6 +15,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.eclipse.rdf4j.federated.evaluation.concurrent.ControlledWorkerScheduler;
 import org.eclipse.rdf4j.federated.structures.QueryInfo;
+import org.eclipse.rdf4j.query.QueryEvaluationException;
 
 /**
  * Execution of union tasks with {@link ControlledWorkerScheduler}. Tasks can be added using the provided functions.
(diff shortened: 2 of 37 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/concurrent/ParallelTask.java
```diff
@@ -44,4 +44,11 @@ default QueryInfo getQueryInfo() {
 	 * Optional implementation to cancel this task on a best effort basis
 	 */
 	void cancel();
+
+	/**
+	 * Optional implementation to close this task on a best effort basis.
+	 */
+	default void close() {
+
+	}
```
hmottestad, line 53: I've added a new method to the interface to close a task. Closing a task is similar to cancelling a task except that if a closed task is run then it just becomes a no-op instead of throwing an exception like it does if it is cancelled.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/union/WorkerUnionBase.java
```diff
@@ -43,4 +43,15 @@ public void addTask(ParallelTask<T> task) {
...
+	@Override
+	public void handleClose() throws QueryEvaluationException {
+		try {
+			for (ParallelTask<T> task : tasks) {
+				task.close();
+			}
+		} finally {
+			super.handleClose();
+		}
+	}
```
hmottestad, line 56: We now try to close all the tasks, since it could be that we have been closed before all the tasks were completed (or even started). There is obviously no guarantee that started tasks can be stopped, so it's all best effort.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/union/ControlledWorkerUnion.java
```diff
@@ -61,4 +61,15 @@ public void toss(Exception e) {
...
+	@Override
+	public void handleClose() throws QueryEvaluationException {
+		try {
+			super.handleClose();
+		} finally {
+			// signal the phaser to close (if currently being blocked)
+			phaser.forceTermination();
+		}
+	}
+
```
hmottestad, line 74: This was the initial issue. The union() method would deadlock because the phaser was never terminated if the close() method had been called before all the tasks had completed.

There are three other classes that use the same approach with phasers and all those have this exact code.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/exception/ExceptionUtil.java
```diff
@@ -60,6 +60,10 @@ public class ExceptionUtil {
 	public static QueryEvaluationException traceExceptionSource(Endpoint endpoint, Throwable ex,
 			String additionalInfo) {
 
+		if (ex instanceof InterruptedException) {
+			Thread.currentThread().interrupt();
+		}
```
hmottestad, line 65: I've tried to improve interrupt handling. Other code in RDF4J will typically re-interrupt the current thread if it needs to swallow or wrap an InterruptedException. This means that other code down the line is still able to abort their current execution, which was what the interrupt was initially signalling.

PR #4116, opened 2022-08-15
Update Apache Felix Version

laeubi:
RDF4J currently uses an old version of the apache felix plugin, this is just a dependency update for this.

By the way maybe you want enable https://github.com/dependabot for this repository to get such notifications automatically?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -708,7 +708,7 @@
 				<plugin>
 					<groupId>org.apache.felix</groupId>
 					<artifactId>maven-bundle-plugin</artifactId>
-					<version>5.1.1</version>
+					<version>5.1.8</version>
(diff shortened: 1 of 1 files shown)


jeenbroekstra:
> RDF4J currently uses an old version of the apache felix plugin, this is just a dependency update for this.

Thanks for spotting this. We typically do (patch) dependency bumps as part of patch releases, so these kinds of PRs can be targeted at the main branch. 

> By the way maybe you want enable https://github.com/dependabot for this repository to get such notifications automatically?

Dependabot _is_ enabled. I'm not entirely sure why it didn't pick this up.

jeenbroekstra (approved):

Issue #4112, opened 2022-08-09
Construct query creating RDFstar statements in memory store does not work anymore with RDF4J 4.1


jetztgradnet:
### Current Behavior

Running a `CONSTRUCT` query in a memory store fails to create RDFstar statements with RDF4J 4.1 with the following error:
```
java.lang.IllegalArgumentException: Node is not a child node: Var (name=s)
```

The example query:
```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX person: <http://example.com/person/>
PREFIX org: <http://example.com/org/>
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
CONSTRUCT {
  ?s ?p ?o.
  << ?s ?p ?o >> <http://example.com/certainty> ?cert.
  << ?s ?p ?o >> <http://example.com/certaintyDeviation> ?certDiv.
  << person:alice foaf:knows person:bob >> <http://example.com/observedBy> person:mike.
}
WHERE {
  {
    SELECT ?s ?p ?o ?cert ?certDiv WHERE {
      VALUES (?s ?p ?o ?cert ?certDiv) {
        (person:alice foaf:knows person:bob "1.0"^^xsd:decimal 0 )
        (person:alice foaf:knows person:carol "0.3"^^xsd:decimal "0.1"^^xsd:decimal)
        (person:carol foaf:knows person:mike "0.7"^^xsd:decimal "0.2"^^xsd:decimal)
        (person:mike foaf:knows person:carol "0.1"^^xsd:decimal UNDEF)
        (person:bob foaf:knows person:carol "0.8"^^xsd:decimal UNDEF)
        (person:alice foaf:knows person:mike "0.6"^^xsd:decimal "0.1"^^xsd:decimal)
        (person:alice foaf:member org:W3C UNDEF UNDEF)
        (person:mike foaf:member org:W3C UNDEF UNDEF)
        (person:alice rdf:type foaf:Person UNDEF UNDEF)
        (person:carol rdf:type foaf:Person UNDEF UNDEF)
        (person:mike rdf:type foaf:Person UNDEF UNDEF)
        (person:bob rdf:type foaf:Person UNDEF UNDEF)
        (org:W3C rdf:type foaf:Organization UNDEF UNDEF)
[...]

PR #4117, opened 2022-08-17
GH-4112 fix query regression for CONSTRUCT query with RDF*

hmottestad:
GitHub issue resolved: #4112 

 - add test
 - add missing replaceChild method (which was the main issue)
 - remove the parent reference cleaner from the standard query optimizers
 - add a parent reference checker to the standard query optimizers if assertions are enabled (so that our tests fail)
 - prevent sharing or reuse of a Var object by checking (assert) that the parent ref was null before setting it

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/ParentReferenceChecker.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/ParentReferenceChecker.java
@@ -1,0 +1,72 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+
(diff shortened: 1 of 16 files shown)


hmottestad:
TODO:

- [ ] Check performance

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/ParentReferenceChecker.java
```diff
@@ -0,0 +1,72 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
```
jeenbroekstra, line 8: Shouldn't this be the new header with the tag in it?

jeenbroekstra (approved):

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/ParentReferenceChecker.java
hmottestad: Yes it should.

Issue #4126, opened 2022-08-24
Clear defaultGraph clears all graphs


maxstolze:
### Current Behavior

We used `con.clear((org.eclipse.rdf4j.model.Resource) null);` to clear the defaultGraph.
Now this clears all graphs in the repository

Only happens with remote repositories. No problem with an in memory store.
### Expected Behavior

Only clear defaultGraph.

### Steps To Reproduce

Using RDF4J 4.1.0 with:
- GraphDB 10.0.0, 10.0.1 & 10.0.2
- Native- & MemoryStore on RDF4J Server 4.1.0

Clear with: `con.clear((org.eclipse.rdf4j.model.Resource) null);`

### Version

4.1.0

hmottestad:
@maxstolze did this work before?

fkleedorfer:
I am pretty certain that the follwing used to work prior to 4.x
```
con.clear(new org.eclipse.rdf4j.model.Resource[] {null});
```

Not so sure about
```
con.clear((org.eclipse.rdf4j.model.Resource) null);
```
(it's a suggestion made by @jeenbroekstra in gitter, which was the last thing we tried before we gave up and it made its way into the issue by accident)

We don't really care about the regression, if it even is one, but we cannot seem to figure out how to clear just the default graph in a remote repo

hmottestad:
I've tried both myself without getting it to work either. I don't think there actually is any difference between the two method calls, they both resolve to an array with a null element. 

I've tried 4.1.0, 4.0.4, 4.0.0 and 3.7.7. None of these work.

If you could find out which version that this used to work for it would be very helpful for me. Testing this is very time consuming for me. I need to know if this used to work before so that I can see what has changed in the code.

fkleedorfer:
When we developed code using this, we were using 3.7.0. We switched directly to 4.0.4 from that. 

I am pretty sure it used to work with remote repos, but I cannot guarantee it. There is a small chance that we overlooked the problem. (We've used this code in the preparation of test cases since January 2022, and in our setup, we can run the tests against a remote repo only manually on the developer machine - the fully automated tests run agains an in-memory db. However, we do run them against the remote repo manually every now and then, and I do not recall having this problem until we switched to 4.x. - recently)

jeenbroekstra:
[...]

PR #4127, opened 2022-08-26
GH-4126 clear default graph remote repo bug

hmottestad:
GitHub issue resolved: #4126 

----

### DIFF
--- a/core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPRepositoryConnection.java
+++ b/core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPRepositoryConnection.java
@@ -22,6 +22,7 @@ import java.io.Reader;
 import java.net.URL;
 import java.net.URLConnection;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -52,6 +53,7 @@ import org.eclipse.rdf4j.model.Statement;
 import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.model.impl.LinkedHashModel;
 import org.eclipse.rdf4j.model.impl.SimpleNamespace;
+import org.eclipse.rdf4j.model.vocabulary.RDF4J;
 import org.eclipse.rdf4j.model.vocabulary.SESAME;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.BooleanQuery;
@@ -85,10 +87,10 @@ import org.eclipse.rdf4j.rio.helpers.StatementCollector;
  * specific RepositoryException subclasses UnautorizedException and NotAllowedException, the semantics of which are
  * defined by the HTTP protocol.
  *
- * @see org.eclipse.rdf4j.http.protocol.UnauthorizedException
- * @see org.eclipse.rdf4j.http.protocol.NotAllowedException
  * @author Arjohn Kampman
  * @author Herko ter Horst
+ * @see org.eclipse.rdf4j.http.protocol.UnauthorizedException
+ * @see org.eclipse.rdf4j.http.protocol.NotAllowedException
  */
 class HTTPRepositoryConnection extends AbstractRepositoryConnection implements HttpClientDependent {
 
@@ -510,12 +512,13 @@ class HTTPRepositoryConnection extends AbstractRepositoryConnection implements H
 	@Override
 	public void add(Resource subject, IRI predicate, Value object, Resource... contexts) throws RepositoryException {
 		if (!isActive()) {
-			logger.debug("adding statement directly: {} {} {} {}",
-					new Object[] { subject, predicate, object, contexts });
+			if (logger.isDebugEnabled()) {
+				logger.debug("adding statement directly: {} {} {} {}", subject, predicate, object, contexts);
+			}
 			// operation is not part of a transaction - just send directly
 			Objects.requireNonNull(contexts,
 					"contexts argument may not be null; either the value should be cast to Resource or an empty array should be supplied");
-			final Model m = new LinkedHashModel();
+			Model m = new LinkedHashModel();
 			m.add(subject, predicate, object, contexts);
 			addModel(m);
 		} else {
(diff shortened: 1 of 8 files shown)


core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPRepositoryConnection.java
```diff
@@ -658,6 +660,10 @@ protected void removeWithoutCommit(Resource subject, IRI predicate, Value object
 		if (object == null) {
 			object = SESAME.WILDCARD;
 		}
+		for (int i = 0; i < contexts.length; i++) {
+			if (contexts[i] == null)
+				contexts[i] = RDF4J.NIL;
+		}
```
jeenbroekstra, line 666: This looks like the right fix to me (I could have sworn we have a util function to do exactly this swap, but can't find it immediately).

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -472,28 +473,40 @@ public WildcardRDFRemover(RepositoryConnection conn) {
...
+			Thread.dumpStack();
 			Resource subject = SESAME.WILDCARD.equals(st.getSubject()) ? null : st.getSubject();
 			IRI predicate = SESAME.WILDCARD.equals(st.getPredicate()) ? null : st.getPredicate();
 			Value object = SESAME.WILDCARD.equals(st.getObject()) ? null : st.getObject();
 
-			// use the RepositoryConnection.clear operation if we're removing
-			// all statements
-			final boolean clearAllTriples = subject == null && predicate == null && object == null;
+			// use the RepositoryConnection.clear operation if we're removing all statements
+			boolean clearAllTriples = subject == null && predicate == null && object == null;
```
jeenbroekstra, line 483: It's not really that important but is there a reason you remove final keywords on these local vars? I tend to deliberately put these in as it clarifies intent that these are constants.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
```diff
@@ -472,28 +473,40 @@ public WildcardRDFRemover(RepositoryConnection conn) {
...
-						conn.remove(subject, predicate, object);
+						System.out.println("conn.remove(subject, predicate, object, ((Resource) null));");
+						conn.remove(subject, predicate, object, ((Resource) null));
 					}
+				} else {
+					assert st.getContext() != null;
+					System.out.println("conn.remove(" + subject + ", " + predicate + ", " + object + ", "
+							+ st.getContext() + ");");
+					conn.remove(subject, predicate, object, st.getContext());
 				}
```
jeenbroekstra, line 510: I don't really like the complexity of the branching logic here - so many nested if-elses and boolean flags. I think if we separate out the logic that determines what the value of the context param should be into a separate method, we can simplify this a bit.

jeenbroekstra (commented):
Good stuff. Also good call introducing a new integration test.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
hmottestad: For short methods I feel that they add clutter to the code. I usually only use constants for class fields and not for code within a method. If the method code is long and hard to understand then it can help to add a final modifier when there are multiple variables where some should not be modified while others are intended to be modified later in the code.

In this case there are also 3 variables above this line that should have be declared as final, if we wanted to be concise.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
hmottestad: Yeah. This is very much work-in-progress. Since I can't debug the code directly I've made it particularly explicit and verbose because I don't trust myself not to make stupid mistakes :P 

I'll see how I can simplify it once I've got all the bugs ironed out.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/transaction/Transaction.java
jeenbroekstra: Java could do with a `val` keyword...

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
```diff
@@ -176,12 +176,20 @@ boolean hasApproved(Resource subj, IRI pred, Value obj, Resource[] contexts) {
...
 		try {
+			if (deprecatedContexts != null) {
+				for (Resource context : contexts) {
+					if (deprecatedContexts.contains(context)) {
+						return true;
+					}
+				}
+			}
+
 			return deprecated.contains(subj, pred, obj, contexts);
```
hmottestad, line 193: There was actually a bug in the Changeset code where it didn't handle deprecating a context correctly. I think this was something that I introduced in 4.1.0 with my Changeset optimizations.

hmottestad:
@maxstolze could you test out this branch to see if it fixes your issue?

maxstolze:
It is still deleting all the graphs.

hmottestad:
> It is still deleting all the graphs.

Could you tell me what you tested so I can try to reproduce it yourself?

maxstolze:
Having a repository with default graph, SHACL constrains graph and two ontologie graphs and after the `con.clear(new org.eclipse.rdf4j.model.Resource[] {null});`, the default graph is cleared, SHACL constrains stay untouched, but ontologie graphs are falsely removed.

jeenbroekstra:
@maxstolze did you update both your client _and_ your server with this branch? Not sure if it was clear but the fix involves changes at both ends of the connection...

maxstolze:
Will double check it now.

maxstolze:
@jeenbroekstra @hmottestad my bad, didn't update everything properly. Sorry for the confusion. 
It is working now! So Issue fixed on our side.

hmottestad:
This should also be back ported to  RDF4J 4.0.x.

Issue #4131, opened 2022-08-28
Support custom aggregate functions


hmottestad:
SPARQL supports aggregate functions like AVG, SUM, COUNT, MIN, MAX and a few more. These are fairly standard for databases and are also supported in relational databases. 

In statistics there are quite a few more aggregate functions that would be useful for our users. Median and standard deviation would be two great additions. 

Our SPARQL Abstract Syntax Tree (AST) already supports custom functions defined with a prefix, the only lacking language feature would be to allow DISTINCT within a custom function. 

Our query processor also supports custom functions [1]. This feature would need to be extended to support custom aggregate functions. 

Jena supports custom aggregate functions [2] and the documentation states that this is allowed by the SPARQL 1.1 spec. 

Task:
- [ ] Extend the AST to support custom aggregate functions in all places that aggregate functions are allowed today. Essentially a new type of custom function that also supports DISTINCT. 
- [ ] Add support for user defined aggregate functions in a similar fashion to how we currently support (non-aggregate) user defined functions
- [ ] Extend the documentation to describe how users can implement their own aggregate functions

1: https://rdf4j.org/documentation/tutorials/custom-sparql-functions/

2: https://jena.apache.org/documentation/query/custom_aggregates.html

PR #4132, opened 2022-08-29
GH-4131 Introduce custom AggregateFunctionCall abstraction

tokovach:
GitHub issue resolved: #4131

Introduced custom AggregateFunctionCall abstraction that allows custom aggregate functions to be introduced through CustomAggregateFunctionRegistry. 

Updated sparql.jjt compile file and subsequently SyntaxTreeBuilder to allow distinct to be parsed for custom aggregate function calls.

Moved around abstract class for aggregate functions (Aggregate and renamed it to AggregateFunction) as well as the AggregateCollector in order to be able to define custom aggregate functions (whose signature can be the same as a standard function call) through a custom function registry (where they are validated inside the TupleExprBuilder).


tokovach:
Will add unit & functional tests covering custom AggregateCustomFunction execution.

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/JJTSyntaxTreeBuilderState.java
```diff
@@ -5,10 +5,8 @@
  * are made available under the terms of the Eclipse Distribution License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/org/documents/edl-v10.php.
- *
- * SPDX-License-Identifier: BSD-3-Clause
- *******************************************************************************/
-/* Generated By:JavaCC: Do not edit this line. JJTSyntaxTreeBuilderState.java Version 7.0.11 */
+ ******************************************************************************/
+/* Generated By:JavaCC: Do not edit this line. JJTSyntaxTreeBuilderState.java Version 5.0 */
```
jeenbroekstra, line 9: Looks like you used an older version of JavaCC than the one we most recently upgraded to - worth rerunning with the newer version?

jeenbroekstra (commented):
I've only given this a brief first glance over, but looks really solid, nice work!

core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/ast/JJTSyntaxTreeBuilderState.java
tokovach: Great, I will rerun it with a newer version.

hmottestad:
## Before
```
```

## After
```
Benchmark                                        Mode  Cnt   Score   Error  Units
SPARQLParseBenchmark.complexPathExpressionQuery  avgt    5  21.765 ± 1.686  ms/op
SPARQLParseBenchmark.selectValuesQuery           avgt    5  20.218 ± 0.265  ms/op
```

hmottestad:
I'm going to follow up on the performance regression in a separate PR.

hmottestad:
Just realised that I merged this into main instead of develop :(

Issue #4134, opened 2022-08-30
Test LmdbStore with the ShaclSail


hmottestad:
> The use of LMDB should allow the LmdbStore to scale better than the NativeStore. When testing the LmdbStore with the ShaclSail it seems that the NativeStore is often faster for smaller datasets when faced with a concurrent workload of small queries. We should try to improve the performance for these workloads by profiling the code and testing out new approaches.

The performance isn't a problem anymore, so I've updated this issue to just simply add the tests that I created.

hmottestad:
That was in fact one issue I found. It's not so much the quality of the statistics but rather that it take a lot of time to calculate them because of the read locks and IO. Adding a cache with a very short eviction helped in my testing.

kenwenzel:
Maybe we could also have a look on zetasketch https://github.com/google/zetasketch

kenwenzel:
https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html

Issue #3982, opened 2022-06-10
LMDB query optimization is wrong for particular queries/data sets


kenwenzel:
### Current Behavior

As discussed in this conversation
https://github.com/eclipse/rdf4j/discussions/3953
the LMDB store seems to compute wrong cardinalities for particular queries and/or data sets resulting in unacceptable query execution times.

### Expected Behavior

The query optimizer should deliver the same results for the LMDB store as for the MemoryStore or NativeStore.

### Version

4.0.1

### Are you interested in contributing a solution yourself?

Yes

PR #4137, opened 2022-09-01
GH-4134 Test LmdbStore with the ShaclSail

hmottestad:
GitHub issue resolved: #4134 

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/ParentReferenceChecker.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/ParentReferenceChecker.java
@@ -56,7 +56,7 @@ public class ParentReferenceChecker implements QueryOptimizer {
 		}
 
 		verifySerializable(tupleExpr);
-		tupleExpr.visit(new ParentCheckingVisitor());
+		tupleExpr.visit(new ParentCheckingVisitor(ParentReferenceChecker.this.previousOptimizerInPipeline));
 	}
 
 	private void verifySerializable(QueryModelNode tupleExpr) {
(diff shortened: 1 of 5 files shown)


hmottestad:
@kenwenzel here are some optimizations I'm testing out.

kenwenzel:
Looks good. I also thought about integrating the estimate functions of libmdbx into LMDB:
https://libmdbx.dqdkfa.ru/group__c__rqest.html

It should be rather easy to patch LMDB but that would also implicate that we have to maintain a patched version of the LWJGL bindings...

hmottestad:
> Looks good. I also thought about integrating the estimate functions of libmdbx into LMDB: https://libmdbx.dqdkfa.ru/group__c__rqest.html
> 
> It should be rather easy to patch LMDB but that would also implicate that we have to maintain a patched version of the LWJGL bindings...

Do you think it would be possible to contribute it to LMDB instead?

Reading through he LMDB documentation I saw that there should be some cursor operations to get the first and last page, if the database is ordered. Maybe there is some way to estimate the size based on the addresses of the first and last page?

I remember you mentioned that you were reading a max of 1000 elements from the cursor when estimating the cardinality, is that still the case? Maybe we could do some optimized caching where we evict cardinality results based on how much resources it took to calculate them. So the cardinality for `null rdf:type null null` would be cached much longer than `:peter rdf:type null null`.

There is still the option of using the code from the ExtensibleStore which uses Hyper Log Log. I think the biggest issue with that code is that it requires refreshing if there is a lot of churn (data added and removed and then added again). During that refresh cycle it needs to read all the data from the database which would probably block all writers.

kenwenzel:
> > Looks good. I also thought about integrating the estimate functions of libmdbx into LMDB: https://libmdbx.dqdkfa.ru/group__c__rqest.html
> > It should be rather easy to patch LMDB but that would also implicate that we have to maintain a patched version of the LWJGL bindings...
> 
> Do you think it would be possible to contribute it to LMDB instead?

I've already posted a question on the OpenLDAP mailing list where I asked them to implement the feature. Unfortunately, I've got no response so far. But maybe I should offer my support regarding this topic...

> 
> Reading through he LMDB documentation I saw that there should be some cursor operations to get the first and last page, if the database is ordered. Maybe there is some way to estimate the size based on the addresses of the first and last page?

This is not possible as the keys are ordered but the b-tree pages are just randomly distributed.

> I remember you mentioned that you were reading a max of 1000 elements from the cursor when estimating the cardinality, is that still the case? Maybe we could do some optimized caching where we evict cardinality results based on how much resources it took to calculate them. So the cardinality for `null rdf:type null null` would be cached much longer than `:peter rdf:type null null`.

Yes, that is still the case. I read a fixed number of entries to determine the medium distance of the keys. Then I skip to the last key and try to extrapolate the whole number based on the distance between the keys. I'm currently working on optimizing this a bit to fix #3982.

> There is still the option of using the code from the ExtensibleStore which uses Hyper Log Log. I think the biggest issue with that code is that it requires refreshing if there is a lot of churn (data added and removed and then added again). During that refresh cycle it needs to read all the data from the database which would probably block all writers.

I also thought about implementing this but it would also require a considerable amount of storage space.

kenwenzel:
FYI - I am also working on improving the cardinality computation itself. Instead of just using the first `n` rows as samples to estimate the cardinality, I divide the whole search space into `b` buckets and then I use samples from these buckets for the estimation. I will probably push a PR this week.

kenwenzel:
@hmottestad I've tested the cardinality cache but can't see any improvments in our standard query benchmarks. Does this only pay off when using multi-threaded access?

kenwenzel:
@hmottestad Do you have any benchmarks with and without the cache?

hmottestad:
> @hmottestad Do you have any benchmarks with and without the cache?

No. I need to create some LMDB + SHACL benchmarks.

hmottestad:
I've looked into this branch again and I can't reproduce the initial performance issues. So I've modified the branch to only add the tests and benchmark.

Issue #3421, opened 2021-11-12
Background parsers may hang if the foreground task disapears


JervenBolleman:
### Problem description

When running big federated queries the service http work is done on a background thread. If the foreground task disapears e.g. due to some exception or timeout being hit. Then the background thread may for ever try to offer a result to the shared queue. 
Blocking critical resources.

### Preferred solution

In the offer loop there should be a check if the foreground task is still available. I propose passing a [WeakReference](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/WeakReference.html) to the caller. When this WeakReference is empty we know the foreground task is no longer available and that the Background task can start offering things to the queue.

### Are you interested in contributing a solution yourself?

Yes

PR #4138, opened 2022-09-01
GH-3421 use cleaner instead of weak reference

hmottestad:
GitHub issue resolved: #3421 

 - The use of WeakReference was buggy because it was linked to objects that did not necessarily stick around for the entire time that the iteration was in use. 
 - Swapped out the main use case for the WeakReference with Java 9 Cleaner.
 - Due to our maven structure I had to duplicate some code.
 - Deprecated as many methods as I could and cleaned up our code to pass null instead of a WeakReference together with an assertion down the line to make sure we always pass in null.

----

### DIFF
--- a/core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/QueueIteration.java
+++ b/core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/QueueIteration.java
@@ -35,13 +35,24 @@ public abstract class QueueIteration<E, T extends Exception> extends LookAheadIt
 
 	private final Queue<Exception> exceptions = new ConcurrentLinkedQueue<>();
 
-	private final WeakReference<?> callerRef;
-
 	/**
 	 * Creates an <var>QueueIteration</var> with the given (fixed) capacity and default access policy.
 	 *
 	 * @param capacity the capacity of this queue
 	 */
+	protected QueueIteration(int capacity) {
+		this(capacity, false);
+	}
+
+	/**
+	 * Creates an <var>QueueIteration</var> with the given (fixed) capacity and default access policy.
+	 *
+	 * @param capacity the capacity of this queue
+	 * @deprecated WeakReference is no longer supported as a way to automatically close this iteration. The recommended
+	 *             approach to automatically closing an iteration on garbage collection is to use a
+	 *             {@link java.lang.ref.Cleaner}.
+	 */
+	@Deprecated(since = "4.1.2", forRemoval = true)
 	protected QueueIteration(int capacity, WeakReference<?> callerRef) {
 		this(capacity, false, callerRef);
 	}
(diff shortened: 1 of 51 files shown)


core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPTupleQuery.java
```diff
@@ -48,8 +48,11 @@ public TupleQueryResult evaluate() throws QueryEvaluationException {
...
 			conn.flushTransactionState(Protocol.Action.QUERY);
+
+			WeakReference<HTTPTupleQuery> callerRef = new WeakReference<>(this);
+
 			return client.sendTupleQuery(queryLanguage, queryString, baseURI, dataset, getIncludeInferred(),
-					getMaxExecutionTime(), new WeakReference<>(this), getBindingsArray());
+					getMaxExecutionTime(), callerRef, getBindingsArray());
 		} catch (IOException | RepositoryException | MalformedQueryException e) {
 			throw new HTTPQueryEvaluationException(e.getMessage(), e);
 		}
```
hmottestad, line 58: @JervenBolleman

Ontotext is experiencing some issues with FedX queries being cancelled before they are done. They have narrowed it down to the introduction of weak references which were intended to close iterations running in a different thread (background) if the caller had forgotten to close the query result. 

I've expanded the code a bit here by introducing a variable for the callerRef. I believe that the core of the issue is that the weak reference wraps the tuple query object, resulting in the iteration being closed if the tuple query is garbage collected.

The code below could potentially see the tuple query object being garbage collected before the tuple query results were fully consumed.


```
	public static void main(String[] args) {
		try (RepositoryConnection connection = repo.getConnection()) {
			try (TupleQueryResult evaluate = getEvaluate(connection)) {
				for (BindingSet bindings : evaluate) {
					System.out.println(bindings);
				}
			}
		}
	}

	private static TupleQueryResult getEvaluate(RepositoryConnection connection) {
		TupleQuery tupleQuery = connection.prepareTupleQuery("select * where {?a ?b ?c}");
		return tupleQuery.evaluate();
	}
```

core/http/client/src/main/java/org/eclipse/rdf4j/http/client/BackgroundResultExecutor.java
```diff
@@ -85,4 +94,151 @@ private void autoCloseRunnable(QueryResult<?> result, Runnable runner) {
...
+
+		@Override
+		public List<String> getBindingNames() throws QueryEvaluationException {
+			return delegate.getBindingNames();
+		}
+
+		private final static class CleanableState implements Runnable {
+
+			private final TupleQueryResult iteration;
+			private boolean closed = false;
```
JervenBolleman, line 141: Should this not be volatile?

hmottestad: It is only used from the `run()` method to reduce the chances that the cleaner closes the underlying iteration superfluously. If one thread closes the query result (which writes `true` to the `closed` field) and garbage collection runs  immediately afterwards and triggers the cleaner, then there is some chance that the write to the `closed` field will not yet be visible. I don't think that it's a scenario that is worth catering to by adding the overhead of a volatile write when closing the query results.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/QueueIteration.java
```diff
@@ -94,11 +94,7 @@ public void put(E item) throws InterruptedException, T {
 					&& !queue.offer(item, 1, TimeUnit.SECONDS)) {
 				// No body, just iterating regularly through the loop conditions to respond to state changes without a
 				// full busy-wait loop
-				if (callerRef.get() == null) {
-					// Whatever called us is gone.
-					// So stop
-					close();
-				}
+				Thread.onSpinWait();
```
JervenBolleman, line 97: This is not a spinlock, so no need to do this.

core/repository/http/src/main/java/org/eclipse/rdf4j/repository/http/HTTPTupleQuery.java
JervenBolleman: Yes, I see the problem and this is a better solution. Still fragile as depending on cleaner tasks. It was a missed opportunity to really fix this for 4.x with a shared variable ;( only so much time in a day.

PR #4145, opened 2022-09-02
GH-4104: fix issue with Instant, turn it into datetimestamp value

### LINKED ISSUE #4104: Values.literal does not work for Instant
### Current Behavior

`Values.literal(TemporalAccessor value)` should allow for an `Instant` , but `Values.literal(Instant.now())`  throws an Exception

```
Exception in thread "main" java.lang.IllegalArgumentException: value <2022-08-01T21:14:38.470534100Z> cannot be represented by an XML Schema date/time datatype
	at org.eclipse.rdf4j.model.base.AbstractLiteral$TemporalAccessorLiteral.<init>(AbstractLiteral.java:712)
	at org.eclipse.rdf4j.model.base.AbstractValueFactory.createLiteral(AbstractValueFactory.java:214)
	at org.eclipse.rdf4j.model.impl.ValidatingValueFactory.createLiteral(ValidatingValueFactory.java:200)
	at org.eclipse.rdf4j.model.util.Values.literal(Values.java:616)
	at org.eclipse.rdf4j.model.util.Values.literal(Values.java:600)
```

### Expected Behavior

A literal with type xsd:dateTime is created

### Steps To Reproduce

_No response_

### Version

4.1.0

Yes

### Anything else?

_No response_

barthanssens:

GitHub issue resolved: #4104 

- add formatter to convert Instant to xsd:dateTimestamp
- added test
- 

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
@@ -14,6 +14,7 @@ package org.eclipse.rdf4j.model.base;
 import static java.lang.Math.abs;
 import static java.time.temporal.ChronoField.DAY_OF_MONTH;
 import static java.time.temporal.ChronoField.HOUR_OF_DAY;
+import static java.time.temporal.ChronoField.INSTANT_SECONDS;
 import static java.time.temporal.ChronoField.MINUTE_OF_HOUR;
 import static java.time.temporal.ChronoField.MONTH_OF_YEAR;
 import static java.time.temporal.ChronoField.NANO_OF_SECOND;
@@ -522,7 +523,7 @@ public abstract class AbstractLiteral implements Literal {
 		private static final ChronoField[] FIELDS = {
 				YEAR, MONTH_OF_YEAR, DAY_OF_MONTH,
 				HOUR_OF_DAY, MINUTE_OF_HOUR, SECOND_OF_MINUTE, NANO_OF_SECOND,
-				OFFSET_SECONDS
+				OFFSET_SECONDS, INSTANT_SECONDS
 		};
 
 		private static final DateTimeFormatter LOCAL_TIME_FORMATTER = new DateTimeFormatterBuilder()
@@ -584,6 +585,12 @@ public abstract class AbstractLiteral implements Literal {
 
 				.toFormatter();
 
+		private static final DateTimeFormatter DATETIMESTAMP_FORMATTER = new DateTimeFormatterBuilder()
+
+				.appendInstant()
+
+				.toFormatter();
+
 		private static final DateTimeFormatter DASH_FORMATTER = new DateTimeFormatterBuilder()
 
 				.appendLiteral("--")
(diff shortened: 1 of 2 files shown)


core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
```diff
@@ -584,6 +585,12 @@ static class TemporalAccessorLiteral extends AbstractLiteral {
 
 				.toFormatter();
 
+		private static final DateTimeFormatter DATETIMESTAMP_FORMATTER = new DateTimeFormatterBuilder()
+
+				.appendInstant()
+
+				.toFormatter();
```
jeenbroekstra, line 592: Are these empty lines deliberate?

core/model/src/test/java/org/eclipse/rdf4j/model/util/ValuesTest.java
```diff
@@ -508,6 +509,18 @@ public void testTemporalAccessorLiteralNull() {
...
+	@Test
+	public void testTemporalAccessorLiteralInstant() {
+		final Instant value = Instant.ofEpochSecond(1234567890, 12);
+		Literal literal = literal(value);
+
+		assertThat(literal).isNotNull();
+		assertThat(literal.temporalAccessorValue()).isEqualTo(value);
+		assertThat(literal.getLabel()).isEqualTo(value.toString());
+		assertThat(literal.getDatatype()).isEqualTo(XSD.DATETIMESTAMP);
+		assertThat(literal.getCoreDatatype()).isEqualTo(CoreDatatype.XSD.DATETIMESTAMP);
```
jeenbroekstra, line 521: What's the reasoning for using dateTimeStamp instead of just dateTime? As far as I'm aware the only difference is that the time zone is not optional in dateTimeStamp, but dateTime seems to be more far more commonly used.

abrokenjester (commented):
Nice work @barthanssens - couple of minor inline comments.

core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
barthanssens: Can be removed, I tried to follow the same empty lines / grouping as in the rest of the file.

jeenbroekstra: Huh yeah, hadn't noticed it was the same for the other files. Odd. I don't particularly think it makes it better or easier to read, but no big deal - happy to leave as-is for now, for consistency's sake at least :)

core/model/src/test/java/org/eclipse/rdf4j/model/util/ValuesTest.java
barthanssens: Unfortunately a bit of a hack... currently the DateTimeFormatter is selected based on the XSD type.

So for xsd:dateTime a DateTimeFormatter is selected that uses ChronoField.YEAR field etc, which is not supported by Instant...

Giving it a second thought.... Perhaps it could be implemented the other way round (always use a formatter using Chronofield.INSTANT_SECONDS for xsd:dateTime), though that would break if a Temporal would not support that field

jeenbroekstra: @barthanssens but regardless of which formatter you use, you could use the `xsd:dateTime` datatype URI for the resulting literal I think? Every `dateTimeStamp` is also a valid `dateTime`...

barthanssens: I have to re-check the code, but I was under the impression that the current code first selects the xsd type on the chronofields of the java class, (where Instant does not support e.g. Year) 
and then that xsd type is used to select the dateformatter

Instead of first do the formatting and then set the xsd datatype.

I'll look into it, I do agree that using dateTime for all cases is a better solution...

Issue #3781, opened 2022-04-08
Create/update namespace endpoint does not validate the prefix and the URI


nikolaykolev:
### Current Behavior

I can create/update a prefix with invalid prefix and URI, there's no validation in the endpoint.

### Expected Behavior

There should be a validation of the URI at least, I'm not sure about the prefix, but I guess creating a prefix with a space or < > in it should not be allowed as well.

### Steps To Reproduce

1. Execute the following:
curl 'http://localhost:7222/repositories/x/namespaces/asd%20c' -X 'PUT' --data-raw 'http://xxx.com'
2. The namespace is created successfully although the prefix has a space in it (asd c)

1. Execute the following
curl 'http://localhost:7222/repositories/x/namespaces/asd' -X 'PUT' --data-raw 'httpxxxcom'
2. The namespace is created with URI httpxxxcom

### Version

4.0.0-M2

damyan-ognyanov:
There is a regression in the `NamespaceController` introduced with the previous fix. The controller accepts invalid prefixes like `22` or `2abc` but reject some valid prefixes, like `abc-de` or `abc.de`

PR #4147, opened 2022-09-03
GH-3781 add namespace validation

patrickwyler:
GitHub issue resolved: #3781

The NamespaceController is now validating the prefix and namespace before updating it.

----

 - [X] I've added tests for the changes I made

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
@@ -10,7 +10,11 @@ package org.eclipse.rdf4j.http.server.repository.namespaces;
 import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 
+import static org.apache.commons.lang3.StringUtils.isAlphanumeric;
+import static org.apache.commons.lang3.StringUtils.isEmpty;
+
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -18,6 +22,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.eclipse.rdf4j.common.io.IOUtil;
+import org.eclipse.rdf4j.common.net.ParsedIRI;
 import org.eclipse.rdf4j.common.webapp.views.EmptySuccessView;
 import org.eclipse.rdf4j.common.webapp.views.SimpleResponseView;
 import org.eclipse.rdf4j.http.server.ClientHTTPException;
@@ -38,11 +43,10 @@ import org.springframework.web.servlet.mvc.AbstractController;
  * @author Arjohn Kampman
  */
 public class NamespaceController extends AbstractController {
-
 	private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
 	public NamespaceController() throws ApplicationContextException {
-		setSupportedMethods(new String[] { METHOD_GET, METHOD_HEAD, "PUT", "DELETE" });
+		setSupportedMethods(METHOD_GET, METHOD_HEAD, "PUT", "DELETE");
 	}
 
 	@Override
@@ -97,10 +101,7 @@ public class NamespaceController extends AbstractController {
 		String namespace = IOUtil.readString(request.getReader());
 		namespace = namespace.trim();
 
-		if (namespace.length() == 0) {
-			throw new ClientHTTPException(SC_BAD_REQUEST, "No namespace name found in request body");
-		}
-		// FIXME: perform some sanity checks on the namespace string
+		validateUpdateNamespaceData(prefix, namespace);
 
 		try (RepositoryConnection repositoryCon = RepositoryInterceptor.getRepositoryConnection(request)) {
 			repositoryCon.setNamespace(prefix, namespace);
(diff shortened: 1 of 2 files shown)


tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
```diff
@@ -38,11 +43,13 @@
  * @author Arjohn Kampman
  */
 public class NamespaceController extends AbstractController {
+	private static final String URL_REGEX = "^(((https?|ftps?|gopher|telnet|nntp)://)(%[0-9A-Fa-f]{2}|[-()_.!~*';*/?:@&#=+$,A-Za-z0-9])+)([).!';/?:,][[:blank:]])?$";
+	private static final Pattern URL_PATTERN = Pattern.compile(URL_REGEX);
```
jeenbroekstra, line 47: This is too strict - it does not allow for URIs other than the ones starting with the schemes you have in there, for one thing. 

It's probably better (for consistency's sake) to leverage the RDF4J `ParsedIRI` class here, which can be used to check that a string is a valid IRI.

jeenbroekstra (changes requested):
I'm not actually convinced that the controller is the right place to fix this. It should up to the repository itself to determine if a namespace declaration is valid or not.

jeenbroekstra:
By the way: thanks for picking this up @patrickwyler! Let's discuss how to improve this fix further.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
patrickwyler: I found this code snipped which validates the base URI. I think it would make sense to have the same validation for the namespace URI too. 
https://github.com/eclipse/rdf4j/blob/b849630bfac9c84e0444390179409456da043b15/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/query/SPARQLOperation.java#L54

..or I could just check if the namespace URI  isn't empty.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
```diff
@@ -121,4 +121,27 @@ private ModelAndView getRemoveNamespaceResult(HttpServletRequest request, String
...
+		}
+
+		if (!isValidNamespaceIri(namespace)) {
+			throw new ClientHTTPException(SC_BAD_REQUEST, "Namespace not valid");
+		}
+	}
+
+	private boolean isValidNamespaceIri(String namespace) {
+		try {
+			return ParsedIRI.create(namespace).isAbsolute();
```
jeenbroekstra, line 141: Looks good. 

The only thing I'm not sure about is if you should be using `ParsedIRI.create` here instead of the constructor (`new ParsedIRI`). The difference between the two is that `create` attempts to percent-encode any unexpected characters it encounters in the input (only throwing a parse error if even after doing this the IRI is still illegal), whereas the constructor just throws an error immediately if the input is not valid.

So for example if your input is `http://example.org/with whitespace`, the constructor would throw an error (whitespaces are not legal in IRIs), but the `#create` method would return `http://example.org/with%20whitespace`.

Which is the right choice here kind of depends on how strict you want to be.

tools/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceControllerTest.java
```diff
@@ -0,0 +1,129 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
```
jeenbroekstra, line 7:
```suggestion
/*******************************************************************************
 * Copyright (c) 2022 Eclipse RDF4J contributors.
 *
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Distribution License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/org/documents/edl-v10.php.
 *
 * SPDX-License-Identifier: BSD-3-Clause
 *******************************************************************************/
```

tools/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceControllerTest.java
```diff
@@ -0,0 +1,129 @@
...
+		final ModelAndView result = controller.handleRequest(request, response);
+
+		// Assert
+		assertThat(result).isNotNull();
+		assertThat(result.getView()).isInstanceOf(EmptySuccessView.class);
+	}
+
+	@ParameterizedTest
+	@ValueSource(strings = { "  ", "\t", "\n", "-", "rdf 4j", "rdf-4j" })
+	void addNamespace_prefix_invalid(String prefix) {
```
jeenbroekstra, line 63: Nicely set up testing!

jeenbroekstra (commented):
Nice work @patrickwyler - couple of comments but nothing major. As for my earlier remark that this should be done in the repository rather than the controller: I think I'm actually fine with this approach after all.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceController.java
patrickwyler: Thanks for the explanation! Didn't know about the difference, I think it would be better to have a stricter validation, just changed my code to the constructor variant and added your two examples to my tests.

tools/server-spring/src/test/java/org/eclipse/rdf4j/http/server/repository/namespaces/NamespaceControllerTest.java
patrickwyler: Thanks :innocent:

patrickwyler:
I think it's ready for the final review @jeenbroekstra.

jeenbroekstra (approved):

jeenbroekstra:
Nice work @patrickwyler - looks good to me! I'll hold off on merging because I think @hmottestad may want to do some cleanup with some other PRs first, but good to go otherwise.

PR #4151, opened 2022-09-04
Update security policy wrt supported patch releases

jeenbroekstra:

Update the list of supported versions for security patch releases.

## open questions:

1. is this the correct list (in particular do we actually want to support doing security patch release for RDF4J 3.7.x?)
2. is this generic replacement easy enough to understand (intent is to make it easier for us to maintain without losing clarity for our users)?

----

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,12 +6,15 @@ Eclipse RDF4J follows the [Eclipse Vulnerability Reporting Policy](https://www.e
 
 Eclipse RDF4J supports security updates for the following releases:
 
+
 | Version | Supported          |
 | ------- | ------------------ |
-| 4.1.x   | ✅ |
-| 4.0.x   | ✅                 |
-| 3.7.x   | ✅                 |
-| < 3.7   | :x:                |
+| current release   | ✅ |
+| latest minor release before the current   | ✅(on request only) |
+| latest major release before the current   | ✅(on request only) |
+| anything older   | :x:                |
+
+For example if the current release is 4.1, we support security patches for 4.1.x (the current release) and 4.0.x (latest minor before current), as well as for 3.7.x (latest major before current), but not for 3.6.x or older. Security patches for the current release are provided proactively by the team, while patches for older supported releases are provided on request only.
 
 ## Reporting a Vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,9 +8,10 @@ Eclipse RDF4J supports security updates for the following releases:
 
 | Version | Supported          |
 | ------- | ------------------ |
-| 3.7.x   | :white_check_mark: |
-| 3.6.x   | :white_check_mark: |
-| < 3.6   | :x:                |
+| 4.1.x   | ✅ |
+| 4.0.x   | ✅                 |
+| 3.7.x   | ✅                 |
+| < 3.7   | :x:                |
 
 ## Reporting a Vulnerability
 


SECURITY.md
```diff
@@ -8,9 +8,10 @@ Eclipse RDF4J supports security updates for the following releases:
```
jeenbroekstra, line 8: - ✅ the most recent stable release
- ✅ the latest minor release _before_ the current one 
- ✅ the latest major release _before_ the current
- ❌ anything older

(for example if the latest release is 4.1, we support security patches for 4.1.x and 4.0.x releases, as well as for 3.7.x, but not for 3.6.x or older)

SECURITY.md
```diff
@@ -6,11 +6,15 @@ Eclipse RDF4J follows the [Eclipse Vulnerability Reporting Policy](https://www.e
...
 | ------- | ------------------ |
-| 3.7.x   | :white_check_mark: |
-| 3.6.x   | :white_check_mark: |
-| < 3.6   | :x:                |
+| current release   | ✅ |
+| latest minor release before the current   | ✅                 |
+| latest major release before the current   | ✅                 |
+| anything older   | :x:                |
+
+For example if the current release is 4.1, we support security patches for 4.1.x (the current release) and 4.0.x (latest minor before current), as well as for 3.7.x (latest major before current), but not for 3.6.x or older.
```
hmottestad, line 17: I would prefer if we only supported the latest version. Backporting is a very manual process at the moment. We could improve that by extending our scripts and by improving our branching process. Maybe have a "security" branch that branches off from the oldest supported version and is automatically merged into main (like how we merge main into develop).

jeenbroekstra: Keep in mind that this is only for security patches, which 9 times out of 10 are dependency fixes, so relatively easy.

I don't think we can get away with not supporting such security patches for at least one older (major) release family than the latest. Not everyone can upgrade to the latest.

We could potentially do "on request" only perhaps?

hmottestad: This is how I went about backporting the previous fix:
 - find the correct place to branch
 - cherry pick the commit
 - fix merge conflict (there were even some in the pom.xml file)
 - run all the tests (I could have created a PR and let github handle this)
 - try to use the script only to find out that the safety mechanisms prevent it from running on any other branch than main
 - manually run through the steps in the release script
 - trigger the jenkins jobs
 - type up the release notes and news item

There is a lot of potential for streamlining and automation of this process, but I don't really want to put effort into it unless we really need to. So I would prefer "on request" to be the default for all older versions. 

Maybe we could also add a sentence to inform users that they can override the version of a transitive dependency from their own project?

jeenbroekstra: Finding the correct place to branch should just be the version tag, no? And of course you're right that cherry-picking and solving merge conflicts gets harder the further back you go - but for dependency fixes at least you could opt to just not do them as cherry-picks at all, but just separate commits. 

The fact that the scripts don't cater for this is a bit of an issue, agreed. We could look into making that hard "only main branch" a bit more an "are you sure you want to release from this branch?". In any case, for now I'm fine with going for an "on request only" model for older releases. I'll put some wording in this PR to that effect. 

> Maybe we could also add a sentence to inform users that they can override the version of a transitive dependency from their own project?

I don't know that this document is the right place for that - this is about our policies as a project, not about technical tips, really. Also: this really only works if/when you _only_ depend on RDF4J directly, as soon as you have transitive dependencies (e.g. if you use the OWL API or something like Any23) this becomes way _way_ harder to do - and not something we should be advising except as a temporary workaround.

PR #4153, opened 2022-09-05
GH-4149 Check for ill-typed literals during SHACL datatype validation

hmottestad:
GitHub issue resolved: #4149 

SHACL spec says that we should check for ill-typed literals if they are the datatype is one of the supported XSD datatypes. Easiest way to do this is to use the `XMLDatatypeUtil.isValidValue(...)` method. To allow this to be used in the SPARQL query we also introduce a custom function in the RSX namespace that calls this method.

----

### DIFF
--- a/core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/RSX.java
+++ b/core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/RSX.java
@@ -39,6 +39,7 @@ public class RSX {
 	public final static IRI targetShape = create("targetShape");
 	public final static IRI dataGraph = create("dataGraph");
 	public final static IRI shapesGraph = create("shapesGraph");
+	public final static IRI valueConformsToXsdDatatypeFunction = create("valueConformsToXsdDatatypeFunction");
 
 	private static IRI create(String localName) {
 		return Vocabularies.createIRI(RSX.NAMESPACE, localName);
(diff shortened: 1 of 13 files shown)


hmottestad:
@jeenbroekstra do you know any built in functionality to easily check the conformance if a literal to its datatype using SPARQL? I've ended up creating a custom SPARQL function, but if there is a better way than I would prefer that!

jeenbroekstra:
> @jeenbroekstra do you know any built in functionality to easily check the conformance if a literal to its datatype using SPARQL? I've ended up creating a custom SPARQL function, but if there is a better way than I would prefer that!

Not behind a machine where I can easily test this, but wouldn't a cast function (which are built in) work for this? `xsd:dateTime(?lit)` will, I think, result in a type error when the input literal is not a valid datetime value.

hmottestad:
> > @jeenbroekstra do you know any built in functionality to easily check the conformance if a literal to its datatype using SPARQL? I've ended up creating a custom SPARQL function, but if there is a better way than I would prefer that!
> 
> Not behind a machine where I can easily test this, but wouldn't a cast function (which are built in) work for this? `xsd:dateTime(?lit)` will, I think, result in a type error when the input literal is not a valid datetime value.

I gave this a shot but I couldn't get it to work for `xsd:date` because we haven't implemented a constructor function for that. The SPARQL spec is a bit obtuse about which constructor functions are required, I feel it would have been simpler if they had just included all the datatypes.

So I'm sticking with my custom function.

jeenbroekstra:
> > > @jeenbroekstra do you know any built in functionality to easily check the conformance if a literal to its datatype using SPARQL? I've ended up creating a custom SPARQL function, but if there is a better way than I would prefer that!
> > 
> > 
> > Not behind a machine where I can easily test this, but wouldn't a cast function (which are built in) work for this? `xsd:dateTime(?lit)` will, I think, result in a type error when the input literal is not a valid datetime value.
> 
> I gave this a shot but I couldn't get it to work for `xsd:date` because we haven't implemented a constructor function for that. The SPARQL spec is a bit obtuse about which constructor functions are required, I feel it would have been simpler if they had just included all the datatypes.

Interesting! Probably something we can look to add ourselves easily enough. 

> So I'm sticking with my custom function.

Yeah seems like a good solution.

Issue #4157, opened 2022-09-07
Regression: Queries with HAVING clause throw AssertionError in Var because parent is already set


jetztgradnet:
### Current Behavior

When parsing the following query with RDF4J 4.1.1 (worked with 4.1.0 and 4.0.0) I get an AssertionError:

```
SELECT ?this
WHERE {
	?this <http://example.org/nestedProperty> ?value .
}
GROUP BY ?this
HAVING ( ( count(?value)  < 2 ) && ( count(?value)  != 0 ) )
```

Removing the `HAVING` clauses makes it parse, so the issue must be somewhere in there. I assume one of the recent optimizations are missing one of the Expr.clone() calls or so?

Stacktrace:
```
java.lang.AssertionError: Unexpected assertion
	at com.metaphacts.rdf4j.RDF4JTest.testVar(RDF4JTest.java:20)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
[...]

PR #4159, opened 2022-09-07
GH-4157 Fix for assertion error due to Var not being cloned

hmottestad:
GitHub issue resolved: #4157 

 - added a test case and fixed the bug
 - also some minor code cleanup

----

### DIFF
--- a/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
+++ b/core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
@@ -1646,9 +1646,7 @@ public class TupleExprBuilder extends AbstractASTVisitor {
 		public void meet(Var var) {
 			if (toBeReplaced.equals(var)) {
 				QueryModelNode parent = var.getParentNode();
-				Var replacementVar = replacement.clone();
-				parent.replaceChildNode(var, replacementVar);
-				assert replacementVar.getParentNode() != null;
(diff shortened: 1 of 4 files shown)


core/queryparser/sparql/src/main/java/org/eclipse/rdf4j/query/parser/sparql/TupleExprBuilder.java
```diff
@@ -2613,7 +2611,7 @@ public void meet(Sum node) throws VisitorException {
 
 		private void meetAggregate(AggregateOperator node) {
 			if (node.equals(operator)) {
-				node.getParentNode().replaceChildNode(node, replacement);
+				node.getParentNode().replaceChildNode(node, replacement.clone());
```
hmottestad, line 2614: @jeenbroekstra I think that might not have previously had any tests covering this code here, otherwise the assertion should have failed I think.

jeenbroekstra: Peculiar - we have tests for SPARQL queries with aggregate operators I'm sure. Not sure why this code path was not hit by them though.

jeenbroekstra (approved):

nguyenm100:
Seeing this in my stuff with the latest on trunk.

image

not sure if this is related but appears to be a similar stack.  I think this question has come up before in that is org.eclipse.rdf4j.testsuite.query.parser.sparql.manifest.SPARQLQueryComplianceTest still valid b/c there are about 1/2 dozen tests that fail.  

The one included here is new.

Issue #4167, opened 2022-09-09
Some of the resources results for SPARQLQueryComplianceTest are wrong


nguyenm100:
### Current Behavior

for: https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQLQueryComplianceTest.java, some of the results are incorrect: https://github.com/eclipse/rdf4j/tree/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1

The tests runs through a bunch of data files that specifiecs data, queries, and results.  I found the following tests to specify incorrect results data and thus fails when running this testsuit:

sparql11-wildcard-cycles-04 - https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1/property-paths/sparql11-wildcard-cycles-04.srx

sparql11-not-in-02 - https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1/builtin/sparql11-not-in-02.srx

sparql11-sum-02  - https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1/aggregates/sparql11-sum-02.srx

### Expected Behavior

The response files to be correct

### Steps To Reproduce

Run SPARQLQueryComplianceTest against these files.

### Version

4.1.2-SNAPSHOT

nguyenm100:
Ontop of which, should we delete: 

sparql11-sequence-04 (https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1/property-paths/sparql11-sequence-04.rq)
sparql11-sequence-05
sparql11-sequence-06

which tests the occurrence specifiers that didn't make it to SPARQL1.1 and thus not implemented in RDF4J

nguyenm100:
PR https://github.com/eclipse/rdf4j/pull/4168

jeenbroekstra:
That set of tests is no longer in use. I don't necessarily object to the quick fix but we should probably just remove that directory.

Issue #4165, opened 2022-09-09
SPARQLQueryComplianceTest Regression: sparql11-sum-03


nguyenm100:
### Current Behavior

Running: https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQLQueryComplianceTest.java

Test: sparql11-sum-03 (https://github.com/eclipse/rdf4j/tree/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1/aggregates)

returning incorrect results as compared to GraphDB 9.7

Total
--
0
0

### Expected Behavior

total
--  
 
"7"^^xsd:integer

### Steps To Reproduce

using dataset:

@prefix : <http://example.org/> .

:a :p "1" .
:a :p 1 .
:b :p 3 .
:b :p 4 .

run query:

PREFIX : <http://example.org/>
SELECT (SUM(?x) AS ?total)
WHERE {
  ?a :p ?x.
}
GROUP BY ?a

### Version

4.1.2-SNAPSHOT 

Though may have been this way for some time.

hmottestad:
Should be fixed once this PR is merged https://github.com/eclipse/rdf4j/pull/4132

Issue #4164, opened 2022-09-09
SPARQLQueryComplianceTest Regression: sparql11-sum-04


nguyenm100:
### Current Behavior

Running: https://github.com/eclipse/rdf4j/blob/main/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQLQueryComplianceTest.java

Test: sparql11-sum-04 (https://github.com/eclipse/rdf4j/tree/main/testsuites/sparql/src/main/resources/testcases-sparql-1.1/aggregates)

returning incorrect results as compared to GraphDB 9.7

A | Total
-- | --
ex:y | 8
ex:z | 10.0
ex:x | 10

### Expected Behavior

GraphDB returns:
a | total |
-- | -- 
ex:x | "10"^^xsd:integer
ex:y |  
 ex:z | "10.0"^^xsd:decimal

### Steps To Reproduce

using dataset:

 @prefix : <http://example.org/> .

:x :p 1, 2, 3, 4 .
:y :p 1, _:b2, 3, 4 .
:z :p 1.0, 2.0, 3.0, 4 .

run query:

PREFIX : <http://example.org/>
SELECT ?a (SUM(?x) AS ?total)
WHERE {
  ?a :p ?x.
}
GROUP BY ?a

### Version

4.1.2-SNAPSHOT

This may have been this way for some time.  I'm just going back through some of the regressions.

hmottestad:
Should be fixed once this PR is merged: https://github.com/eclipse/rdf4j/pull/4132

PR #4168, opened 2022-09-09
GH-4167 - Fixing datafiles for SPARQLQueryComplianceTest

nguyenm100:
Results of some tests were incorrect and always failing when compared to actual from SPARQLQueryComplianceTests.  This PR is to patch those results.

### DIFF
--- a/testsuites/sparql/src/main/resources/testcases-sparql-1.1/aggregates/sparql11-sum-02.srx
+++ b/testsuites/sparql/src/main/resources/testcases-sparql-1.1/aggregates/sparql11-sum-02.srx
@@ -7,10 +7,5 @@
     <variable name="totalPrice"/>
   </head>
   <results>
-    <result>
-      <binding name="totalPrice">
-      	<literal datatype="http://www.w3.org/2001/XMLSchema#integer">0</literal>
-      </binding>
-    </result>
   </results>
 </sparql>
--- a/testsuites/sparql/src/main/resources/testcases-sparql-1.1/builtin/sparql11-not-in-02.srx
+++ b/testsuites/sparql/src/main/resources/testcases-sparql-1.1/builtin/sparql11-not-in-02.srx
@@ -16,13 +16,5 @@
 		  <literal datatype="http://www.w3.org/2001/XMLSchema#integer">41</literal>
 		</binding>
 	 </result>
-	 	 <result>
-		<binding name="book">
-		  <uri>http://example.org/book4</uri>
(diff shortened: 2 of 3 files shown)


hmottestad:
Could you help me out a bit by explaining how I can easily run these tests locally on my laptop?

nguyenm100:
Hi, I'm using a fork of Halyard (by @pulquero) who has wired up a lot of these tests.  You can see them here: https://github.com/pulquero/Halyard/blob/latest/strategy/src/test/java/com/msd/gin/halyard/strategy/HalyardSPARQLQueryTest.java

jeenbroekstra:
The problem is that the testcases in the `testcases-sparql-1.1` directory are no longer in use in RDF4J itself. It was originally intended as an extension of official W3C test cases with our own additions for as we discovered them, but it quickly become too cumbersome to maintain that. 

We have switched our own testing framework for SPARQL over to the `RepositorySPARQLComplianceTestSuite`, which is not based on an manifest-based set of test data. `SPARQLQueryComplianceTest` is only still in use for running the W3C-approved manifest suites for SPARQL 1.0 and 1.1 compliance (which are in the `testcases-sparql-1.0-w3c` and `testcases-sparql-1.1-w3c` directories).

I have no objection to this PR (though I haven't checked the tests yet to see if the fixes make sense) but the real fix is that we just get rid of this old datadir and advise anyone still using it to switch to `RepositorySPARQLComplianceTestSuite` (which is much easier to work with anyway).

jeenbroekstra (approved):

nguyenm100:
Is RepositorySPARQLComplianceTestSuite wired up as part of the rdf4j build and does it carry over the tests from test-cases-sparql-1.1?  I ask b/c the tests in that directory flagged some issues that were recently fixed see: https://github.com/eclipse/rdf4j/issues/4165 & https://github.com/eclipse/rdf4j/issues/4164

and another issue I'll be reporting in a sec.

nguyenm100:
see: https://github.com/eclipse/rdf4j/issues/4209

jeenbroekstra:
> Is RepositorySPARQLComplianceTestSuite wired up as part of the rdf4j build and does it carry over the tests from test-cases-sparql-1.1? I ask b/c the tests in that directory flagged some issues that were recently fixed see: #4165 & #4164

It is wired up as part of the build. Whether it completely covers all possible test cases in that old directory I don't know, but probably not. If you spot lack of coverage it might be good to port them over.

PR #4194, opened 2022-09-14
4.1.3-release-notes

hmottestad:
Javadocs, release-notes and news item for 4.1.3


jeenbroekstra (approved):
Nice work :) 

Just a reminder: don't forget to create the release in github (right hand column) as well. 
It's something I often forget myself and while not a huge deal, I think it looks sloppy if that column does not show the actual latest release.

Issue #4245, opened 2022-10-24
Regression RDF4J 4.2: NPE when control flow computes hashCode of "COUNT(*)" algebra node


aschwarte10:
### Current Behavior

One of our unit tests fails with a NullpointerException after upgrading RDF4J to 4.2: 

the `hashCode` function of `UnaryValueOperator` invokes `arg.hashCode`, where there exist cases of `arg` being `null`.

An example is the algebra reprentation of `COUNT(*) AS ?cnt`  in a projection.

In our case the hashCode is invoke when rendering the query string (see below), but it may be invoked also by other optimizers or parts of the query engine.

Stack Trace for my test scenario (see below)

```
java.lang.NullPointerException
	at org.eclipse.rdf4j.query.algebra.UnaryValueOperator.hashCode(UnaryValueOperator.java:96)
	at org.eclipse.rdf4j.query.algebra.AbstractAggregateOperator.hashCode(AbstractAggregateOperator.java:65)
	at org.eclipse.rdf4j.query.algebra.Count.hashCode(Count.java:39)
	at org.eclipse.rdf4j.query.algebra.GroupElem.hashCode(GroupElem.java:95)
	at java.base/java.util.ArrayList.hashCodeRange(ArrayList.java:627)
	at java.base/java.util.ArrayList.hashCode(ArrayList.java:614)
	at org.eclipse.rdf4j.query.algebra.Group.hashCode(Group.java:164)
	at org.eclipse.rdf4j.query.algebra.UnaryTupleOperator.hashCode(UnaryTupleOperator.java:104)
	at org.eclipse.rdf4j.query.algebra.Extension.hashCode(Extension.java:124)
	at org.eclipse.rdf4j.query.algebra.UnaryTupleOperator.hashCode(UnaryTupleOperator.java:104)
	at org.eclipse.rdf4j.query.algebra.Projection.hashCode(Projection.java:109)
	at java.base/java.util.HashMap.hash(HashMap.java:339)
	at java.base/java.util.HashMap.put(HashMap.java:607)
	at org.eclipse.rdf4j.queryrender.sparql.experimental.ParsedQueryPreprocessor.meet(ParsedQueryPreprocessor.java:718)
	at org.eclipse.rdf4j.query.algebra.Projection.visit(Projection.java:80)
	at org.eclipse.rdf4j.query.algebra.UnaryTupleOperator.visitChildren(UnaryTupleOperator.java:82)
[...]

PR #4250, opened 2022-10-27
GH-4245 fix SPARQL COUNT hashCode() NullPointerException

hmottestad:
GitHub issue resolved: #4245 

 - add hashCode() checks into the main query parsing/evaluating tests
 - implement hashCode() and equals() in various classes so that the tests work
 - fix the NullPointerException by first checking if the arg is null

----

### DIFF
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryTupleOperator.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryTupleOperator.java
@@ -17,18 +17,10 @@ import java.util.Set;
  */
 public abstract class UnaryTupleOperator extends AbstractQueryModelNode implements TupleExpr {
 
-	/*-----------*
-	 * Variables *
-	 *-----------*/
-
 	/**
 	 * The operator's argument.
 	 */
 	protected TupleExpr arg;
-
-	/*--------------*
-	 * Constructors *
-	 *--------------*/
 
 	protected UnaryTupleOperator() {
 	}
@@ -41,10 +33,6 @@ public abstract class UnaryTupleOperator extends AbstractQueryModelNode implemen
 	protected UnaryTupleOperator(TupleExpr arg) {
 		setArg(arg);
 	}
-
-	/*---------*
-	 * Methods *
-	 *---------*/
 
 	/**
 	 * Gets the argument of this unary tuple operator.
--- a/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryTupleOperator.java
+++ b/core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryTupleOperator.java
@@ -90,13 +90,17 @@ public abstract class UnaryTupleOperator extends AbstractQueryModelNode implemen
 	}
 
 	@Override
-	public boolean equals(Object other) {
-		if (other instanceof UnaryTupleOperator) {
-			UnaryTupleOperator o = (UnaryTupleOperator) other;
-			return arg.equals(o.getArg());
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (!(o instanceof UnaryTupleOperator)) {
+			return false;
 		}
 
-		return false;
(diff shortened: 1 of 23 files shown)


aschwarte10 (commented):
Looks good on code level. I will do one more confirmation tomorrow by putting the queryalgebra jar into our codebase for testing.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryTupleOperator.java
```diff
@@ -90,13 +90,17 @@ public void replaceChildNode(QueryModelNode current, QueryModelNode replacement)
...
+			return true;
 		}
+		if (!(o instanceof UnaryTupleOperator)) {
+			return false;
+		}
+
+		UnaryTupleOperator that = (UnaryTupleOperator) o;
 
-		return false;
+		return arg.equals(that.arg);
```
jeenbroekstra, line 103: I think you may need a null-check here as well - perhaps use `Objects.equals(arg, that.arg)`?

jeenbroekstra: And if we're being consistent, should this class `hashCode` also do a null check? 

Fwiw I think that generally the way we handle this in these classes is a little inconsistent. We have a no-arg constructor (which effectively allows the `arg` to be `null`), but then the setter has assertions to ensure it can not be set to `null`. It would probably be cleaner to come down one way or the other. But not necessarily in this PR, happy with a quick fix :)

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryValueOperator.java
```diff
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra;
 
+import java.util.Objects;
+
```
aschwarte10, line 14: If you look at the javadoc, this is now not consistent with the implementation. Maybe you want to revisit?

```
An abstract superclass for unary value operators which, by definition, has one argument.
```

At least for the `Count` node this is not always the case.

aschwarte10 (commented):
I have done a local test in our environment and applied the compiled queryalgebra-model jar in our codebase. All tests are green again.

From my side if @jeenbroekstra 's comment is discussed / addressed this is good to go.

Please see also one remark inline on the javadoc.

core/queryalgebra/model/src/main/java/org/eclipse/rdf4j/query/algebra/UnaryTupleOperator.java
hmottestad: I took a look at `UnaryTupleOperator` and I couldn't find any scenarios where `arg` ends up being `null`. The values that I see for `arg` are all some form of a query body, and even and empty query body won't result in `null`. For `SELECT * WHERE { }` we use a `SingletonSet` as the query body. 

The issues with `UnaryValueOperator` is that it's not supposed to handle multiple arguments, but it still needs to handle `*`, even the SPARQL spec lists it as a special case. Our approach to handling `*` in `SELECT * WHERE {?a a ?type}` is to expand `*` into `?a ?type`. That's intentionally not allowed with the `UnaryValueOperator` because it's not allowed to have multiple arguments with something like `COUNT`.

Maybe for 5.0.0 we could consider two things for the algebra:
- Introduce something to represent `*` in `UnaryValueOperator` (maybe a new class or just a constant).
- Remove all the empty constructors from the classes that actually do require arguments.

hmottestad:
> I have done a local test in our environment and applied the compiled queryalgebra-model jar in our codebase. All tests are green again.
> 
> From my side if @jeenbroekstra 's comment is discussed / addressed this is good to go.
> 
> Please see also one remark inline on the javadoc.

Thanks @aschwarte10 ! 

I've added some more to Jeen's comment now. What do you think?

There are a few bug fixes on the main branch already that would be nice to release fairly soon. I don't have any specific date in mind, just that I'm fairly busy for the next few days. If someone else has more time and wants to create a release then that's all good by me :)

jeenbroekstra (approved):

Issue #4256, opened 2022-10-30
SPARQL Scoping issue for filters inside EXISTS


hmottestad:
### Current Behavior

See: https://github.com/eclipse/rdf4j/issues/4169

```
PREFIX : <http://example/>
SELECT * WHERE {
	?a :p ?n
	FILTER NOT EXISTS {
		?a :q ?m .
		FILTER(?n = ?m)
	}
}
```

`?n` inside the `FILTER` clause is considered a new scope.

### Expected Behavior

`?n` inside the `FILTER` clause should be in scope with the `?n` in `?a :p ?n`.

### Version

4.x

### Are you interested in contributing a solution yourself?

Yes

hmottestad:
We use the following code to decide if a variable is in scope for a FILTER operation:

```
	public static boolean isPartOfSubQuery(QueryModelNode node) {
		if (node instanceof SubQueryValueOperator) {
			return true;
		}

		QueryModelNode parent = node.getParentNode();
		if (parent == null) {
			return false;
		} else {
			return isPartOfSubQuery(parent);
		}
	}
```

And EXISTS is among the effective sub classes of `SubQueryValueOperator`:

<img width="766" alt="Screenshot 2022-10-30 at 07 35 02" src="https://user-images.githubusercontent.com/797185/198865808-d9ea08ab-6860-4a37-94c7-0298e604097b.png">

PR #4257, opened 2022-10-30
GH-4256 sparql exists filter scope

hmottestad:
GitHub issue resolved: #4256 

From the spec: 

> Variables in the pattern that are bound in the current [solution mapping](http://www.w3.org/TR/rdf-sparql-query/#defn_sparqlSolutionMapping) take the value that they have from the solution mapping. Variables in the pattern pattern that are not bound in the current solution mapping take part in pattern matching.

https://www.w3.org/TR/sparql11-query/#func-filter-exists

I've removed the check for `isPartOfSubQuery` because it seems to be related to some legacy code that is no longer in use. 

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
@@ -517,18 +517,12 @@ public class StrictEvaluationStrategy implements EvaluationStrategy, FederatedSe
 	}
 
 	protected QueryEvaluationStep prepare(Filter node, QueryEvaluationContext context) throws QueryEvaluationException {
-
-		if (FilterIterator.isPartOfSubQuery(node)) {
-			context = new FilterIterator.RetainedVariableFilteredQueryEvaluationContext(node, context);
-		}
 		QueryEvaluationStep arg = precompile(node.getArg(), context);
 		QueryValueEvaluationStep ves;
 		try {
 			ves = precompile(node.getCondition(), context);
 		} catch (QueryEvaluationException e) {
-			// If we have a failed compilation we always return false.
-			// Which means empty. so let's short circuit that.
-//			ves = new QueryValueEvaluationStep.ConstantQueryValueEvaluationStep(BooleanLiteral.FALSE);
+			// If we have a failed compilation we always return false. Which means empty.
 			return QueryEvaluationStep.EMPTY;
 		}
 		// if the query evaluation is constant it is either FILTER(true) or FILTER(false)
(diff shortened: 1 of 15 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/StrictEvaluationStrategy.java
```diff
@@ -517,18 +517,12 @@ protected QueryEvaluationStep prepare(Service service, QueryEvaluationContext co
 	}
 
 	protected QueryEvaluationStep prepare(Filter node, QueryEvaluationContext context) throws QueryEvaluationException {
-
-		if (FilterIterator.isPartOfSubQuery(node)) {
-			context = new FilterIterator.RetainedVariableFilteredQueryEvaluationContext(node, context);
-		}
```
hmottestad, line 523: This is the change that fixes the issue. It looks like the original reason for this code was to handle some legacy stuff, I presume something related to `CompareAll` and `CompareAny` since these classes are not in use anymore.

hmottestad: Here is the code for `FilterIterator.isPartOfSubQuery(...)`:
```
	public static boolean isPartOfSubQuery(QueryModelNode node) {
		if (node instanceof SubQueryValueOperator) {
			return true;
		}

		QueryModelNode parent = node.getParentNode();
		if (parent == null) {
			return false;
		} else {
			return isPartOfSubQuery(parent);
		}
	}
```

hmottestad: `SubQueryValueOperator` is subclasses by `In`, `Exists`, `CompareAll` and `CompareAny`. It is also subclassed by a the abstract class `CompareSubQueryValueOperator`, which seems insignificant.

JervenBolleman (approved):
I spend some time re-reading the specification for `exist` and I think this fix is correct. I definitely did not think through this code evaluation wise when moving it into a QueryEvaluationStep.  
The example is shown in the spec in the section comparing MINUS and EXISTS and I understand  it failed before: So I think this should go in.

hmottestad:
I should add a test for MINUS then.

```
PREFIX : <http://example/>
SELECT * WHERE {
        ?x :p ?n
        MINUS {
                ?x :q ?m .
                FILTER(?n = ?m)
        }
}
```

The `?n` above the MINUS section is not in scope for the `?n` inside the FILTER. 

This is a difference between EXISTS and MINUS.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/TripleSourceBase.java
```diff
@@ -138,27 +138,32 @@ public CloseableIteration<BindingSet, QueryEvaluationException> getStatements(
...
-			} else if (bindings.size() > 0) {
-				res = new InsertBindingsIteration(res, bindings);
-			}
 
-			resultHolder.set(new ConsumingIteration(res, federationContext.getConfig().getConsumingIterationMax()));
+				resultHolder.set(new ConsumingIteration(res, federationContext.getConfig().getConsumingIterationMax()));
+			} catch (Throwable t) {
+				res.close();
+				throw t;
+			}
```
hmottestad, line 166: This is an unrelated issue, but it was causing a test to fail intermittently and I thought I might fix it while I was at it.

The `if (!res.hasNext()) {` line can throw an exception if the iteration was interrupted. To make the code simple and to handle any other cases that might be problematic I've gone ahead and wrapped most of the code that uses the `withConnection(...)` method with try/catch.

Issue #2749, opened 2021-01-05
Adding a Model to a repository doesn't transfer its namespaces


Dzeri96:
When adding a Model to a Repository with `conn.add(model)`, its namespaces get lost.
Writing the repo to a file like
```
RDFWriter writer = Rio.createWriter(RDFFormat.TURTLE, outputStream);
conn.export(writer);
```
produces a very unreadable Turtle file with no prefixes.

Only after manually re-adding them like this
```
for (Namespace ns : ontologyNamespaces) {
    conn.setNamespace(ns.getPrefix(), ns.getName());
}
```
does the output actually contain prefixes. It definitely the act of adding the model and not the RDFwriter's fault.

If this is a bug, I'd be happy to contribute. If this is expected behavior (which is counter-intuitive in my opinion), is there a nicer way to include the namespaces when importing?

_VERSION: 3.4.4_

barthanssens:
It's not really a bug IMHO, since the RepositoryConnection will use the `add(Iterable<Statement>)` method, and Statements themselves don't know about namespaces ...
IIRC it's the same behavior when adding a Model to another Model.
(Note that prefixes in Model/Repository A may clash with those in Model/Repository B)

Maybe @jeenbroekstra can tell if this was a deliberate design choice, or just something that stuck through the years.

jeenbroekstra:
I suggest we call this an improvement: it's not necessarily a bug in the sense that we never defined the behavior, but I do agree that it is counter-intuitive (and inconsistent when compared to adding data directly from a File or an InputStream).

A simple fix would be to check (in the add method) if the Iterable also implements `NamespaceAware`, and if so, copy over namespace definitions. 

To resolve potential clashes, for now we can just stick with the same behavior as when directly adding a file to a Repository (e.g. by supplying an inputstream): check if the repository already has a namespace mapping with the same prefix, and _only_ if it does not, add the namespace from the model. 

@Dzeri96  if you're interested in providing a fix for this, that'd be great! Before you dive in, have a look at our [contributor guidelines](https://github.com/eclipse/rdf4j/blob/master/.github/CONTRIBUTING.md), in particular the points on how to sign the Eclipse ECA. Let me know if you need any help with any of it.

Dzeri96:
If others agree, I'd like to work on this when I have some free time. ETA for a PR is mid-february if the issue is not too complicated

barthanssens:
Sure, would be great if you could work on this, thanks.

[...]

PR #4262, opened 2022-11-01
GH-2749 Include Namespaces in Statement imports from NamespaceAware collections

Dzeri96:
Signed-off-by: Dzerom Dzenkins <dzeri96@proton.me>

GitHub issue resolved: #2749

As suggested by @jeenbroekstra, I've added a call to `setNamespace` for all new namespaces coming from the imported `Statement`s, and I've removed the appropriate overrides in the SPARQL repository, as it seems like the code was almost identical. Please correct me if I'm wrong.

Since this is a WIP, there are some things to clear up:
1. Where do I put the unit tests for this? I couldn't find an appropriate place with similar tests.
2. What should we do with the deprecated `Iteration` overload of the `add` method? I suspect this might get removed with the next release so I haven't added any changes there.
3. What about the `remove` method? Should we prune unused `Namespace`s?

As soon as these questions are cleared up, and the flaky tests are fixed, I'll continue working on this PR!

----

### DIFF
--- a/core/repository/api/src/main/java/org/eclipse/rdf4j/repository/base/AbstractRepositoryConnection.java
+++ b/core/repository/api/src/main/java/org/eclipse/rdf4j/repository/base/AbstractRepositoryConnection.java
@@ -15,9 +15,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.net.URL;
-import java.util.HashSet;
 import java.util.Objects;
-import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.rdf4j.common.iteration.Iteration;
@@ -370,20 +368,19 @@ public abstract class AbstractRepositoryConnection implements RepositoryConnecti
 				"contexts argument may not be null; either the value should be cast to Resource or an empty array should be supplied");
 
 		boolean localTransaction = startLocalTransaction();
-		Set<Namespace> newNamespaces = new HashSet<>();
 
 		try {
 			for (Statement st : statements) {
-				if (st instanceof NamespaceAware) {
-					newNamespaces.addAll(((NamespaceAware) st).getNamespaces());
-				}
 				addWithoutCommit(st, contexts);
 			}
 
-			for (Namespace newNamespace : newNamespaces) {
-				String nsPrefix = newNamespace.getPrefix();
-				if (getNamespace(nsPrefix) == null) {
-					setNamespace(nsPrefix, newNamespace.getName());
+			if (statements instanceof NamespaceAware) {
+				var newNamespaces = ((NamespaceAware) statements).getNamespaces();
+				for (Namespace newNamespace : newNamespaces) {
+					String nsPrefix = newNamespace.getPrefix();
+					if (getNamespace(nsPrefix) == null) {
+						setNamespace(nsPrefix, newNamespace.getName());
+					}
 				}
 			}
 
@@ -426,15 +423,6 @@ public abstract class AbstractRepositoryConnection implements RepositoryConnecti
 				"contexts argument may not be null; either the value should be cast to Resource or an empty array should be supplied");
 
 		addWithoutCommit(st, contexts);
-
-		if (st instanceof NamespaceAware) {
-			for (Namespace newNamespace : ((NamespaceAware) st).getNamespaces()) {
-				String nsPrefix = newNamespace.getPrefix();
-				if (getNamespace(nsPrefix) == null) {
-					setNamespace(nsPrefix, newNamespace.getName());
-				}
-			}
-		}
 
 		conditionalCommit(localTransaction);
 	}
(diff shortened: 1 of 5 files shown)


core/repository/api/src/main/java/org/eclipse/rdf4j/repository/base/AbstractRepositoryConnection.java
```diff
@@ -366,10 +370,23 @@ public void add(Iterable<? extends Statement> statements, Resource... contexts)
 				"contexts argument may not be null; either the value should be cast to Resource or an empty array should be supplied");
 
 		boolean localTransaction = startLocalTransaction();
+		Set<Namespace> newNamespaces = new HashSet<>();
+
 		try {
 			for (Statement st : statements) {
+				if (st instanceof NamespaceAware) {
+					newNamespaces.addAll(((NamespaceAware) st).getNamespaces());
+				}
```
jeenbroekstra, line 379: I don't think this will work - it's not `Statement` that implements `NamespaceAware`, but the collection of statements.

It's also potentially a bit of a performance burden doing the check this way. I'd suggest that instead, you do a single check outside of the loop, and then just transfer all namespaces in one go, something like this.:

```java
if (statements instanceof NamespaceAware) {
   var namespaces = ((NamespaceAware)statements).getNamespaces();
   for (namespace: namespaces) {
     setNamespace(namespace.getPrefix(), namespace.getName());
   } 
}
```

I also don't think there's need to check if the namespace already exists in the repo or not. `setNamespace` will just overwrite if it exists (which I think is the behavior we want here).

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -642,32 +642,6 @@ public void add(Reader reader, String baseURI, RDFFormat dataFormat, Resource...
...
-			addWithoutCommit(st, contexts);
-		}
-		try {
-			conditionalCommit(localTransaction);
-		} catch (RepositoryException e) {
-			conditionalRollback(localTransaction);
-			throw e;
-		}
-	}
-
```
jeenbroekstra, line 670: I don't think you should remove this. For one thing, `SPARQLConnection` does not actually implement `setNamespace` (because there is no SPARQL command for this), so we can't support the transfer of namespace in this way when adding a Model. 

If this is something we want to support for `SPARQLConnection` as well we should probably look at ensuring that namespaces in the supplied model make it into the serialized SPARQL update that will be sent to the endpoint, somehow. This is handled in the `createInsertDataCommand` method. But to be fair even then I'm not sure you can guarantee it. There is no general standard on whether a remote SPARQL endpoint stores/reuses the the namespace definitions you supply when you do an update.

TL;DR I'd revert this change for now, and just not worry about SPARQLConnection for the time being.

jeenbroekstra (commented):
Thanks for supplying this PR @Dzeri96 ! I've left a couple of suggestions for improvements. Also: can you make sure that you add some tests to cover the desired behavior? The `RepositoryConnectionTest` class (in `rdf4j-repository-testsuite`) is probably a good spot for this.

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/base/AbstractRepositoryConnection.java
Dzeri96: I'll get to this today, however, didn't you suggest that we skip existing namespaces in the original issue I created?

Which option should I go with?

jeenbroekstra: > I'll get to this today, however, didn't you suggest that we skip existing namespaces in the original issue I created?
> 
> Which option should I go with?

Ah good call I had forgotten that. Yes, let's go with skipping existing for now, as it's more consistent with the behavior when adding a new file.

Dzeri96:
@jeenbroekstra Please have a look at the second iteration of my implementation. I've added a simple test, however some implementations of the `RepositoryConnectionTest` fail to execute on my machine, even on the latest `main` branch. These are:
- `ElasticSearchStoreConnectionIT`
- `HTTPStoreConnectionTest`
- `RDFSchemaRepositoryConnectionTest`
- `SPARQLStoreConnectionTest`

I've nevertheless added a skipping override for the test in the `SPARQLStoreConnectionTest`

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/base/AbstractRepositoryConnection.java
```diff
@@ -366,10 +368,22 @@ public void add(Iterable<? extends Statement> statements, Resource... contexts)
...
+
+			if (statements instanceof NamespaceAware) {
+				var newNamespaces = ((NamespaceAware) statements).getNamespaces();
+				for (Namespace newNamespace : newNamespaces) {
+					String nsPrefix = newNamespace.getPrefix();
+					if (getNamespace(nsPrefix) == null) {
+						setNamespace(nsPrefix, newNamespace.getName());
+					}
+				}
+			}
```
jeenbroekstra, line 385: Nice!

jeenbroekstra (approved):
Nice work @Dzeri96 - this looks good to me!

Not sure why some tests fail in your local setup, but the Github build succeeded without failure so I don't think it's your code that's the problem.

Issue #4279, opened 2022-11-09
Weakness/bug in FilterOptimizer 


pulquero:
### Current Behavior

I've encountered a bug in a custom sail, so out of courtesy I'm making you aware in case you run into similar problems with your implementations.

Note the current code travels all the way down to the bottom filter, then merges upwards. This means as it unrolls upwards, it revisits a filter that has been merged into a previous lower filter. As clone() is being used, we are now in an inconsistent state as we are unrolling up the old branch not the result of what has been modified.
```
		public void meet(Filter filter) {
			super.meet(filter);
			if (filter.getParentNode() instanceof Filter && filter.getParentNode().getParentNode() != null) {

				Filter parentFilter = (Filter) filter.getParentNode();
				QueryModelNode grandParent = parentFilter.getParentNode();
				And merge = new And(filter.getCondition().clone(), parentFilter.getCondition().clone());

				Filter newFilter = new Filter(filter.getArg().clone(), merge);
				grandParent.replaceChildNode(parentFilter, newFilter);
				meet(newFilter);
			}
		}
```

Fix is straightforward, shift everything by a level, so that we are only ever modifying things at or below us:
```
		public void meet(Filter filter) {
			super.meet(filter);
			if (filter.getArg() instanceof Filter && filter.getParentNode() != null) {

				Filter childFilter = (Filter) filter.getArg();
				QueryModelNode parent = filter.getParentNode();
				And merge = new And(childFilter.getCondition().clone(), filter.getCondition().clone());

				Filter newFilter = new Filter(childFilter.getArg().clone(), merge);
				parent.replaceChildNode(filter, newFilter);
				meet(newFilter);
			}
		}
```

### Expected Behavior

RDF4J codebase provides a good foundation/examples from which to write custom sails.

### Steps To Reproduce

Write a custom sail...

### Version

4.2.0

PR #4280, opened 2022-11-10
GH-4279 FilterOptimizer should not modify the query tree above the node it is currently on

hmottestad:
GitHub issue resolved: #4279 

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/FilterOptimizer.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/FilterOptimizer.java
@@ -110,15 +110,14 @@ public class FilterOptimizer implements QueryOptimizer {
 		@Override
 		public void meet(Filter filter) {
 			super.meet(filter);
-			if (filter.getParentNode() instanceof Filter && filter.getParentNode().getParentNode() != null) {
+			if (filter.getArg() instanceof Filter && filter.getParentNode() != null) {
 
-				Filter parentFilter = (Filter) filter.getParentNode();
-				QueryModelNode grandParent = parentFilter.getParentNode();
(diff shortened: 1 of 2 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/optimizer/FilterOptimizer.java
```diff
@@ -110,15 +110,14 @@ private FilterMerger() {
...
+				QueryModelNode parent = filter.getParentNode();
+				And merge = new And(childFilter.getCondition().clone(), filter.getCondition().clone());
 
-				Filter newFilter = new Filter(filter.getArg().clone(), merge);
-				grandParent.replaceChildNode(parentFilter, newFilter);
-				meet(newFilter);
+				Filter newFilter = new Filter(childFilter.getArg().clone(), merge);
+				parent.replaceChildNode(filter, newFilter);
 			}
 		}
```
hmottestad, line 122: @pulquero your solution works well. I wasn't able to create a test that reproduces the issue. I found that when I removed the line with `meet(newFilter);` then I got some failing tests. Your solution made those tests pass again.

I think that `meet(newFilter);` was masking the issue. We would first recurse down to the innermost filter, then optimise it by swapping out the parent node. When returning back up we would visit the old parent node and if that node also needed to be optimised we would actually end up optimising the child nodes again.

pulquero: Test case i think would be join, filter, filter, filter, sp for left arg and sp for right arg. Or the other way round. At the bottom it'll swap join with filter, but as it unwinds a 2nd time, the local filter var will be the old one it had on the way down, not the new one that it has been replaced by.

jeenbroekstra (approved):

PR #4285, opened 2022-11-19
small tweaks to Dockerfile and build script

jeenbroekstra:
- use apt-get instead of apt in Dockerfile
- ensure the target of the COPY command ends on a / to indicate it's a directory
- remove parallel build setting from maven to ensure stable build

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,7 @@
 # Temp to reduce image size
 FROM ubuntu:jammy AS temp
 
-RUN apt update && apt install unzip
+RUN apt-get update && apt-get install unzip
 
 COPY ignore/rdf4j.zip /tmp/rdf4j.zip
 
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN	rm -rf /usr/local/tomcat/webapps/* && \
 	chown -R tomcat /var/rdf4j /usr/local/tomcat && \
 	chmod a+x /usr/local/tomcat /usr/local/tomcat/bin /usr/local/tomcat/bin/catalina.sh
 
-COPY --from=temp /tmp/eclipse-rdf4j*/war/*.war /usr/local/tomcat/webapps
+COPY --from=temp /tmp/eclipse-rdf4j*/war/*.war /usr/local/tomcat/webapps/
```
jeenbroekstra, line 26: I'm not sure why this doesn't trip anyone else up when using this Dockerfile, but on my machine it consistently fails on this command when there's no trailing slash. I'm running Docker version 20.10.21, btw.

barthanssens: Hmz indeed, according to the documentation it actually shouldn't work without the slash ...
FWIW: also using 20.10.21, on Debian

barthanssens (approved):
LGTM

PR #4325, opened 2022-12-16
GH-4324 update pr verify actions

### LINKED ISSUE #4324: Check formatting in separate github action and update java maven cache action
### Problem description

Would make the github actions a bit faster

### Preferred solution

_No response_

Yes

_No response_

### Anything else?

_No response_

JervenBolleman:
GitHub issue resolved: #4324 

Use a new java action that has maven caching build in.
Also move the formatting check into one place that is named after it.

----

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -3,6 +3,33 @@ name: PR verify
 on: pull_request
 
 jobs:
+  formatting:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jdk: 18
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.jdk }}
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-verify.yml
```diff
@@ -3,6 +3,25 @@ name: PR verify
...
+        cache: 'maven'
+    - name: Check formatting
+      run: mvn -B formatter:validate impsort:check xml-format:xml-check
+    - name: Cancel workflow on failure
+      uses: vishnudxb/cancel-workflow@v1.2
+      if: failure()
+      with:
+        repo: eclipse/rdf4j
+        workflow_id: ${{ github.run_id }}
+        access_token: ${{ github.token }}
```
jeenbroekstra, line 24: I think that if we make this its own job, we can get rid of this cancel step.

JervenBolleman: Yes, but don't we want to cancel the other jobs because they will need to run with the next push that has a formatting fix? So cancelling now saves some energy?

jeenbroekstra: Ah good point. I had it in my head that this would only cancel the current job - but if it also kills the parallel jobs then it's definitely worth leaving in.

abrokenjester (approved):

aschwarte10:
I think this old PR is obsolete by now: we do have  an action in place

grafik

I propose to cleanup and close this old PR

If you think there is something worth keeping here, please reopen and rebase against latest develop

PR #4326, opened 2022-12-16
GH-2061 on top of GH-3843, new collection factory on top of MapDb3 wi…

JervenBolleman:
GitHub issue resolved: #2061
This introduces a new collection factory backed by mapdb3.
However, at this point it is a new option and not yet configured or used by any specific sail.

----

### DIFF
--- a/core/collection-factory/api/src/main/java/org/eclipse/rdf4j/collection/factory/api/CollectionFactory.java
+++ b/core/collection-factory/api/src/main/java/org/eclipse/rdf4j/collection/factory/api/CollectionFactory.java
@@ -19,7 +19,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.function.ToIntBiFunction;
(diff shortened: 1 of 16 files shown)


JervenBolleman:
@barthanssens as you looked at upgrading to MapDB v3 last, this is part of how the upgrade would look using the collection-factory-api approach.

barthanssens:
Thanks Jerven. Last time I've checked, the performance of MapDB 3.0 was less than v1.0 for our purposes... is this still the case ?

JervenBolleman:
With a bit more code this can be made about twice as fast. Mostly by being smarter in the serialization logic.
At this point I wanted to show a minimal patch before adding all that complication.

JervenBolleman:
@barthanssens  would you mind having a look now? MapDb3 is not enabled by default now. But I think with the serializer changes this will be faster than before.

JervenBolleman:
So the main difference in performance in mapdb1/3 is the change of per default not using a mmap'ed backing file.

Re-enabling those is the difference is in favor of the MapDB3 version as implemented.

JervenBolleman:
@hmottestad FYI I hope to merge this once 4.3.0 is out of the door. While not enabled by default, I think it is best to not include last minute additions in the 4.3.0 release.

Issue #4290, opened 2022-11-21
Not being able to evaluate query when COUNT and ORDER BY used together in a single statement.


MartinaDimova:
### Current Behavior

Following query:
```
select (count(*) as ?c) where { 
	?s ?p ?o .
} 
order by (?s)
```
is not evaluated.

### Expected Behavior

A behavior of one of the functions /COUNT or ORDER BY/ is expected.

### Version

4.2.0

### Anything else?

A test class is created to check the behavior of a such a query, both tests are failing with NullPointerException in rdf4j project.

```
public class AggregateTestCountOrderBy {

	private static SailRepository rep;

	private String query = "select (count(<http://example/book7>) as ?c) where { <http://example/book7> ?p ?o . } order by ?c ";


	@BeforeAll
	public static void setUp() throws IOException {
		rep = new SailRepository(new MemoryStore());
		addData();
	}

	private static void addData() throws IOException {
		try (RepositoryConnection conn = rep.getConnection()) {
			conn.add(new StringReader(" <http://example/book1> <urn:n> \"12.5\"^^xsd:float .\n"
					+ "    <http://example/book1> <urn:n1> \"12.5\"^^xsd:float .\n"
					+ "    <http://example/book1> <urn:n1> \"12.6\"^^xsd:float .\n"
					+ "    <http://example/book2> <urn:n> \"6\"^^xsd:int .\n"
					+ "    <http://example/book3> <urn:n>  \"12.5\"^^xsd:double .\n"
					+ "    <http://example/book4> <urn:n>  31.3 .\n"
					+ "    <http://example/book5> <urn:n>  \"311.11241cawda3\" .\n"
					+ "    <http://example/book5> <urn:n1> 311 .\n"
					+ "    <http://example/book6> <urn:n>  0.090000200001 .\n"
					+ "    <http://example/book7> <urn:n>  31.3 .\n"
					+ "    <http://example/book7> <urn:n>  29.243564 .\n"
					+ "    <http://example/book8> <urn:n>  3 ."), "", RDFFormat.TURTLE);
		}
	}

	@Test
	public void testCountOrderBy_ShouldItEvaluateByOrderBy() {
		try (RepositoryConnection conn = rep.getConnection()) {
			TupleQuery tupleQuery = conn.prepareTupleQuery(query);
			try (TupleQueryResult result = tupleQuery.evaluate()) {
				assertThat(result.next().getValue("c").stringValue()).isEqualTo("http://example/book7");
[...]

PR #4327, opened 2022-12-17
GH-4290 fix for count and order by with an implicit group clause

jeenbroekstra:
GitHub issue resolved: #4290  

- the logic to reorder extensions for SELECT expressions and the order clause now takes into account that a new implicit group node may have been introduced
- added regression test
- rewrote all tests in AggregateTest to use AssertJ.

----

### DIFF
--- a/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/AggregateTest.java
+++ b/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/AggregateTest.java
@@ -11,12 +11,11 @@
 package org.eclipse.rdf4j.testsuite.sparql.tests;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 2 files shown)


jeenbroekstra:
@JervenBolleman I think I managed to find a fix for the problem. Would appreciate your eyes on it. I'd still be happy to catch up live and talk through it if you want - let me know.

testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/AggregateTest.java
```diff
@@ -432,48 +422,71 @@ public void testDistinctMax() {
...
+
+		TupleQuery preparedQuery = conn.prepareTupleQuery(query);
+
+		List<BindingSet> result = QueryResults.asList(preparedQuery.evaluate());
+		assertThat(result).hasSize(1);
+
+		BindingSet bs = result.get(0);
+		assertThat(bs.size()).isEqualTo(1);
+		assertThat(getIntValue(bs.getValue("c"), 0)).isEqualTo(19);
 	}
```
jeenbroekstra, line 455: I got a bit carried away refactoring the tests, so the diff is a bit larger than ideal. This bit here is the actual new regression test.

JervenBolleman (approved):
LGTM, @jeenbroekstra  thank you. This is similar to what I thought the solution should be so no need to talk me through it.

PR #4330, opened 2022-12-18
GH-4232 performance issue in VALUES clause query

### LINKED ISSUE #4232: Extreme Performance Degradation in Large TupleQueryResult Iteration in 3.7.4/4.1.0 vs 2.5.5
### Current Behavior

In versions 3.7.4 and 4.1.0, iterating over a TupleQueryResult with large results is extremely slow when compared to older RDF4j versions (2.5.5).

With newer RDF4j versions, running the following query over[ the CHEBI Lite ontology](https://www.ebi.ac.uk/chebi/downloadsForward.do) stored in a graph takes over an hour to process results whereas in RDF4j 2.5.5 it only takes 4s.

```
prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix owl: <http://www.w3.org/2002/07/owl#>

select ?parent ?child
where {
    ?parent rdf:type owl:Class .
    optional {
        ?child rdfs:subClassOf ?parent ;
               rdf:type owl:Class .
    }
}
```

See https://github.com/daltontc/rdf4jTest/tree/bug/tuplequeryresults_slow_iteration for an example

### Expected Behavior

The time it takes to iterate over large query results should be relatively consistent between versions, if not better in later versions.

### Steps To Reproduce

1. Clone https://github.com/daltontc/rdf4jTest/tree/bug/tuplequeryresults_slow_iteration (requires Git LFS due to large trig file)
2. Run Main with 2.5.5
3. Results are iterated over in 4s
```
Total query time: 4228
Parent Count: 302455
Child Count: 122258
```
4. Switch version to 3.7.4 or 4.1.0
5. Run Main
6. Results take ages to completely iterate over

### Version

3.7.4, 4.1.0

_No response_

### Anything else?

_No response_

abrokenjester:
GitHub issue resolved: #4232 

-  ensure VALUES clause always covers entire group graph pattern. This is the generically correct way to parse VALUES clauses. An optimizer can potentially look at the ordering in the algebra to push the values clause down into the join tree (by inspecting which parts of the tree have variables bound in the VALUES clause).
- adding benchmarks 

----

### DIFF
--- a/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/benchmark/SPARQLValuesClauseBenchmark.java
+++ b/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/benchmark/SPARQLValuesClauseBenchmark.java
@@ -1,0 +1,169 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *******************************************************************************/
+
+package org.eclipse.rdf4j.sail.nativerdf.benchmark;
+
+import static org.eclipse.rdf4j.model.util.Values.iri;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.io.IOUtils;
+import org.eclipse.rdf4j.common.transaction.IsolationLevels;
+import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.model.util.ModelBuilder;
+import org.eclipse.rdf4j.model.vocabulary.OWL;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.repository.sail.SailRepository;
+import org.eclipse.rdf4j.repository.sail.SailRepositoryConnection;
+import org.eclipse.rdf4j.sail.nativerdf.NativeStore;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/**
+ * @author Jeen Broekstra
+ */
+@State(Scope.Benchmark)
+@Warmup(iterations = 5)
+@BenchmarkMode({ Mode.AverageTime })
+@Fork(value = 1, jvmArgs = { "-Xms1G", "-Xmx1G" })
+//@Fork(value = 1, jvmArgs = {"-Xms1G", "-Xmx1G", "-XX:+UnlockCommercialFeatures", "-XX:StartFlightRecording=delay=60s,duration=120s,filename=recording.jfr,settings=profile", "-XX:FlightRecorderOptions=samplethreads=true,stackdepth=1024", "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"})
+@Measurement(iterations = 5)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class SPARQLValuesClauseBenchmark {
+
+	@Rule
+	public TemporaryFolder tempDir = new TemporaryFolder();
+
+	private SailRepository repository;
+
+	private static final String query_with_values_clause;
+	private static final String query_without_values_clause;
+
+	static {
+		try {
+			query_with_values_clause = IOUtils.toString(
+					getResourceAsStream("benchmarkFiles/query-values.qr"), StandardCharsets.UTF_8);
+
+			query_without_values_clause = IOUtils.toString(
+					getResourceAsStream("benchmarkFiles/query-without-values.qr"), StandardCharsets.UTF_8);
+
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	public static void main(String[] args) throws RunnerException, IOException, InterruptedException {
+		Options opt = new OptionsBuilder()
+				.include("SPARQLValuesClauseBenchmark") // adapt to run other benchmark tests
+				// .addProfiler("stack", "lines=20;period=1;top=20")
+				.forks(1)
+				.build();
+
+		new Runner(opt).run();
+
+	}
+
+	@Setup(Level.Trial)
+	public void beforeClass() throws IOException, InterruptedException {
+		tempDir.create();
+		File file = tempDir.newFolder();
+
+		repository = new SailRepository(new NativeStore(file, "spoc,posc,cspo,opsc"));
+
+		int numberOfItems = 10_000;
+		int numberOfChildren = 10_000;
+
+		try (var conn = repository.getConnection()) {
+			conn.begin(IsolationLevels.NONE);
+			for (int i = 0; i < numberOfItems; i++) {
+
+				var parent = iri("http://example.org/parent_" + i);
+
+				Model m = new ModelBuilder().setNamespace(OWL.NS)
+						.setNamespace("ex", "http://example.org/")
+						.subject(parent)
+						.add(RDF.TYPE, OWL.CLASS)
+						.add(RDF.TYPE, RDFS.CLASS)
+						.add(RDFS.LABEL, "parent " + i)
+						.build();
+				conn.add(m);
+				if (i % 2 == 0) {
+					for (int j = 0; j < numberOfChildren; j++) {
+						m = new ModelBuilder().setNamespace(OWL.NS)
+								.setNamespace("ex", "http://example.org/")
+								.subject("ex:child_" + i + "_" + j)
+								.add(RDF.TYPE, OWL.CLASS)
+								.add(RDF.TYPE, RDFS.CLASS)
+								.add(RDFS.SUBCLASSOF, parent)
+								.add(RDFS.LABEL, "child of " + i)
+								.build();
+						conn.add(m);
+					}
+				}
+			}
+			conn.commit();
+		}
+	}
+
+	@TearDown(Level.Trial)
+	public void afterClass() {
+		repository.shutDown();
+	}
+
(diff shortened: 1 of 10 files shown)


core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/benchmark/SPARQLValuesClauseBenchmark.java
```diff
@@ -0,0 +1,180 @@
...
+@State(Scope.Benchmark)
+@Warmup(iterations = 5)
+@BenchmarkMode({ Mode.AverageTime })
+@Fork(value = 1, jvmArgs = { "-Xms1G", "-Xmx1G" })
+//@Fork(value = 1, jvmArgs = {"-Xms1G", "-Xmx1G", "-XX:+UnlockCommercialFeatures", "-XX:StartFlightRecording=delay=60s,duration=120s,filename=recording.jfr,settings=profile", "-XX:FlightRecorderOptions=samplethreads=true,stackdepth=1024", "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"})
+@Measurement(iterations = 5)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class SPARQLValuesClauseBenchmark {
+
+	@Rule
```
hmottestad, line 64: This is a junit annotation and won't be picked up by JMH. I hadn't actually noticed that this was used in all the other benchmarks too, looks like @barthanssens cleaned up the unit tests and accidentally applied the same cleanup to the benchmarks.

hmottestad:
Results from my laptop:

```
Benchmark                                          Mode  Cnt    Score   Error  Units
SPARQLValuesClauseBenchmark.simpleEquivalentQuery  avgt    5    0.606 ± 0.007  ms/op
SPARQLValuesClauseBenchmark.valuesOptionalQuery    avgt    5  298.065 ± 6.825  ms/op
```

The query with the VALUES clause uses ~300ms per query while the other query uses ~0.6ms.

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/benchmark/SPARQLValuesClauseBenchmark.java
hmottestad: Maybe we can use `weakTempFile​(String suffix)` from here: https://javadoc.io/static/org.openjdk.jmh/jmh-core/1.36/org/openjdk/jmh/util/FileUtils.html

jeenbroekstra: Oooh good spot! I had mindlessly copy pasted this .

jeenbroekstra:
On my laptop:

```
Benchmark                                          Mode  Cnt    Score    Error  Units
SPARQLValuesClauseBenchmark.simpleEquivalentQuery  avgt    5    0.956 ±  0.093  ms/op
SPARQLValuesClauseBenchmark.valuesOptionalQuery    avgt    5  452.636 ± 23.386  ms/op
```

Looks like you have a slightly better laptop :)

jeenbroekstra:
Query explanations at Executed level:

Simple query:

````
Projection (resultSizeActual=505)
╠══ProjectionElemList
║     ProjectionElem "parent"
║     ProjectionElem "child"
╚══LeftJoin (LeftJoinIterator) (resultSizeActual=505)
   ├──Join (JoinIterator) (resultSizeActual=504)
   │  ╠══BindingSetAssignment ([[type=http://www.w3.org/2002/07/owl#Class]]) (costEstimate=1, resultSizeEstimate=1, resultSizeActual=1)
   │  ╚══StatementPattern (costEstimate=504, resultSizeEstimate=504, resultSizeActual=504)
   │        Var (name=parent)
   │        Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │        Var (name=type, value=http://www.w3.org/2002/07/owl#Class)
   └──Join (JoinIterator) (resultSizeActual=2)
      ╠══StatementPattern (costEstimate=11, resultSizeEstimate=508, resultSizeActual=254.0K)
      ║     Var (name=child)
      ║     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
      ║     Var (name=type)
      ╚══StatementPattern (costEstimate=1, resultSizeEstimate=9.9K, resultSizeActual=2)
            Var (name=child)
            Var (name=_const_4592be07_uri, value=http://www.w3.org/2000/01/rdf-schema#subClassOf, anonymous)
            Var (name=parent)

```

Values optional query:

```
Projection (resultSizeActual=505)
╠══ProjectionElemList
║     ProjectionElem "parent"
║     ProjectionElem "child"
╚══LeftJoin (LeftJoinIterator) (resultSizeActual=505)
   ├──Join (JoinIterator) (resultSizeActual=504)
   │  ╠══BindingSetAssignment ([[type=http://www.w3.org/2002/07/owl#Class]]) (costEstimate=1, resultSizeEstimate=1, resultSizeActual=1)
   │  ╚══StatementPattern (costEstimate=504, resultSizeEstimate=504, resultSizeActual=504)
   │        Var (name=parent)
   │        Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │        Var (name=type, value=http://www.w3.org/2002/07/owl#Class)
   └──Join (JoinIterator) (resultSizeActual=2)
      ╠══StatementPattern (costEstimate=11, resultSizeEstimate=508, resultSizeActual=254.0K)
      ║     Var (name=child)
      ║     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
      ║     Var (name=type)
      ╚══StatementPattern (costEstimate=1, resultSizeEstimate=9.9K, resultSizeActual=2)
            Var (name=child)
            Var (name=_const_4592be07_uri, value=http://www.w3.org/2000/01/rdf-schema#subClassOf, anonymous)
            Var (name=parent)
```

jeenbroekstra:
Query explanation simple query:

```
Projection (resultSizeActual=505)
╠══ProjectionElemList
║     ProjectionElem "parent"
║     ProjectionElem "child"
╚══LeftJoin (LeftJoinIterator) (resultSizeActual=505)
   ├──StatementPattern (resultSizeEstimate=504, resultSizeActual=504)
   │     Var (name=parent)
   │     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │     Var (name=_const_7e619dcc_uri, value=http://www.w3.org/2002/07/owl#Class, anonymous)
   └──Join (JoinIterator) (resultSizeActual=2)
      ╠══StatementPattern (costEstimate=50, resultSizeEstimate=9.9K, resultSizeActual=2)
      ║     Var (name=child)
      ║     Var (name=_const_4592be07_uri, value=http://www.w3.org/2000/01/rdf-schema#subClassOf, anonymous)
      ║     Var (name=parent)
      ╚══StatementPattern (costEstimate=1, resultSizeEstimate=504, resultSizeActual=2)
            Var (name=child)
            Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
            Var (name=_const_7e619dcc_uri, value=http://www.w3.org/2002/07/owl#Class, anonymous)
```

Query explanation values clause query:

```
Projection (resultSizeActual=505)
╠══ProjectionElemList
║     ProjectionElem "parent"
║     ProjectionElem "child"
╚══LeftJoin (LeftJoinIterator) (resultSizeActual=505)
   ├──Join (JoinIterator) (resultSizeActual=504)
   │  ╠══BindingSetAssignment ([[type=http://www.w3.org/2002/07/owl#Class]]) (costEstimate=1, resultSizeEstimate=1, resultSizeActual=1)
   │  ╚══StatementPattern (costEstimate=504, resultSizeEstimate=504, resultSizeActual=504)
   │        Var (name=parent)
   │        Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │        Var (name=type, value=http://www.w3.org/2002/07/owl#Class)
   └──Join (JoinIterator) (resultSizeActual=2)
      ╠══StatementPattern (costEstimate=11, resultSizeEstimate=508, resultSizeActual=254.0K)
      ║     Var (name=child)
      ║     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
      ║     Var (name=type)
      ╚══StatementPattern (costEstimate=1, resultSizeEstimate=9.9K, resultSizeActual=2)
            Var (name=child)
            Var (name=_const_4592be07_uri, value=http://www.w3.org/2000/01/rdf-schema#subClassOf, anonymous)
            Var (name=parent)
```

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/benchmark/SPARQLValuesClauseBenchmark.java
barthanssens: autch, sorry about that @hmottestad

jeenbroekstra:
I still can't see exactly what goes wrong, other than that it just seems to be a pathological case in terms of the data shape. I'm starting to wonder if we should have an optimizer that just duplicates the bindingsetassignment clause into the right arg of the left join. We can't pre-bind values in the right arg of the left join, but perhaps if we optimize to something like this:

```
Projection (resultSizeActual=505)
╠══ProjectionElemList
║     ProjectionElem "parent"
║     ProjectionElem "child"
╚══LeftJoin (LeftJoinIterator) (resultSizeActual=505)
   ├──Join (JoinIterator) (resultSizeActual=504)
   │  ╠══BindingSetAssignment ([[type=http://www.w3.org/2002/07/owl#Class]]) (costEstimate=1, resultSizeEstimate=1, resultSizeActual=1)
   │  ╚══StatementPattern (costEstimate=504, resultSizeEstimate=504, resultSizeActual=504)
   │        Var (name=parent)
   │        Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
   │        Var (name=type, value=http://www.w3.org/2002/07/owl#Class)
   └──Join (JoinIterator) (resultSizeActual=2)
      ╠══ Join
           -> BindingSetAssignment ([[type=http://www.w3.org/2002/07/owl#Class]]) (costEstimate=1, resultSizeEstimate=1, resultSizeActual=1)
           ╠══StatementPattern (costEstimate=11, resultSizeEstimate=508, resultSizeActual=254.0K)
           ║     Var (name=child)
           ║     Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
           ║     Var (name=type)
           ╚══StatementPattern (costEstimate=1, resultSizeEstimate=9.9K, resultSizeActual=2)
                   Var (name=child)
                  Var (name=_const_4592be07_uri, value=http://www.w3.org/2000/01/rdf-schema#subClassOf, anonymous)
                 Var (name=parent)
```

(ugly manual editing by me to get the idea across)

I'm not yet sure if that is something that is generally legal tbh. Just recording my thoughts for the next time I revisit.

hmottestad:
In this case I think it's an issue with the join optimizer. The join optimizer doesn't really need to be limited by scoping issues when it calculates cardinality, as long as any optimizations are legal. 

The join optimizer could assume that the binding set assignment will apply and use it to both calculate the size and to ignore the positive effect that binding the `?type` variable early will have.

jeenbroekstra:
> In this case I think it's an issue with the join optimizer. The join optimizer doesn't really need to be limited by scoping issues when it calculates cardinality, as long as any optimizations are legal.
> 
> The join optimizer could assume that the binding set assignment will apply and use it to both calculate the size and to ignore the positive effect that binding the `?type` variable early will have.

I don't quite understand what you're getting at. As far as I can tell the problem is that the join optimizer calculates an initial cost for the `(?child, type, ?type)` pattern that is lower than it should be due to its base cardinality being lower than that of the `(?child, subClassOf, ?parent)` pattern. I'm struggling to see how to accurately compensate here though, to be honest.

PR #4367, opened 2023-01-17
WIP: Migrate javax -> jakarta (Jakarta EE 9)

erikgb:
GitHub issue resolved: # 

----

### DIFF
--- a/.github/workflows/develop-status.yml
+++ b/.github/workflows/develop-status.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        jdk: [11, 19]
+        jdk: [17, 19]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/main-status.yml
+++ b/.github/workflows/main-status.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        jdk: [11, 19]
+        jdk: [17, 19]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        jdk: [11, 19]
+        jdk: [17, 19]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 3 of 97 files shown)


.github/workflows/develop-status.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        jdk: [11, 19]
+        jdk: [17, 19]
```
jeenbroekstra, line 14: I don't think that's a migration we can do at this point - there's significant subset of our users that require support for Java 11.

erikgb: I see and can understand why. 😃 Could it be an option to require Java 17 to use the modules with spring/spring-boot dependencies - if I can find a way to configure it?

JervenBolleman: With RDF4j V5 we can all go to java 17 IMHO.  Maybe it is better I open that in a discussion topic though.

erikgb: > With RDF4j V5 we can all go to java 17 IMHO. Maybe it is better I open that in a discussion topic though.

I hope so. I can start a discussion about this. It is a requirement for the javax -> jakarta migration that blocks a lot of other upgrades like servlet containers (Jetty, Tomcat), Spring/Spring Boot+++

jetztgradnet: why would you need to bump this to Java 17? RDF4J compiled for Java 11 can still be *used* with Java 17, no?

erikgb: Spring Boot 3.x and Spring 6.x requires Java 17. And without upgrading Spring we cannot migrate from javax to jakarta. This is WIP. We need to find a smart way of doing this....

jetztgradnet: Sorry, I had a misunderstanding: I was thinking that some (external) Spring app using Java 17 would not require RDF4J also being compiled for Java 17. But you are actually talking about the RDF4J workbench using Spring inside, so RDF4J depending on Spring, not the other way round, right? 

So maybe we create a separate module that is using a Spring6/Spring Boot 3/Java 17 variant of the workbench?

erikgb: The problem is the dependencies. You must choose, either javax or jakarta. Never both.

Issue #3559, opened 2021-12-30
Move to Jakarta EE


hmottestad:
We've already moved JAXB, but there are still a few references to other EE stuff that should be moved to Jakarta. IntelliJ has a migration for this which will hopefully help.

hmottestad:
We need to update:
 - jaxb to 3.0.1

hmottestad:
After giving this a shot I've decided it's far too much work to make sure that we don't break compatibility with existing application servers.

erikgb:
@hmottestad Is there any chance we could bring this back on the table, probably in a new major version of rdf4j? I am trying to migrate a spring-boot project (with rdf4j dependencies) to spring-boot 3.x, and it requires the javax -> jakarta migration. So this issue is a total blocker.....

hmottestad:
@erikgb Last I looked into this it seemed like a fairly large task. Would you be interested in contributing a fix yourself?

Do you know which modules your code depends on? Maybe there is a chance that the scope can be reduced to a smaller part of the RDF4J codebase.

erikgb:
> Would you be interested in contributing a fix yourself?

I could give it a shot.

> Do you know which modules your code depends on? Maybe there is a chance that the scope can be reduced to a smaller part of the RDF4J codebase.

We depend on quite a lot, and I don't think this migration can be limited to a part of the codebase. Here is an example of the stack-traces I get:

```
java.lang.NoClassDefFoundError: javax/xml/bind/annotation/adapters/HexBinaryAdapter
	at org.eclipse.rdf4j.rio.helpers.AbstractRDFParser.createNode(AbstractRDFParser.java:413)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.createNode(RDFXMLParser.java:802)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.getNodeResource(RDFXMLParser.java:514)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.processNodeElt(RDFXMLParser.java:414)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.startElement(RDFXMLParser.java:331)
	at org.eclipse.rdf4j.rio.rdfxml.SAXFilter.reportDeferredStartElement(SAXFilter.java:303)
	at org.eclipse.rdf4j.rio.rdfxml.SAXFilter.startElement(SAXFilter.java:234)
```

erikgb:
[...]

PR #4372, opened 2023-01-18
refactor: eliminate commons-fileupload dependency

erikgb:
and replace it with multipart support added in Servlet 3.0.

GitHub issue resolved: # 

commons-fileupload hasn't had any release since 2018, and is not really required - as Java EE added support for multipart handling in Servlet 3.0, ref. https://docs.oracle.com/javaee/7/tutorial/servlets011.htm

Removing the dependency to commons-fileupload is a pre-requirement to the javax -> jakarta migration proposed in https://github.com/eclipse/rdf4j/issues/3559.

I did the simplest thing to remove the dependency, and IMO the affected code should be refactored more, but I am leaving this to a follow-up PR.

----

### DIFF
--- a/tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/base/TransformationServlet.java
+++ b/tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/base/TransformationServlet.java
@@ -19,6 +19,7 @@ import java.util.Map;
 
 import javax.servlet.ServletConfig;
 import javax.servlet.ServletException;
+import javax.servlet.annotation.MultipartConfig;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
@@ -31,6 +32,7 @@ import org.eclipse.rdf4j.workbench.util.WorkbenchRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@MultipartConfig
 public abstract class TransformationServlet extends AbstractRepositoryServlet {
 
 	protected static final ParserConfig NON_VERIFYING_PARSER_CONFIG;
--- a/pom.xml
+++ b/pom.xml
@@ -391,11 +391,6 @@
 				<version>1.4</version>
 			</dependency>
 			<dependency>
-				<groupId>commons-fileupload</groupId>
-				<artifactId>commons-fileupload</artifactId>
-				<version>1.3.3</version>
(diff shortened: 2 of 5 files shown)


tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/base/TransformationServlet.java
```diff
@@ -31,6 +32,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@MultipartConfig
```
erikgb, line 35: I suspect this annotation is not inherited, and would have to be added to all subclasses. But hoping tests and/or maintainers can verify if this works.

jeenbroekstra: I'm not familiar with this annotation. What does it do?

jeenbroekstra (commented):
Looks good to me! I haven't been able to test it yet (Can you perhaps add a few test to cover this behavior? I'm aware test coverage on the workbench is currently poor), but the idea looks solid enough. And eliminating an unneeded/obsolete dependency is a good idea.

erikgb:
> Looks good to me! I haven't been able to test it yet (Can you perhaps add a few test to cover this behavior? I'm aware test coverage on the workbench is currently poor), but the idea looks solid enough. And eliminating an unneeded/obsolete dependency is a good idea.

Thanks. Apart from verifying the new multipart setup, I have tried to do a safe refactoring, so I do not expect any issues when multipart request works at all.

Sorry, I probably don't have the bandwidth to add additional tests now. According to @hmottestad there are some tests that could be enabled in CI, ref. https://github.com/eclipse/rdf4j/discussions/4370#discussioncomment-4717478

tools/workbench/src/main/java/org/eclipse/rdf4j/workbench/base/TransformationServlet.java
erikgb: It adds multipart request support to the servlet. Can also be configured in web.xml. I am happy to do that instead, but I need to know which servlets need it - as this is configured per servlet. See https://docs.oracle.com/javaee/7/tutorial/servlets011.htm for additional details.

erikgb:
Superseeded by https://github.com/eclipse/rdf4j/pull/4375, closing.

Issue #4403, opened 2023-02-05
Migrate WireMock tests to JUnit 5


erikgb:
### Problem description

We still use JUnit 4 integration with WireMock. This is unfortunate, as we still need to have module dependencies to vintage JUnit. This should be fairly easy to fix, ref. https://wiremock.org/docs/junit-jupiter/.

### Preferred solution

Migrate to the WireMock JUnit 5 integration.

### Are you interested in contributing a solution yourself?

Yes

PR #4404, opened 2023-02-05
GH-4403: Migrate WireMock tests to JUnit 5

erikgb:
GitHub issue resolved: #4403 

This migrates WireMock tests to JUnit 5 using the documented approach: https://wiremock.org/docs/junit-jupiter/.

Also upgrading WireMock to latest stable version to make the JUnit 5 integration available.

----

### DIFF
--- a/core/repository/manager/src/test/java/org/eclipse/rdf4j/repository/manager/RepositoryManagerTest.java
+++ b/core/repository/manager/src/test/java/org/eclipse/rdf4j/repository/manager/RepositoryManagerTest.java
@@ -24,8 +24,8 @@ import org.eclipse.rdf4j.repository.Repository;
 import org.eclipse.rdf4j.repository.RepositoryException;
 import org.eclipse.rdf4j.repository.config.RepositoryConfig;
 import org.eclipse.rdf4j.repository.config.RepositoryConfigException;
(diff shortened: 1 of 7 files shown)


core/repository/manager/src/test/java/org/eclipse/rdf4j/repository/manager/RepositoryManagerTest.java
```diff
@@ -36,7 +36,7 @@ public class RepositoryManagerTest {
 
 	protected RepositoryManager subject;
 
-	@Before
+	@BeforeEach
```
erikgb, line 39: Should this test class be redeclared as abstract and this setUp method deleted? I think so, but there might be a reason for having it like this? A bit confusing now with methods that are semi-overidden....

hmottestad:
I think that maybe this should be targeted at develop instead of main since it's not strictly a bug fix.

abrokenjester:
> I think that maybe this should be targeted at develop instead of main since it's not strictly a bug fix.

This one doesn't touch any production/release code though, so arguably it can fit in either. But given that the other PR touches the testsuites, which are released artifacts, it might be good for consistency sake to merge _all_ JUnit 5 stuff to develop instead of main.

Issue #4421, opened 2023-02-13
Remove dependency to javax.xml.bind from core libraries


jetztgradnet:
### Problem description

We are in process of moving our application to JakartaEE and Jetty11. One blocker is the use of `javax.xml.bind` in the RDF4J core libraries (we do not use the workbench or server parts of RDF4J). 

A full move to JakartaEE is a huge topic (see #3559 and the WIP PR #4367) and staying compatible with both JavaEE and JakartaEE at the same time impossible or at least very difficult with additional variants.

We are looking for a minimal short-term solution in the 4.2.x/4.3 timeframe within the next few weeks.

### Preferred solution

We would like to get rid of the few uses of `javax.xml.bind` in the core libraries.

There are currently only 3 uses of `javax.xml.bind`:
* `org.eclipse.rdf4j.rio.helpers.AbstractRDFParser` uses `javax.xml.bind.annotation.adapters.HexBinaryAdapter` which (here) is a thin wrapper around `javax.xml.bind.DatatypeConverter.printHexBinary(byte[])`
* `org.eclipse.rdf4j.http.protocol.transaction.TransactionSAXParser` uses `javax.xml.bind.DatatypeConverter.parseBase64Binary(String)` (fully qualified class reference in line 96, not using an import)
* `org.eclipse.rdf4j.http.protocol.transaction.TransactionWriter` uses `javax.xml.bind.DatatypeConverter.printBase64Binary(byte[])`

Example of an exception at runtime (also see #3559):
```
java.lang.NoClassDefFoundError: javax/xml/bind/annotation/adapters/HexBinaryAdapter
	at org.eclipse.rdf4j.rio.helpers.AbstractRDFParser.createNode(AbstractRDFParser.java:413)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.createNode(RDFXMLParser.java:802)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.getNodeResource(RDFXMLParser.java:514)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.processNodeElt(RDFXMLParser.java:414)
	at org.eclipse.rdf4j.rio.rdfxml.RDFXMLParser.startElement(RDFXMLParser.java:331)
	at org.eclipse.rdf4j.rio.rdfxml.SAXFilter.reportDeferredStartElement(SAXFilter.java:303)
[...]

PR #4424, opened 2023-02-13
GH-4421: Eliminate dependencies to javax.xml.bind from core libraries

erikgb:
GitHub issue resolved: #4421

This PR eliminates dependencies to` javax.xml.bind` from core libraries. Main motivation being to ease the javax->jakarta migration, ref. #3559.

- Base64 encoding now uses the standard Java library support added in Java 8
- To perform Hex encoding, I am proposing to use commons-codec for now,. But it should be fairly simple to write our own utility method if that is an unwanted dependency. And migrate to the standard Java library support added in Java 17 - once Java 17 becomes the minimum supported version.

----

Kudos to @aschwarte10 for the test additions.

### DIFF
--- a/core/http/protocol/pom.xml
+++ b/core/http/protocol/pom.xml
@@ -31,9 +31,9 @@
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
-			<groupId>jakarta.xml.bind</groupId>
-			<artifactId>jakarta.xml.bind-api</artifactId>
-			<version>${jaxb.version}</version>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-params</artifactId>
+			<scope>test</scope>
 		</dependency>
 	</dependencies>
 </project>
--- a/core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/transaction/TransactionSAXParser.java
+++ b/core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/transaction/TransactionSAXParser.java
@@ -11,6 +11,7 @@
 package org.eclipse.rdf4j.http.protocol.transaction;
 
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -93,7 +94,7 @@ class TransactionSAXParser extends SimpleSAXAdapter {
 			String encoding = atts.get(TransactionXMLConstants.ENCODING_ATT);
 
 			if (encoding != null && "base64".equalsIgnoreCase(encoding)) {
-				text = new String(javax.xml.bind.DatatypeConverter.parseBase64Binary(text));
+				text = new String(Base64.getDecoder().decode(text));
 			}
 			Literal lit;
 			if (lang != null) {
--- a/core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/transaction/TransactionWriter.java
+++ b/core/http/protocol/src/main/java/org/eclipse/rdf4j/http/protocol/transaction/TransactionWriter.java
@@ -13,8 +13,7 @@ package org.eclipse.rdf4j.http.protocol.transaction;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-
-import javax.xml.bind.DatatypeConverter;
+import java.util.Base64;
 
 import org.eclipse.rdf4j.common.xml.XMLUtil;
 import org.eclipse.rdf4j.common.xml.XMLWriter;
@@ -299,7 +298,7 @@ public class TransactionWriter {
 
 			if (!valid) {
 				xmlWriter.setAttribute(TransactionXMLConstants.ENCODING_ATT, "base64");
-				label = DatatypeConverter.printBase64Binary(label.getBytes(StandardCharsets.UTF_8));
+				label = Base64.getEncoder().encodeToString(label.getBytes(StandardCharsets.UTF_8));
 			}
 
 			xmlWriter.textElement(TransactionXMLConstants.LITERAL_TAG, label);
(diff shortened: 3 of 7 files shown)


aschwarte10 (commented):
Thanks @erikgb for driving this 👍 

For me the approach chosen generally makes sense.

You are currently targeting the `main` branch (i.e. a patch release like 4.2.x). I am not sure whether we can add a new dependency to one of the modules in a patch release. We for instance do resolution of transitive dependencies manually by listing them explicitly (i.e. we do not rely on the modules POM). Maybe it would be better to target the develop branch and go for a 4.3.0 release?

In addition: did you check that unit test cases exist? At least for  `AbstractRDFParser#createNode(String nodeID)` it should be straight forward to create them (to check before / after)

Thanks

jetztgradnet:
Thanks, @erikgb, for approaching this so quickly!
I like your approach, using the JDK where possible and Apache Commons Codec is a good approach for this. As @aschwarte10 mentioned, we might rather target 4.3.0 as this now requires a new (or different) dependency which was not used for this module before.

@hmottestad @abrokenjester any chance of getting this into a 4.3.0 branch and milestone build rather sooner than later?

erikgb:
> You are currently targeting the `main` branch (i.e. a patch release like 4.2.x). I am not sure whether we can add a new dependency to one of the modules in a patch release. We for instance do resolution of transitive dependencies manually by listing them explicitly (i.e. we do not rely on the modules POM). Maybe it would be better to target the develop branch and go for a 4.3.0 release?

I see. I am happy to do whatever the maintainers think is best. To avoid the new dependency, it should be pretty straight-forward to write a custom method to do hex-encoding.

> In addition: did you check that unit test cases exist? At least for `AbstractRDFParser#createNode(String nodeID)` it should be straight forward to create them (to check before / after)

No, I did not, but good point. I assumed this piece of code had test coverage. Ideally, tests should be added before this PR. Anything you could contribute? I am not sure how the tests would look like....

hmottestad:
The changes look good to me! Thanks for handling this @erikgb . 

As for test coverage, I'm not sure how this exactly impacts things...so it would be great if @jetztgradnet could contribute some tests.

I also agree that this should be merged into the develop branch to be targeted at next minor build.

erikgb:
> I also agree that this should be merged into the develop branch to be targeted at next minor build.

PR updated to target the develop branch.

aschwarte10:
@erikgb how about adding a test as follows to the existing `AbstractParserTest` class? This one is working on current develop and my understanding is that hashing is stable

```
@Test
	public void testCreateNode() throws Exception {

		// node ids look like "genid_.*-suffix
		assertTrue(parser.createNode("someid").stringValue().endsWith("-someid"));

		// some long id (length > 32) => suffix is hashed
		String longNodeId = "someverylongnodeidwithmorethan32characters";
		assertTrue(parser.createNode(longNodeId).stringValue().endsWith("2A372A91878F0980C8F53341D2D8A944"));

	}
```

erikgb:
> @Test
> 	public void testCreateNode() throws Exception {
> 
> 		// node ids look like "genid_.*-suffix
> 		assertTrue(parser.createNode("someid").stringValue().endsWith("-someid"));
> 
> 		// some long id (length > 32) => suffix is hashed
> 		String longNodeId = "someverylongnodeidwithmorethan32characters";
> 		assertTrue(parser.createNode(longNodeId).stringValue().endsWith("2A372A91878F0980C8F53341D2D8A944"));
> 
> 	}

Thanks @aschwarte10! I've added this test now, and it revealed a bug. 😅  Commons-codec seems to encode hex with lowercase characters by default.

Do you have any input to testing the other changes?

aschwarte10:
Testing the `SAXParser` in a unit test is not easy, but for the `TransactionWriter` you could do something like

```
public class TransactionWriterTest {

	@Test
	public void testSerializeLiteral() throws Exception {

		assertSerializedLiteralEquals(
				"<literal datatype='http://www.w3.org/2001/XMLSchema#string'>Hello World</literal>",
				Values.literal("Hello World"));

		assertSerializedLiteralEquals(
				"<literal datatype='http://www.w3.org/2001/XMLSchema#int'>42</literal>",
				Values.literal(42));
		
		// with XML characters
		assertSerializedLiteralEquals(
				"<literal datatype='http://www.w3.org/2001/XMLSchema#string'>with special &lt;&gt; char</literal>",
				Values.literal("with special <> char"));

		// with non valid characters
		char c = '\u0001';
		assertSerializedLiteralEquals(
				"<literal datatype='http://www.w3.org/2001/XMLSchema#string' encoding='base64'>bm9uIHZhbGlkIGNoYXIB</literal>",
				Values.literal("non valid char" + c));
	}

	void assertSerializedLiteralEquals(String expected, Literal literal) throws Exception {
		TransactionWriter writer = new TransactionWriter();

		StringWriter sw = new StringWriter();
		XMLWriter xmlWriter = new XMLWriter(sw);
		writer.serialize(literal, xmlWriter);

		Assertions.assertEquals(expected, sw.toString());
	}
}

```

Can you also please have a look at why the java 19 test is failing?

erikgb:
> Can you also please have a look at why the java 19 test is failing?

I am pretty sure it is a flake. We'll see on the next push, as I don't see a way to retrigger CI jobs. Runs fine locally, and seems totally unrelated to the change.

erikgb:
Thanks again for the test input, @aschwarte10! ❤️ I have added the test cases now, using JUnit 5 parametrized tests (more compact).

erikgb:
For the flaky test, I created https://github.com/eclipse/rdf4j/issues/4434 after observing the same test failing in another PR.

aschwarte10 (approved):
From my side everything looking good. Maybe @hmottestad wants an additional look at this?

jetztgradnet:
Thanks, @erikgb, for your implementation! I will test this in my current migration project.

jetztgradnet:
I tested this branch in my migration project and this solved my issued with the dependencies to `javax.xml.bind`
Would be great to have this merged soon and ideally with a 4.3.0-M1 milestone build published soon after :-)

Issue #3507, opened 2021-12-11
Clean up test server implementations


jeenbroekstra:
The test rig currently has at least four separate implementations of a Jetty-based test server:

- org.eclipse.rdf4j.http.server.TestServer (in rdf4j-http-server unit tests)
- org.eclipse.rdf4j.repository.http.HTTPMemServer (in rdf4j-repository-compliance)
- org.eclipse.rdf4j.query.parser.sparql.SPARQLEmbeddedServer (in rdf4j-sparql-compliance)
- org.eclipse.rdf4j.federated.server.SPARQLEmbeddedServer (in rdf4j-tools-federation)

All four do essentially the same thing: spin up an RDF4J Server on localhost with a few test repositories to run tests over HTTP against. We should unify these into a single reusable test utility.

barthanssens:
Seems to be duplicate of #2933

jeenbroekstra:
> Seems to be duplicate of #2933

Oops! So it is, well spotted. The description on this issue is more up to date so I'll close the other one.

erikgb:
@abrokenjester @barthanssens WDYT about adopting something like https://mjeanroy.dev/junit-servers/ instead of building our own test utility? I did a small POC, and seems to work well.

erikgb:
Here is an MVP POC showing how this could look like: https://github.com/eclipse/rdf4j/pull/4437

PR #4437, opened 2023-02-17
WIP: POC using JUnit server

erikgb:
This is a POC demostrating how https://mjeanroy.dev/junit-servers/jetty/jupiter could fix https://github.com/eclipse/rdf4j/issues/3507.


tools/server/pom.xml
```diff
@@ -113,6 +85,12 @@
 			<artifactId>opencsv</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.github.mjeanroy</groupId>
```
hmottestad, line 89: Licensing could potentially be an issue. This is using the maven test scope, so that means that a superficial license check is all that is needed.

erikgb: Is MIT license ok? https://github.com/mjeanroy/junit-servers/blob/master/LICENSE

abrokenjester: That should be fine (doesn't hurt to double-check with the dash-license tool). 

On a related note, I think the dash-license tool now has a maven plugin that we can add into our workflow so these things are automatically checked and reported. https://github.com/eclipse/dash-licenses

erikgb: It seems like the dash maven plugin is not published anywhere, so it has to be installed locally. But that is a minor issue. I can check it, or does it have to be done by the project maintainers?

erikgb: The new license-check workflow is green: https://github.com/eclipse/rdf4j/pull/4437/checks

But it probably doesn't check test-scoped dependencies, or what? The introduced dependency is not listed in the summary. Anything more to do regarding this? I'm planning to move this POC into a proper PR, ref. https://github.com/eclipse/rdf4j/issues/4454

erikgb:
Closing i favor of https://github.com/eclipse/rdf4j/pull/4464

Issue #4474, opened 2023-03-03
Avoid transitive dependency management in modules using test server


erikgb:
### Problem description

While working on https://github.com/eclipse/rdf4j/pull/4464, I noticed that the two compliance modules that currently use a test server explicitly manage transitive dependencies (from `rdf4j-http-server` module). This is far from optimal and should be improved, and is also a semi-blocker for doing improvements to the test server duplication, ref. https://github.com/eclipse/rdf4j/issues/3507.

### Preferred solution

I suggest to publish a ["classes" attachment](https://maven.apache.org/plugins/maven-war-plugin/war-mojo.html#attachClasses) from the `rdf4j-http-server` that can be used to avoid this duplication of dependency management.

### Are you interested in contributing a solution yourself?

Yes

PR #4475, opened 2023-03-03
GH-4474: Publish classes attachment from war modules

erikgb:
GitHub issue resolved: #4474

By publishing "classes" artifacts from war modules we avoid the duplicated dependency management in modules using the war to run tests with a test server. In this PR I fix the two compliance modules running tests using a test server. The federation module desperately need some care, but I'll leave this for a follow-up PR. I suspect we need to establish a dedicated module for the federation tests to fix this - as the compliance modules are for their corresponding functional module.

----

### DIFF
--- a/compliance/repository/pom.xml
+++ b/compliance/repository/pom.xml
@@ -16,54 +16,7 @@
 	<dependencies>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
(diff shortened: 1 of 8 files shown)


abrokenjester:
Interesting. Not quite sure I fully understand the problem this solves but it certainly looks cleaner this way.

abrokenjester (approved):

PR #4481, opened 2023-03-07
Revert "Bump snakeyaml from 1.26 to 2.0 in /compliance/elasticsearch"

abrokenjester:
Reverts eclipse/rdf4j#4476


abrokenjester:
Reverting because the version bump appears to cause a test failure in the elasticsearch compliance tests.

Issue #4482, opened 2023-03-09
AbstractLiteral toString is comparing a CoreDatatype with an IRI


ate47:
### Current Behavior

at the line 195 we have `datatype.equals(CoreDatatype.XSD.STRING`, but datatype is of type IRI and CoreDatatype.XSD.STRING of type CoreDatatype, making it false all the time.

https://github.com/eclipse/rdf4j/blob/78771b7161f6b5a84b237db597b53a8842935860/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java#L195

Making the toString value of an IRI leading to that:

```turtle
"abdc"^^<http://www.w3.org/2001/XMLSchema#string>
```

### Expected Behavior

printing the literal without the XSD.STRING datatype like with all the other version of the toString
```turtle
"abdc"
```

### Steps To Reproduce

I'm using a custom ValueFactory using the Abstract one, but this one is also doing it:

```java
System.out.println(
  new AbstractValueFactory() {}
     .createLiteral(("abcd"))
);
```

### Version

4.2.3

### Are you interested in contributing a solution yourself?

Yes

abrokenjester:
Thanks @ate47 , good find!

PR #4483, opened 2023-03-09
GH-4482 Replace Datatype IRI to CoreDatatype in AbstractLiteral

ate47:
GitHub issue resolved: #4482

I replace the IRI by CoreDatatype, I hope it is fine without any test knowing how trivial it is.

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/AbstractLiteral.java
@@ -188,14 +188,8 @@ public abstract class AbstractLiteral implements Literal {
 
 				.map(language -> label + '@' + language)
 
-				.orElseGet(() -> {
-
-					final IRI datatype = getDatatype();
-
-					return datatype.equals(CoreDatatype.XSD.STRING) ? label
-							: label + "^^<" + datatype.stringValue() + ">";
-
-				});
+				.orElseGet(() -> CoreDatatype.XSD.STRING == getCoreDatatype() ? label
+						: label + "^^<" + getDatatype().stringValue() + ">");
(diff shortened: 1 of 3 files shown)


abrokenjester (commented):
The fix looks good to me. Given that this is a bug fix could you rebase your fix on the main branch so we can merge it into main instead of develop?

ate47:
> The fix looks good to me. Given that this is a bug fix could you rebase your fix on the main branch so we can merge it into main instead of develop?

done :)

abrokenjester (approved):
Nice work! I'm keeping it open for @hmottestad to take a look as well, since he's more knowledgable than me on how the the CoreDatatype functionality works.

hmottestad:
This could potentially be a performance issue for literals that lazy load the CoreDatatype. I don't think it's a big problem though since the code essentially replaces a string equals. 

Btw. You can use == instead of .equals(...) since it's not allowed to return null as a CoreDatatype.

ate47:
> This could potentially be a performance issue for literals that lazy load the CoreDatatype. I don't think it's a big problem though since the code essentially replaces a string equals.
> 
> Btw. You can use == instead of .equals(...) since it's not allowed to return null as a CoreDatatype.

I've replaced the equals to ==, but you're not understanding the issue I think, the issue wasn't the performances, the issue was as explained in #4482 that the previous code was comparing the datatype IRI to the core datatype. So it was false all the time and was displaying `"aaa"^^xsd:string` with a call to toString

hmottestad:
I understood the issue. It's a good fix. I was just mentioning a possible performance implication, but the alternative is to compare the datatype IRI instead, and that typically results in a string comparison. So I think sticking to CoreDatatype is the best, just like what you have done :)

Issue #4359, opened 2023-01-08
Remove Iteration interface


hmottestad:
Related to #3907

Remove the Iteration interface and migrate to the CloseableIteration interface.

kenwenzel:
@hmottestad I've noticed that `CloseableIteration` is marked as deprecated since 4.1.0. That leads to many warnings in the IDE. Is this due to the discussion if the interface should be kept at all or eventually replaced by an interface directly extending `Iterator`?

KOMMA does also just extend the iterator interface (borrowed from Jena):
https://github.com/komma/komma/blob/master/bundles/core/net.enilink.commons.iterator/src/main/java/net/enilink/commons/iterator/IExtendedIterator.java

hmottestad:
@abrokenjester and I were considering removing the support for checked exceptions (as java generics) and that wouldn't make it backwards compatible. If we are doing that we might as well extend Iterator too.

hmottestad:
### Develop
```
Benchmark                                                     Mode  Cnt     Score    Error  Units
QueryBenchmark.complexQuery                                   avgt    5     1.492 ±  0.032  ms/op
QueryBenchmark.different_datasets_with_similar_distributions  avgt    5     0.767 ±  0.021  ms/op
QueryBenchmark.groupByQuery                                   avgt    5     0.966 ±  0.040  ms/op
QueryBenchmark.long_chain                                     avgt    5   260.424 ±  8.599  ms/op
QueryBenchmark.lots_of_optional                               avgt    5    54.358 ±  0.439  ms/op
QueryBenchmark.minus                                          avgt    5  2350.765 ± 16.433  ms/op
QueryBenchmark.nested_optionals                               avgt    5    74.068 ±  0.110  ms/op
QueryBenchmark.pathExpressionQuery1                           avgt    5     8.003 ±  0.123  ms/op
QueryBenchmark.pathExpressionQuery2                           avgt    5     1.116 ±  0.033  ms/op
QueryBenchmark.query_distinct_predicates                      avgt    5    73.386 ±  0.367  ms/op
QueryBenchmark.simple_filter_not                              avgt    5     3.237 ±  0.061  ms/op
```

PR #4500, opened 2023-04-05
GH-4359 remove iteration interface

hmottestad:
GitHub issue resolved: #4359 

----


compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/ElasticsearchSailTest.java
```diff
@@ -117,7 +117,7 @@ public void testSnippetLimitedToPredicate()
 	}
 
 	@Test
-	public void testGraphQuery() throws QueryEvaluationException, MalformedQueryException, RepositoryException {
+	public void testGraphQuery() throws MalformedQueryException, RepositoryException {
```
hmottestad, line 120: This is some general code cleanup.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
```diff
@@ -29,13 +31,9 @@
...
  *
- * @deprecated In the future this interface will stop extending {@link Iteration} and instead declare the same interface
- *             methods directly. The interface will also stop requiring implementations to automatically close when
- *             exhausted, instead making this an optional feature and requiring the user to always call close. This
- *             interface may also be removed.
  */
-@Deprecated(since = "4.1.0")
-public interface CloseableIteration<E, X extends Exception> extends Iteration<E, X>, AutoCloseable {
+//@Deprecated(since = "4.1.0")
+public interface CloseableIteration<E> extends AutoCloseable, Iterator<E> {
```
hmottestad, line 36: Removed the support for specifying exceptions with generics. We don't use this for checked exceptions anymore, so we're not loosing anything here.

We can now extend Iterator directly. 

One question is if we should rename this class to CloseableIterator and rename all use of "iteration" to "iterator". Since this class is used in so many places we could do a _soft_ rename and keep the old CloseableIteration around for a while more, but have it extend CloseableIterator.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/AbstractCloseableIteration.java
```diff
@@ -57,8 +57,8 @@ public final void close() throws X {
 	 * Called by {@link #close} when it is called for the first time. This method is only called once on each iteration.
 	 * By default, this method does nothing.
 	 *
-	 * @throws X
+	 *
 	 */
-	protected void handleClose() throws X {
+	protected void handleClose() {
 	}
```
hmottestad, line 63: One issue I've seen when profiling is that the `handleClose()` method is overridden multiple times in a class hierarchy. This reduces the performance because the JVM struggles with knowing which method to actually call and needs to use a method stub per object before it can optimize the call site. 

So I would like to duplicate up the code instead of having it extend this class.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/CloseableIteration.java
abrokenjester: In favor of renaming. It's a big refactor whichever way we go, I'd rather get it in a clean/good state in one attempt.

core/common/iterator/src/main/java/org/eclipse/rdf4j/common/iteration/AbstractCloseableIteration.java
abrokenjester: Do we even still need this abstract class, really? Given that the only thing it really adds is a private boolean flag, its reuse value is very limited.

Issue #4512, opened 2023-04-17
TupleQueryResultView does not report back errors


pulquero:
### Current Behavior

If a QueryEvaluationException occurs, RDF4J server still sends a HTTP 200 code.

### Expected Behavior

RDF4J server should send 500.

### Version

4.2.3

### Are you interested in contributing a solution yourself?

None

### Anything else?

TupleQueryResultView does have the correct error handling code
```
			} catch (QueryEvaluationException e) {
				logger.error("Query evaluation error", e);
				response.sendError(SC_INTERNAL_SERVER_ERROR, "Query evaluation error: " + e.getMessage());
```
Unfortunately it has an "auto-close" try not a standard try, so the output is sent 200 regardless. This
```
try (OutputStream out = response.getOutputStream()) {
```
should be something like
```
try {
OutputStream out = response.getOutputStream();
.
.
.
out.close();
```
Note out.close() is purposely not in a finally block, so it only commits the response on success.

pulquero:
Here's the commit that broke it:
https://github.com/eclipse/rdf4j/commit/217be508c4e7485dad7062516ad7d819b0e57aed#

pulquero:
Or can use an autoclose-try, but need to have a separate nested try..catch inside it.

abrokenjester:
I agree there is a problem here with response statuses, but I think this might be an issue that was even there before the refactor you mention, and therefore not related to the introduction of the try-with-resources. 

The root of the problem seems to me to be that in the error handling, we are trying to change the response status code from 200 to 500, but this happens after the response has already been committed (i.e. data has already started being sent to the client).

In any case, I am having some trouble coming up with a quick test that reproduces the problem - this might be easier to tackle in the develop branch where we've got some better testing library support for things related to HTTP client/server interactions.

abrokenjester:
I cannot actually reproduce the issue. I've added some additional test coverage but everything indicates this works as expected.

pulquero:
I get correct behaviour if an exception occurs during DefaultQueryRequestHandler.evaluateQuery(). But not if the exception occurs during QueryResults.report() in TupleQueryResultView.

abrokenjester:
> I get correct behaviour if an exception occurs during DefaultQueryRequestHandler.evaluateQuery(). But not if the exception occurs during QueryResults.report() in TupleQueryResultView.

[...]

PR #4524, opened 2023-04-29
GH-4512 error code must be set on uncommitted response - do not autoclose the outputstream

abrokenjester:
GitHub issue resolved: #4512  

- change try-with-resource to regular try-catch-finally to ensure closing the outputstream (and committing the response) happens _after_ potentially setting an error code (in the catch clause)

----


abrokenjester:
@pulquero can you take a look and let me know if this looks like ok?

pulquero:
That might work but I think strictly the close() should be inside the try. The stream is for writing the response body. If there is an error you want to not write/close whatever body you have, abandon it and just write error headers instead.

hmottestad:
Wouldn't it be easier to keep the try-with-resource and instead add a try/catch inside that to handle the error?

abrokenjester:
> Wouldn't it be easier to keep the try-with-resource and instead add a try/catch inside that to handle the error?

Hadn't thought of that - might be cleaner. I'll take a look and push an update.

PR #4533, opened 2023-05-02
GH-4532 fix breaking changes

hmottestad:
GitHub issue resolved: #4532 

 - fix merge issues causing NullPointerException in QueryJoinOptimizer
 - use both the new and the old config vocabulary to maintain backwards compatibility with those who read/manipulate the models directly
 - prioritize legazy predicates

----


hmottestad:
@abrokenjester I was testing 4.3.0-SNAPSHOT with GraphDB and found out that it uses the config models directly and crashes because it assumes that the deprecated predicates are still in use. Fixing this is kinda fiddly since we need to assume that users can use both the new and deprecated vocabularies for both reading and updating.

In 5.0.0 we can drop support for those who read/manipulate the models directly by changing back to only using the new vocabularies when writing to the model.

abrokenjester:
> @abrokenjester I was testing 4.3.0-SNAPSHOT with GraphDB and found out that it uses the config models directly and crashes because it assumes that the deprecated predicates are still in use. Fixing this is kinda fiddly since we need to assume that users can use both the new and deprecated vocabularies for both reading and updating.

Isn't the Configurations util a big help for that? They'd basically just need to use those same methods when reading a config model to allow both old and new config files to co-exist.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Configurations.java
```diff
@@ -47,17 +50,17 @@ public class Configurations {
 	@InternalUseOnly
 	public static Optional<Resource> getResourceValue(Model model, Resource subject, IRI property,
 			IRI legacyProperty) {
-		var result = Models.objectResource(model.getStatements(subject, property, null));
+		var result = Models.objectResource(model.getStatements(subject, legacyProperty, null));
 		if (result.isPresent()) {
 			return result;
 		}
-		return Models.objectResource(model.getStatements(subject, legacyProperty, null));
+		return Models.objectResource(model.getStatements(subject, property, null));
```
abrokenjester, line 57: Why do you prefer prioritising legacy predicates?

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/config/AbstractRepositoryImplConfig.java
```diff
@@ -66,6 +67,8 @@ public Resource export(Model model) {
 
 		if (type != null) {
 			model.add(implNode, CONFIG.Rep.type, literal(type));
+			model.add(implNode, REPOSITORYTYPE, literal(type));
```
abrokenjester, line 70: I'm really not a fan of this: we're now creating config files that duplicate data, and we re-introduce the deprecated vocabulary into new files. The setup we went for was to make sure that any new files only use the new vocabulary, and provide helpers and utilities to make reading of both old and new files doable. 

I'd like to understand a little better why just tweaking GraphDB to work with the new vocabulary (or at least support both) on _its_ side isn't an option here.

hmottestad: The issue isn't really GraphDB, but that the introduction of the new vocabulary amounts to a breaking change for anyone who is using the current (legacy) vocabulary.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Configurations.java
hmottestad: Users who modify the model would be using the current (legacy) vocabulary and if there is any discrepancy between what the new vocabulary says and the current (legacy) vocabulary then we should pick the current (legacy) vocabulary.

A very simple example:
```
  [] legacy:enabled true;
     new:enabled false.
```

hmottestad:
> > @abrokenjester I was testing 4.3.0-SNAPSHOT with GraphDB and found out that it uses the config models directly and crashes because it assumes that the deprecated predicates are still in use. Fixing this is kinda fiddly since we need to assume that users can use both the new and deprecated vocabularies for both reading and updating.
> 
> Isn't the Configurations util a big help for that? They'd basically just need to use those same methods when reading a config model to allow both old and new config files to co-exist.

It'll definitely help once I get around to migrating to the new vocabulary. At the moment though I consider it a breaking change that it's not possible to upgrade to 4.3.0 unless that change is used consistently throughout the code.

hmottestad:
@abrokenjester I did come up with an idea during the night. We could introduce a setting somewhere (env maybe) that turns off the support for the current (legacy) vocabulary. So users who want to migrate to the new vocabulary and be ready for 5.0.0 can flip that switch.

In 5.0.0 we will stop duplicating everything by using both the new and legacy vocabulary. We will also flip over the priority from the legacy vocabulary to the new one in  the Configurations util, and maybe add some assertions to stop conflicts from happening. 

For anyone who isn't working with the models directly the change should be seamless. If they migrate 4.x.0 -> 5.0.0 they will never see any duplicate use of the new/legacy vocabulary. If they migrate 4.x.0 -> 4.3.0 -> 5.0.0 the config files will have duplicates in 4.3.0, but those will be cleaned up in 5.0.0.

hmottestad:
A completely different alternative would be to provide the new vocabulary and Configurations util in 4.3.0, but only start using it in 5.0.0. 

We would still use the Configurations util when reading the config model to support those who want to start migrating now, but we would only write using the current (legacy) vocabulary.

hmottestad:
I should really have gotten more involved when this was still in the PR phase, but I'm not sure I would have seen the issues that I've discovered now.

abrokenjester:
You're completely right of course that if this breaks existing apps, then we have to rethink it. I think the main thing for me is that I still don't fully understand how this is a breaking change though: any Config class we have is set up to read either old or new properties (so it can deal with legacy data as well as new vocabulary), and the methods that write these configs to file have all been fixed to consistently produce new vocabulary. Can you give a code example of how this is used in existing code (in either GraphDB or somewhere else) that would break with this setup? I must be overlooking some obvious use case.

hmottestad:
Here is one place where there is an issue: 

Screenshot 2023-05-03 at 17 27 30

I don't know what this code does, or why it needs to use the vocabulary directly. I don't see anything particularly wrong with this code though. It assumes that the model uses the current (deprecated) vocabulary and fails now that we only use the new vocabulary.

abrokenjester:
That looks like it might be from a config class in GraphDB itself? Just like our own config classes have a parse method.

If we stuck with the current approach, this code would need to change to use the Configurations fallback method we are currently using ourselves, or something better/easier to use.

The question is: do we consider this sufficiently a breaking change to roll it back, or to introduce something that leaves us with ugly config files that duplicate info? Or can we consider this sufficiently niche that providing some upgrade notes to adapt such cases is good enough?

Fwiw if we decide it should be rolled back I'd be in favour of just completely removing from 4.3, and adding to the 5.0 branch instead.

hmottestad:
It's a kind of breaking change that is really hard to debug because it ends up causing a strange exception way down the line with no indication that it's caused by that code until you look into the deprecation warnings. 

Stability of namespaces and vocabulary is a big part of the semantic web, so it feels a like more of a fundamental breaking change when we change it up in a minor release.

abrokenjester:
Ok - I had originally though that this wouldn't be such a big issue because the use cases of directly interacting with the RDF data for config are vanishingly rare. But I agree that I may have underestimated that and we really should be treating this as a breaking change.

I suggest that, instead of trying to "fix" the change, we just take it out of the 4.3 release, and instead apply it on the 5.0 branch.

abrokenjester:
Update: that may not be as simple as I expected. I hadn't realized this was already merged more than 2 months ago. Github won't even let me create a revert PR from the original PR, too much has changed since then.

Ok let's suck it up, and just accept that for 4.3 we'll have ugly config files with duplicate data. I hate it, it's bad design and a bad experience for most of our users, but a breaking change is a breaking change.

hmottestad:
I'll take a jab at cleaning it up.

abrokenjester:
@hmottestad I've just realised that we've already got a milestone out with these changes in, and I think @aschwarte10 has already prepped their product to work with this.

I don't think we can back out anymore. We either go with what we have and provide upgrade guidance, or we go with some halfway approach to make it backward compatible .

hmottestad:
Milestone builds are not releases, so we can roll back the changes if we want to. It's not the best thing to do though, in general.

aschwarte10:
> Could you comment @aschwarte10 and let us know if you've already migrated over to the new config vocabulary that was included in the milestone build?

The change with the repository configuration schema has not been part of 4.3.0 M1 and thus we have not yet done any migration. We required 4.3.0 M1 mostly for the Jakarta changes in the core modules (related to Jaxb).

And even if we already did migrations:  as @hmottestad said - this is milestone releases. For me it is perfectly fine to roll back things if they do not work out. It is just important that the GA releases are in good shape.

abrokenjester:
Cool, I can live with that: fully agree that we can do this for milestones, just wanted to make sure we weren't creating a heap of extra work for metaphacts. @hmottestad really appreciate you giving a revert a go. Hopefully you can do it in a way that we can merge it into the 5.0 branch instead. Shout if you need a hand.

core/model/src/main/java/org/eclipse/rdf4j/model/util/Configurations.java
```diff
@@ -21,21 +21,27 @@
...
  *
  * @author Jeen Broekstra
- *
  * @implNote intended for internal use only.
  */
 @InternalUseOnly
 public class Configurations {
 
+	private static final boolean USE_CONFIG = "true"
+			.equalsIgnoreCase(System.getProperty("org.eclipse.rdf4j.model.vocabulary.experimental.enableConfig"));
```
hmottestad, line 37: I've introduced a system property to control if we should use the new CONFIG stuff.
 - "true" -> use CONFIG when adding data to the models and ignore legacy vocabulary when reading from a model
 - anything else -> use the legacy vocabulary when adding data to the models, but support both the lagacy and CONFIG vocabularies when reading

core/repository/api/src/main/java/org/eclipse/rdf4j/repository/config/AbstractDelegatingRepositoryImplConfig.java
```diff
@@ -69,7 +72,11 @@ public Resource export(Model model) {
 
 		if (delegate != null) {
 			Resource delegateNode = delegate.export(model);
-			model.add(resource, CONFIG.delegate, delegateNode);
+			if (USE_CONFIG) {
+				model.add(resource, CONFIG.delegate, delegateNode);
+			} else {
+				model.add(resource, RepositoryConfigSchema.DELEGATE, delegateNode);
+			}
```
hmottestad, line 79: The code is a bit bloated, but it should make it simpler to "revert" back to just using the CONFIG vocabulary.

hmottestad:
@abrokenjester I've introduced a system property that switches the code over to only using the new CONFIG vocabulary. I've only reverted some of the test code, other than that I've kept all the CONFIG code to make it easier to switch over in 5.0.0.

core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/CONFIG.java
```diff
@@ -20,6 +21,7 @@
  *
  * @since 4.3.0
  */
+@Experimental
 public class CONFIG {
```
hmottestad, line 25: And I marked the CONFIG class as experimental so that we can make changes in 5.0.0 if we want. The idea is to allow users to test out the new CONFIG code in 4.3.0 and provide us with feedback. Then we still have the option to integrate that feedback into the code for the 5.0.0 release.

abrokenjester (approved):

PR #4582, opened 2023-05-21
Gh 4581 remove jerven design mistakes

### LINKED ISSUE #4581: Remove my own design errors with v5
### Current Behavior

There are a few errors in design that are already deprecated but where originally introduced my myself. With the wisdom of a few years more I would like to remove them.
Chief among them are the LimitedSize strategy, and makeSet methods on the EvaluationStrategy interface.

### Expected Behavior

Remove the deprecated code.

### Steps To Reproduce

_No response_

### Version

4.x

Yes

### Anything else?

_No response_

JervenBolleman:
GitHub issue resolved: # 

Remove LimitedSize evaluation strategy code, as well as a few methods in the EvaluationStrategy API that was replaced by the collection factory API.

Closes #4581

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/DefaultEvaluationStrategy.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/DefaultEvaluationStrategy.java
@@ -533,16 +533,15 @@ public class DefaultEvaluationStrategy implements EvaluationStrategy, FederatedS
 	}
 
 	protected QueryEvaluationStep prepare(Group node, QueryEvaluationContext context) throws QueryEvaluationException {
-		return bindings -> new GroupIterator(DefaultEvaluationStrategy.this, node, bindings,
-				iterationCacheSyncThreshold,
-				context);
+		return bindings -> new GroupIterator(DefaultEvaluationStrategy.this, node, bindings, context);
 	}
 
 	protected QueryEvaluationStep prepare(Intersection node, QueryEvaluationContext context)
 			throws QueryEvaluationException {
 		QueryEvaluationStep leftArg = precompile(node.getLeftArg(), context);
 		QueryEvaluationStep rightArg = precompile(node.getRightArg(), context);
-		return new IntersectionQueryEvaluationStep(leftArg, rightArg, this.getCollectionFactory().get());
+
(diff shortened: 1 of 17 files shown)


JervenBolleman:
@hmottestad could this still be included in the next 5.x Milestone build?

hmottestad:
> @hmottestad could this still be included in the next 5.x Milestone build?

Yeah. I haven't started on the build process yet. I saw you requested a review from me, so I'll take a look asap.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/DefaultEvaluationStrategy.java
```diff
@@ -710,10 +709,26 @@ protected QueryEvaluationStep prepare(DescribeOperator node, QueryEvaluationCont
...
+				final CloseableIteration<BindingSet> evaluate = child.evaluate(bindings);
+				return new DistinctIteration<BindingSet>(evaluate,
+						cf.createSetOfBindingSets()) {
+
+					@Override
+					protected void handleClose() throws QueryEvaluationException {
+						try {
+							cf.close();
+						} catch (QueryEvaluationException e) {
+							super.handleClose();
```
hmottestad, line 726: Should this maybe be try-finally or is super.handleClose() only supposed to be called when there is an exception?

JervenBolleman: This should be a try-finally, although right now super.handleClose is a noop. That could change in the future.
will-fix.

hmottestad:
From what I understand looking through the code you've moved the set creation to a higher level in the code (EvaluationFactory?) so that it's easier to override in the specific stores. 

Only issue I found in your code was the try-catch instead of try-finally. If you're happy with everything then you can go ahead and merge.

JervenBolleman:
> From what I understand looking through the code you've moved the set creation to a higher level in the code (EvaluationFactory?) so that it's easier to override in the specific stores.
> 
> Only issue I found in your code was the try-catch instead of try-finally. If you're happy with everything then you can go ahead and merge.

Yes. Exactly, now with collection factories one does not need to extend the store. One can just inject the limited size collections. Also now there are fewer and fewer places left where collections are not coming from a collection factory that can spill to disk. So the original goal I had with these LimitedSize* all those years ago is no longer valid.

Issue #4574, opened 2023-05-19
Spring boot component for a HTTP sparql server does not support multiple default graphs


naturzukunft:
### Problem description

Default graph in SPARQL is specified using the FROM clause. Users can specify multiple FROM clauses, just as they can specify multiple FROM NAMED clauses. The default graph is the union of all the graphs specified by the FROM clauses.

The SparqlQueryEvaluator only supports a single string for the default graph.

See: https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#specifyingDataset

See discussion in #4562

### Preferred solution

```
	void evaluate(EvaluateResult result, Repository repository, String query, String acceptHeader,
			String[] defaultGraphUri,
			String[] namedGraphUris) throws MalformedQueryException, IllegalStateException, IOException;
```

### Are you interested in contributing a solution yourself?

Yes

PR #4587, opened 2023-05-22
GH4574 - Extending SparqlQueryEvaluator defaultGraphUri -> Array

naturzukunft:
GitHub issue resolved: #4574

The Parameter `defaultGraphUri` in `SparqlQueryEvaluator.evaluate` is now a array. So it's possible to pass more than one  defaultGraphUri.

----

### DIFF
--- a/spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/sparql/SparqlQueryEvaluator.java
+++ b/spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/sparql/SparqlQueryEvaluator.java
@@ -24,13 +24,14 @@ public interface SparqlQueryEvaluator {
 	 * @param repository      the repository against which the query is to be executed
 	 * @param query           The query to be evaluated
 	 * @param acceptHeader    needed to find the best response format.
-	 * @param defaultGraphUri see {@link Dataset#getDefaultGraphs()}
-	 * @param namedGraphUris  see {@link Dataset#getNamedGraphs()}
+	 * @param defaultGraphUri The graphs that are embedded in the query with FROM {@link Dataset#getDefaultGraphs()}
+	 * @param namedGraphUris  The graphs that are embedded in the query with FROM NAMED see
+	 *                        {@link Dataset#getNamedGraphs()}
 	 * @throws MalformedQueryException If the supplied query is malformed.
 	 * @throws IOException             if there is a problem with the {@link EvaluateResult#getOutputstream()}
 	 * @throws IllegalStateException   if no acceptHeader is present
 	 */
 	void evaluate(EvaluateResult result, Repository repository, String query, String acceptHeader,
-			String defaultGraphUri,
+			String[] defaultGraphUri,
 			String[] namedGraphUris) throws MalformedQueryException, IllegalStateException, IOException;
 }
--- a/spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/QueryResponder.java
+++ b/spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/QueryResponder.java
@@ -54,7 +54,7 @@ public class QueryResponder {
 			HttpServletRequest request, HttpServletResponse response) throws IOException {
 		try {
 			EvaluateResultHttpResponse result = new EvaluateResultHttpResponse(response);
-			sparqlQueryEvaluator.evaluate(result, repository, query, acceptHeader, defaultGraphUri,
+			sparqlQueryEvaluator.evaluate(result, repository, query, acceptHeader, toArray(defaultGraphUri),
 					toArray(namedGraphUri));
 		} catch (MalformedQueryException | IllegalStateException | IOException e) {
 			response.sendError(HttpServletResponse.SC_BAD_REQUEST);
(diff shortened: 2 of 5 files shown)


assembly-descriptors/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.rdf4j</groupId>
 		<artifactId>rdf4j</artifactId>
-		<version>4.3.1-SNAPSHOT</version>
+		<version>4.3.1-GH4574-SNAPSHOT</version>
```
hmottestad, line 7: You probably meant to clean up this before committing.

spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/sparql/SparqlQueryEvaluator.java
```diff
@@ -24,13 +24,14 @@ public interface SparqlQueryEvaluator {
...
+	 * @param namedGraphUris  The graphs that are embedded in the query with FROM NAMED see
+	 *                        {@link Dataset#getNamedGraphs()}
 	 * @throws MalformedQueryException If the supplied query is malformed.
 	 * @throws IOException             if there is a problem with the {@link EvaluateResult#getOutputstream()}
 	 * @throws IllegalStateException   if no acceptHeader is present
 	 */
 	void evaluate(EvaluateResult result, Repository repository, String query, String acceptHeader,
-			String defaultGraphUri,
+			String[] defaultGraphUri,
 			String[] namedGraphUris) throws MalformedQueryException, IllegalStateException, IOException;
```
hmottestad, line 36: This is a change to a public interface. It's not backwards compatible. 

@abrokenjester Do you know why japicmp doesn't complain about this? I get a feeling that japicmp is disabled by default and needs to be enabled in each module.

JervenBolleman:
Well spotted @naturzukunft I had no idea it would be legal to have more than one  default-graph-uri parameter! But it makes sense as this is an override of the FROM concept and that allows multiple ones.

I think this is fine to go into 5 without further ado, as it is on a method with a limited scope for overriding.

hmottestad:
> Well spotted @naturzukunft I had no idea it would be legal to have more than one default-graph-uri parameter! But it makes sense as this is an override of the FROM concept and that allows multiple ones.
> 
> I think this is fine to go into 5 without further ado, as it is on a method with a limited scope for overriding.

That fine by me too!

At the moment the PR is targeted at the main branch, but if we target it at the develop branch then it will end up in RDF4J 5.0.0.

spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/sparql/SparqlQueryEvaluator.java
naturzukunft: i will change that, so that there are 2 methods. the "old" one can be marked as deprecated.

naturzukunft:
> > Well spotted @naturzukunft I had no idea it would be legal to have more than one default-graph-uri parameter! But it makes sense as this is an override of the FROM concept and that allows multiple ones.
> > I think this is fine to go into 5 without further ado, as it is on a method with a limited scope for overriding.
> 
> That's fine by me too!
> 
> At the moment the PR is targeted at the main branch, but if we target it at the develop branch then it will end up in RDF4J 5.0.0.

i would prefere to have it in an earlier version!

assembly-descriptors/pom.xml
naturzukunft: i will change that. But it's always a problem to use the same version in feature branches. maybe rdf4j should think about using the maven gitflow plugin ?

spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/sparql/SparqlQueryEvaluator.java
hmottestad: Ok. The new interface method needs to have a default implementation.

naturzukunft: I do not understand the reason for this. What should the default implementation do ?

naturzukunft:
```
	default void evaluate(EvaluateResult result, Repository repository, String query, String acceptHeader,
			String[] defaultGraphUris,
			String[] namedGraphUris) throws MalformedQueryException, IllegalStateException, IOException {
		String defaultGraphUri = defaultGraphUris != null ? defaultGraphUris[0] : null;
		evaluate(result, repository, query, acceptHeader, defaultGraphUri, namedGraphUris);
	}
```

assembly-descriptors/pom.xml
naturzukunft: done!

spring-components/spring-boot-sparql-web/src/main/java/org/eclipse/rdf4j/http/server/readonly/sparql/SparqlQueryEvaluator.java
naturzukunft: done

naturzukunft:
What are the next steps here ?

hmottestad:
I'm planning a milestone build of RDF4J 5.0 and I think it would be good to include this PR in that build.

naturzukunft:
> I'm planning a milestone build of RDF4J 5.0 and I think it would be good to include this PR in that build.

@hmottestad  when will the milestone build be ready ?

hmottestad:
Hopefully within the next week or two.

Issue #4589, opened 2023-05-22
Enable japicmp for the entire codebase starting from 5.0.0


hmottestad:
For 4.x.x japicmp was only enabled on a selection of modules. Starting with 5.0.0 we should enable japicmp by default and instead disable it in modules where we do not strictly stick to semantic versioning.

Issue #4588, opened 2023-05-22
Bump japicmp version to 5.0.0 for the develop branch


hmottestad:
Bump japicmp version to 5.0.0 for the develop branch so that it doesn't complain about breaking changes.

PR #4590, opened 2023-05-22
GH-4589 Enable japicmp for all modules by default

hmottestad:
~~GitHub issue resolved: #4588~~
GitHub issue resolved: #4589 

----

### DIFF
--- a/core/model-api/pom.xml
+++ b/core/model-api/pom.xml
@@ -29,10 +29,6 @@
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
-				<groupId>com.github.siom79.japicmp</groupId>
-				<artifactId>japicmp-maven-plugin</artifactId>
-			</plugin>
 		</plugins>
 	</build>
 </project>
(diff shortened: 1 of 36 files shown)


hmottestad:
Exceptions to japicmp:
 - compliance
 - tools/runtime-osgi
 - tools/server
 - tools/workbench

These are all primarily disabled because japicmp complained that the jar file wasn't available. The compliance and workbench code I agree should not have strict backwards compatibility requirements, osgi should also be fine since it's just a wrapper to enable osgi. The server module, on the other hand, should probably be stricter wrt. backwards compatibility...but then we would need to figure out why japicmp is complaining about the missing jar file.

core/model-api/pom.xml
```diff
@@ -23,16 +23,13 @@
 				<artifactId>maven-jar-plugin</artifactId>
 				<executions>
 					<execution>
+						<phase>compile</phase>
```
hmottestad, line 26: I thought this would fix an issue with building the test-jar on time, but it might have made it worse.

PR #4593, opened 2023-05-22
GH-4592 Replace a Junit4 Suite with v5 DynamicTest

### LINKED ISSUE #4592: Migrate RepositorySPARQLComplianceTestSuite to JUnit5
### Problem description

It currently is a JUnit4 Suite

### Preferred solution

Junit5 testcases

Yes

_No response_

### Anything else?

_No response_

JervenBolleman:
Also replaces a static so more tests should be runnable in parallel in the future.

GitHub issue closes #4592

----

### DIFF
--- a/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQL10QueryComplianceTest.java
+++ b/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQL10QueryComplianceTest.java
@@ -15,6 +15,7 @@ import java.util.Collection;
 import java.util.List;
 
 import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
 
 /**
  * A test suite that runs the W3C Approved SPARQL 1.0 query tests.
@@ -51,6 +52,7 @@ public abstract class SPARQL10QueryComplianceTest extends SPARQLQueryComplianceT
 		}
 	}
 
+	@TestFactory
 	public Collection<DynamicTest> tests() {
 		return getTestData("testcases-sparql-1.0-w3c/data-r2/manifest-evaluation.ttl");
 	}


JervenBolleman:
@abrokenjester I think I got the upgrade of the manifest based tests to junit5 done. I removed some old deprecated test classes now that we are going to go to v5 as they are not compatible. They would need a similar non compatible rewrite to make work.

hmottestad:
@JervenBolleman I'll try to take at look at this as soon as I can.

JervenBolleman:
@hmottestad I am rebasing this so that you don't need to deal with the mergeconflicts.

hmottestad:
> @hmottestad I am rebasing this so that you don't need to deal with the mergeconflicts.

Thanks!

JervenBolleman:
@hmottestad the verify/integration-tests stage passes on my laptop. Running it on a second desktop machine to see if I can reproduce locally the.

JervenBolleman:
@kenwenzel would you mind having a look at this failure. The tests pass for native and memory on all my systems. However, on my desktop and the github action I have strange errors. One difference is that my laptop is fedora while desktop and github action are ubuntu based.

kenwenzel:
> @kenwenzel would you mind having a look at this failure. The tests pass for native and memory on all my systems. However, on my desktop and the github action I have strange errors. One difference is that my laptop is fedora while desktop and github action are ubuntu based.

Hi @JervenBolleman,

the NPEs are due to lazy values. Maybe the hash code function tries to initialize these values after the store has already been closed.
I've improved the handling of lazy values in this commit:
https://github.com/eclipse-rdf4j/rdf4j/commit/9b0622a4224587dfc539e9c5ca948f29d1461b4c

This should ensure that values do not leave the SAIL in an uninitialized state.
It may be the case that I've overseen something.

JervenBolleman:
@kenwenzel @hmottestad I found that part of the reason is that these tests now run concurrently and where sharing state. I fixed just this and will push a new commit.

testsuites/geosparql/src/main/java/org/eclipse/rdf4j/testsuite/query/algebra/geosparql/GeoSPARQLManifestTest.java
```diff
@@ -10,50 +10,14 @@
...
-			SPARQLQueryTestManifest manifest = new SPARQLQueryTestManifest(pop, null, false);
-			tests.addAll(manifest.getTests());
-			manifests.addAll(manifest.getSubManifests());
-		}
-
-		Object[][] result = new Object[tests.size()][6];
-		tests.toArray(result);
-
-		return result;
+	public Collection<DynamicTest> tests() {
```
hmottestad, line 20: I believe that this needs the @TestFactory annotation.

hmottestad:
Screenshot 2023-11-10 at 12 56 05

This is pretty sweet! I can debug a single test in IntelliJ!

testsuites/geosparql/src/main/java/org/eclipse/rdf4j/testsuite/query/algebra/geosparql/GeoSPARQLManifestTest.java
hmottestad: Probably applies to other similar methods too.

testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/BindTest.java
```diff
@@ -266,4 +270,14 @@ public void testGH4499BindFilterNotExist2() {
...
+
+	public Stream<DynamicTest> tests() {
+		return Stream.of(makeTest("GH4499BindFilterNotExist2", this::testGH4499BindFilterNotExist2),
+				makeTest("GH4499BindFilterNotExist1", this::testGH4499BindFilterNotExist1),
+				makeTest("GH3696Bind", this::testGH3696Bind), makeTest("SelectBindOnly", this::testSelectBindOnly),
+				makeTest("SES2250BindErrorsInPath", this::testSES2250BindErrorsInPath),
+				makeTest("SES2250BindErrors", this::testSES2250BindErrors),
+				makeTest("BindScopeUnion", this::testBindScopeUnion), makeTest("BindScope", this::testBindScope),
+				makeTest("BindError", this::testBindError));
+	}
```
hmottestad, line 282: Is there no way to stick with the regular @Test annotation? Or maybe generate this list of tests in a more automatic way?

testsuites/geosparql/src/main/java/org/eclipse/rdf4j/testsuite/query/algebra/geosparql/GeoSPARQLManifestTest.java
JervenBolleman: I think it is actually an error in the manifest file itself. There is a query that looks for `mf:include` but the manifest file for geosparql uses `mf:entities` and even a different IRI prefix for `mf:`. Let me have a look.

testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/BindTest.java
JervenBolleman: I couldn't figure it out from the documentation :(

testsuites/geosparql/src/main/java/org/eclipse/rdf4j/testsuite/query/algebra/geosparql/GeoSPARQLManifestTest.java
JervenBolleman: Looking at a current develop integration test job I think these tests are not run at this time either...
```
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- maven-failsafe-plugin:3.1.2:verify (verify) @ rdf4j-geosparql-compliance ---
```

JervenBolleman: Reenabled these tests. Lucky they all pass :)

JervenBolleman:
@hmottestad  sorry for the force push to undo the merge commit :(

testsuites/geosparql/src/main/java/org/eclipse/rdf4j/testsuite/query/algebra/geosparql/GeoSPARQLManifestTest.java
hmottestad: There are several other similar methods that were missing the @TestFactory annotation. Did you look over the rest of the code?

JervenBolleman:
I believe that the missing annotation was not the problem. For me other
classes with eclipse it works with just the annotation on the super super
class. I am going to make sure they ran in the logs to be sure.

JervenBolleman:
@hmottestad  you where completely right I had missed a `@testfactory` and it was needed! thanks and have a good weekend.

testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/BindTest.java
hmottestad: ChatGPT recommends this:

```java
import java.lang.reflect.InvocationTargetException;
import java.lang.reflect.Method;
import java.util.Arrays;
import java.util.stream.Stream;

import org.junit.jupiter.api.DynamicTest;
import org.junit.jupiter.api.function.Executable;

public class BindTest extends AbstractComplianceTest {

    // ... existing methods ...

    public Stream<DynamicTest> tests() {
        return Arrays.stream(BindTest.class.getDeclaredMethods())
                     .filter(method -> shouldRunAsTest(method))
                     .map(this::createDynamicTestFromMethod);
    }

    private boolean shouldRunAsTest(Method method) {
        // Use your own logic here to determine which methods are test methods
        // For example, you might check for certain annotations or naming patterns
        return method.getName().startsWith("test");
    }

    private DynamicTest createDynamicTestFromMethod(Method method) {
        return DynamicTest.dynamicTest(method.getName(), new Executable() {
            @Override
            public void execute() throws Throwable {
                try {
                    method.invoke(BindTest.this);
                } catch (IllegalAccessException | InvocationTargetException e) {
                    throw e.getCause();
                }
            }
        });
    }
}
```

hmottestad: Dunno how pretty that is though.

JervenBolleman: That's going back to junit v2 ;) I don't mind, the official way is to generate some kind of test template. But it is a lot of work for a little gain. @hmottestad Is it ok if we merge as is and leave that as a todo?

hmottestad: I guess we can always change it up later, and your approach does allow IntelliJ to run a single test instead of having to run the entire suite.

hmottestad (approved):
@JervenBolleman thanks for these fixes! You can go ahead and merge when you want.

Issue #4511, opened 2023-04-17
MemStatementList can safely make assumptions about how it is used


hmottestad:
The MemStatementList uses a lock-free approach for adding and retrieving items. When adding items the MemStatementList will expand its underlying array if it is full. Support lock-free expanding and removal of items at the same time is far from trivial, so I've used a StampedLock to make sure that the remove method runs exclusively. 

The remove method is only used from the snapshot cleanup thread, so we can assume single threaded access. Further more, when adding an item, we can assume that if the item was removed by the snapshot cleanup thread then it doesn't matter if we add the item back again or not.

When inserting an item into the array we need to consider the following scenarios:
- The place in the array was already taken
- The array was being expanded while we inserted our item
- The item was removed immediately after we inserted it

The first scenario is handled by using CAS when writing to the array. The second scenario is handled by checking if the array is the same array after we've written our item, and if not then we can check if our item is still there. 

The third scenario can safely be ignored because we can assume that the snapshot cleanup thread doesn't clean any active snapshots. Furthermore we can assume that if it did, then adding the item back would be safe since it would just be cleaned up in the next cleanup run.

abrokenjester:
Possibly related to #4468 ?

hmottestad:
Hopefully yes.

hmottestad:
I should pick this back up again.

Issue #4468, opened 2023-02-26
MemStatementListTestIT should not be flaky


erikgb:
I have seen this test failing a couple of times on unrelated PRs. Example: https://github.com/eclipse/rdf4j/pull/4464/checks

```
[INFO] Running org.eclipse.rdf4j.sail.memory.model.MemStatementListTestIT
Error:  Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 244.829 s <<< FAILURE! - in org.eclipse.rdf4j.sail.memory.model.MemStatementListTestIT
Error:  org.eclipse.rdf4j.sail.memory.model.MemStatementListTestIT.addRemoveConsistentMultipleThreads  Time elapsed: 120.123 s  <<< ERROR!
java.util.concurrent.TimeoutException: addRemoveConsistentMultipleThreads() timed out after 120 seconds
```

PR #4596, opened 2023-05-23
GH-4511 MemStatementList concurrency during cleanup

hmottestad:
GitHub issue resolved: #4511 #4468 

We can assume that there is only one thread that can remove statements from the statement list, because there is only 1 cleanup thread in the MemorySailStore. That means we can remove the StampedLock and instead change to being 100% lock free!

I also ended up doing a bit of a rewrite of parts of the code and also fixing a bug I hadn't noticed before.

A lot of time was spent on trying to create a decent test. That test made me discover the bug when the code checks if the write was lost due to the array growing.

----

### DIFF
--- a/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementList.java
+++ b/core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementList.java
@@ -77,6 +77,8 @@ public class MemStatementList {
 			MemStatement[] statements = getStatements();
 			int length = statements.length;
 
+			boolean shouldGrowArray = true;
+
 			if (length > (int) SIZE.getAcquire(this)) {
 
 				int previouslyInsertedIndex = (int) PREVIOUSLY_INSERTED_INDEX.getOpaque(this);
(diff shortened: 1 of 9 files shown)


core/sail/memory/src/main/java/org/eclipse/rdf4j/sail/memory/model/MemStatementList.java
```diff
@@ -75,69 +72,73 @@ public void add(MemStatement st) throws InterruptedException {
...
 
-								// check if the statements array has been swapped out (because it was grown) while we
-								// were
-								// inserting into it
-								MemStatement[] statementsAfterInsert = getStatements();
-								if (statementsAfterInsert != statements
-										&& STATEMENTS_ARRAY.getAcquire(statements, i) != st) {
-									// we wrote into an array while it was growing and our write was lost
-									break;
-								}
```
hmottestad, line 109: This is actually a bug that could lead to a statement not getting inserted if there is a highly concurrent write load.

hmottestad:
### Main branch

Screenshot 2023-05-23 at 23 01 53

### This branch

Screenshot 2023-05-23 at 23 02 26

This branch is ever so slightly faster.

Issue #4603, opened 2023-05-27
Upgrade to JavaEE/JakartaEE 8


erikgb:
### Problem description

Java EE has become Jakarta EE, and we have to migrate to Jakarta EE to be able to upgrade our "EE" dependencies like servlet, jsp and jaxb.

Jakarta EE 8 is the first release of Jakarta EE, and is fully compatible with Java EE 8 - which means that we can migrate dependencies to jakarta artifacts before addressing the massive code migration of javax to jakarta. I am assuming https://github.com/eclipse/rdf4j/issues/3559 is addressing the latter and will only be resolved by an upgrade to Jakarta EE 9. Jakarta EE also releases a BOM, which will make it significantly easier to manage Jakarta EE dependencies.

### Preferred solution

I have worked on a draft PR for a long time already (https://github.com/eclipse/rdf4j/pull/4397), and I suggest creating a new PR with the changes proposed in the WIP PR.

One of the main issues seems to be that our Solr/ES compatibility tests pull in dependencies to a legacy version of Jetty that blocks us from using a newer version of Jetty (compatible with JavaEE 8) in this module.

### Are you interested in contributing a solution yourself?

Yes

hmottestad:
Superseded by: https://github.com/eclipse-rdf4j/rdf4j/issues/5717

Issue #4252, opened 2022-10-27
Upgrade Jetty 9 to 10/11/12


barthanssens:
### Problem description

Jetty 9 is EOL, so we need to move to a newer version

### Preferred solution

Use a more recent version, e.g. 10 or 11.

### Anything else?

Version to be used may also depend on other dependencies (Elastic/Solr backend)

erikgb:
WireMock will need to be upgraded, ref. https://github.com/wiremock/wiremock/issues/1760. This has been fixed in https://github.com/wiremock/wiremock/pull/1942, and will be released in the new major version of WireMock. But if I am not mistaken, this will require the javax -> jakarta migration, ref. https://github.com/eclipse/rdf4j/issues/3559.

erikgb:
Some relevant blockers: https://github.com/wiremock/wiremock/issues/1967 https://github.com/wiremock/wiremock/issues/1980

Seems like we need to bundle a lot of changes with the javax->jakarta API namespace move.....

erikgb:
https://github.com/eclipse/rdf4j/pull/4397 contains an upgrade to Jetty 10 (almost). It seems like Solr/ElasticSearch might be a potential blocker.

JervenBolleman:
Solr 9.2, now ships jetty 10. so the elastic search one might be the last problematic one?

aschwarte10:
Support for newer application servers has recently been added to RDF4J 6.0 develop branch

As part of  https://github.com/eclipse-rdf4j/rdf4j/issues/5801 we also update the docker images to latest version of application servers.

Note that RDF4J 6.0 M1 is compatible with latest application servers (i.e. the web applications can be deployed in Jetty 12 and tomcat 11)

I propose to close this ticket as duplicate. If something is missing, please reopen

PR #4604, opened 2023-05-27
GH-4603 Upgrade to JavaEE/JakartaEE 8

erikgb:
GitHub issue resolved: #4603 #4252

Upgrades JavaEE/JakartaEE dependencies to version 8 using the Jakarta dependencies managed by the Jakarta EE 8 BOM. Jetty also needs to be upgraded to version 10 to be compatible with JavaEE/JakartaEE, and is also included in this PR.

----

### DIFF
--- a/compliance/solr/pom.xml
+++ b/compliance/solr/pom.xml
@@ -10,6 +10,10 @@
 	<name>RDF4J: Solr Sail Tests</name>
 	<description>Tests for Solr Sail.</description>
 	<properties>
+		<!-- FIXME: Delete; Jetty 10.x requires Solr 9.x -->
+		<jetty.version>9.4.50.v20221201</jetty.version>
+		<!-- FIXME: Delete; Jetty 10.x requires Solr 9.x -->
+		<enforce-no-javax-dependencies.fail>false</enforce-no-javax-dependencies.fail>
 		<!-- used to set up embedded solr server in integration tests -->
 		<test.solr.home>${project.basedir}/solr</test.solr.home>
 	</properties>
@@ -62,6 +66,11 @@
 			<version>${solr.version}</version>
 			<scope>test</scope>
 			<exclusions>
+				<!-- FIXME: Uncomment; Jetty 10.x requires Solr 9.x -->
+				<!--exclusion>
+					<groupId>javax.servlet</groupId>
+					<artifactId>javax.servlet-api</artifactId>
+				</exclusion -->
 				<exclusion>
 					<groupId>org.apache.logging.log4j</groupId>
 					<artifactId>*</artifactId>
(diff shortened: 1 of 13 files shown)


compliance/solr/pom.xml
```diff
@@ -10,6 +10,10 @@
 	<name>RDF4J: Solr Sail Tests</name>
 	<description>Tests for Solr Sail.</description>
 	<properties>
+		<!-- FIXME: Delete; Jetty 10.x requires Solr 9.x -->
+		<jetty.version>9.4.50.v20221201</jetty.version>
```
hmottestad, line 14: I'm still not sure what this means for us.

compliance/solr/pom.xml
```diff
@@ -62,6 +66,11 @@
 			<version>${solr.version}</version>
 			<scope>test</scope>
 			<exclusions>
+				<!-- FIXME: Uncomment; Jetty 10.x requires Solr 9.x -->
+				<!--exclusion>
+					<groupId>javax.servlet</groupId>
+					<artifactId>javax.servlet-api</artifactId>
+				</exclusion -->
```
hmottestad, line 73: Also not sure what this means for us.

hmottestad (commented):
I'm not sure what the implications of the jetty issue is. Does it mean that we are shipping with one version in one part of our code and another version somewhere else?

compliance/solr/pom.xml
erikgb: It means what it says in the comment. 😉 Some of the Solr dependencies used for testing are hard-locked to Jetty 9. I hope we can fix this by upgrading Solr before rdf4j 5 is released. Or even better: find a way to avoid being dictated by legacy transitive dependencies.

compliance/solr/pom.xml
erikgb: We must choose to have jakarta.* **or** javax.* dependencies. Never both. As there is a BOM for Jakarta dependencies, I would prefer to use that one.

erikgb:
> I'm not sure what the implications of the jetty issue is. Does it mean that we are shipping with one version in one part of our code and another version somewhere else?

If this is merged and released, we would use Jetty 9 in the Solr compliance module and Jetty 10 in the other modules. I am assuming a very little percentage of our users are actually using the compliance modules, as it appears to be a mostly internal module containing tests. But please correct me if I am wrong.

hmottestad:
If we need to use Jetty 9 in the Solr compliance module, wouldn't our users run into issues if they rely on the Solr module (not compliance) together with another module? Would they end up with a having to exclude Jetty 10 to get Solr working?

erikgb:
> If we need to use Jetty 9 in the Solr compliance module, wouldn't our users run into issues if they rely on the Solr module (not compliance) together with another module? Would they end up with a having to exclude Jetty 10 to get Solr working?

I think you are right that this might be an issue. The embedded Solr pulls in Jetty dependencies. I thought it was just test dependencies, but sadly not. Any suggestions for a workaround? Doing everything in one PR does not seem right to me.

hmottestad:
Probably need to update Solr, unfortunately. Which entails updating Lucene and probably Elasticsearch. 

I think it's worth double checking how jetty impacts our use of Solr. I assume that our Solr code is just client stuff and that the server stuff is only used for the embedded Solr during testing. Maybe the client code doesn't depend so heavily on Jetty?

Issue #4610, opened 2023-06-01
FedX runs into Phaser > 65535


qootec:
### Current Behavior

Perform "simple" query on the FedX (see below for more info)
```
select * where { 
    ?x a :X;
       :value ?value;
       :relation1 ?p1.

    ?y a :Y;
       :relation2 ?p1;
       :relation1 ?p2.
    
    ?z a :Z;
       :relation3 ?p2.
}
```

Result is: 
```
Query evaluation error: java.lang.IllegalStateException: Attempt to register more than 65535 parties for java.util.concurrent.Phaser@1707c5f1[phase = 0 parties = 65535 arrived = 0] (HTTP status 500)
```

### Expected Behavior

Normal query response.
The following reduction on the query's complexity avoids the issue:
```
select * where { 
    ?x a :X;
       :value ?value;
       :relation1 ?p1.

    ?y a :Y;
       :relation2 ?p1;
       :relation1 ?p2.
}
```

### Steps To Reproduce

Environment:
* GraphDB 10.2.1, using RDF4j 4.2.3
* FedX repo consisting of two small repos (number of triples between ())
  - Repo1: Local GraphDB repo (10946)
  - Repo2: Remote SPARQL repo pointing to an Ontop repo (1968)

Repo contents:
* Repo1: Concepts X and Z and the relations related to them
* Repo2: Concept Y and the relations related to them
* X, Y, Z all inherit from a parent Concept.

Performed the query in the GraphDB query window.

### Version

4.2.3

### Are you interested in contributing a solution yourself?

None

### Anything else?

* I can provide a reproduction environment if desired.
* Looks like #3106 , but that one is quite old and seems to be fixed.

aschwarte10:
Thanks for filing up this issue and the example scenario. I also thought we configured the phaser correctly (to use the nesting of phasers) with https://github.com/eclipse/rdf4j/issues/3106

Maybe some case there was missed or something else changed meanwhile.

I will try to look into this and come up with a reproducible test environment. However, I would need a bit of time as I am quite busy these days. Will come back here once I have something

qootec:
Thanks for your effort.
Two additional points for your information:
* The problem is not occurring when I re-order the statements in the first query as follows:
```
select * where { 
    ?y a :Y;
       :relation2 ?p1;
       :relation1 ?p2.
    
    ?x a :X;
       :value ?value;
       :relation1 ?p1.

    ?z a :Z;
       :relation3 ?p2.
}
```
[...]

PR #4613, opened 2023-06-04
GH-4610: fix FedX issue with large join (Phaser > 65535)

aschwarte10:
GitHub issue resolved: #4610 

Instead of registering to the root phaser, we need to register to the current phaser. Otherwise we may reach the limit of 65535 for large joins.

Note that in ControlledWorkerJoin and ControlledWorkerLeftJoin this was done correctly.

----


aschwarte10:
Unfortunately I cannot serve with a proper unit test. I had some environment (derived from the issue description) where I had the exception occurring (with the help of local debug tweaking). The issue very much depends on the concurrency. With this fix it is gone

However, I am convinced that this is the fix for the issue, especially when comparing to `ControlledWorkerJoin` and `ControlledWorkerLeftJoin`, see also the original PR https://github.com/eclipse/rdf4j/pull/3108/files#diff-13164a6c512de096d5d4ce745086d18f73ae0fc12610dff908d4491aaac16387

hmottestad (approved):

PR #4624, opened 2023-06-08
GH-4622 SHACL Validation Report preserve blank node IDs with remote repositories

### LINKED ISSUE #4622: SHACL - Blank node IDs should be preserved for remote validation reports
Blank node IDs should be preserved for remote validation reports because users need to be able to look up data in the repository based on the validation report model. 

One typical use case is for retrieving the full SHACL Shape from the repository. PropertyShape(s) are usually blank nodes, and when the report refers to a PropertyShape it uses the blank nodes. 

Another use case is where the validation report refers to a blank node from the users data that is causing the validation violation.

hmottestad:
GitHub issue resolved: #4622 

 - switch to RDF4J Binary RDF since this format doesn't rename blank node IDs
 - add parser config option to preserve blank node IDs
 - add test

----

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ProtocolExceptionResolver.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ProtocolExceptionResolver.java
@@ -10,7 +10,6 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.http.server;
 
-import java.io.StringWriter;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -19,9 +18,10 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.eclipse.rdf4j.common.exception.ValidationException;
 import org.eclipse.rdf4j.common.webapp.views.SimpleResponseView;
-import org.eclipse.rdf4j.model.Model;
+import org.eclipse.rdf4j.http.server.repository.statements.ValidationExceptionView;
 import org.eclipse.rdf4j.rio.RDFFormat;
-import org.eclipse.rdf4j.rio.Rio;
+import org.eclipse.rdf4j.rio.RDFWriterFactory;
+import org.eclipse.rdf4j.rio.RDFWriterRegistry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.web.servlet.HandlerExceptionResolver;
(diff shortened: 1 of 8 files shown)


tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ProtocolExceptionResolver.java
```diff
@@ -78,15 +78,14 @@ public ModelAndView resolveException(HttpServletRequest request, HttpServletResp
 
 			StringWriter stringWriter = new StringWriter();
 
-			// We choose NQUADS because we want to support streaming in the future, and because there could be a use for
-			// different graphs in the future
-			Rio.write(validationReportModel, stringWriter, RDFFormat.NQUADS);
+			// We choose RDFJSON because this format doesn't rename blank nodes.
+			Rio.write(validationReportModel, stringWriter, RDFFormat.RDFJSON);
```
hmottestad, line 82: So RDFXML, N-QUADS, N-TRIPLES, Turtle, etc. all have to use a specific format for blank nodes which means that we can't preserve blank node identifiers with those formats.

abrokenjester (commented):
It all looks technically correct to me, but I'm puzzled by why you'd need blank node preservation here in the first place. 

It's normally something that we only use in _very_ specific internal client-server comms use cases (e.g. in supporting remote transactions). I don't think we ever use it (or recommend using it) for somehow allowing a _user_ to reference a bnode by id remotely. 

Also: we tend to use RDF Binary everywhere else to support this: why choose RDF/JSON in this case?

hmottestad:
> It all looks technically correct to me, but I'm puzzled by why you'd need blank node preservation here in the first place.
> 
> It's normally something that we only use in _very_ specific internal client-server comms use cases (e.g. in supporting remote transactions). I don't think we ever use it (or recommend using it) for somehow allowing a _user_ to reference a bnode by id remotely.
> 
> Also: we tend to use RDF Binary everywhere else to support this: why choose RDF/JSON in this case?

I haven't tried using RDF Binary. I'll test it out.

The reason I need to preserve blank node identifiers is that this is a transfer of the validation exception from the server to the client. The user would then typically want to find out more about what went wrong by retrieving the full shape from the server. It's common practice to use blank nodes for PropertyShape(s), so that's why we need to preserve the blank node identifiers.

hmottestad:
### TODO

- [ ] Add test to make sure the client still works with the old server

hmottestad:
I think that this should be split into two branches. One with a simple fix that just changes out the format to binary, and a second that changes things up a bit by using the http body. The second branch should also bump the protocol version and target the develop branch.

Issue #4625, opened 2023-06-13
Some QueryModelNode implementations are not serializable (e.g. StatementPattern)


pulquero:
### Current Behavior

```
import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.ObjectOutputStream;

import org.eclipse.rdf4j.query.algebra.StatementPattern;
import org.junit.jupiter.api.Test;

public class QuickTest {
	@Test
	public void test() throws IOException {
		StatementPattern sp = new StatementPattern();
		sp.getBindingNames();
		new ObjectOutputStream(new ByteArrayOutputStream()).writeObject(sp);
	}
}

```

### Expected Behavior

Test passes.

### Version

4.3.1

### Are you interested in contributing a solution yourself?

None

### Anything else?

SmallStringSet is not Serializable and thus breaks the serialization of StatementPattern.

hmottestad:
Could you explain why the StatementPattern needs to be serializable since we don't have any tests that cover this feature?

hmottestad:
Ok. I remember that the group by operator or something spilled to disk using serialization, so I had thought that maybe you had run into an edge case in the code where a StatementPattern object would be serialized. 

Shouldn’t be too hard to fix it, but I think we need to add a serialization/de-serialization step to our tests.

hmottestad:
Btw. I think that this issue has been present since 4.1.0. Does that sound correct to you?

pulquero:
Yes, some 4.x version for sure, because I didnt have any issues with 3.x.

hmottestad:
I have a PR open for this now, but there are more issues with serialization than just StatementPattern. 

https://github.com/eclipse/rdf4j/pull/4626

PR #4626, opened 2023-06-17
GH-4625 Make all QueryModelNode serializable 

hmottestad:
GitHub issue resolved: #4625 

----


JervenBolleman:
I think the code is fine as bug fix. On the other hand perhaps it is a good idea to discuss serialization for v5 and later.

JervenBolleman (approved):

hmottestad:
> I think the code is fine as bug fix. On the other hand perhaps it is a good idea to discuss serialization for v5 and later.

I would prefer to drop it, but I did try to quickly serialise the objects to json using Gson which failed due to some cycles in the abstract query tree. Having some way of persisting the abstract query tree would be very useful if we want to cache query plans. I would also like to be able to transfer the query plans from the server to the client for debugging query performance issues.

Issue #4659, opened 2023-07-07
Query reusing graph clause variable in statement pattern returns incorrect result


nguyenm100:
### Current Behavior

running the following query for a particular dataset produces different results on rdf4j (4.2.3) vs graphdb (10.2.2 which uses rdf4j 4.2.3)

query:
```
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix owl: <http://www.w3.org/2002/07/owl#>

select distinct ?modeliri ?modelLabel {
    graph ?modeliri {
        ?modeliri a owl:Ontology ;
        rdfs:label ?modelLabel .
    }
}
```

data:
```
<http://something/s> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2002/07/owl#Ontology> <http://something/s> .
<http://something/s> <http://www.w3.org/2000/01/rdf-schema#label> "s-label-something"@en <http://something/s> .
<http://something/s> <http://www.w3.org/2000/01/rdf-schema#label> "s-label-somethingelse"@en <http://somethingelse/c> .
```

### Expected Behavior

graphdb appears to be correct.

### Steps To Reproduce

1. rdf4j running on wsl docker, graphdb desktop on windows
2. insert quads:
```
<http://something/s> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2002/07/owl#Ontology> <http://something/s> .
<http://something/s> <http://www.w3.org/2000/01/rdf-schema#label> "s-label-something"@en <http://something/s> .
<http://something/s> <http://www.w3.org/2000/01/rdf-schema#label> "s-label-somethingelse"@en <http://somethingelse/c> .
```
3. run query:
```
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix owl: <http://www.w3.org/2002/07/owl#>

select distinct ?modeliri ?modelLabel {
    graph ?modeliri {
        ?modeliri a owl:Ontology ;
        rdfs:label ?modelLabel .
    }
}
```
4. expect 1 row on rdf4j 4.2.3 but received 2.
```
<http://something/s> | "s-label-something"@en
<http://something/s> | "s-label-somethingelse"@en
```
graphdb returns 1 (as expected)
```
<http://something/s> | "s-label-something"@en


```

### Version

4.2.3

### Are you interested in contributing a solution yourself?

Perhaps?

### Anything else?

[...]

Issue #514, opened 2016-09-07
SPARQL 1.1 Conformance test subquery/sq03 fails


abrokenjester:
(Migrated from https://openrdf.atlassian.net/browse/SES-937)

Cause is the way in which context scoping is handled as part of the statement pattern in Sesame, as opposed to a separate Graph operator, as in the SPARQL algebra.

The query:

{code}
prefix ex:  http://www.example.org/schema#
prefix in:  http://www.example.org/instance#

select ?x where {
graph ?g {
  {select ?x where {?x ?p ?g}}
}
}
{code}

On this data:

{code}
:graph1 {
  :a :p :b .
  :c :p :graph1 .
}
{code}

Sesame returns:

||?x||
|:c|

But the correct answer is:

||?x||
|:c|
|:a|

The core of the problem is variable {{?g}}, which is used both as a statement pattern var and a graph selection var. Sesame binds {{?g}} to the graph name and then evaluates the inner query, thereby eliminating a possible binding of {{?x}} (because {{?g}} is already bound). Since the statement pattern is inside a projection, the binding of {{?g}} in the graph clause should not influence the binding of {{?g}} in the statement pattern.

pulquero:
graph ?g {...} only makes sense for statement patterns, so we shouldnt bind ?g to anything that isnt a statement pattern. It should just be dropped. So the algebra should be
`Projection(?x,
    Projection(?x,
        StatementPattern(?x, ?p, ?g, null, Scope.NAMED)
    )
)`

PR #4677, opened 2023-07-17
GH-4659 Fix for shared var name between context and subject in StatementPatternQueryEvaluationStep

hmottestad:
GitHub issue resolved: #4659 

 - add test
 - reorder code so that the getters are always created in the StatementPatternQueryEvaluationStep, even if they are duplicates
 - also slightly improve the human readable format of the query plans explanation

----

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/explanation/GenericPlanNode.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/explanation/GenericPlanNode.java
@@ -39,6 +39,8 @@ public class GenericPlanNode {
 	private final static String uniqueIdPrefix = UUID.randomUUID().toString().replace("-", "");
 	private final static AtomicLong uniqueIdSuffix = new AtomicLong();
 
+	private static final String spoc[] = { "s", "p", "o", "c" };
+
 	private final static String newLine = System.getProperty("line.separator");
 
 	private final String id = "UUID_" + uniqueIdPrefix + uniqueIdSuffix.incrementAndGet();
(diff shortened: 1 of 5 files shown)


hmottestad:
So I've found out that this doesn't fix https://github.com/eclipse/rdf4j/issues/514

hmottestad:
The query from that old test:

```sparql 
prefix ex:  http://www.example.org/schema#
prefix in:  http://www.example.org/instance#

select ?x where {
  graph ?g {
    {
      select ?x where { ?x ?p ?g }
    }
  }
}

```

hmottestad:
That query is really odd. Does the graph scope apply to all everything within it? Would the same be the case if it contained something like a FILTER NOT EXISTS? 

I assume it's illegal to have two GRAPH declarations, but I kinda wonder what would happen if the inner SELECT contained its own GRAPH declaration.

hmottestad:
I'm actually thinking that it's a variable scoping issue.

```sparql
prefix ex:  http://www.example.org/schema#
prefix in:  http://www.example.org/instance#

select ?x where {
  graph ?g {
    {
      select ?x where { ?x ?p ?g }
    }
  }
}
```

Since SPARQL has bottom-up semantics the inner SELECT query should be executed without knowledge of the outer variables. It's only when the results from the SELECT query bubble up that they get "filtered" by any variables that are bound in the outer scope.

Since `?g` isn't declared in the projection of the inner SELECT query then it should not be affected by the outer `?g` bound to the named graph.

Consider the following query:

```sparql
prefix ex:  http://www.example.org/schema#
prefix in:  http://www.example.org/instance#

select ?x ?a ?b ?c where {
  graph ?g {
    {
      select ?x ?g where { ?x ?p ?g }
    }
    ?a ?b ?c.
  }
}
```

Here the inner SELECT produces results for both `?x` and `?g`. The outer `?g` in the graph declaration will be bound by the bindings from the inner SELECT, which in turn will affect the `?a ?b ?c.` line.

If we remove the `?a ?b ?c.` line then the `graph ?g` declaration doesn't really affect anything.

Issue #4769, opened 2023-09-04
Scoping issue with undefined variable in FILTER


fkleedorfer:
### Current Behavior

We use the memory store for integration tests. The store is instantiated once and  then data is added to and removed from it, with calls to commit() after each batch of changes.

We noticed intermittent test failures (expected query results are sometimes not returned). We isolated the failing tests and let one of them repeat until failure. Observation: it reproducibly succeeds 19 times and then fails. 
Adding a `Thread.sleep(x)` at the beginning of the test increases the number of times it succeeds when x > 300. at x > 1000 the failures seem to go away completely.

I put a breakpoint at MemorySailStore.java:638 (at beginning of the `flush()` method), which, after several calls during the instantiation of the store (where we add some background data), is hit once every time the test is executed. At that point, if I call the `cleanSnapshots()` method every 10 or so test runs, the failure never occurs. 

Without intervention throught the thread debugger, the `cleanSnapshots()` method never gets called before the test fails in execution 20, roughly 1.8 seconds after the first invocation of the test. 

When set the sleep timeout mentioned above to 1000 millis, the `cleanSnapshots()` methods does reproducibly get called every few test runs, and the test failure never occurs.

Here is the query that yields results normally, and yields nothing under the circumstances described.
```
PREFIX mms: <https://mms.researchstudio.at/mms#>
PREFIX s: <http://schema.org/>
SELECT DISTINCT ?entityId ?entityType ?entityDateCreated ?entityDateModified
WHERE { ?entityId <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?entityType ;
    s:dateCreated ?entityDateCreated ;
    s:dateModified ?entityDateModified .
?entityId s:editor ?entityEditor .
{ ?entityId <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?entityType .
FILTER ( ( ?entityEditor NOT IN ( "ax533d70-0cd5-4425-beda-080c56b8b820" ) ) ) }
FILTER ( ?entityType IN ( mms:Standard, mms:Project, mms:Organization, mms:Feature, mms:FeatureGroup, mms:PropertySet, mms:Mapping ) ) }
ORDER BY DESC( ?entityDateModified )
```

At the point of failing, the stores' statementList has 40692 statements, including those from old snapshots, and 24844 'null' slots, the size of the array is 65536. the activeConnections Map only has one connection (the one we are using for the tests).

[...]

PR #4773, opened 2023-09-06
GH-4769 Variable scoping issue with FILTER

hmottestad:
GitHub issue resolved: #4769 

----

### DIFF
--- a/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/FilterScopeTest.java
+++ b/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/FilterScopeTest.java
@@ -1,0 +1,267 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ ******************************************************************************/
+
+package org.eclipse.rdf4j.testsuite.sparql.tests;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.eclipse.rdf4j.model.BNode;
+import org.eclipse.rdf4j.model.util.Values;
+import org.eclipse.rdf4j.model.vocabulary.FOAF;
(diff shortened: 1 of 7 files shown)


hmottestad:
@abrokenjester Could you take a look at this scoping issue? @fkleedorfer accidentally ran into it and I have managed to make some tests to reproduce the issue. I had initially tracked it back to https://github.com/eclipse-rdf4j/rdf4j/issues/3601 and reverting the changes to the FilterIterator seemed to fix the issue. However while I was working on the tests to reproduce it locally I managed to accidentally create yet another scoping scenario that was not solved by simply reverting the FilterIterator.

testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/sparql/tests/FilterScopeTest.java
```diff
@@ -0,0 +1,267 @@
...
+		String query = String.join("\n", "",
+				"PREFIX foaf: <" + FOAF.NAMESPACE + ">",
+				"SELECT * WHERE {",
+				"  ?person a ?type ; ",
+				"  foaf:age ?age ;",
+				"  foaf:knows ?person2 .",
+				"  FILTER(?person != <http://example.org/2> && ?type = foaf:Person)",
+				"	{",
+				"      ?person foaf:age ?age .",
+				"      FILTER(?type NOT IN (foaf:NotPerson)) #?type is not in scope, so should produce an error. ",
```
abrokenjester, line 49: Yep - the error here that should happen is a type error when evaluating the `NOT IN` operator, since its operand is unbound. The net result of this should be that the `FILTER` condition is effectively always `false`.

abrokenjester (commented):
@hmottestad The test cases as you describe them look sensible to me, and it indeed sounds like it might be related to the PR you found. I haven't fully grokked what _additional_ issue you've found here beyond what Florian already described. I am happy to take a closer look this weekend, but also trust that what you've got here makes sense and is inline with the SPARQL 1.1 spec as I understand it.

JervenBolleman:
If I remove the SameTermFilterOptimizer from the flow then the test passes.

JervenBolleman:
Ok, it's a double bug :( one indeed in the SameTermFilterOptimizer that should stop messing with variables out of scope. The other with filtering variables that are not in scope.

hmottestad:
Thanks for figuring this out @JervenBolleman !

JervenBolleman:
@hmottestad I haven't figured out a fix yet. The original work-around that was previously in the code is not sufficient :(

hmottestad:
@JervenBolleman great work! You want to clean up the commented out code and then merge this?

Issue #4784, opened 2023-09-27
Lost Connections could cause deadlock


hmottestad:
The connection cleanup during shutdown can cause a deadlock if one lost connections is holding an exclusive lock while another connection is waiting for the same lock.

hmottestad:
We should consider backporting this fix

PR #4787, opened 2023-09-29
GH-4784 interrupt threads using connections when forcefully closing

hmottestad:
GitHub issue resolved: #4784 #4790 

This is mainly an issue with the MemoryStore, but has been detected when using the ShaclSail since it uses memory stores for tracking the changes in a transaction.

When one connection holds the MemorySailStore `txnLockManager` then another thread can get stuck waiting for the same lock if both threads are forcefully closed. Two issues are causing this, the first is that the `txnLockManager` can't be forcefully unlocked by another thread, the second is that the `txnLockManager` is being locked without support for interruptions.

The fix is to track the owner thread of the connection and interrupt that thread. Since connections shouldn't be shared between threads we can assume that the thread that created the connection is the owner.

----

### DIFF
--- a/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
+++ b/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
@@ -574,7 +574,7 @@ class LmdbSailStore implements SailStore {
 													if (!running.get()) {
 														logger.warn(
 																"LmdbSailStore was closed while active transaction was waiting for the next operation. Forcing a rollback!");
-														opQueue.add(ROLLBACK_TRANSACTION);
+														rollback();
 													} else if (Thread.interrupted()) {
 														throw new InterruptedException();
 													} else {
--- a/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
+++ b/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
@@ -12,7 +12,6 @@ package org.eclipse.rdf4j.sail.lmdb;
 
 import java.io.File;
 import java.io.IOException;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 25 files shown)


core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
```diff
@@ -561,14 +571,28 @@ private void startTransaction(boolean preferThreading) throws SailException {
...
-													Thread.yield();
+													if (!running.get()) {
+														logger.warn(
+																"LmdbSailStore was closed while active transaction was waiting for the next operation. Forcing a rollback!");
+														opQueue.add(ROLLBACK_TRANSACTION);
+													} else if (Thread.interrupted()) {
+														throw new InterruptedException();
+													} else {
+														Thread.yield();
+													}
```
hmottestad, line 582: @kenwenzel I've made some tests to fix some deadlock scenarios in the MemoryStore. The tests open two connections and start two transactions, in two separate threads. One of the tests then tries to call shutdown() on the store, the other test tries to close the connections directly. I ran into an issue with the async transaction handling in the LmdbSailStore that I've tried to handle here.

The issue is that when the executor service is shutdown() it waits indefinitely for the async transaction handling to complete. The async transaction handling can at that point be stuck in a `while(true)` loop that is waiting for the transaction to either be committed or rolled back. For some reason it doesn't seem to get rolled back, but adding a check for `running` and then rolling back the transaction seems to work.

What do you think @kenwenzel ?

kenwenzel: Thank you for Investigating this. I am not quite sure why this happens as I thought that it is handled by the running flag. But maybe I was wrong.

testsuites/sail/src/main/java/org/eclipse/rdf4j/testsuite/sail/SailConcurrencyTest.java
```diff
@@ -307,6 +312,108 @@ public void testGetContextIDs() throws Exception {
...
+			} else {
+				connection1.get().close();
+				connection2.get().close();
+			}
+		} catch (SailException ignored) {
+		}
+
+		store.shutDown();
+	}
+
```
hmottestad, line 416: @kenwenzel These are the tests.

Issue #4819, opened 2023-10-18
Support for merge join and ordered indexes


hmottestad:
Some data sources can return statements in a particular order. Data stored in a B-tree has a natural order due to the natur of a B-tree. 

The current join strategy for RDF4J is the iterate over the left data source and for each row create a new iterator from the right data source. This takes advantage the fact that we can bind the result from the left iterator when we create the right iterator, which makes it very selective.

For larger datasets the above approach can cause a lot of IO requests. An example could be that we are joining all `?a a foaf:Person` with `?a foaf:age ?age`. If we have 100 000 000 people, then that ends up being 1 000 000 000 new iterators for retrieving the age.

Instead we can take advantage of the ordering of the indexes and retrieve:
 - `?a a foaf:Person` from the OPSC index (O and P are constant)
 - `?a foaf:age ?age` from the PSOC index (P is constant)

These two iterators would both be ordered on the subject (S) variable (?a). We can then do a merge join, which is very efficient both IO wise and memory wise.

This issue aims to:
 - support inner join using merge join
 - allow for specifying the order when retrieving statements
 - inform the query join optimizer about which statement orderings that are available
 - pick merge join when possible (and appropriate)
 - support ordered indexes in the ExtensibleStore

kenwenzel:
This would be really cool.

kenwenzel:
@hmottestad How will this compare to the existing hash joins?

hmottestad:
The hash joins are used for negation, so it's a bit of a different use case. Merge join is meant to be an alternative to regular join, but faster for large joins. 

Our current approach is akin to indexed nested loop join, we loop through the left side and do indexed lookups on the right side. With smart read ahead and caching it would be possible to reduce the IO, but worst case we end up with O(N log(N) ) reads from disk with a B+ tree index.

With merge join we would do two range queries against the index, so basically O(2) reads. It's not quite true to call it O(2) because the range query reads one block at a time from the B+ tree, so we could call it O(N/K) where K is the size of each block.

So it's basically a more optimal way of joining large amounts of data when the underlying store has ordered indexes. 

Does LMDB have ordered indexes?

hmottestad:
(The hash joins are sometimes used for joins where there are scoping issues, so it's not only for negation.)

kenwenzel:
Yes, LmdbStore uses the same indexing scheme as the NativeStore.

[...]

PR #4822, opened 2023-10-19
GH-4819 Merge Join

hmottestad:
GitHub issue resolved: #4819 

----


core/common/ordering/src/main/java/org/eclipse/rdf4j/common/ordering/StatementOrder.java
```diff
@@ -0,0 +1,21 @@
...
+package org.eclipse.rdf4j.common.ordering;
+
+public enum StatementOrder {
+
+	S,
+	P,
+	O,
+	C;
+
+}
```
hmottestad, line 21: The enum used to specify what order the statements are in.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDataset.java
```diff
@@ -79,6 +83,26 @@ public interface SailDataset extends SailClosable {
...
+	 *                       is optional. If no contexts are supplied the method operates on all contexts.
+	 * @return An iterator over the relevant statements.
+	 * @throws SailException If the triple source failed to get the statements.
+	 */
+	default CloseableIteration<? extends Statement> getStatements(StatementOrder statementOrder, Resource subj,
+			IRI pred, Value obj,
+			Resource... contexts) throws SailException {
+		throw new SailException("Statement ordering not supported by this store");
+	}
+
```
hmottestad, line 105: A method on the SailDataset to retrieve statements in a particular order.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDataset.java
```diff
@@ -94,4 +118,12 @@ default CloseableIteration<? extends Triple> getTriples(Resource subj, IRI pred,
 		throw new SailException("RDF-star triple retrieval not supported by this store");
 	}
 
+	default Set<StatementOrder> getAvailableOrderings(Resource subj, IRI pred, Value obj, Resource... contexts) {
+		return Set.of();
+	}
```
hmottestad, line 123: A way for the SailDataset to inform upstream code about which orderings it can provide for different statement patterns.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDataset.java
```diff
@@ -94,4 +118,12 @@ default CloseableIteration<? extends Triple> getTriples(Resource subj, IRI pred,
...
+
+	default Comparator<? extends Statement> getComparator(StatementOrder statementOrder) {
+		throw new SailException("Statement ordering not supported by this store");
+	}
```
hmottestad, line 127: A way to provide a comparator for the Union and Merge Join code. The comparator doesn't need to be the ValueComparator that we have which orders the data based on the SPARQL spec. Most likely the comparator will be something closer to a string based comparator.

hmottestad: And I also see now that we need a more generic approach to the comparator because it will need to work on BindingSets too.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -394,6 +395,29 @@ public final CloseableIteration<? extends Statement> getStatements(Resource subj
 		}
 	}
 
+	@Override
+	public final CloseableIteration<? extends Statement> getStatements(StatementOrder order, Resource subj, IRI pred,
+			Value obj, boolean includeInferred, Resource... contexts) throws SailException {
```
hmottestad, line 400: This is at the SailConnection level. It might be better to have the `order` argument be after `includeInferred` and before `Resource... contexts`.

hmottestad:
@abrokenjester @kenwenzel I'm working on support merge join. The first step is to add support for the sail to return ordered statements. I would really appreciate some feedback on how to make my changes as consistent as possible with the existing interfaces. Any suggestions on naming is also very welcome.

hmottestad:
Multi variable joins might be problematic:

```
 ?a a ?b.
 ?a ex:type ?b.
```

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
```diff
@@ -891,5 +893,21 @@ public CloseableIteration<? extends Statement> getStatements(Resource subj, IRI
...
+
+		@Override
+		public Set<StatementOrder> getSupportedOrders(Resource subj, IRI pred, Value obj, Resource... contexts) {
+			return Set.of();
+		}
+
+		@Override
+		public Comparator<Value> getComparator() {
+			throw new UnsupportedOperationException("Not implemented yet");
+		}
```
hmottestad, line 911: @kenwenzel here are the three methods you need to implement to get merge join working.

hmottestad:
> @hmottestad I can offer to add support for ordered indexes to LmdbStore. Do you already have any performance figures?

This would be really awesome. I've tagged you above in a comment to show what needs to be implemented. That should hopefully be all that is needed. One thing I'm not certain about is transaction isolation, so it might be best to test it out with IsolationLevels.NONE. There is also some work left to be done on the DualUnionIteration, let me know if you run into the UnsupportedOperationException in that class.

I haven't made any benchmarks, mostly because I've only implemented this in the ExtensibleStore with a backing data structure sorts the data for each request.  That is just so I can test things out locally. The end goal is for https://github.com/the-qa-company/qEndpoint to use it for analytical queries. They have a really innovative product that uses HDT for storing the data on disk and can work with massive datasets using very little memory.

The queries that we want to support are ones that would anyway need to read most of the data from a range query on the index.

For instance:
```
select (count(?a) as ?count) where {
    ?a a foaf:Person;
       foaf:age ?age.
}
```

If you want to see what if the query planner chooses merge join you can use the query explanation:

```
Explanation explain = connection
    .prepareTupleQuery(query)
    .explain(Explanation.Level.Executed);
System.out.println(explain);
```

And you should see `Join (InnerMergeJoinIterator)` in the query explanation. 

Here is an example for a completely different query:
```
Projection (resultSizeActual=5)
╠══ ProjectionElemList
║     ProjectionElem "s"
║     ProjectionElem "o"
║     ProjectionElem "o2"
║     ProjectionElem "s2"
║     ProjectionElem "o3"
╚══ Join (JoinIterator) (resultSizeActual=5)
   ├── Join (InnerMergeJoinIterator) (resultSizeActual=5) [left]
   │  ╠══ Join (InnerMergeJoinIterator) (resultSizeActual=5) [left]
   │  ║  ├── StatementPattern [statementOrder: S]  (costEstimate=0, resultSizeEstimate=0, resultSizeActual=5) [left]
   │  ║  │     s: Var (name=s)
   │  ║  │     p: Var (name=_const_9285ccfc_uri, value=http://www.w3.org/2000/01/rdf-schema#label, anonymous)
   │  ║  │     o: Var (name=o)
   │  ║  └── StatementPattern [statementOrder: S]  (costEstimate=0, resultSizeEstimate=0, resultSizeActual=14) [right]
   │  ║        s: Var (name=s)
   │  ║        p: Var (name=_const_44b0da67_uri, value=http://www.w3.org/2000/01/rdf-schema#comment, anonymous)
   │  ║        o: Var (name=o2)
   │  ╚══ StatementPattern [statementOrder: S]  (costEstimate=0, resultSizeEstimate=0, resultSizeActual=14) [right]
   │        s: Var (name=s)
   │        p: Var (name=_const_531c5f7d_uri, value=http://xmlns.com/foaf/0.1/knows, anonymous)
   │        o: Var (name=s2)
   └── StatementPattern (costEstimate=0, resultSizeEstimate=0, resultSizeActual=5) [right]
         s: Var (name=s2)
         p: Var (name=_const_44b0da67_uri, value=http://www.w3.org/2000/01/rdf-schema#comment, anonymous)
         o: Var (name=o3)
```

hmottestad:
This all gives me an idea for making the query explanation smarter by including what index is being used and a recommendation for the most optimal index so the user can configure the best indexes for their particular queries.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
kenwenzel: @hmottestad  I've added an initial implementation in:
https://github.com/kenwenzel/rdf4j/tree/GH-4819-merge-join

Unfortunately, `QueryBenchmark.complexQuery()` directly fails with an Exception. Maybe it helps to locate the problem.

hmottestad: I fixed it, was an optimisation I had made where I hadn't quite accounted for all the edge cases.

hmottestad:
@kenwenzel try this query:

```
PREFIX ex: <http://example.com/ns#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX sh: <http://www.w3.org/ns/shacl#>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
PREFIX dcat: <http://www.w3.org/ns/dcat#>
PREFIX dc: <http://purl.org/dc/terms/>
PREFIX skos:  <http://www.w3.org/2004/02/skos/core#>
PREFIX foaf:  <http://xmlns.com/foaf/0.1/>
PREFIX dct: <http://purl.org/dc/terms/>

SELECT  (count(?a) as ?count) where {
        ?a  dct:identifier ?identifier ;
            dct:description ?description ;
            dcat:contactPoint ?contactPoint .
}
```

On my machine this is 3x faster with merge join. 

PS: You can disable merge join by changing the set of orders that are returned:
```
		@Override
		public Set<StatementOrder> getSupportedOrders(Resource subj, IRI pred, Value obj, Resource... contexts) {
// 			return supportedOrders;
			return Set.of();
		}
```

kenwenzel:
@hmottestad Probably the merge join will show its strength when the IO performance degrades due to memory mapping and/or slow storage systems. This should be the case if the databases grow larger than RAM.

hmottestad:
@kenwenzel collaborating on merge join would be a lot simpler if you were a committer. Would you be interested in me nominating you?

kenwenzel:
> @kenwenzel collaborating on merge join would be a lot simpler if you were a committer. Would you be interested in me nominating you?

I can't make guarantees regarding contributions but I don't want to refuse the offer a second time (Jeen already asked a while ago). So yes and thank you for asking.

JervenBolleman:
@hmottestad I was asked by the team (@D063520) behind [the qEndpoint](https://github.com/the-qa-company/qEndpoint/) to help with implementing merge joins :) and I see you are already working on it.

hmottestad:
> @hmottestad I was asked by the team (@D063520) behind [the qEndpoint](https://github.com/the-qa-company/qEndpoint/) to help with implementing merge joins :) and I see you are already working on it.

They are sponsoring this feature :)

JervenBolleman:
We should think about how we can pass this about information about sorted iterators on to `GROUP BY` and `ORDER BY` operators.

e.g.
```sparql
SELECT ?o
WHERE {?s ?p ?o}
ORDER BY ?o
```
If the statementpattern evaluation step/iterarator provides information to order step/iterator we can avoid redoing a lot of work. 

Same for the GROUP BY, if encounter order is fixed or partially fixed we might avoid building the large hash table.

D063520:
@JervenBolleman: we reached out different people and I understood you are too busy, @hmottestad had some free time and he was super reactive

JervenBolleman:
> @JervenBolleman: we reached out different people and I understood you are too busy, @hmottestad had some free time and he was super reactive

I think it is wonderful you/the qa company are sponsoring this feature and @hmottestad is much better at this kind of work then I am :) win win all round.

hmottestad:
### TODO:

- [ ] Make tests for edge cases when using the new unmark feature
- [ ] Add method to iterators to return which variable the iterator is sorted by, if any

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/AbstractQueryPreparer.java
```diff
@@ -0,0 +1,338 @@
...
+import org.eclipse.rdf4j.query.parser.ParsedBooleanQuery;
+import org.eclipse.rdf4j.query.parser.ParsedGraphQuery;
+import org.eclipse.rdf4j.query.parser.ParsedTupleQuery;
+import org.eclipse.rdf4j.query.parser.ParsedUpdate;
+import org.eclipse.rdf4j.query.parser.impl.AbstractParserQuery;
+import org.eclipse.rdf4j.query.parser.impl.AbstractParserUpdate;
+import org.eclipse.rdf4j.rio.RDFHandler;
+import org.eclipse.rdf4j.rio.RDFHandlerException;
+
+public abstract class AbstractQueryPreparer implements QueryPreparer {
```
hmottestad, line 52: This class was deleted by accident in 5.0.0-M2.

hmottestad:
### Develop branch


### This branch
```
Benchmark                                                     Mode  Cnt    Score    Error  Units
QueryBenchmark.complexQuery                                   avgt    5    1.039 ±  0.030  ms/op
QueryBenchmark.different_datasets_with_similar_distributions  avgt    5    0.452 ±  0.003  ms/op
QueryBenchmark.groupByQuery                                   avgt    5    0.575 ±  0.026  ms/op
QueryBenchmark.long_chain                                     avgt    5  155.290 ±  9.074  ms/op
QueryBenchmark.lots_of_optional                               avgt    5   41.770 ±  1.891  ms/op
QueryBenchmark.minus                                          avgt    5  870.915 ± 46.426  ms/op
QueryBenchmark.nested_optionals                               avgt    5   51.305 ±  2.648  ms/op
QueryBenchmark.pathExpressionQuery1                           avgt    5    5.495 ±  1.240  ms/op
QueryBenchmark.pathExpressionQuery2                           avgt    5    0.493 ±  0.006  ms/op
QueryBenchmark.query_distinct_predicates                      avgt    5   51.940 ±  4.232  ms/op
QueryBenchmark.simple_filter_not                              avgt    5    1.909 ±  0.665  ms/op
```

hmottestad:
Doesn't seem to be any performance degradation.

Issue #4847, opened 2023-11-17
ArrayBindingSetIterator advances on both hasNext and next


JervenBolleman:
### Current Behavior

```
index =0;
@Override
public boolean hasNext() {
	for (; index < values.length; index++) {
		if (whichBindingsHaveBeenSet[index] && values[index] != null) {
			return true;
		}
	}
	return false;
}
```
should be
```
index =0;
@Override
public boolean hasNext() {
	for (int at = index; at < values.length; at++) {
		if (whichBindingsHaveBeenSet[at] && values[at] != null) {
			return true;
		}
	}
	return false;
}
```

### Expected Behavior

hasNext() should not advance the iterator.

### Version

4.7.8

### Are you interested in contributing a solution yourself?

Yes

PR #4848, opened 2023-11-17
GH-4847 ArrayBindingSetIterator should not advance on hasNext, also a…

JervenBolleman:
…void LinkedHashSet althoughether

GitHub issue resolved: #4847 

Fix the iterator logic to not advance when it should not.

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/ArrayBindingSet.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/ArrayBindingSet.java
@@ -10,11 +10,13 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation;
 
+import java.util.AbstractSet;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/ArrayBindingSet.java
```diff
@@ -295,8 +330,8 @@ public ArrayBindingSetIterator() {
 
 		@Override
 		public boolean hasNext() {
-			for (; index < values.length; index++) {
-				if (whichBindingsHaveBeenSet[index] && values[index] != null) {
+			for (int at = index; at < values.length; at++) {
+				if (whichBindingsHaveBeenSet[at] && values[at] != null) {
 					return true;
 				}
 			}
```
hmottestad, line 337: Doesn't the old solution only increments if the current position is false/null?

JervenBolleman: No the problem is that both hasNext and next advance the index. So with a four set values,
`hasNext, hasNext, next, hasNext, hasNext, next` consumes the iterator instead of  
`hasNext, hasNext, next, hasNext, hasNext, next , hasNext, hasNext, next, hasNext, hasNext, next`.
Note the required double hasNext call required to trigger the bug.

hmottestad: I see you maybe a test. I’ll try to debug it to understand what’s going on, from just reading it I can’t quite understand how it can increment a second time.

JervenBolleman: @hmottestad you are right it didn't I miss diagnosed the issue.

Issue #4870, opened 2024-01-17
Make MapDB3 CollectionFactory the default over the current MapDB3 one


JervenBolleman:
### Problem description

In 4.x the MapDb v1 based collection factory is the default, we should move to the faster MapDb3 as a default in RDF4j v5.

### Are you interested in contributing a solution yourself?

Yes

PR #4871, opened 2024-01-17
GH-4870 MapDB3 based collection factory is now the default instead of…

JervenBolleman:
… an MapDB1 based one

GitHub issue resolved: #GH-4870

Make the prior work on a MapDB3 based collection factory the default.


hmottestad:
Do you know if we actually have any tests that cover the use of MapDB? I assume it's used when overflowing to disk when low on memory, but I wonder if that is possible to easily test for.

JervenBolleman:
The code get's triggered in a number of query evaluation and conformance tests. Plus the unit tests just for this code.

hmottestad (approved):

Issue #4886, opened 2024-01-26
Problem with use of @TestFactory and @TempDir


hmottestad:
Using `@TempFactory` bypasses a lot of the lifecycle hooks in JUnit 5 which means that `@TempDir` is only instantiated once for the entire life of the test class, instead of for each test.

Issue #4806, opened 2023-10-10
TimeLimitIteration calls close() asynchronous from another thread


kenwenzel:
### Current Behavior

Instead of just setting the `isInterrupted` flag, the `close()` method is directly called from `interrupt()`.
The latter method is executed within an asynchronous thread which leads to race conditions if the iterator expects non-multithreaded usage (which is the usual case).

This seems to affect the `LmdbRecordIterator` that causes a `A heap has been corrupted` exception at least under Windows.

### Expected Behavior

The `close()` method should not be called asynchronous from another thread.

Another solution would be to explicitly support the asynchronous invocation of `close()` in `LmdbRecordIterator`.

### Version

4.2.2

### Are you interested in contributing a solution yourself?

Perhaps?

kenwenzel:
A possibly related stack trace:
```
"TimeLimitIteration" #347 daemon prio=5 tid=0x00000162c5e1e800 nid=0x314c runnable [0x0000000408afe000]
   java.lang.Thread.State: RUNNABLE
   JavaThread state: _thread_in_native
 - org.lwjgl.util.lmdb.LMDB.nmdb_cursor_close(long) @bci=0 (Compiled frame; information may be imprecise)
 - org.lwjgl.util.lmdb.LMDB.mdb_cursor_close(long) @bci=12, line=1765 (Compiled frame)
 - org.eclipse.rdf4j.sail.lmdb.LmdbRecordIterator.close() @bci=11, line=192 (Compiled frame)
 - org.eclipse.rdf4j.sail.lmdb.LmdbStatementIterator.handleClose() @bci=8, line=87 (Compiled frame)
 - org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close() @bci=13, line=52 (Compiled frame)
 - org.eclipse.rdf4j.common.iteration.Iterations.closeCloseable(org.eclipse.rdf4j.common.iteration.Iteration) @bci=11, line=189 (Compiled frame)
 - org.eclipse.rdf4j.common.iteration.UnionIteration.handleClose() @bci=98, line=113 (Compiled frame)
 - org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close() @bci=13, line=52 (Compiled frame)
 - org.eclipse.rdf4j.common.iteration.DualUnionIteration.close() @bci=28, line=142 (Interpreted frame)
[...]

PR #4887, opened 2024-01-26
GH-4886 problem with use of @TestFactory and @TempDir

hmottestad:
GitHub issue resolved: #4886 

----

### DIFF
--- a/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQLComplianceTest.java
+++ b/testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQLComplianceTest.java
@@ -67,18 +67,14 @@ public abstract class SPARQLComplianceTest {
 			this.name = name;
 		}
 
-		public void test() {
+		public void test() throws Exception {
 			assumeThat(getIgnoredTests().contains(getName())).withFailMessage("test case '%s' is ignored", getName())
 					.isFalse();
 			try {
 				setUp();
 				runTest();
-			} catch (Exception e) {
-				try {
-					tearDown();
-				} catch (Exception e2) {
-
-				}
+			} finally {
+				tearDown();
 			}
 		}
 
--- a/compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbPARQL11UpdateComplianceTest.java
+++ b/compliance/sparql/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbPARQL11UpdateComplianceTest.java
@@ -11,6 +11,7 @@
 package org.eclipse.rdf4j.sail.lmdb;
 
 import java.io.File;
+import java.util.UUID;
 
 import org.eclipse.rdf4j.repository.Repository;
 import org.eclipse.rdf4j.repository.dataset.DatasetRepository;
@@ -29,7 +30,11 @@ public class LmdbPARQL11UpdateComplianceTest extends SPARQL11UpdateComplianceTes
 
 	@Override
 	protected Repository newRepository() throws Exception {
+
+		var temp = new File(folder, UUID.randomUUID().toString());
+		temp.mkdir();
+
 		return new DatasetRepository(
-				new SailRepository(new LmdbStore(folder, new LmdbStoreConfig("spoc"))));
+				new SailRepository(new LmdbStore(temp, new LmdbStoreConfig("spoc"))));
 	}
 }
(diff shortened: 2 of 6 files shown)


testsuites/sparql/src/main/java/org/eclipse/rdf4j/testsuite/query/parser/sparql/manifest/SPARQLComplianceTest.java
```diff
@@ -67,18 +67,14 @@ public DynamicSparqlComplianceTest(String displayName, String testURI, String na
...
 				runTest();
-			} catch (Exception e) {
-				try {
-					tearDown();
-				} catch (Exception e2) {
-
-				}
+			} finally {
+				tearDown();
 			}
```
hmottestad, line 78: This code was swallowing exceptions and also the tearDown() method was only called if there was an exception.

hmottestad:
@JervenBolleman I've run into a more general issue with your rewrite of the `SPARQLQueryComplianceTest` class et al. because the JUnit 5 `@TestFactory` annotation doesn't work very well with the `@TempDir` annotation. In fact `@TestFactory` doesn't work well with any of the lifecycle annotations, including before each. Essentially the combination of `@TestFactory` and `@TempDir` means that we use the same temporary directory for all the tests in a class, instead of a new sail and repository with a new temporary directory for each and every test.

This issue has been masked by the test code catching exceptions. Pretty much none of the NativeStore tests have been running. I managed to get the NativeStore tests working now, but the LmdbStore doesn't seem so happy about being continuously re-instantiated on top of the directory for an existing LmdbStore. @kenwenzel maybe you could take a look at why the LmdbStore gets upset after a while when being shutdown and then recreated on the same directory. You should be able to reproduce the issues by running `LmdbPARQL11UpdateComplianceTest`.

JervenBolleman:
Oh no, I am so sorry for this!

hmottestad:
"Aldri så galt at det ikke er godt for noe" - Norwegian way to say that even when bad things happen they still bring something good. The ExtensibleStore test failing led to discovering the issue with the tests here, and that led to discovering the issue with the LmdbStore :)

Maybe a quick fix is to add a File or Path param to the creeateRepo method and manually handle the creating and deletion of temporary directories.

Another approach would be to try to use the ParamaterizedTest annotation with the method source option. I do this with the ShaclSail where I read the directory structure to generate the tests at runtime.

It's still really nice to be able to run just a single test from IntelliJ though! Instead of always having to run every single one like I had to before.

JervenBolleman:
I tried to create a dir per dynamic test. Still have the failing extensiblestore test. Native and Lmdb tests seem to run?
The code is in [pull:4888](https://github.com/eclipse-rdf4j/rdf4j/pull/4888)

hmottestad:
Yeah, I think the ExtensibleStore failure is unrelated. It uses a ConcurrentHashMap as the main data structure when running the tests, so maybe there are some changes to that class in jdk 21 that is causing the issue. Not sure what it could be, and it's particularly annoying that I can't reproduce it in IntelliJ.

kenwenzel:
@hmottestad I could imagine that something is not correctly closed when unsing the LmdbStore. Have you checked that everything gets closed after running the tests?

hmottestad:
From what I could see everything is closed correctly, but I haven't done anything more than look over the code.

kenwenzel:
Is the store instantiated with an empty directory or are there any artifacts left from the previous run?
It might also be the case that the store is not correctly closed due to some issue like
https://github.com/eclipse-rdf4j/rdf4j/issues/4806

I'll try to take a look within the next days.

hmottestad:
> Is the store instantiated with an empty directory or are there any artifacts left from the previous run?
> 
> It might also be the case that the store is not correctly closed due to some issue like
> 
> https://github.com/eclipse-rdf4j/rdf4j/issues/4806
> 
> 
> 
> I'll try to take a look within the next days.

You can checkout this commit here: 5d825520e1c268222d6cb0163724c26ffa94f539

The issues occurs when the store is shutdown() and then recreated with the same directory. It doesn't happen immediately but after ~6 times.

kenwenzel:
> > Is the store instantiated with an empty directory or are there any artifacts left from the previous run?
> > It might also be the case that the store is not correctly closed due to some issue like
> > #4806
> > I'll try to take a look within the next days.
> 
> You can checkout this commit here: [5d82552](https://github.com/eclipse-rdf4j/rdf4j/commit/5d825520e1c268222d6cb0163724c26ffa94f539)
> 
> The issues occurs when the store is shutdown() and then recreated with the same directory. It doesn't happen immediately but after ~6 times.

@hmottestad I nailed this down to an issue with the LMDB ValueStore and its recycling of IDs. I hope that I can fix this within the next days.

Issue #4921, opened 2024-02-29
Turtle writer does not respect namespaces in IRIs


reckart:
### Current Behavior

The turtle writer does not consider the namespace provided by an IRI. Instead, it turns the IRI into a string and then tries to derive a namespace from that string and then looks up that namespace in the namespace table.

### Expected Behavior

If the IRI defines a namespace, that should first be used to look up the prefix in the namespace table. Interpreting the IRI as a string should only be a fallback.

### Steps To Reproduce

```
	@Test
	public void testIriNamespace() throws Exception {
	    Model model = new DynamicModelFactory().createEmptyModel();
	    String prefix = "foo-bar";
        String ns = "foo:this.is.my.bar.";
        model.setNamespace(prefix, ns);
	    model.add(vf.createIRI(ns, "lala"), vf.createIRI(ns, "lulu"), vf.createIRI(ns, "lolo"));

        StringWriter stringWriter = new StringWriter();
	    Rio.write(model, stringWriter, RDFFormat.TURTLE);
	    
	    assertThat(stringWriter.toString()).contains("foo-bar:lala foo-bar:lulu foo-bar:lolo .");
	}
```

### Version

4.3.9

### Are you interested in contributing a solution yourself?

Yes

reckart:
PR provided: https://github.com/eclipse-rdf4j/rdf4j/pull/4922

hmottestad:
Not seen this before. Can you point to where this is defined in the turtle specification?

reckart:
I have just ported code from Jena to RDF4J. In the Jena version, all IRIs I used were properly abbreviated when a namespace prefix was defined. Not so in RDF4J. I tracked this down to two issues:

1) SimpleIRI does not retain namespace and localname separately (I could work around this using a custom subclass of AbstractIRI)
2) TurtleWriter did not consider the explicit namespace in the IRI

Have a look at the provided test case.

As far as I can see during a quick scan of the Turtle specs, they deal mostly with parsing the turtle format. However, what I am trying to fix here is the serialization. The current serialization done by RDF4J provides valid Turtle, but it is unnecessarily verbose. My change makes more effective use of the namespace prefix declarations, leading to a serialized form that is more compact and easier to read.

PR #4922, opened 2024-02-29
GH-4921 Turtle writer does not respect namespaces in IRIs

reckart:
GitHub issue resolved: #4921

- SimpleValueFactory.createIRI(String, String) how actually properly respects the specified namespace and localname
- TurtleWriter tries to use the namespace encoded in the IRI unless the localname contains characters which are not valid in prefixed notation
- Added test

----

### DIFF
--- a/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleUtil.java
+++ b/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleUtil.java
@@ -460,14 +460,17 @@ public class TurtleUtil {
 	}
 
 	public static boolean isValidPrefixedName(String s) {
-		final int numberOfCodePoints = s.codePointCount(0, s.length());
-		for (int i = 1; i < numberOfCodePoints; i++) {
-			final int codePoint = s.codePointAt(i);
-			if (!isPN_CHARS(codePoint)) {
-				return false;
-			}
+		if (s == null || s.isEmpty()) {
+			return false;
 		}
-		return true;
+
+		if (!isPN_CHARS_BASE(s.codePointAt(0))) {
+			return false;
+		}
+
+		return s.codePoints() //
+				.skip(1) // Skip the first code point
+				.allMatch(TurtleUtil::isPN_CHARS);
 	}
 
 	/**
--- a/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleUtil.java
+++ b/core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleUtil.java
@@ -459,6 +459,17 @@ public class TurtleUtil {
 		return s;
 	}
 
+	public static boolean isValidPrefixedName(String s) {
+		final int numberOfCodePoints = s.codePointCount(0, s.length());
+		for (int i = 1; i < numberOfCodePoints; i++) {
+			final int codePoint = s.codePointAt(i);
+			if (!isPN_CHARS(codePoint)) {
+				return false;
+			}
+		}
+		return true;
+	}
+
 	/**
 	 * Decodes an encoded Turtle string. Any \-escape sequences are substituted with their decoded value.
 	 *
--- a/core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleWriterTest.java
+++ b/core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleWriterTest.java
@@ -212,7 +212,8 @@ public class TurtleWriterTest extends AbstractTurtleWriterTest {
 		var actual = Rio.parse(new StringReader(stringWriter.toString()), "", RDFFormat.TURTLE);
 		assertThat(Models.isomorphic(expected, actual)).as("isomorphic").isTrue();
 
-		assertThat(stringWriter.toString()).isEqualTo(data);
+		// Requires https://github.com/eclipse-rdf4j/rdf4j/issues/4929 to be fixed
+		// assertThat(stringWriter.toString()).isEqualTo(data);
 	}
 
 	@Test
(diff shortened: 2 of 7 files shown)


reckart:
Hm... need to check why the build fails...

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleUtil.java
```diff
@@ -459,6 +459,17 @@ public static String encodeURIString(String s) {
...
+		final int numberOfCodePoints = s.codePointCount(0, s.length());
+		for (int i = 1; i < numberOfCodePoints; i++) {
+			final int codePoint = s.codePointAt(i);
+			if (!isPN_CHARS(codePoint)) {
+				return false;
+			}
+		}
+		return true;
+	}
+
```
hmottestad, line 472: Maybe you could use the codepoints() method?

```
public static boolean isValidPrefixedName(String s) {
    return s.codePoints().skip(1) // Skip the first code point
           .allMatch(TurtleUtil::isPN_CHARS);
}
```

core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleWriterTest.java
```diff
@@ -672,6 +674,20 @@ public void testBlankNodeInlining_indirectCircularReferenceWithIRI() throws Exce
 		assertTrue(Models.isomorphic(expected, actual));
 	}
 
+	@Test
+	public void testIriNamespace() throws Exception {
+		Model model = new DynamicModelFactory().createEmptyModel();
+		String prefix = "foo-bar";
+		String ns = "foo:this.is.my.bar.";
+		model.setNamespace(prefix, ns);
+		model.add(vf.createIRI(ns, "lala"), vf.createIRI(ns, "lulu"), vf.createIRI(ns, "lolo"));
```
hmottestad, line 683: So the full IRI of the subject would be `foo:this.is.my.bar.lala`. I'm still not convinced that this is a valid IRI, but I'll try to take a closer look at the spec again.

hmottestad: Had a look now and found the example section of the URI spec. Since all URIs are valid IRIs then this should hold.

https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.2

This example here is similar to your test case: `news:comp.infosystems.www.servers.unix`

hmottestad:
Looks good in general. There are some test failures, I triggered the CI again to see if the tests still fail.

Any chance you could create a test that writes -> parses -> writes to see if the output stays the same?

core/rio/turtle/src/main/java/org/eclipse/rdf4j/rio/turtle/TurtleUtil.java
reckart: Will be included in PR update.

reckart:
@hmottestad I have added two additional tests that check to which extend the format is preserved in a round trip. While the data is isomorphic, the actual serialization is different.

E.g. if I parse this:

```
prefix server-news: <news:comp.infosystems.www.servers.> .
server-news:unix rdfs:label "News on Unix" .
server-news:windows rdfs:label "News on Windows" .
```

and serialize it out again, it comes out as:

```
@prefix server-news: <news:comp.infosystems.www.servers.> .
<news:comp.infosystems.www.servers.unix> rdfs:label "News on Unix" .
<news:comp.infosystems.www.servers.windows> rdfs:label "News on Windows" .
```

That seems to be because the `AbstractRDFParser` fails to consider the namespace declarations from the Turtle when generating the IRI instances.

While I think that should also be fixed, it seems to me a) worth a new issue, b) a larger change and for me personally, I'm currently more interested in the serialization than in the parsing - so I'd defer that.

So if we decide to keep these tests, at least the string comparison on the Turtle serialization would need to be removed or commented out until the parser properly respects the prefix declarations in the source file.

reckart:
So a the reason why `testPerformance(org.eclipse.rdf4j.rio.rdfxml.RDFXMLPrettyWriterBackgroundTest)` fails the is that the RDFXML writer actually respects prefix/namespace/localname delarations in IRIs, but because the test actually declares a IRI with an invalid localname, the test breaks due to the change that `SImpleIRI` now actually preserves this properly.

**org.eclipse.rdf4j.rio.RDFWriterTest.RDFWriterTest(RDFWriterFactory, RDFParserFactory)**
```
		uri5 = vf.createIRI(exNs, "uri5/you");
```

Before the improvement to `SimpleIRI` to properly preserve namespace and localname, this was serialized as
```
<you xmlns="http://example.org/uri5/" rdf:about="http://example.org/49/a111"/>
```

Now, this comes out as

```
<ex:uri5/you rdf:about="http://example.org/49/a111"/>
```

kenwenzel:
I am not sure if it is desirable to preserve the initial namespace-localname-split in SimpleIRI. How does this influence the behavior of the generic IRI interface? 
If the IRI is stored in NativeStore or LmdbStore in between then the original components will be lost.

reckart:
@kenwenzel for my use-case, I do not really care if prefixed and namespaces are always preserved. But what I care for is that if I build a model in which I explicitly declare prefixes/namespaces and also if I create IRIs with explicit namespaces, then this should be respected.

In general, I would think it would be sensible to also respect declared namespaces and prefixes when reading IRIs from a store and then serializing them into a file. If the file has prefix declarations, then it would be sensible to try matching the IRI against these prefixes and then use the best match. But this would be slower then just doing an IRI namespace lookup in the prefix table, so I opted for the moment against implementing that.

hmottestad:
@reckart this fix here and the other fix you've contributed are released in 4.3.11.

reckart:
Wow, a release specifically with my contributions :) That is a great easter present 👍 Thanks.

Issue #4952, opened 2024-04-17
FedX - Allow for FedXConfig options to be overriden in repo config templates


vtermanis:
### Problem description

We (at [Iotics](https://github.com/Iotic-Labs)) run RDF4J as a container and configure FedX at deploy time using [a config template](https://rdf4j.org/documentation/programming/federation/#advanced-federation-using-a-repository-config-template). It would be useful to be able to override defaults of `FedXConfig` options such as `withEnforcemaxQueryTime` and `withSourceSelectionCacheSpec` in the template.

### Preferred solution

Be able to specify `FedXConfig` options in the template so it's possible e.g. change the maximum query time and source selection cache configuration, e.g.:

```turtle
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix rep: <http://www.openrdf.org/config/repository#>.
@prefix fedx: <http://rdf4j.org/config/federation#>.

[] a rep:Repository ;
   rep:repositoryImpl [
      rep:repositoryType "fedx:FedXRepository" ;
      fedx:config [
         fedx:enforceMaxQueryTime 42 ;
         fedx:enableMonitoring true ;
      ];
   ];
   rep:repositoryID "my-federation" ;
   rdfs:label "FedX Federation" .
```

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

Because `FedXRepositoryConfig` already handles FedX repo config templates, we have not considered any alternatives.

### Anything else?

We don't know whether it makes sense to expose all or just a subset of the available options. (See https://github.com/eclipse-rdf4j/rdf4j/pull/4953)

PR #4953, opened 2024-04-17
GH-4952 - Introduce FedXConfig overrides for FedXRepositoryConfig

vtermanis:
GitHub issue resolved: #4952 

Introduce `FedXConfig` option overrides for `FedXRepositoryConfig` so that repo templates can also access options which previously were only available programmatically.

We've raised the PR against the `main` branch because from our understanding it's not clear yet when v5.0.0 will be released and we'd be looking to use this feature in production ourselves. Looking at what's changed between 4 latest & 5 (against the files affected by this change) I think it should be trivial to port.

----

    - **I will squash all the `fixup` commits once review is complete.**


tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
```diff
@@ -187,6 +203,14 @@ public void parse(Model m, Resource implNode) throws RepositoryConfigException {
 			Models.objectLiteral(m.getStatements(implNode, DATA_CONFIG, null))
 					.ifPresent(value -> setDataConfig(value.stringValue()));
 
+			Models.objectResource(m.getStatements(implNode, FEDX_CONFIG, null))
+					.ifPresent(res -> {
+						if (getConfig() == null) {
```
vtermanis, line 208: Note: I chose to do this so the previous behaviour is preserved:
1. If no `FedXConfig` has been provided (programmatically or via template) then it will be `null`
2. When exporting, `FedXConfig` is not included (when `null`) rather than being expanded into defaults (see `export` above)

vtermanis:
Note: Integration test pipeline has failed due to an unrelated reason:
> compliance/rio/src/test/java/org/eclipse/rdf4j/rio/n3/N3ParserTest.java:28 | Expected '.', found '3' [line 2]

hmottestad:
I've set it to rerun the integration test. I've not seen this test being flaky before, but maybe it is.

hmottestad:
It's not an issue with your code. It's a test that is marked as disabled, but has somehow started to run again. Not sure how that happened.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
```diff
@@ -152,6 +163,11 @@ public Resource export(Model m) {
 			m.add(implNode, DATA_CONFIG, vf.createLiteral(getDataConfig()));
 		}
 
+		if (getConfig() != null) {
```
aschwarte10, line 166: This will mean: any externally set `FedXConfig` (e.g. via code) will be serialized into the repository configuration, e.g. when saving it.

I think overall (from history) there is a bit of technical design issue in the `FedXRepositoryConfig` and `FedXConfig`: the `FedXConfig` mixes "string" like options and things that can only be set via code (e.g. the TaskWrapper). 

For our product use-case: we are having an extension of FedXConfig (with some proprietary options, and also a TaskWrapper) and we are setting this config via code. We generally would not want to see it exported to the repository config.

So maybe to retain compatibility on that level, we can use one of the following options:

a) introduce another boolean field (say `isExternalConfig`)  and only export if it is not externally provided
b) put the handling of the FedXConfig for both parsing and exporting to a protected method, which can be overridden in customizations of the repository config

I think option b) is cleaner an will allow most flexibility and still retain compatibility.

So the idea would be to move your new exporting/parsing part of the code from this class to new protected methods

* `protected void exportFedXConfig(Model model, Resource implNode)`
* `protected void parseFedXConfig(Model model, Resource implNode)`

and call those in the current places.

Does this make sense to you?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
aschwarte10: Good considerations! Please see my comments for the exporting

aschwarte10 (commented):
Thanks for the initiative and contribution. I see the value of making certain settings of the federation easier configurable via the repository configuration file (e.g. managed from the RDF4J workbench). Actually I was also looking for such feature.

Overall the functional change looks clean and good. Thanks for the unit test coverage.

I have one comment though inline on compatibility (at least with the use FedX in our product). Can you please have a look at that?

In addition, users would benefit from a small documentation adjustment of https://rdf4j.org/documentation/programming/federation/#fedx-configuration (stating with a small example that settings can be added in the repository configuration). Likely the example from the javadoc is sufficient already. As far as I remember the documentation is also built from git, the location should be `rdf4j/site/content/documentation/programming/federation.md`

Regarding the contribution procedures I'd also like to see a confirmation from @hmottestad

aschwarte10:
@vtermanis did you have a chance to look at the above feedback? I think it would be ready with a minimal technical change of moving some pieces of code to protected methods.

vtermanis:
@aschwarte10  - sorry about the delay, we've been busy. I will update the PR based your recommendations. Thank you for the initial review.
Follow up Q: It occurred to me that it'll be useful to us to also be able to tweak the join block size and maybe the worker thread counts. To that effect: Is it worth exposing **all** (apart from from TaskWrapper) options? Are there any which you don't think people should touch?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
vtermanis: Yes - that sounds good. I'll have a go.

vtermanis: On second thought - the proposed method signatures allow sub-classes to do whatever they want (be it config-related or not). Would it not make more sense to pass them the node which contains the config (like is done in the separate class) - separation of concerns and all that.
And e.g. allow `new FedXConfig(ConfigReader)` to inject a custom or sub-classed implementation?

(But I also appreciate that this probably doesn't need to be over-engineered.)

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
vtermanis: Changed with the other suggestion: 5a160dfd1bfcaa48e3c5

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXConfigParser.java
```diff
@@ -0,0 +1,129 @@
...
+ * @author Iotic Labs
+ */
+public class FedXConfigParser {
+
+	private static final ValueFactory vf = SimpleValueFactory.getInstance();
+
+	/**
+	 * IRI of the property populating {@link FedXConfig#getEnforceMaxQueryTime()}
+	 */
+	public static final IRI CONFIG_ENFORCE_MAX_QUERY_TIME = vf.createIRI(NAMESPACE, "enforceMaxQueryTime");
```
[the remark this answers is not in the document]
vtermanis: 5a160dfd1bfcaa48e3c5

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
vtermanis: 5a160dfd1bfcaa48e3c5

vtermanis:
I've added the remaining conf options: f72c49a030b1f244

vtermanis:
> In addition, users would benefit from a small documentation adjustment of https://rdf4j.org/documentation/programming/federation/#fedx-configuration (stating with a small example that settings can be added in the repository configuration).

I've added that in a separate commit - see what you think.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
```diff
@@ -205,4 +292,120 @@ public void parse(Model m, Resource implNode) throws RepositoryConfigException {
...
+	 * @param implNode the node to which to write the config reference (i.e. {@link #FEDX_CONFIG}) to
+	 */
+	protected void exportFedXConfig(Model model, Resource implNode) {
+		if (getConfig() == null) {
+			return;
+		}
+
+		BNode confNode = Values.bnode();
+
+		model.add(confNode, CONFIG_JOIN_WORKER_THREADS, vf.createLiteral(config.getJoinWorkerThreads()));
```
aschwarte10, line 374: the downside of the current implementation is that exporting will also write the FedXConfig default values, i.e. values that may not have been present in the repository configuration ttl before.

Ideally the parse / export logic would be consistent, i.e. running parse and export produces the same result.

I think I could live with this, but I would like hear @hmottestad's opinion whether this is in line with the RDF4J design and other repo configurations.

aschwarte10 (commented):
Thanks for addressing the comments.

Overall the change now looks clean and solid. I am fine to expose all configuration settings also as part of the repository config.

I have one comment / question inline, where I'd like to hear @hmottestad 's feedback.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
vtermanis: > Ideally the parse / export logic would be consistent, i.e. running parse and export produces the same result.

But isn't the issue here that we don't know for sure whether the user explicitly set values. Also, isn't being explicit better than implicit?
(Otherwise I suppose one could, for each individual value, check the default first)

site/content/documentation/programming/federation.md
```diff
@@ -299,9 +299,10 @@ FedX provides various means for configuration. Configuration settings can be def
 |Property | Description |
 |---------|-------------|
 |prefixDeclarations | Path to prefix declarations file, see [PREFIX Declarations](#prefix-declarations) |
-|cacheLocation | Location where the memory cache gets persisted at shutdown, default _cache.db_ |
+|sourceSelectionCacheSpec | Cache specification for the `SourceSelectionMemoryCache`, default _maximumSize=1000,expireAfterWrite=6h_ |
```
conbon, line 302: missing includeInferredDefault / consumingIterationMax

vtermanis: 290d78ce00

tools/federation/src/main/java/org/eclipse/rdf4j/federated/repository/FedXRepositoryConfig.java
aschwarte10: I would leave it up to @hmottestad as he has likely the broadest overview at this point.

As a side note: we are usually solving the parsing / export consistency by having "null" state for the field (i.e. Boolean, Integer, ...). In the current design this does not really fit as it uses the API level FedXConfig. To achieve this properly something similar one would need a repository level FedXConfig and some translation to the api level FedXConfig.

hmottestad:
@aschwarte10 I've created a CQ for the license issue, but you don't need to wait for that before you merge this PR.

aschwarte10 (approved):
I had the chance to have a conversation with @hmottestad and we are both fine with the exporting of defaults. Sorry for the delay

Hence, for me the change looks good to me. Thanks for the clean and concise contribution. 

We will see this change as part of the 5.0 release of rdf4j.

vtermanis:
> Sorry for the delay

No, problem. Thanks for the review!

Issue #4968, opened 2024-04-30
server-sprint - AbstractQueryRequestHandler does not close result when an exception is raised


vtermanis:
### Current Behavior

During handling of a query in [AbstractQueryRequestHandler](https://github.com/eclipse-rdf4j/rdf4j/blob/4.3.11/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java#L57), if an exception occurs [between evaluation and return of the model & view](https://github.com/eclipse-rdf4j/rdf4j/blob/4.3.11/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java#L86-L100), the result is not closed and thus, during closing of the connection it is possible that [another exception is raised](https://github.com/eclipse-rdf4j/rdf4j/blob/4.3.11/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java#L296) due to this.

### Expected Behavior

Exceptions thrown in the handler should be passed on to the client (e.g. `HTTP/406` for an unsupported content type) instead of sometimes resulting in an `HTTP/500` error.

### Steps To Reproduce

- :information_source:  Versions: 4.3.11 / 5.0.0M2 (no significant changes spotted for affected code between these two)

We have observed this when users perform a FedX query with incompatible content type: e.g. `DESCRIBE` with `Accept: application/sparql-results+json`. Most of the time we just see [this warning](https://github.com/eclipse-rdf4j/rdf4j/blob/4.3.11/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/CleanerIteration.java#L68-L69) together with the expected `HTTP/406` response but **sometimes** instead an `HTTP/500` is returned (due to the aforementioned exception).

### Version

4.3.11

### Are you interested in contributing a solution yourself?

Yes.

### Anything else?

One could close the result on exception, just like the connection (see [draft PR](https://github.com/eclipse-rdf4j/rdf4j/pull/4969)). This solves it for our use-case, I believe.
[...]

PR #4969, opened 2024-04-30
GH-4968: server-spring - Close query result on pre-render exception

vtermanis:
GitHub issue resolved: #4968

- Introduce explicit result closing (in addition to existing repo close) when an exception occurs between the evaluation & rendering stages.
- :information_source: I have **not** added any tests yet since I'd like to get feedback on whether this is the right approach (and there weren't any other pre-existing tests for the `AbstractQueryRequestHandler` class.

----

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java
@@ -58,6 +58,7 @@ public abstract class AbstractQueryRequestHandler implements QueryRequestHandler
 			HttpServletResponse response) throws HTTPException, IOException {
 
 		RepositoryConnection repositoryCon = null;
+		Object queryResponse = null;
 
 		try {
 			Repository repository = repositoryResolver.getRepository(request);
@@ -75,9 +76,6 @@ public abstract class AbstractQueryRequestHandler implements QueryRequestHandler
 			boolean distinct = isDistinct(request);
 
 			try {
-
-				Object queryResponse;
-
 				if (headersOnly) {
 					queryResponse = null;
 				} else {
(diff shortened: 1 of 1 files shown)


tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java
```diff
@@ -114,17 +112,25 @@ public ModelAndView handleQueryRequest(HttpServletRequest request, RequestMethod
...
-			if (repositoryCon != null) {
-				repositoryCon.close();
+			// only close the response & connection when an exception occurs. Otherwise, the QueryResultView will take
+			// care of closing it.
+			try {
+				if (queryResponse != null) {
+					queryResponse.close();
+				}
+			} catch (Exception qre) {
+				logger.warn("Query response closing error", qre);
```
vtermanis, line 122: Question: I thought it might be sensible here to log the result-close failure so the client still receives the original failure reason. Is that the right approach? And: Would it make sense to do the same for `repositoryCon` below?

JervenBolleman (approved):
I think this is an improvement and can be merged as is.

vtermanis:
It looks like the integration test failure is unrelated (in that this from my understanding doesn't use `server-spring`):
```
testsuites/repository/src/main/java/org/eclipse/rdf4j/testsuite/repository/optimistic/ModificationTest.java:180 | Malformed query result from server
```

hmottestad:
I'm rerunning that GitHub action to see if it is just temperamental. I haven't seen it being flaky before though.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java
```diff
@@ -114,17 +112,25 @@ public ModelAndView handleQueryRequest(HttpServletRequest request, RequestMethod
...
 
 		} catch (Exception e) {
-			// only close the connection when an exception occurs. Otherwise, the QueryResultView will take care of
-			// closing it.
-			if (repositoryCon != null) {
-				repositoryCon.close();
+			// only close the response & connection when an exception occurs. Otherwise, the QueryResultView will take
+			// care of closing it.
+			try {
+				if (queryResponse != null) {
```
hmottestad, line 118: Maybe you can simplify your code down to an instanceOf AutoClosable check here without having to introduce the Boolean query stuff.

hmottestad:
Test is still failing. I suspect it's related to the Boolean query changes you've introduced. I added a proposal to your code for a simple instanceOf AutoCloseable check instead.

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/repository/handler/AbstractQueryRequestHandler.java
vtermanis: Ifb01e2a2681ddbd150f

vtermanis:
I just realised that I'm probably doing the whole "add more commits during review" thing wrong: Internally we use `!fixup` commits and once the review is complete - we do a `git rebase --auto-squash` to push everything together.

What's your preference?
1. Add extra commits (but history looks strange like e.g. here where I e.g undid some changes)
2. Force-push to replace the existing commit(s) with updated versions
3. Something else?

hmottestad (approved):

vtermanis:
Squashed

vtermanis:
Ah wait - should the target branch also be `develop` instead of `main`? (like with the other PR I submitted)

Issue #4899, opened 2024-02-06
MapDB3 based collection factory should provide a queue and set implementation


JervenBolleman:
### Problem description

[qEndpoint](https://github.com/the-qa-company/qEndpoint/issues/349) has an issue where a path query triggers an out of memory exception. See the makeQuey and makeSet calls in [PathIteration](https://github.com/eclipse-rdf4j/rdf4j/blob/develop/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/PathIteration.java). 

For the MapDb3 implementation we should be able to [add a queue](https://github.com/eclipse-rdf4j/rdf4j/blob/3996cd843dbd9b65f130df4839fb627755c1595e/core/collection-factory/mapdb3/src/main/java/org/eclipse/rdf4j/collection/factory/mapdb/MapDb3CollectionFactory.java#L175) that spills to disk.

### Preferred solution

Implement a java.util.Queue wrapper for a mapdb3 QueuLong that should be able to work here.

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

Spilling to disk might make existing queries much slower. Perhaps take a much higher switch point from in memory to mapdb then the MemoryTillSizeXSet default.

kenwenzel:
Do you think it makes sense to consider an LMDB-based persistent data structure?
I've added a PersistentSet to the LmdbStore module.

JervenBolleman:
An lmdb based collection factory is still in my would like to do list.

JervenBolleman:
MapDb3 QueueLong is undocumented and does not work like I assumed. Have a simple wrapper around a map where head and tail are maintained as keys.

PR #4970, opened 2024-05-01
Gh 4899 allow for storage specialization in path iteration

JervenBolleman:
GitHub issue resolved: #4899

The PathIterartion has  a data structure for a pair of values. I made this extend `BindingSet` so that we can use the capability for implementing specialized collections also for this value pairs.

While doing so I noticed a bug due to a path name not being the same in places. Avoided this in the future by using methods to generate the same string for a variable name by reducing code duplication.

----


core/collection-factory/api/src/main/java/org/eclipse/rdf4j/collection/factory/api/CollectionFactory.java
```diff
@@ -116,6 +116,25 @@ public Set<BindingSet> createSetOfBindingSets(Supplier<MutableBindingSet> create
 	 */
 	public Queue<Value> createValueQueue();
 
+	/**
+	 * @return a new queue that may be optimized and may use the functions passed in.
+	 */
+	public default Queue<BindingSet> createBindingSetQueue(Supplier<MutableBindingSet> create,
```
hmottestad, line 122: Could you mark this as Experimental? So that we can change it later if we need.

core/collection-factory/api/src/main/java/org/eclipse/rdf4j/collection/factory/api/CollectionFactory.java
```diff
@@ -116,6 +116,25 @@ public Set<BindingSet> createSetOfBindingSets(Supplier<MutableBindingSet> create
...
+			Function<String, Predicate<BindingSet>> getHas, Function<String, Function<BindingSet, Value>> getget,
+			Function<String, BiConsumer<Value, MutableBindingSet>> getSet) {
+		return createQueue();
+	}
+
+	/**
+	 * @return a new queue optimized for bindingsets
+	 */
+	public default Queue<BindingSet> createBindingSetQueue() {
```
hmottestad, line 131: Could you mark this as Experimental? So that we can change it later if we need.

core/collection-factory/mapdb3/src/main/java/org/eclipse/rdf4j/collection/factory/mapdb/MapDb3CollectionFactory.java
```diff
@@ -259,6 +259,22 @@ public Queue<Value> createValueQueue() {
 		}
 	}
 
+	@Override
+	public Queue<BindingSet> createBindingSetQueue(Supplier<MutableBindingSet> create,
```
hmottestad, line 263: Could you mark this as Experimental? So that we can change it later if we need.

hmottestad (approved):
LGTM. Can't say I exactly understand what's going on. Only thing I would like to change is to mark the new public methods and interface methods as experimental so that we can change them later without waiting for 6.0.0. Up to you though.

JervenBolleman:
I turned on auto-merge if passes.

PR #4972, opened 2024-05-05
GH-4920 SPARQLConnection.size() now uses count query

### LINKED ISSUE #4920: SPARQLConnection size method counts client side instead of server side
### Current Behavior

```java
	@Override
	public long size(Resource... contexts) throws RepositoryException {
		try (RepositoryResult<Statement> stmts = getStatements(null, null, null, true, contexts)) {
			long i = 0;
			while (stmts.hasNext()) {
				stmts.next();
				i++;
			}
			return i;
		}
	}
```

### Expected Behavior

Should be using a simple SPARQL query with a count.

### Steps To Reproduce

Run [void-generator](https://github.com/JervenBolleman/void-generator) see all triples from an endpoint being downloaded.

### Version

4.3.8

Yes

### Anything else?

_No response_

JervenBolleman:
GitHub issue resolved #4920

SPARQLConnection.size() method should not fetch every statement in the repository. Just send a count query instead.

----

 - [x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-chxanges-pull-requests-and-commits)

### DIFF
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -42,6 +42,8 @@ import org.eclipse.rdf4j.model.Value;
 import org.eclipse.rdf4j.model.impl.DynamicModelFactory;
 import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.model.util.Literals;
+import org.eclipse.rdf4j.model.vocabulary.RDF4J;
+import org.eclipse.rdf4j.model.vocabulary.SESAME;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.BooleanQuery;
 import org.eclipse.rdf4j.query.GraphQuery;
@@ -313,12 +315,13 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 	String sizeAsTupleQuery(Resource... contexts) {
 		String query = COUNT_EVERYTHING;
 		if (contexts != null && isQuadMode()) {
-			if (contexts.length == 1 && contexts[0].isIRI()) {
-				query = "SELECT (COUNT(*) AS ?count) WHERE { GRAPH <" + ((IRI) contexts[0]).stringValue()
+			if (contexts.length == 1 && isExposableGraphIri(contexts[0])) { // in case the context is null we want the
+																			// default graph.
+				query = "SELECT (COUNT(*) AS ?count) WHERE { GRAPH <" + contexts[0].stringValue()
 						+ "> { ?s ?p ?o}}";
 			} else if (contexts.length > 0) {
 				String graphs = Arrays.stream(contexts)
-						.filter(Resource::isIRI)
+						.filter(SPARQLConnection::isExposableGraphIri)
 						.map(Resource::stringValue)
 						.map(s -> "FROM <" + s + ">")
 						.collect(Collectors.joining(" "));
@@ -326,6 +329,17 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 			}
 		}
 		return query;
+	}
+
+	/**
+	 * For the sparql protocol a context must be an IRI However we can't send out the RDF4j intenral default graph IRIs
+	 *
+	 * @param resource to test if it can be the IRI for a named graph
+	 * @return true if it the input can be a foreign named graph.
+	 */
+	private static boolean isExposableGraphIri(Resource resource) {
+		// We use the instanceof test to avoid any issue with a null pointer.
+		return resource instanceof IRI && RDF4J.NIL != resource && SESAME.NIL != resource;
 	}
 
 	@Override
--- a/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
+++ b/core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
@@ -313,33 +313,40 @@ public class SPARQLConnection extends AbstractRepositoryConnection implements Ht
 	}
 
 	String sizeAsTupleQuery(Resource... contexts) {
-		String query = COUNT_EVERYTHING;
-		if (contexts != null && isQuadMode()) {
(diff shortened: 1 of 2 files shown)


JervenBolleman:
@hmottestad did you see this license failure before?

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -288,14 +292,40 @@ public boolean isEmpty() throws RepositoryException {
...
+			}
+		} catch (QueryEvaluationException e) {
+			throw new RepositoryException(e);
+		}
+		return 0;
+	}
+
+	String sizeAsTupleQuery(Resource... contexts) {
+		String query = COUNT_EVERYTHING;
+		if (contexts != null && isQuadMode()) {
```
hmottestad, line 315: I think contexts should never be null, but contexts[0] can probably be null if the user is asking for the default context.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -288,14 +292,40 @@ public boolean isEmpty() throws RepositoryException {
...
+			if (contexts.length == 1 && contexts[0].isIRI()) {
+				query = "SELECT (COUNT(*) AS ?count) WHERE { GRAPH <" + ((IRI) contexts[0]).stringValue()
+						+ "> { ?s ?p ?o}}";
+			} else if (contexts.length > 0) {
+				String graphs = Arrays.stream(contexts)
+						.filter(Resource::isIRI)
```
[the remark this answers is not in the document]
JervenBolleman: I SPARQL named graphs are not allowed to be blank nodes. So it could have worked if the stars aligned, but never reliably.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
hmottestad: RDF4J has the rdf4j:nil context and also a sesame one for addressing the default context. I believe it's a general issue with SPARQL that there is no way to address just the triples in the unnamed context. Jena also has their own: https://jena.apache.org/documentation/tdb/datasets.html#special-graph-names

Maybe we can test if either one of these returns results, and if not fallback to the old method?

hmottestad: It's a problem only when the default graph in SPARQL is the union of all graphs, but this is what I've found most sensible for all applications that I've used.

hmottestad:
> @hmottestad did you see this license failure before?

I haven't seen that one before, but I have once experienced an old dependency suddenly failing the license check. I didn't figure out what was wrong last time, ended up just upgrading the dependency. Let me try to submit it to clearlydefined. Might take some time though :(

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
JervenBolleman: I will check that a null context[0] will work as well as the rdf4j:nil or semame:nil. For the cases where the remote uses a non union default graph the old case would also have returned 0.

hmottestad:
> > @hmottestad did you see this license failure before?
> 
> I haven't seen that one before, but I have once experienced an old dependency suddenly failing the license check. I didn't figure out what was wrong last time, ended up just upgrading the dependency. Let me try to submit it to clearlydefined. Might take some time though :(

CQ: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14675

core/repository/sparql/src/test/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnectionTest.java
```diff
@@ -100,6 +114,26 @@ public void testAddSingleContextHandling() throws Exception {
 		assertThat(sparqlUpdate).containsPattern(expectedAddPattern).containsPattern(expectedRemovePattern);
 	}
 
+	@Test
+	public void testSizeQuery() throws Exception {
```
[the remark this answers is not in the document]
aschwarte10: I think in FedX federation part we have spin up a SPARQL embedded server for integration testing. Maybe you can borrow some ideas from there 

https://github.com/eclipse-rdf4j/rdf4j/blob/main/tools/federation/src/test/java/org/eclipse/rdf4j/federated/server/SPARQLEmbeddedServer.java

hmottestad:
I'm still a bit unsure about this. We want to be able to correctly count the statements in the default/unnamed graph. I don't know how well the previous solution handled it, but I would assume that it would be able to check for if the context is null on each statement.

core/repository/sparql/src/main/java/org/eclipse/rdf4j/repository/sparql/SPARQLConnection.java
```diff
@@ -288,14 +294,52 @@ public boolean isEmpty() throws RepositoryException {
...
+
+	/**
+	 * For the sparql protocol a context must be an IRI However we can't send out the RDF4j intenral default graph IRIs
+	 *
+	 * @param resource to test if it can be the IRI for a named graph
+	 * @return true if it the input can be a foreign named graph.
+	 */
+	private static boolean isExposableGraphIri(Resource resource) {
+		// We use the instanceof test to avoid any issue with a null pointer.
+		return resource instanceof IRI && RDF4J.NIL != resource && SESAME.NIL != resource;
```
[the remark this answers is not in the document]
JervenBolleman: Fixed

JervenBolleman:
> I'm still a bit unsure about this. We want to be able to correctly count the statements in the default/unnamed graph. I don't know how well the previous solution handled it, but I would assume that it would be able to check for if the context is null on each statement.

The behavior is the same between this code and previous regarding the default graph.
Before:
```sparql
select * where {?s ?p ?o}
```

New
```sparql
select (count(*) as ?c) where {?s ?p ?o}
```

The logic is a bit more robust in the case of using the internal default graph IRI's of RDF4j. Which should not be send over the wire.

Issue #4973, opened 2024-05-08
Improve performance of InnerMergeJoin


hmottestad:
Improve performance of InnerMergeJoin.

PR #4975, opened 2024-05-10
GH-4973 improve performance og inner merge join

hmottestad:
GitHub issue resolved: #4973 

----

### DIFF
--- a/assembly-descriptors/pom.xml
+++ b/assembly-descriptors/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.rdf4j</groupId>
 		<artifactId>rdf4j</artifactId>
-		<version>5.0.0-SNAPSHOT</version>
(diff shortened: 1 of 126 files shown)


kenwenzel:
I am not sure if something like this could be also helpful in some situations:
https://github.com/linkedfactory/linkedfactory-pod/blob/main/bundles/io.github.linkedfactory.core/src/main/java/io/github/linkedfactory/core/rdf4j/common/query/CompositeBindingSet.java

The binding set is helpful if only few additional bindings are generated via a join. I use this for evaluating joins over time series data.

hmottestad:
Ahh. So you can keep multiple references to a single binding set and for each of them add another value. 

I once made a StatementBindingSet that wrapped a statement and transparently mapped the variable names to the correct component of the statement. That in combination with your suggestion could really make for a much more lightweight binding set.

hmottestad:
Squash and merge!

Issue #4983, opened 2024-05-21
Fine tune bind-join configuration in FedX to modern environments


aschwarte10:
### Problem description

The original bind join implementation (c.f. `CthontrolledWorkerBoundJoin`) was developed when there was no SPARQL 1.1 was around and the implementation was using UNION constructs to implement the bind join.

A few years back the implementation was changed to make use of the VALUES clause.

In validations we found that nowadays the original tuning parameters (bind-join size etc) are not the optimal values anymore. Note that the bind-join size impacts the number of sub-queries for large joins, and hence the use of threads from the join thread pool.

Hence we recommend to fine-tune the implementation, to increase overall performance and query throughput.

### Preferred solution

We propose to increase the default bind join size to a value of 25.

In addition we propose to remove the original implementation of making the first few bind-join sub-queries with small blocks. In the current implementation it was: 

- 1st binding of intermediate result is send in non-bound way
- then there is 3 sub-queries with bind join size 3 (hard coded)
- only then the configurable bind join size applies

Especially the hard-coded sub-queries with bind join size 3 occupy threads from the pool which are not available for other parallel queries.

### Are you interested in contributing a solution yourself?

Yes

PR #4984, opened 2024-05-21
GH-4983: fine-tune bind join implementation in FedX engine

aschwarte10:
GitHub issue resolved: #4983

This change applies fine-tuning to the bind join implementation on
certain parameters.

Note that the original implementations and benchmarks were done when
there was no SPARQL 1.1 around, and FedX was using UNION constructs for
the bind join implementation. Since a few year back it is using VALUES
clauses.

Changes:

- default bind join size is set to 25 (also reflected in documentation,
some unit tests require adjustments)
- no longer send small hard-coded sub-queries of size 3 for the first
requests and instead rely on configured bind join size
- no longer send the first request in a non-bound way by default (Note:
kept as internal configuration parameter in the implementation to allow
modifications in extensions)
- support to influence the bind join size in customizations

-----

Note: already in a quick local validation using `MediumConcurrencyTest` with 1000 queries, we find an improvement in performace. Without these changes the test executes in ~5.5s (averaged) while after this changes it executes in ~4.8s (averaged). This is impacted by the number of reduced remote requests, and thereby the occupation of threads from the pool

In case of specific questions on code changes, please reach out. I'd like to ideally get this in for M3 for easier use in our product and therefore more thorough evaluation

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConfig.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConfig.java
@@ -40,7 +40,7 @@ public class FedXConfig {
 
 	private int leftJoinWorkerThreads = 10;
 
-	private int boundJoinBlockSize = 15;
+	private int boundJoinBlockSize = 25;
 
 	private int enforceMaxQueryTime = 30;
 
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
@@ -34,7 +34,7 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Execute the nested loop join in an asynchronous fashion, using grouped requests, i.e. group bindings into one SPARQL
- * request using the UNION operator.
+ * request using a VALUES clause.
  *
  * The number of concurrent threads is controlled by a {@link ControlledWorkerScheduler} which works according to the
  * FIFO principle and uses worker threads.
@@ -49,11 +49,20 @@ public class ControlledWorkerBoundJoin extends ControlledWorkerJoin {
 
 	private static final Logger log = LoggerFactory.getLogger(ControlledWorkerBoundJoin.class);
 
+	/**
+	 * Whether to submit the first intermediate result immediately in a non-bound way
+	 */
+	private boolean submitFirstResultImmediately = false;
+
 	public ControlledWorkerBoundJoin(ControlledWorkerScheduler<BindingSet> scheduler, FederationEvalStrategy strategy,
 			CloseableIteration<BindingSet> leftIter,
 			TupleExpr rightArg, BindingSet bindings, QueryInfo queryInfo)
 			throws QueryEvaluationException {
 		super(scheduler, strategy, leftIter, rightArg, bindings, queryInfo);
+	}
+
+	protected void setSubmitFirstResultImmediately(boolean flag) {
+		this.submitFirstResultImmediately = flag;
 	}
 
 	@Override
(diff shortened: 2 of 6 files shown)


aschwarte10:
@hmottestad or @JervenBolleman : could you give this a high level pair of eyes to check (it with the explanation it makes sense to you) such that we can merge this for 5.0?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
```diff
@@ -73,27 +82,17 @@ protected void handleBindings() throws Exception {
...
-					expr = new CheckStatementPattern(stmt, queryInfo);
-					taskCreator = new CheckJoinTaskCreator(strategy, (CheckStatementPattern) expr);
-				}
-			} else if (expr instanceof FedXService) {
-				taskCreator = new FedXServiceJoinTaskCreator(strategy, (FedXService) expr);
-			} else {
-				throw new RuntimeException("Expr is of unexpected type: " + expr.getClass().getCanonicalName()
-						+ ". Please report this problem.");
+		if (submitFirstResultImmediately) {
+			// first item is always sent in a non-bound way
```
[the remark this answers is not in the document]
JervenBolleman: @aschwarte10 could you confirm if this comment is still correct because it jars with how I understand the the explanation. Otherwise it looks good to me.

JervenBolleman (changes requested):

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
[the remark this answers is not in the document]
aschwarte10: when inside this if branch (i.e. inside the `submitFirstResultImmediately`)  the comment is correct. It is essentially also the same code as before - only the creation of the task creator is extracted into a helper method for reuse.

So yes: inside the if branch (which in this implementation is deactivated, but supported to be enabled in sub-classes) the first item is always sent in a non bound way.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
```diff
@@ -157,6 +147,47 @@ public void handleClose() throws QueryEvaluationException {
...
 
+	protected TaskCreator determineTaskCreator(TupleExpr expr, BindingSet bs) {
+		final TaskCreator taskCreator;
+		if (expr instanceof StatementTupleExpr) {
+			StatementTupleExpr stmt = (StatementTupleExpr) expr;
+			if (stmt.hasFreeVarsFor(bs)) {
+				taskCreator = new BoundJoinTaskCreator(strategy, stmt);
+			} else {
+				expr = new CheckStatementPattern(stmt, queryInfo);
+				taskCreator = new CheckJoinTaskCreator(strategy, (CheckStatementPattern) expr);
```
hmottestad, line 158: Instead of using a variable, would it be possible to just return directly?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
```diff
@@ -106,27 +105,18 @@ protected void handleBindings() throws Exception {
...
-			}
+			// determine the bind join block size
+			nBindings = getNextBindJoinSize(nBindingsCfg, totalBindings);
 
 			bindings = new ArrayList<>(nBindings);
 
 			int count = 0;
 			while (!isClosed() && count < nBindings && leftIter.hasNext()) {
-				bindings.add(leftIter.next());
+				var bs = leftIter.next();
```
hmottestad, line 115: It's nice with var isn't it :)

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConfig.java
```diff
@@ -40,7 +40,7 @@ public class FedXConfig {
 
 	private int leftJoinWorkerThreads = 10;
 
-	private int boundJoinBlockSize = 15;
+	private int boundJoinBlockSize = 25;
```
hmottestad, line 43: Does this kinda become the opposite of parallelism? As in if this variable is infinite then it would only use a single thread? 

Would it make sense to increase this variable at runtime? Start it off at 1 and then double it for every join until it reaches some kind of configurable max?

hmottestad (approved):
LGTM. Added some comments and questions, but nothing critical.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
aschwarte10: yes, I also really got used to some other newer language features, e.g. multi-line strings for tests.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/FedXConfig.java
aschwarte10: > Does this kinda become the opposite of parallelism? As in if this variable is infinite then it would only use a single thread?

Exactly: the join operator blocks until a block is filled up prior to submitting it to the scheduler and executing it as sub-query.

So, this is really a tuning parameter of a federation. When FedX was developed some 10 years ago, in the benchmarks a value of 15 produced the best results. Nowadays it looks like higher values seem to produce better performance. In our scenarios we are currently looking for a value somwhere in the range 25 to 50.

> Would it make sense to increase this variable at runtime? Start it off at 1 and then double it for every join until it reaches some kind of configurable max?

Also this is the right observation. In fact this change is already doing a bit of preparation into this direction with moving the "getNextBindJoinSize" into a helper method. My idea is to do evaluations there first in specialized implementations. I think especially for large joins (with more than 1000 intermediate results) it pays off to increase the join size after the 1000 first results have been processed. I do not see this in 5.0 though, but a bit later once have more experience (or ideally objective benchmark results for some query mix)

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBoundJoin.java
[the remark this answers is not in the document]
aschwarte10: @JervenBolleman does this answer help? Do you want to update your "changes requested" from the review?

Issue #4505, opened 2023-04-12
Add GeoSPARQL vocabulary constants


abrokenjester:
### Discussed in https://github.com/eclipse/rdf4j/discussions/4504

<div type='discussions-op-text'>

<sup>Originally posted by **luisramosbosch** April 11, 2023</sup>
Dear All, 

I am trying to develop a minimum example in geosparql, following this documentation:

https://rdf4j.org/documentation/programming/geosparql/

I have the idea, that I should be able to build the example using code, but when I tried to do it, 
I found some method (object property, and class defition) I would expect were no present,
like geo:hasGeometry or geof:Point, which would be neccesary to build the model of the example.

model.add(eiffelTower, GEO.hasGeometry, coordinateset );
model.add(coordinateset, RDF.TYPE, GEOF.Point );

Could you tell me if building the model from code is possible?, or if I have to edit the model somewhere else, 
and use rdf4j only for querying?.

Best regards

Luis Ramos
</div>

hmottestad:
Fixed by: @linnaung

PR #4988, opened 2024-05-26
GH-4505: Add GeoSPARQL vocabulary constants

linnaung:
GitHub issue resolved: #4505 

The following item is addressed:

- add GeoSPARQL vocabulary constants

Also addressed: 
- update javadoc GeoSPARQL version to 1.1
- update copyright year to current year

----

### DIFF
--- a/core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/GEO.java
+++ b/core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/GEO.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
+ * Copyright (c) 2024 Eclipse RDF4J contributors.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Distribution License v1.0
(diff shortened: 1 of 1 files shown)


core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/GEO.java
```diff
@@ -11,25 +11,538 @@
...
+	 * @see <a href="http://www.opengis.net/ont/geosparql#wktLiteral">The geo:wktLiteral property</a>
+	 */
 	public static final IRI WKT_LITERAL;
 
 	public static final String DEFAULT_SRID = "http://www.opengis.net/def/crs/OGC/1.3/CRS84";
 
 	static {
+
+		// classes
+		FEATURE = Vocabularies.createIRI(NAMESPACE, "Feature");
```
hmottestad, line 482: Could you inline these? Makes it easier to see that things correspond to the iri.

core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/GEO.java
```diff
@@ -11,25 +11,538 @@
...
+		FEATURE_COLLECTION = Vocabularies.createIRI(NAMESPACE, "FeatureCollection");
```
hmottestad, line 483: I've also stopped using all caps for some of the newer vocabularies. I know it makes things a bit ugly when one vocabulary uses all caps and another doesn't, but it's just so much easier to understand what it what when using the exact same name for the variable in Java and in RDF.

@JervenBolleman Maybe you can chime in with your opinion?

JervenBolleman: I agree with trying to use the same "spelling" as much as possible for constants. I had started a script to make it consistent for all (Deprecating the UPPER_SNAKE_CASE) but I need to look for this and see if I can finish it.

core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/GEO.java
linnaung: done.

core/model-vocabulary/src/main/java/org/eclipse/rdf4j/model/vocabulary/GEO.java
linnaung: done. Please note that I leave the following constant properties as it is:
- `geo:asWKT`
- `geo:wktLiteral`

hmottestad:
Technically this should be merged into develop, since it's kinda a new feature, but since it's such a minor thing and since it would probably be nice to have in the 4.x release I'll merge it into main and it'll be part of the next bug fix release.

Issue #4990, opened 2024-05-29
FedX federation fails with query evaluation exception when evaluating zero length property path


aschwarte10:
### Current Behavior

We found that the FedX query engine currently fails with the following exception when the query contains a zero length path (e.g. `?myClass rdfs:subClassOf* ?parentClass`)

The exact exception message is: 

> Federation Strategy does not support org.openrdf.query.algebra.evaluation.TripleSource#getStatements. If you encounter this exception, please report it.

This is because the federation engine falls back to the `DefaultEvaluationStrategy` implementation for property paths, which in case of zero-path is not tailored to the use in federation.

Note that path with length minimum 1 (i.e. the + operator) are working fine.

### Expected Behavior

The federation engine should support correct semantics of zero-length path

### Steps To Reproduce

Can be reproduced with a query like (assuming data being present)

```
SELECT * WHERE {
    ?subClass (rdfs:subClassOf*) ?myClass
```

### Version

4.3.x

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

We would like to see this addressed in 4.3.x and eventually in in the 5.0 release

conbon:
@aschwarte10 We have also encountered this issue recently on the 4.3.9 branch and carried out some initial debugging. 
I'll provide what we have as additional input.

On a single fedx repository, the following was inserted:
```
INSERT DATA {
  GRAPH ex:graph {
    :a foaf:knows :b .
    :a a foaf:Thing .
    :b foaf:friend :c .
    :b a foaf:Thing .
    :a foaf:knows :c .
    :d foaf:friend :e .
    :d a foaf:Thing .
    :e foaf:knows :f .
    :e a foaf:Thing .
    :f foaf:knows :e .
    :f a foaf:Thing .
    :f foaf:name "test" .
    :a foaf:homepage :h .
  }
}
```

**Queries**
All of these are wrapped in SELECT * { } .

| Query | Result | Comments |
|-------| -------| ------------|
| `?s (foaf:knows\|foaf:friend)+ ?o .`  `?s foaf:knows+ ?o .` |  🟢  | ArbitraryLengthPath -> NUnion -> StatementSourcePattern |
[...]

PR #4991, opened 2024-05-29
GH-4990: support for zero-length path expression in FedX

aschwarte10:
GitHub issue resolved: #4990 

This change adds support for zero-length path expressions in the FedX
engine and also fixes source selection for other path expressions.

Note that the implementation reuses the implementation from
PathIteration and ZeroLengthPathIteration. Unfortunately the existing
implementations cannot be reused without larger refactoring right now.

Functionality is covered with unit tests

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
@@ -333,7 +333,7 @@ public abstract class FederationEvalStrategy extends StrictEvaluationStrategy {
 			identifiedMembers = new HashSet<>(members);
 		} else {
 			StatementPattern checkStmt = new StatementPattern(stmt.getScope(), new Var("subject"),
-					stmt.getPredicateVar().clone(), new Var("object"), stmt.getContextVar());
+					clone(stmt.getPredicateVar()), new Var("object"), clone(stmt.getContextVar()));
 			@SuppressWarnings("unused") // only used as artificial parent
 			HolderNode holderParent = new HolderNode(checkStmt);
 
@@ -362,6 +362,13 @@ public abstract class FederationEvalStrategy extends StrictEvaluationStrategy {
 		}
 
 		return identifiedMembers;
+	}
+
+	private Var clone(Var var) {
+		if (var == null) {
+			return null;
+		}
+		return var.clone();
 	}
 
 	protected void optimizeJoinOrder(TupleExpr query, QueryInfo queryInfo, GenericInfoOptimizer info) {
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
@@ -26,24 +26,32 @@ import org.eclipse.rdf4j.federated.FederationContext;
 import org.eclipse.rdf4j.federated.algebra.CheckStatementPattern;
 import org.eclipse.rdf4j.federated.algebra.ConjunctiveFilterExpr;
 import org.eclipse.rdf4j.federated.algebra.EmptyResult;
+import org.eclipse.rdf4j.federated.algebra.EmptyStatementPattern;
 import org.eclipse.rdf4j.federated.algebra.ExclusiveGroup;
+import org.eclipse.rdf4j.federated.algebra.ExclusiveStatement;
 import org.eclipse.rdf4j.federated.algebra.ExclusiveTupleExpr;
 import org.eclipse.rdf4j.federated.algebra.ExclusiveTupleExprRenderer;
+import org.eclipse.rdf4j.federated.algebra.FedXArbitraryLengthPath;
 import org.eclipse.rdf4j.federated.algebra.FedXLeftJoin;
 import org.eclipse.rdf4j.federated.algebra.FedXService;
+import org.eclipse.rdf4j.federated.algebra.FedXZeroLengthPath;
 import org.eclipse.rdf4j.federated.algebra.FederatedDescribeOperator;
 import org.eclipse.rdf4j.federated.algebra.FilterExpr;
 import org.eclipse.rdf4j.federated.algebra.FilterValueExpr;
+import org.eclipse.rdf4j.federated.algebra.HolderNode;
 import org.eclipse.rdf4j.federated.algebra.NJoin;
 import org.eclipse.rdf4j.federated.algebra.NUnion;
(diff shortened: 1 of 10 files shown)


aschwarte10:
Note: the initial commit provides the failing unit tests. I'll provide the implementation / fix in the course of the next days.

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/FedXZeroLengthPathIteration.java
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * An iteration to evaluated {@link FedXZeroLengthPath}
+ *
+ * @see ZeroLengthPathIteration
+ */
+public class FedXZeroLengthPathIteration extends LookAheadIteration<BindingSet, QueryEvaluationException> {
+
+	/*
+	 * IMPL NOTE:
```
aschwarte10, line 51: please see this comment: we are using the same implementation as in ZeroLengthPathIteration, but adjust it for the federation. I will add another comment at the relevant place, the rest of the logic is 1:1 copied

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/FedXZeroLengthPathIteration.java
```diff
@@ -0,0 +1,220 @@
...
+	 */
+
+	private static final Literal OBJECT = SimpleValueFactory.getInstance().createLiteral("object");
+
+	private static final Literal SUBJECT = SimpleValueFactory.getInstance().createLiteral("subject");
+
+	// Note: in contrast to the original zero length path iteration we use "_" instead of "-"
+	// as we need variable names valid in SELECT queries
+
+	private static final String ANON_SUBJECT_VAR = "zero_length_internal_start";
```
aschwarte10, line 68: here I had to use "_" instead of "-" as in the federation I generate SELECT queries to be sent to the members. "-" is not a valid character in variable names

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/FedXZeroLengthPathIteration.java
```diff
@@ -0,0 +1,220 @@
...
+
+		StatementPattern subjects;
+		if (contextVar != null) {
+			subjects = new StatementPattern(Scope.NAMED_CONTEXTS, startVar, predicate, endVar, contextVar.clone());
+		} else {
+			subjects = new StatementPattern(startVar, predicate, endVar);
+		}
+
+		// specialization for federation: we need to attach statement sources
+		// to the precompiled expr
```
aschwarte10, line 124: this is the only specialization: for the federation we need to send the statement pattern to the identified relevant members. Hence we use a fedx structure which is aware of the sources

Remainder of the code is 1:1

tools/federation/src/test/java/org/eclipse/rdf4j/federated/PropertyPathTests.java
```diff
@@ -105,20 +110,12 @@ public void testPropertyPath_ExclusiveGroup() throws Exception {
 	}
 
 	@Test
-	public void testPropertyPath_ExclusivePath() throws Exception {
+	public void testPropertyPath_Combination() throws Exception {
```
aschwarte10, line 113: this test from the query plan was actually not correct in the general case. Note that I no longer compare the query plan, but the results

aschwarte10 (commented):
Please see inline a few comments that explain and point to the relevant bits of code in this PR.

Note that most of this PR code is unfortunately copy & paste of code - which without heavier refactoring cannot be simply reused. We may invest into this refactoring when integrating the change to 5.0 - for a bug fix release I prefer to keep as is

If anyone needs help in the review (e.g. pair review with explanation) please give me a ping, I should be also available short-term tomorrow

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
```diff
@@ -271,11 +280,88 @@ protected Set<Endpoint> performSourceSelection(List<Endpoint> members, SourceSel
...
+		// for now the strategy is to do source selection on a statement pattern containing the predicate
+		// of the original path expression.
+		// Note: for zero-length path we add all members (as there does not have to exist a path with the
+		// predicate)
+		Set<Endpoint> identifiedMembers;
+		if (pathExpr.getMinLength() == 0) {
+			identifiedMembers = new HashSet<>(members);
+		} else {
+			StatementPattern checkStmt = new StatementPattern(stmt.getScope(), new Var("subject"),
+					stmt.getPredicateVar().clone(), new Var("object"), stmt.getContextVar());
```
hmottestad, line 336: Should the context var be cloned too?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/iterator/FedXZeroLengthPathIteration.java
hmottestad: We should probably change that in the original implementation too. I'll take a look.

hmottestad:
This look good to me @aschwarte10 with the only possible issue I found being the cloning of that context var. That's probably not coming up as an issue if the context var is null in all the tests.

aschwarte10:
> This look good to me @aschwarte10 with the only possible issue I found being the cloning of that context var. That's probably not coming up as an issue if the context var is null in all the tests.

Good catch @hmottestad . Fixed in the second commit (incl. unit test coverage). Can you take another look?

aschwarte10:
@hmottestad a second question: what is your feeling about refactoring the original `PathIteration` and `ZeroLengthPathIteration` a bit to make most of the logics reusable? I would not do this in 4.3.x, but maybe in in 5.0? What would need to be done is to move some initialization out of the constructor and to introduce one protected method for the construction of one item. This would allow to do the small specialization in a federated version of the implementation

hmottestad:
Thanks @aschwarte10 I'll merge this when all the tests are done and then start on the bug fix release.

hmottestad:
> @hmottestad a second question: what is your feeling about refactoring the original `PathIteration` and `ZeroLengthPathIteration` a bit to make most of the logics reusable? I would not do this in 4.3.x, but maybe in in 5.0? What would need to be done is to move some initialization out of the constructor and to introduce one protected method for the construction of one item. This would allow to do the small specialization in a federated version of the implementation

That would probably be a good idea. We have some issues with the current path iteration code not spilling to disk if it runs out of memory. @JervenBolleman started looking into it I think. So if we can have a single implementation instead of two, I think it would be easier to implement spilling to disk.

Issue #4997, opened 2024-05-31
VALUES for SparqlBuilder


fkleedorfer:
### Problem description

The sparqlbuilder module does not have a way of providing VALUES.

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

I already have that. Will open a PR that contributes it.

PR #5002, opened 2024-06-01
GH-4997: Enable SparqlBuilder to create VALUES clauses

fkleedorfer:
GitHub issue resolved: #4997 

Adds a builder for VALUES clauses and integrates them with GraphPattern and Select

----

### DIFF
--- a/core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Values.java
+++ b/core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Values.java
@@ -1,0 +1,217 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *******************************************************************************/
+package org.eclipse.rdf4j.sparqlbuilder.constraint;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.sparqlbuilder.core.Variable;
+import org.eclipse.rdf4j.sparqlbuilder.graphpattern.GraphPattern;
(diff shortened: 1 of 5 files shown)


fkleedorfer:
Happy to rebase that onto develop and change the target branch to develop, if needed.

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/constraint/Values.java
```diff
@@ -0,0 +1,217 @@
+/* *****************************************************************************
+ * Copyright (c) 2015 Eclipse RDF4J contributors, Aduna, and others.
```
hmottestad, line 2: Could you update this to the current copyright template?

You can find it here: https://github.com/eclipse/rdf4j/blob/main/CONTRIBUTING.md

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/core/query/Query.java
```diff
@@ -246,7 +256,7 @@ public String getQueryString() {
 		if (offset >= 0) {
 			query.append(OFFSET + " ").append(offset).append("\n");
 		}
-
+		SparqlBuilderUtils.appendAndNewlineIfPresent(values, query);
```
hmottestad, line 259: Doesn't the VALUES clause usually go at the beginning of the WHERE clause?

hmottestad:
> Happy to rebase that onto develop and change the target branch to develop, if needed.

Yeah, this needs to target the develop branch.

Btw. Have you considered supporting VALUES inside UNION clauses? It's a common scenario for getting better performance since the scoping of the UNION means that any VALUES clauses outside the UNION clause will be evaluated after the UNION clauses. Duplicating the VALUES clause inside each branch of the UNION clause fixes the scoping issue so the VALUES clause can be used when executing the joins inside each UNION clause.

fkleedorfer:
You can add it to any GraphPattern, so you can also add it to a Union pattern

fkleedorfer:
... But we might be able to make that more explicit if you like

core/sparqlbuilder/src/main/java/org/eclipse/rdf4j/sparqlbuilder/core/query/Query.java
fkleedorfer: No.  https://www.w3.org/TR/sparql11-query/#sparqlGrammar says:

``` 
[2] | Query  ::=  Prologue
                  ( SelectQuery | ConstructQuery | DescribeQuery | AskQuery )
                  ValuesClause
```

Issue #5000, opened 2024-05-31
rdf4j-spring: Suppress logging of exceptions if they are thrown on.


fkleedorfer:
### Problem description

RDF4JTemplate and friends produce exceptions such as the UnexpectedResultSetSizeException, which are expected to be caught downstream, at least in some settings. The fact that these are logged with level WARN at the framework level bloats our logs unnecessarily.

### Preferred solution

Log the exception message on DEBUG level, definitely not the stacktrace.

### Are you interested in contributing a solution yourself?

Yes

PR #5006, opened 2024-06-01
GH-5000: Tone down WARN log output for frequent, non-critical events

fkleedorfer:
GitHub issue resolved: #5000

Reduces loglevel for two very frequent, not-so-serious situations

----

 - [~] I've added tests for the changes I made

### DIFF
--- a/spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/support/operation/TupleQueryResultConverter.java
+++ b/spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/support/operation/TupleQueryResultConverter.java
@@ -28,6 +28,7 @@ import java.util.stream.Collector;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.eclipse.rdf4j.common.exception.RDF4JException;
 import org.eclipse.rdf4j.query.BindingSet;
 import org.eclipse.rdf4j.query.TupleQueryResult;
 import org.eclipse.rdf4j.spring.dao.support.BindingSetMapper;
@@ -56,8 +57,10 @@ public class TupleQueryResultConverter {
 		try {
 			consumer.accept(tupleQueryResult);
 		} catch (Exception e) {
-			logger.debug("Caught execption while processing TupleQueryResult", e);
-			throw mapException("Error processing TupleQueryResult", e);
+			logger.debug("Caught execption while processing TupleQueryResult: {}", e.getMessage());
+			RDF4JException mapped = mapException("Error processing TupleQueryResult", e);
+			logger.debug("Re-throwing as {} ", mapped.getClass().getSimpleName());
+			throw mapped;
 		} finally {
 			tupleQueryResult.close();
 			tupleQueryResult = null;
@@ -71,8 +74,10 @@ public class TupleQueryResultConverter {
 		try {
 			return function.apply(tupleQueryResult);
 		} catch (Exception e) {
-			logger.warn("Caught execption while processing TupleQueryResult", e);
-			throw mapException("Error processing TupleQueryResult", e);
+			logger.debug("Caught execption while processing TupleQueryResult: {}", e.getMessage());
+			RDF4JException mapped = mapException("Error processing TupleQueryResult", e);
+			logger.debug("Re-throwing as {} ", mapped.getClass().getSimpleName());
+			throw mapped;
 		} finally {
 			tupleQueryResult.close();
 			tupleQueryResult = null;


spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/tx/TransactionalRepositoryConnectionFactory.java
```diff
@@ -95,7 +95,7 @@ public void closeConnection() {
 		} finally {
 			try {
 				if (con != null && con.isActive()) {
-					logger.warn(
```
hmottestad, line 98: Can you explain why this should be debug instead of warn? Seems like a fairly critical issue with the users code.

fkleedorfer: On second thought, I agree with you. 

I wanted to reduce the loglevel because I've seen this a lot in my logs without it being of any use, as there were other log messages generated in the same transaction by the problem eventually also leading to this. However, It's just one line, and it is, as you say, pretty serious (the explicit commit or rollback must have failed somehow), so let's leave it in. It might actually be worth digging into the cases in which the message shows up.

fkleedorfer: reverted that change.

spring-components/rdf4j-spring/src/main/java/org/eclipse/rdf4j/spring/dao/support/operation/TupleQueryResultConverter.java
```diff
@@ -71,8 +74,10 @@ public <T> T applyToResult(Function<TupleQueryResult, T> function) {
 		try {
 			return function.apply(tupleQueryResult);
 		} catch (Exception e) {
-			logger.warn("Caught execption while processing TupleQueryResult", e);
-			throw mapException("Error processing TupleQueryResult", e);
+			logger.debug("Caught execption while processing TupleQueryResult: {}", e.getMessage());
```
hmottestad, line 77: Do these exceptions happen so often that we can't log the exception as WARN or at least DEBUG?

fkleedorfer: The result processing pipeline that is wrapped here will throw recoverable, client-controlled exceptions like the IncorrectResultSetSizeException. Ie eg the client says, I expect one result here and I expect an exception to be thrown if that is not the case.

These things can happen a lot, especially in unit tests, where you want edge cases covered as well. 

Also, the exception is thrown on, so it will end up in the logs one way or the other, unless it is actively swallowed down the line, which is for the client to decide.

I have been using this code on a daily basis for years and have never had any advantage whatsoever from any of the many thousand stacktraces I have seen logged here. They just clog my logs. I noticed very quickly I had made a mistake putting that in back in the day.

hmottestad: Are the exceptions used as some type of signaling? That's usually fairly expensive unless you make sure that your exceptions don't contain a stack trace.

fkleedorfer: Not really. Consider this example:
```
public Organization getOrganizationById(IRI id) {
    return getNamedTupleQuery(Keys.QUERY_ORGANIZATIONS.name())
            .withBinding(Organization._id, id)
            .evaluateAndConvert()
            .toSingleton(resultMapper);
}
```
In this case, the client decided to use `toSingleton(resultMapper)` because they expect exactly one result. If it is not one but 0 or more than 1, an exception is thrown. In such a case, nothing went wrong on the RDF4J level, there just isn't a result. If something goes wrong on the repo/connection/query execution level, that creates an exception earlier in the chain (during the exection of `evaluateAndConvert()`.

However, if the client used `. toSingletonOptional(resultMapper)`, the exception would only be thrown if there are more than one results. 

If the client was ok with getting any number of results, they would use other methods, such as `toSet(resultMapper)` or `toList(resultMapper`. 

So, these exceptions are thrown by the framework because it was asked by the client to do so in certain situations. It is not its job to log them. (Maybe even logging it on with level debug is too much. ) We might as well ask for a log output to be put in `Objects.requireNonNull`.

Issue #5030, opened 2024-06-14
FedX does not correctly reduce federation members, when the query contains a FROM clause


aschwarte10:
### Current Behavior

FedX supports interpret the Dataset being defined on the query: if it is of type `FedXDataset`, the  federation is reduced to the members defined by `FedXDataset#getEndpoints()`.

This is a nice means to evaluate a query on a subset of the federation and is generally working fine.

When the query contains a FROM clause, however, the externally passed `Dataset` is wrapped into a `FallbackDataset` (which internally keeps the FedXDataset as primary one.

### Expected Behavior

The federation engine correctly respects an externally passed `FedXDataset`to reduce the federation

### Steps To Reproduce

See test case in the provided PR

### Version

4.8.12

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

Would like to see this merged still as part of 5.0

PR #5031, opened 2024-06-14
GH-5030: fix FedXDataset support for queries with FROM clauses

aschwarte10:
GitHub issue resolved: #5030 

FedX supports interpret the Dataset being defined on the query: if it is of type FedXDataset, the federation is reduced to the members defined by FedXDataset#getEndpoints().

This is a nice means to evaluate a query on a subset of the federation and is generally working fine.

When the query contains a FROM clause, however, the externally passed Dataset is wrapped into a FallbackDataset (which internally keeps the FedXDataset as primary one.

This change now inspects the FallbackDataset and unwraps it to reduce the members. Note that for execution still the "FallbackDataset" is used.

----

### DIFF
--- a/core/query/src/main/java/org/eclipse/rdf4j/query/impl/FallbackDataset.java
+++ b/core/query/src/main/java/org/eclipse/rdf4j/query/impl/FallbackDataset.java
@@ -113,4 +113,12 @@ public class FallbackDataset implements Dataset, Serializable {
 		}
 	}
 
+	public Dataset getPrimary() {
+		return primary;
+	}
+
+	public Dataset getFallback() {
+		return fallback;
+	}
+
 }
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
@@ -121,6 +121,7 @@ import org.eclipse.rdf4j.query.algebra.evaluation.util.QueryEvaluationUtil;
 import org.eclipse.rdf4j.query.algebra.helpers.TupleExprs;
 import org.eclipse.rdf4j.query.algebra.helpers.collectors.VarNameCollector;
 import org.eclipse.rdf4j.query.impl.EmptyBindingSet;
+import org.eclipse.rdf4j.query.impl.FallbackDataset;
 import org.eclipse.rdf4j.repository.RepositoryException;
 import org.eclipse.rdf4j.repository.sparql.federation.CollectionIteration;
 import org.eclipse.rdf4j.repository.sparql.federation.RepositoryFederatedService;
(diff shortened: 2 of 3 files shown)


core/query/src/main/java/org/eclipse/rdf4j/query/impl/FallbackDataset.java
```diff
@@ -113,4 +113,12 @@ private void appendURI(StringBuilder sb, IRI uri) {
 		}
 	}
 
+	public Dataset getPrimary() {
```
aschwarte10, line 116: do you think it is OK to make the primary dataset accessible for use in FedX?

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
```diff
@@ -179,19 +180,8 @@ public TupleExpr optimize(TupleExpr expr, EvaluationStatistics evaluationStatist
...
-		if (dataset instanceof FedXDataset) {
-			// run the query against a selected set of endpoints
-			FedXDataset ds = (FedXDataset) dataset;
-			members = federationContext.getEndpointManager().getEndpoints(ds.getEndpoints());
-		} else {
-			// evaluate against entire federation
-			FedX fed = federationContext.getFederation();
-			members = fed.getMembers();
-		}
+		List<Endpoint> members = getMembersFromContext(stats);
```
aschwarte10, line 184: Note: the logic is now moved into a helper method. Potentially allowing derived extensions to do additional pruning of members in the query context

tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
```diff
@@ -270,6 +260,34 @@ public TupleExpr optimize(TupleExpr expr, EvaluationStatistics evaluationStatist
...
+	 *
+	 * @param evaluationStatisticss to keep the current query context
+	 * @return
+	 */
+	protected List<Endpoint> getMembersFromContext(FederationEvaluationStatistics evaluationStatisticss) {
+		Dataset queryDataset = evaluationStatisticss.getDataset();
+
+		// if the query uses a FROM clause, the external dataset
+		// is wrapped as primary dataset in FallbackDataset
+		if (queryDataset instanceof FallbackDataset) {
```
aschwarte10, line 276: this is actually the only change in logic: we unwrap the primary dataset from the FallbackDataset

aschwarte10:
@hmottestad can I ask for a review to have this bug fix still in 5.0? It is relevant for the use of FedX in our product, where we have means to reduce the federation for a specific query.

Also: can you comment on the failed "dash-license" check? Is this something we need to look after? In my change I have not done anything with dependencies.

Issue #5033, opened 2024-06-18
FedX: Missing limit pushing of ASK queries with single statement pattern causes poor query performance 


aschwarte10:
### Current Behavior

Currently for ASK queries having a single statement pattern the limit 1 is not pushed into the sub query, while it is implement for select queries.

The effect is poor behavior.

Example:

```
ASK { ?person a foaf:Person }
```

Currently the implementation of the federation will fetch all persons available in the federation members (and locally checks that there is at least one binding), though we are just interested in the existence

Note: for SELECT queries FedX already has an optimizer that pushes the limit for such trivical cases into the sub-query.

### Expected Behavior

The limit is pushed and the same optimizations as for SELECT queries with a single statements is applied.

### Steps To Reproduce

To seee poor performance, run an ASK query on a large database with millions of instances.

The difference in the query plan is the "Upper Limit: N" attached to the statement pattern

```
QueryRoot
   Slice (limit=1)
      StatementSourcePattern
         Var (name=person)
         Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
         Var (name=_const_e1df31e0_uri, value=http://xmlns.com/foaf/0.1/Person, anonymous)
         StatementSource (id=sparql_localhost:18080_repositories_endpoint1, type=REMOTE)
         StatementSource (id=sparql_localhost:18080_repositories_endpoint2, type=REMOTE)
         Upper Limit: 1
```

### Version

4.3.12

### Are you interested in contributing a solution yourself?

Yes

PR #5034, opened 2024-06-18
GH-5033: fix pushing of limits for simple ASK queries in FedX

aschwarte10:
GitHub issue resolved: #5033 

This change makes sure to push limits for simple ASK queries with a single statement patterns into the query.

The optimization is the same as applied for simple SELECT queries with a LIMIT.

Rational: if the limit is not pushed, the federation engine will first fetch all data for the statement pattern and only then locally check if there is at least one, i.e it will cause performance issues and memory pressure when there are many triples matching the statement pattern (for instance millions of persons).

Query plan of a simple ASK query after this fix (note the upper limit):

```
QueryRoot
   Slice (limit=1)
      StatementSourcePattern
         Var (name=person)
         Var (name=_const_f5e5585a_uri, value=http://www.w3.org/1999/02/22-rdf-syntax-ns#type, anonymous)
         Var (name=_const_e1df31e0_uri, value=http://xmlns.com/foaf/0.1/Person, anonymous)
         StatementSource (id=sparql_localhost:18080_repositories_endpoint1, type=REMOTE)
         StatementSource (id=sparql_localhost:18080_repositories_endpoint2, type=REMOTE)
         Upper Limit: 1
```

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/optimizer/LimitOptimizer.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/optimizer/LimitOptimizer.java
@@ -65,6 +65,18 @@ public class LimitOptimizer extends AbstractSimpleQueryModelVisitor<Optimization
 			applicableLimitInScope = node.getLimit();
 		}
 		super.meet(node);
+
+		TupleExpr expr = node.getArg();
+		// if the top most element is a statement (e.g. for an ASK query with single statement pattern),
+		// i.e. no join, union or
+		// any other complex pattern, we can push the limit
+		// => this case typically represents a query with a single BGP
+		if (expr instanceof FedXStatementPattern) {
+			if (applicableLimitInScope > 0) {
+				pushLimit((FedXStatementPattern) expr, applicableLimitInScope);
+			}
+		}
+
 		applicableLimitInScope = -1;
 
 	}
--- a/tools/federation/src/test/java/org/eclipse/rdf4j/federated/LimitTests.java
+++ b/tools/federation/src/test/java/org/eclipse/rdf4j/federated/LimitTests.java
@@ -1,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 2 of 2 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/optimizer/LimitOptimizer.java
```diff
@@ -65,6 +65,18 @@ public void meet(Slice node) throws OptimizationException {
 			applicableLimitInScope = node.getLimit();
 		}
 		super.meet(node);
+
+		TupleExpr expr = node.getArg();
+		// if the top most element is a statement (e.g. for an ASK query with single statement pattern),
+		// i.e. no join, union or
+		// any other complex pattern, we can push the limit
+		// => this case typically represents a query with a single BGP
+		if (expr instanceof FedXStatementPattern) {
```
aschwarte10, line 74: this is applying the same logic as a few lines below.

The difference is that in case of a SELECT query, we find the single statement pattern below a `Projection` node and not directly as argument of the `Slice`

tools/federation/src/test/java/org/eclipse/rdf4j/federated/LimitTests.java
```diff
@@ -0,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Eclipse RDF4J contributors.
```
hmottestad, line 2: Could you update the year to 2024?

aschwarte10: ah, again copy & paste from a parallel file. Will do immediately. Good spot.

Have squashed it into the existing commit

hmottestad:
```
Error:    ServiceTests.test5:150->FedXBaseTest.execute:101 » QueryEvaluation java.lang.NullPointerException
```

aschwarte10:
> ```
> Error:    ServiceTests.test5:150->FedXBaseTest.execute:101 » QueryEvaluation java.lang.NullPointerException
> ```

Hmm, I had a closer look at this test. It cannot be related to this change (=> the change in this code is only active when there is a LIMIT in the query (i.e. only when the query plan contains a SLICE operator the LimitOptimizer is run. This is not the case for SeviceTest#test5

I have no idea why it is failing here and personally have not seen before.

Note: it was also not failing when pushing the first commit of this PR (i.e. before the "year" change). Local test-reruns also do not indicate problems.

I have now retriggered the verify build. Let's see what comes out of that

@hmottestad hope that we can still make this one for 5.0

Issue #5048, opened 2024-06-23
Upgrade spring to 5.3.37


barthanssens:
### Current Behavior

Congrats on releasing 5.0 ;-)

I've noticed, when releasing the docker workbench image, there are a few vulnerabilities in spring framework itself (which may or may not affect RDF4J workbench)

### Expected Behavior

Upgrading to the latest (patch) release of spring should fix most (but probably not all) reported CVEs for spring dependencies

### Version

5.0.0

### Are you interested in contributing a solution yourself?

Yes

PR #5049, opened 2024-06-23
GH-5048: updated spring dependency to fix CVEs

barthanssens:
GitHub issue resolved: #5048  

Just a bump in (patch release) version number for Spring Framework

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -371,7 +371,7 @@
 		<lucene.version>8.9.0</lucene.version>
 		<solr.version>8.9.0</solr.version>
 		<elasticsearch.version>7.15.2</elasticsearch.version>
-		<spring.version>5.3.30</spring.version>
+		<spring.version>5.3.37</spring.version>
 		<guava.version>32.1.3-jre</guava.version>
 		<jmhVersion>1.37</jmhVersion>
 		<servlet.version>3.1.0</servlet.version>


barthanssens:
urlrewrite CQ request: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/15467

pom.xml
```diff
@@ -371,7 +371,7 @@
 		<lucene.version>8.9.0</lucene.version>
 		<solr.version>8.9.0</solr.version>
 		<elasticsearch.version>7.15.2</elasticsearch.version>
-		<spring.version>5.3.30</spring.version>
```
hmottestad, line 374: The spring version should probably match the spring version used in spring-boot. You can update spring-boot to 2.7.18 which uses spring 5.3.31. I think that that version also fixes the CVE.

Spring-boot version should probably be specified in the root pom. At the moment it's defined here: https://github.com/eclipse-rdf4j/rdf4j/blob/52b4106356a10584d63fb710d98654908c661e91/spring-components/pom.xml#L16

barthanssens: good idea

Though probably not the only place, since the rdf4j-server.war also includes spring 5.x jars in the WEB-INF/lib, but not springboot (do we actually used Spring there ?)

hmottestad: Spring is used in the server. Spring-boot is not used in either the server or workbench. But spring-boot is used in its own sub-module where there are some spring-boot related things and also an example app. 

Most people I know only use spring-boot now a days. Some might use a different framework, but I've not heard of anyone that has started a new project with just spring and not spring boot. So it's a good thing to have.

hmottestad: I didn't know that we have our own jars in the web-inf for spring. That doesn't sound like a good idea.

Issue #5060, opened 2024-06-28
Upgrade zookeeper to 3.7.x 


barthanssens:
### Current Behavior

I've noticed, when releasing the docker workbench image, there are a few vulnerabilities in zookeeper (which may or may not affect RDF4J workbench), which is used by solr

- CVE-2023-44981

### Expected Behavior

Upgrading to a newer release of zookeeper fix the reported CVE for zookeeper dependency

### Version

5.0.0

### Are you interested in contributing a solution yourself?

Yes

barthanssens:
Was used by Solr Sail (which was removed in RDF4J 6.0)

PR #5062, opened 2024-06-28
GH-5060: use more recent version of zookeeper to fix CVE

barthanssens:
GitHub issue resolved: # 5060

- use a newer version of zookeeper for fix CVE

----

### DIFF
--- a/core/sail/solr/pom.xml
+++ b/core/sail/solr/pom.xml
@@ -12,8 +12,22 @@
 	<properties>
 		<!-- FIXME: Support for embedded Solr server require non-provided Java EE dependencies -->
 		<enforce-javaee-provided.fail>false</enforce-javaee-provided.fail>
+		<!-- use at least 3.7 to fix CVE -->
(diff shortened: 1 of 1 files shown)


hmottestad:
What version of zookeeper are we using today? And where is zoo-keeper being used do you know?

barthanssens:
> What version of zookeeper are we using today? And where is zoo-keeper being used do you know?

Via solr-sorlj version 3.6.2 is used (or at least, zookeeper is distributed in the 5.0 SDK), not sure if it is actually used (i.e. started) when using the Solr Index Sail ...

hmottestad:
Maven is supposed to choose the highest version number when there is a conflict, so hopefully the next time we update solr then it will choose the zookeeper version from solr if it's higher. 

I don't really trust that though. So maybe you can add a comment to the Pom to explain that we are only specifying the version here because solr version x.x.x has a transitive dependency for zookeeper that is has a vulnerability?

hmottestad (approved):

Issue #5055, opened 2024-06-26
LMDB throwing java.lang.NoClassDefFoundError: org/mapdb/DBException on workbench: Explore >> Types invocation


nguyenm100:
### Current Behavior

when you select Explore>>Types from workbench while a LMDB repo is open, an exception is thrown.

Type Exception Report

Message Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: org/mapdb/DBException

Description The server encountered an unexpected condition that prevented it from fulfilling the request.

Exception

org.eclipse.rdf4j.repository.http.HTTPQueryEvaluationException: Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: org/mapdb/DBException
	org.eclipse.rdf4j.repository.http.HTTPTupleQuery.evaluate(HTTPTupleQuery.java:55)
	org.eclipse.rdf4j.workbench.commands.TypesServlet.service(TypesServlet.java:32)
	org.eclipse.rdf4j.workbench.base.TupleServlet.service(TupleServlet.java:55)
	org.eclipse.rdf4j.workbench.base.TupleServlet.service(TupleServlet.java:48)
	org.eclipse.rdf4j.workbench.base.TransformationServlet.service(TransformationServlet.java:100)
	org.eclipse.rdf4j.workbench.base.AbstractServlet.service(AbstractServlet.java:129)
	org.eclipse.rdf4j.workbench.proxy.ProxyRepositoryServlet.service(ProxyRepositoryServlet.java:100)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchServlet.service(WorkbenchServlet.java:214)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchServlet.handleRequest(WorkbenchServlet.java:136)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchServlet.service(WorkbenchServlet.java:111)
	org.eclipse.rdf4j.workbench.proxy.WorkbenchGateway.service(WorkbenchGateway.java:119)
	org.eclipse.rdf4j.workbench.base.AbstractServlet.service(AbstractServlet.java:129)
	org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
	org.apache.catalina.filters.SetCharacterEncodingFilter.doFilter(SetCharacterEncodingFilter.java:115)
	org.eclipse.rdf4j.workbench.proxy.CacheFilter.doFilter(CacheFilter.java:64)
[...]

PR #5067, opened 2024-06-29
GH-5055 make mapdb3 default for the FedX engine as well

JervenBolleman:
GitHub issue resolved: #5055 

Also use MapDB3 for FedX

Hand tested that this resolves the reported problem

### DIFF
--- a/tools/federation/pom.xml
+++ b/tools/federation/pom.xml
@@ -92,7 +92,7 @@
 		</dependency>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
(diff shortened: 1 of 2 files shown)


hmottestad (approved):
LGTM! Thanks for fixing this. 

Any chance you'd like to create a test so that we can increase our test coverage of the workbench?

https://github.com/eclipse-rdf4j/rdf4j/blob/main/e2e/tests/workbench.spec.js

We use Playwright: https://playwright.dev

It's OK if you don't have the time.

JervenBolleman:
I will try regarding the test case. I also want to mark the mapdb1 as
deprecated.

Issue #5064, opened 2024-06-28
Docker: use jetty and/or newer verssion of Tomcat


barthanssens:
### Problem description

Tomcat 8.5 is EOL 31 March 2024 (https://tomcat.apache.org/tomcat-85-eol.html), so we either need to move to a more recent version or switch to another servlet engine for the server/workbench Docker image

### Preferred solution

Use Jetty 9.4, since it's already being used by other dependencies (bonus: Jetty is an Eclipse project) and uses the same Servlet API 3.1 as Tomcat. Probably supported until the end of 2025 (https://github.com/jetty/jetty.project/issues/7958)

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

Eventually, we'll need to move to a newer version of the servlet API

aschwarte10:
> Preferred solution
> 
Use Jetty 9.4, since it's already being used by other dependencies (bonus: Jetty is an Eclipse project) and uses the same Servlet API 3.1 as Tomcat. Probably supported until the end of 2025 (https://github.com/jetty/jetty.project/issues/7958

@barthanssens can you explain why you are not considering to go to Jetty 11? Is it because of the Jakarta switch that is technically blocking it?

barthanssens:
Well, Jetty 11 (or even 12) should be the next step.

Due to various other dependencies, we already have jetty 9.4 in the list of dependencies, so this should not take much effort and (if all goes well) no code changes, just a different Docker makefile that I'm preparing in a branch.

For Jetty 11, we may need to look into the workbench / rdf4j-server (IIRC they both use Servlet API 3.1, not sure if that API works on Jetty 11 without changes).

barthanssens:
@aschwarte10  turns out Jetty 12 does indeed still support older Servlet APIs when enabling some modules :-)

barthanssens:
Problem in jetty 12 : https://github.com/eclipse-rdf4j/rdf4j/issues/5092

PR #5078, opened 2024-07-18
GH-5064: added docker makefile for Jetty 9

barthanssens:
GitHub issue resolved: #5064 

- add docker makefile for Jetty 9
- updated docker makefile for Tomcat 9 (instead of 8.5)
- updated e2e tests to run the tests for both Jetty and Tomcat
- minor updates of documentation
- added leading slash in overview.jsp to redirect to correct path (Jetty)

----


hmottestad:
The end-to-end tests probably just build the default dockerfile. Could you have it run against both the default/current and the jetty one?

barthanssens:
Ah, good thinking, forgot about that

barthanssens:
Workbench seems to work, but still an issue with 404 on overview.view on the server

e2e/run.sh
```diff
@@ -12,23 +12,31 @@
...
-exit $?
+  npx playwright install --with-deps # install browsers
+  npx playwright test
+
+  cd ..
+  cd docker
+  ./shutdown.sh 
+done
 
+exit $?
```
hmottestad, line 42: What happens if the test fail for tomcat but not jetty? Does it still error out?

barthanssens: hmz yes, that's on the TODO list  (and it probably doesn't hurt to update the docker documentation as well...)

barthanssens: fixed

PR #5091, opened 2024-07-26
GH-5090 Lucene 9 version of the Lucene SAIL

### LINKED ISSUE #5090: Lucene 9 version of the Lucene SAIL
### Problem description

The current Lucene SAIL only works with Lucene 8.

### Preferred solution

Have another implementation of the Lucene SAIL that is compatible with version 9.

Yes

Upgrade the current implementation to Lucene 9. However, there would be incompatibilities with with other libraries used in the workbench if doing so.

### Anything else?

https://github.com/eclipse-rdf4j/rdf4j/discussions/5037

reckart:
GitHub issue resolved: #5090

- Copied the Lucene SAIL implementation and upgraded it to Lucene 9

----


reckart:
@hmottestad build fails due to the Lucene 9 dependencies apparently not being on the whitelist.

hmottestad:
There is a license issue that the Eclipse IP-team needs to check for us. I'll try to look into that when I have some time. 

The second seems to be a Javadocs issue. That might just be that we need to whitelist something, but I'm not sure.

hmottestad:
I'm still not very comfortable with having two Lucene sails. Mixing two versions of the same library has previously led to issues down the road. If someone was depending on the parent pom I would assume they would get both versions of the Lucene library.

hmottestad:
Some relevant discussion: https://github.com/eclipse-rdf4j/rdf4j/discussions/5037

reckart:
>  If someone was depending on the parent pom I would assume they would get both versions of the Lucene library.

The maven resolver can only resolve a particular JAR to a single version, not to multiple. 

If you **inherit** from the parent POM, you get the Lucene version defined there. Nobody should really **depend** on the parent POM. They should use the `rdf4j-bom` instead. But even if they depended on the parent POM, I don't see how they'd get Lucene version as a dependency and not even as a managed version.

The situation is somewhat similar to the `rdf4j-rio-jsonld` and `rdf4j-rio-jsonld-legacy` - the user has to make sure to only have one of those in their dependencies.

The new module locally overrides the Lucene version property - it won't leak. Presently, I didn't even add a managed dependency for the module to the BOM - something that actually should still be done.

hmottestad:
The json-ld side is a bit different. There we are depending on two completely different libraries. The only issue that would occur if you depend on both is that you wouldn't be sure which one you end up getting through the Java Service Provider Interface. 

Here with Lucene we are depending on two versions of the same library.

aschwarte10:
@reckart in the develop branch for RDF4J 6.0 we have moved to Lucene 10. Do you explicitly require lucene 9? If not I would propose to close this PR

reckart:
I am actually still stuck on 5.1.3 because of https://github.com/eclipse-rdf4j/rdf4j/issues/5444 - and that's also where I currently run my Lucene 9 fork of the sail

Going forward, I can do Lucene 10... any hopes that RDF4J 6 will fix #5444?

Issue #5148, opened 2024-10-08
Make ValueStore more resilient against data corruption


ebner:
### Problem description

I am currently facing a situation where a repository somehow got corrupted (still investigating) and where I am unable to process the uncorrupted data due to an unhandled error situation in ValueStore. When iterating over all triples in the repo, some statement cannot be loaded properly and that causes the whole RepositoryConnection to be terminated prematurely. The stack trace is as follows:

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at org.eclipse.rdf4j.sail.nativerdf.ValueStore.data2value(ValueStore.java:529)
	at org.eclipse.rdf4j.sail.nativerdf.ValueStore.getValue(ValueStore.java:192)
	at org.eclipse.rdf4j.sail.nativerdf.NativeStatementIterator.getNextElement(NativeStatementIterator.java:70)
	at org.eclipse.rdf4j.sail.nativerdf.NativeStatementIterator.getNextElement(NativeStatementIterator.java:28)
	at org.eclipse.rdf4j.common.iteration.LookAheadIteration.lookAhead(LookAheadIteration.java:78)
	at org.eclipse.rdf4j.common.iteration.LookAheadIteration.hasNext(LookAheadIteration.java:53)
	at org.eclipse.rdf4j.common.iteration.DualUnionIteration.lookaheadWithoutOrder(DualUnionIteration.java:176)
	at org.eclipse.rdf4j.common.iteration.DualUnionIteration.lookAhead(DualUnionIteration.java:119)
	at org.eclipse.rdf4j.common.iteration.DualUnionIteration.hasNext(DualUnionIteration.java:93)
	at org.eclipse.rdf4j.common.iteration.IterationWrapper.hasNext(IterationWrapper.java:67)
	at org.eclipse.rdf4j.sail.base.SailClosingIteration.hasNext(SailClosingIteration.java:83)
	at org.eclipse.rdf4j.common.iteration.IterationWrapper.hasNext(IterationWrapper.java:67)
	at org.eclipse.rdf4j.sail.helpers.SailBaseIteration.hasNext(SailBaseIteration.java:43)
	at org.eclipse.rdf4j.sail.helpers.CleanerIteration.hasNext(CleanerIteration.java:42)
	at org.eclipse.rdf4j.repository.sail.SailCloseableIteration.hasNext(SailCloseableIteration.java:40)
[...]

PR #5150, opened 2024-10-10
GH-5148 Introduce "soft fail" for corrupt ValueStore

ebner:
GitHub issue resolved: #5148

 - Improved use of a corrupted ValueStore by handling error situations in `ValueStore.data2value()` with SailException (previously it was possible to get an `ArrayIndexOutOfBoundsException` and `IllegalArgumentException`).
 - Introduction of a new system property `org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptData` to enable "soft failure mode" and to avoid exceptions as described above. Instead, an instance of the newly introduced CorruptValue will be returned.
 - This solves the problem of not being able to iterate (and in consequence, backup) over the NativeStore in order to save uncorrupted statements. Without "soft failure mode" the `RepositoryConnection` would be forcefully closed when hitting the first corrupted value.

----

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
@@ -33,6 +33,7 @@ import org.eclipse.rdf4j.model.vocabulary.RDF;
 import org.eclipse.rdf4j.model.vocabulary.XSD;
 import org.eclipse.rdf4j.sail.SailException;
 import org.eclipse.rdf4j.sail.nativerdf.datastore.DataStore;
+import org.eclipse.rdf4j.sail.nativerdf.model.CorruptValue;
 import org.eclipse.rdf4j.sail.nativerdf.model.NativeBNode;
 import org.eclipse.rdf4j.sail.nativerdf.model.NativeIRI;
 import org.eclipse.rdf4j.sail.nativerdf.model.NativeLiteral;
@@ -123,6 +124,11 @@ public class ValueStore extends SimpleValueFactory {
 	 */
 	private final ConcurrentCache<String, Integer> namespaceIDCache;
 
+	/**
+	 * Do not throw an exception in case a value cannot be loaded, e.g. due to a corrupt value store.
+	 */
+	private final boolean softFailOnCorruptData;
+
 	/*--------------*
 	 * Constructors *
 	 *--------------*/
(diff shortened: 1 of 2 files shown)


core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
```diff
@@ -534,7 +555,10 @@ private NativeValue data2value(int id, byte[] data) throws IOException {
 		case LITERAL_VALUE:
 			return data2literal(id, data);
 		default:
-			throw new IllegalArgumentException("Invalid type " + data[0] + " for value with id " + id);
+			if (softFailOnCorruptData) {
+				return new CorruptValue(revision, id);
```
hmottestad, line 559: Could potentially include data[0] at this point.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/model/CorruptValue.java
```diff
@@ -0,0 +1,91 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Eclipse RDF4J contributors, Aduna, and others.
```
hmottestad, line 2: This is an old copyright header. See https://github.com/eclipse-rdf4j/rdf4j/blob/main/CONTRIBUTING.md#code-formatting

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/model/CorruptValue.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @author Hannes Ebner
+ */
+public class CorruptValue implements NativeValue {
+
+	/*-----------*
+	 * Constants *
+	 *-----------*/
+
+	private static final long serialVersionUID = 8829067881854394802L;
```
hmottestad, line 31: Todo: Double check that this is a unique number.

hmottestad (changes requested):

hmottestad:
Any chance you would be interested in adding some info about this to our documentation? 

I'm not sure where, but maybe at the end of the NativeStore section here: https://github.com/eclipse-rdf4j/rdf4j/blob/d05b5113549bd7c24c365ccfbc05e6a8870afbda/site/content/documentation/programming/repository.md?plain=1#L60

Could we also try to create a test? Maybe create a NativeStore, add data, verify that the data is there. Do a shutdown() and then modify the files on disk to make them corrupted. Then init() the store again and try to read the data.

PR #5157, opened 2024-10-23
GH-5148 Introduce "soft fail" for corrupt ValueStore

hmottestad:
GitHub issue resolved: #5148 

----

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
@@ -10,7 +10,7 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.nativerdf;
 
-import static org.eclipse.rdf4j.sail.nativerdf.NativeStore.SOFT_FAIL_ON_CORRUPT_DATA;
+import static org.eclipse.rdf4j.sail.nativerdf.NativeStore.SOFT_FAIL_ON_CORRUPT_DATA_AND_REPAIR_INDEXES;
 
 import java.io.File;
 import java.io.IOException;
@@ -223,11 +223,11 @@ public class ValueStore extends SimpleValueFactory {
 		NativeValue resultValue = getValue(id);
 
 		if (resultValue != null && !(resultValue instanceof Resource)) {
-			if (SOFT_FAIL_ON_CORRUPT_DATA && resultValue instanceof CorruptValue) {
+			if (SOFT_FAIL_ON_CORRUPT_DATA_AND_REPAIR_INDEXES && resultValue instanceof CorruptValue) {
 				return (T) new CorruptIRIOrBNode(revision, id, ((CorruptValue) resultValue).getData());
 			}
 			logger.warn(
-					"Possible corrupt data consider setting the system property org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptData to true");
+					"Possible corrupt data consider setting the system property org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptDataAndRepairIndexes to true");
 		}
 
 		return (T) resultValue;
@@ -245,14 +245,14 @@ public class ValueStore extends SimpleValueFactory {
 		NativeValue resultValue = getValue(id);
 
 		if (resultValue != null && !(resultValue instanceof IRI)) {
-			if (SOFT_FAIL_ON_CORRUPT_DATA && resultValue instanceof CorruptValue) {
+			if (SOFT_FAIL_ON_CORRUPT_DATA_AND_REPAIR_INDEXES && resultValue instanceof CorruptValue) {
 				if (resultValue instanceof CorruptIRI) {
 					return (T) resultValue;
 				}
 				return (T) new CorruptIRI(revision, id, null, ((CorruptValue) resultValue).getData());
 			}
 			logger.warn(
-					"Possible corrupt data consider setting the system property org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptData to true");
+					"Possible corrupt data consider setting the system property org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptDataAndRepairIndexes to true");
 		}
 
 		return (T) resultValue;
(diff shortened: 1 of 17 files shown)


ebner:
Thanks for the fixes @hmottestad! I haven't had time yet to look into this, but I can contribute to the documentation if you want.

Following up on another comment from your review of my PR: adding a test for an artificially corrupted store may be difficult, I don't really know how to do this reliably. Perhaps it is possible to manipulate a NativeValue instance in the ValueStore to hold an illegal state that triggers the new soft fail functionality. Any other suggestions?

hmottestad:
I'm actually working on it right now. I have a test that sets each byte in the values.dat file to zero, one at a time, then tries to read all the data.

core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/NativeSailStoreCorruptionTest.java
```diff
@@ -0,0 +1,224 @@
...
+			overwriteByteInFile(new File(dataDir, "values.dat"), i, 0x0);
+
+			repo.init();
+
+			List<Statement> list = getStatements();
+			assertEquals(6, list.size());
+
+		}
+
+	}
```
hmottestad, line 198: This is actually the main test that tests every single byte, except the header bytes, in the values.dat file.

site/content/documentation/programming/repository.md
```diff
@@ -98,6 +98,9 @@ import org.eclipse.rdf4j.sail.nativerdf.NativeStore;
 Repository repo = new SailRepository(new NativeStore());
 ```
 
+In the unlikely event of corruption the system property `org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptData` can be set to `true` to
+allow the NativeStore to output CorruptValue/CorruptIRI/CorruptIRIOrBNode/CorruptLiteral objects.
+
```
hmottestad, line 103: Documentation.

hmottestad:
@ebner I've added some test coverage now.

I think it might be best to also try to see what happens if the data is written to a trig or json-ld file, to test the very simple corrupt value implementations i added.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/btree/BTree.java
```diff
@@ -292,6 +292,12 @@ public BTree(File dataDir, String filenamePrefix, int blockSize, int valueSize,
 			this.valueSize = buf.getInt();
 			this.rootNodeID = buf.getInt();
 
+			if (rootNodeID == 0) {
+				if (nioFile.size() > blockSize) {
```
hmottestad, line 296: This is a somewhat brittle approach and it's really only relevant when using the soft fail property. So maybe this should be wrapped with the property too and throw something other than an illegal state exception.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
```diff
@@ -191,12 +200,62 @@ public NativeValue getValue(int id) throws IOException {
...
+	public <T extends NativeValue & Resource> T getResource(int id) throws IOException {
+
+		NativeValue resultValue = getValue(id);
+
+		if (resultValue != null && !(resultValue instanceof Resource)) {
+			if (SOFT_FAIL_ON_CORRUPT_DATA_AND_REPAIR_INDEXES && resultValue instanceof CorruptValue) {
+				return (T) new CorruptIRIOrBNode(revision, id, ((CorruptValue) resultValue).getData());
+			}
+			logger.warn(
+					"Possible corrupt data consider setting the system property org.eclipse.rdf4j.sail.nativerdf.softFailOnCorruptDataAndRepairIndexes to true");
```
aschwarte10, line 230: this message sounds a bit strange (though it is clear what it means. Maybe you can phrase it in two sentences, same in line 255 and line 629

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ValueStore.java
```diff
@@ -534,49 +601,76 @@ private NativeValue data2value(int id, byte[] data) throws IOException {
 		case LITERAL_VALUE:
 			return data2literal(id, data);
 		default:
-			throw new IllegalArgumentException("Invalid type " + data[0] + " for value with id " + id);
+			if (SOFT_FAIL_ON_CORRUPT_DATA_AND_REPAIR_INDEXES) {
+				logger.error("Soft fail on corrupt data: Invalid type {} for value with id {}", data[0], id);
+				return new CorruptUnknownValue(revision, id, data);
+			}
+			throw new SailException("Invalid type " + data[0] + " for value with id " + id
```
aschwarte10, line 608: Maybe also here, two sentences.

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/DataFile.java
```diff
@@ -197,29 +203,50 @@ public byte[] getData(long offset) throws IOException {
 				(data[2] << 8) & 0x0000ff00 |
 				(data[3]) & 0x000000ff;
 
-		// We have either managed to read enough data and can return the required subset of the data, or we have read
-		// too little so we need to execute another read to get the correct data.
-		if (dataLength <= data.length - 4) {
+		// If the data length is larger than 750MB, we are likely reading the wrong data. Probably data corruption.
```
aschwarte10, line 206: how did you come up with 750MB? Is this a special size?

aschwarte10 (commented):
I am too distant from the native store implementation and its index structures to give a proper review. From what I can tell with going through the PR the logic seems fine: I can follow the idea and the new setting is opt-in. 
Please see inline for minor (optional) comments on some messages

core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/datastore/DataFile.java
hmottestad: It was what I experienced when testing.

I wouldn't think that a single literal would realistically be 750 MB. But I'll double check that the logic only applies if the system property is activated.

hmottestad: I added some more context to the comment

Issue #5159, opened 2024-10-27
GroupIterator doesn't close child iterator


frensjan:
### Current Behavior

When a query times out, the current iteration is closed from a timer thread. However, if GroupIterator is closed while it is building it's entries (in `buildEntries`), this close signal doesn't propagate to it's argument.

### Expected Behavior

The iterator from which the entries of GroupIterator should be closed.

### Version

4.3.8

### Are you interested in contributing a solution yourself?

Yes, see [PR](https://github.com/eclipse-rdf4j/rdf4j/pull/5160)

PR #5160, opened 2024-10-28
GH-5159 close child of GroupIterator

frensjan:
GitHub issue resolved: #5159

I've added the child argument's iterator as (volatile) field in the GroupIterator so that it can be closed while building entires.

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
@@ -30,6 +30,7 @@ import org.eclipse.rdf4j.collection.factory.api.BindingSetKey;
 import org.eclipse.rdf4j.collection.factory.api.CollectionFactory;
 import org.eclipse.rdf4j.collection.factory.impl.DefaultCollectionFactory;
 import org.eclipse.rdf4j.common.iteration.AbstractCloseableIteratorIteration;
+import org.eclipse.rdf4j.common.iteration.CloseableIteration;
 import org.eclipse.rdf4j.common.transaction.QueryEvaluationMode;
 import org.eclipse.rdf4j.model.Literal;
 import org.eclipse.rdf4j.model.Value;
@@ -89,6 +90,9 @@ public class GroupIterator extends AbstractCloseableIteratorIteration<BindingSet
 
 	private final QueryEvaluationStep arguments;
 
+	// The iteration of the arguments, stored while building entries for allowing premature closing
+	private volatile CloseableIteration<BindingSet> argumentsIter;
+
 	private final ValueFactory vf;
 
 	private final CollectionFactory cf;
@@ -129,7 +133,13 @@ public class GroupIterator extends AbstractCloseableIteratorIteration<BindingSet
 
 	@Override
 	public void handleClose() throws QueryEvaluationException {
-		cf.close();
+		try {
+			cf.close();
+		} finally {
+			var iter = argumentsIter;
+			if (iter != null)
+				iter.close();
+		}
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
```diff
@@ -129,7 +133,11 @@ public GroupIterator(EvaluationStrategy strategy, Group group, BindingSet parent
 
 	@Override
 	public void handleClose() throws QueryEvaluationException {
-		cf.close();
+		try {
+			cf.close();
+		} finally {
+			argumentsIter.close();
```
hmottestad, line 139: Looks like you need to add:

```java
var argumentsIter = this.argumentsIter;
if(argumentsIter != null) argumentsIter.close() 
```

hmottestad:
Screenshot 2024-10-28 at 09 08 57

Should we check if the current iterator is closed within the loop? We have the isClosed() method we can call. For performance reasons the backing variables isn't volatile, but it should still get updated eventually. The variable is set before the handleClose() method is called, so it would be a nice safety net in case the argumentsIter doesn't close correctly for some reason.

hmottestad:
Thanks for this fix. Very good catch and a nice and clean solution!

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
frensjan: Obviously! Thanks Håvard.

frensjan:
> Should we check if the current iterator is closed within the loop? We have the isClosed() method we can call. For performance reasons the backing variables isn't volatile, but it should still get updated eventually. The variable is set before the handleClose() method is called, so it would be a nice safety net in case the argumentsIter doesn't close correctly for some reason.

Yes, that's good to add.

JervenBolleman:
LGTM! thank you @frensjan

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
```diff
@@ -129,7 +133,13 @@ public GroupIterator(EvaluationStrategy strategy, Group group, BindingSet parent
 
 	@Override
 	public void handleClose() throws QueryEvaluationException {
-		cf.close();
+		try {
+			cf.close();
+		} finally {
+			if (argumentsIter != null) {
+				argumentsIter.close();
```
hmottestad, line 140: Cant argumwntsIter still be null? It could be set to null between this line and the null check above.

hmottestad (changes requested):

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIteratorTest.java
```diff
@@ -260,6 +251,46 @@ public void testCustomAggregateFunction_WrongIri() throws QueryEvaluationExcepti
...
+		};
+
+		Group group = new Group(new EmptySet());
+		GroupIterator groupIterator = new GroupIterator(evaluator, group, EmptyBindingSet.getInstance(), context);
+
+		Thread iteratorThread = new Thread(groupIterator::next);
+		iteratorThread.start();
+		assertThat(iterating.await(5, TimeUnit.SECONDS)).isTrue();
+		groupIterator.close();
+		assertThat(closed.await(5, TimeUnit.SECONDS)).isTrue();
```
hmottestad, line 291: Can we also assert that the iteratorThread is no longer alive? So we are sure that we're not leaking a thread.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
frensjan: Yes, indeed. Missed that in your initial comment.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIteratorTest.java
frensjan: The test _was_ in fact leaking a Thread. Just fixed it.

hmottestad:
Thanks @frensjan !

I've merged the PR and it'll ve part of the next bug fix release which is due any day now.

Issue #5151, opened 2024-10-17
Refine bind join implementation for "check joins" to use VALUES clause


aschwarte10:
### Problem description

The current implementation of the bind join uses a UNION construct for the case when in the join all variables are bound.

Note: this is not the typical join case, where we already use a VALUES clause, but rather a special case

The issue with the approach is that the query construct is error prone (e.g. for optimizations in some DB engines) and actually to verbose  and not required to eb that complex

### Preferred solution

Use the VALUES clause implementation which can be simply applied also for check joins

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

This is a technically a minor change in the code with quite a bit of impact

PR #5165, opened 2024-10-29
GH-5151: use VALUES clause for FedX bind join with no free vars

aschwarte10:
GitHub issue resolved: #GH-5151

For evaluation of bind joins the implementation for quite some time
makes use of a VALUES clause query.

Except for one code-path: for bind joins - where in the join all
arguments are bound - it was still using the old UNION query approach.
This approach is error prone and no longer required, i.e. the check join
can be executed with the same logic as the regular VALUES clause.

Note: an additional unit test for covering bind joins with no free vars
is added.

This change also marks a number of methods and classes used for the old
UNION based approach as deprecated. The implementations are internal to
the FedX engine and can be removed in the next major release.

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBindJoin.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBindJoin.java
@@ -45,12 +45,7 @@ public class ControlledWorkerBindJoin extends ControlledWorkerBindJoinBase {
 		final TaskCreator taskCreator;
 		if (expr instanceof StatementTupleExpr) {
 			StatementTupleExpr stmt = (StatementTupleExpr) expr;
-			if (stmt.hasFreeVarsFor(bs)) {
-				taskCreator = new BoundJoinTaskCreator(strategy, stmt);
-			} else {
-				expr = new CheckStatementPattern(stmt, queryInfo);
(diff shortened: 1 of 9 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/join/ControlledWorkerBindJoin.java
```diff
@@ -45,12 +45,7 @@ protected TaskCreator determineTaskCreator(TupleExpr expr, BindingSet bs) {
 		final TaskCreator taskCreator;
 		if (expr instanceof StatementTupleExpr) {
 			StatementTupleExpr stmt = (StatementTupleExpr) expr;
-			if (stmt.hasFreeVarsFor(bs)) {
-				taskCreator = new BoundJoinTaskCreator(strategy, stmt);
-			} else {
-				expr = new CheckStatementPattern(stmt, queryInfo);
-				taskCreator = new CheckJoinTaskCreator(strategy, (CheckStatementPattern) expr);
-			}
+			taskCreator = new BoundJoinTaskCreator(strategy, stmt);
```
aschwarte10, line 48: this is the relevant functional change: the VALUES based implementation can also be used for bind joins without free variables without functional change.

Note that in very early versions (where the bind join was implemented with a UNION construct) there was a need to distinguish the cases. This is a left over

hmottestad (approved):

Issue #5208, opened 2024-11-26
Add Notion of Request ID to Make Logging More Transparent


benherber:
### Problem description

For larger deployments of RDF4J with multiple users, it has proven slightly difficult to trace request execution through logs without adding some notion of a 'request ID' (the thread name contains some information but not enough to fully differentiate requests). 

Having such a notion in logs make the request execution more transparent by behind able to trace through all subsystems for any particular execution. Moreover, it also allows for log aggregation tools to group query execution information and glean insights that are useful in debugging and performance optimization.

### Preferred solution

Add a 'request Id' to MDC such that it can be (optionally) logged if desired.

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

Tried to wrangle Tomcat and Logback for implementing a similar feature but came up short. [Looks like Jakarta has tooling](https://jakarta.ee/specifications/servlet/6.0/apidocs/jakarta.servlet/jakarta/servlet/servletrequest#getRequestId()) to aid in this when we upgrade, but that is out of reach for now it seems.

PR #5209, opened 2024-11-26
GH-5208 Enable request tracing through logs using new associated request id

benherber:
GitHub issue resolved: #5208 

This PR adds a unique request identifier to MDC in order to allow for more intelligent log parsing and aggregation as well as aid in server debugging.

----

### DIFF
--- a/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ServerInterceptor.java
+++ b/tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ServerInterceptor.java
@@ -10,9 +10,13 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.http.server;
 
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.slf4j.MDC;
 import org.springframework.web.servlet.HandlerInterceptor;
 
 /**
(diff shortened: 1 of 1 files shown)


tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ServerInterceptor.java
```diff
@@ -24,13 +27,15 @@
...
 
+	private static final String REQUEST_ID = "requestId";
 	private volatile String origThreadName;
 
 	@Override
 	public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler)
 			throws Exception {
 		origThreadName = Thread.currentThread().getName();
 		Thread.currentThread().setName(getThreadName());
+		MDC.put(REQUEST_ID, UUID.randomUUID().toString());
```
hmottestad, line 38: I've also used MDC for this in other dev projects. It's probably still the best option. 

From experience it can get quite expensive to generate a lot of UUIDs especially if the transactions are very small. Can we have the UUID as a string at startup and then an AtomicLong that we increment and append?

benherber: Good point! Should be changed. Introduced a process UUID which is postponed with an `AtomicLong` in a format which should be parsable by some tooling if need be.

Let me know you thoughts.

hmottestad (approved):

tools/server-spring/src/main/java/org/eclipse/rdf4j/http/server/ServerInterceptor.java
```diff
@@ -24,13 +28,23 @@
  */
 public abstract class ServerInterceptor implements HandlerInterceptor {
 
+	private static final String REQUEST_ID_KEY = "requestId";
```
hmottestad, line 31: I'm having some second doubts about the naming. We have users that extend the RDF4J server code and it could be that they are already doing something similar to this. Do you think we could call ours `rdf4jRequestId` just to be sure that we don't overwrite someone else's request ID?

benherber: Ahh fair point. I've also seen the package name convention used. Something like `org.eclipse.rdf4j.requestId` so there would never be the possibility of a clash unless we did it to ourselves?

hmottestad: Package name looks very nice!

hmottestad (approved):

Issue #5210, opened 2024-11-26
Allow Child Sails to Benefit from `AbstractSailConnection`'s Buffering Scheme


benherber:
### Problem description

`AbstractSailConnection` manages a lot of concurrency and buffering for implementing sails; however, when it goes to flush the buffer, it does so one triple at a time instead of allowing the implementing sail to flush the buffer itself. This causes the implementing sail to either have to pay a possible higher write cost since the writes are decoupled from each other or implement buffering itself which is counterintuitive to the expressed design of the parent class.

### Preferred solution

Ideally I would like just to be handed a `Collection<Statement>` that I can write to storage media en masse similar to how the snapshot store works. For backwards compatibility it can default to the existing implementation. The sail implementation being made is incompatible with how snapshot store is designed for a few reasons I won't get into. It has proven just much easier to extend from `AbstractSailConnection` instead for what I'm trying to do.

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

I looked into a few solutions such as overriding methods in `AbstractSailConnection` but ran into concurrency issues since child sails don't have access to all the internals. I also experimented with my own buffer scheme in the child sail, but ended up just reimplementing what was in `AbstractSailConnection` more-or-less.

nguyenm100:
@hmottestad can we get some feedback on this when you get a sec? we'd prefer not having to fork internally where possible. tx

PR #5211, opened 2024-11-26
GH-5210 Enable implementing sails to benefit from parent sail buffering

benherber:
GitHub issue resolved: #5210 

This PR adds an overridable internal method `bulkAddStatementsInternal` which implementing SAILs can use to leverage the existing multithreading-safe, buffering implemented in `AbstractSailConnection`. This change should be fully backwards compatible as it defaults to the exiting implementation.

----

### DIFF
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
@@ -24,6 +24,7 @@ import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.LockSupport;
 import java.util.concurrent.locks.ReentrantLock;
 
+import org.eclipse.rdf4j.common.annotation.Experimental;
 import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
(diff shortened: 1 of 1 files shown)


core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -973,6 +971,13 @@ protected void prepareInternal() throws SailException {
 	protected abstract void addStatementInternal(Resource subj, IRI pred, Value obj, Resource... contexts)
 			throws SailException;
 
+	protected void bulkAddStatementsInternal(final Collection<? extends Statement> statements)
```
hmottestad, line 974: Could you annotate this with @Experimental so that we can adjust the signature down the line if we need to?

hmottestad (changes requested):

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
benherber: Sure thing!

Issue #5229, opened 2025-01-15
FedX: left bind join (i.e. OPTIONAL) may produce incorrect results for a single left bindingset as input


aschwarte10:
### Current Behavior

It can happen that the left bind join, where the left argument produces a single bindingset, can cause an incorrect (empty) result. The situation can occur for advanced source selection cases.

### Expected Behavior

The federation engine correctly evaluates OPTIONAL clauses. For the concrete case a non-empty result is expected

### Steps To Reproduce

Unit test reproducing the issue

```
@ParameterizedTest
	@ValueSource(booleans = { true, false })
	public void test_leftBindJoin_emptyLeftArgumentAsExclusiveGroup(boolean bindLeftJoinOptimizationEnabled)
			throws Exception {

		var endpoints = prepareTest(
				Arrays.asList("/tests/basic/data_emptyStore.ttl", "/tests/basic/data_emptyStore.ttl"));

		Repository repo1 = getRepository(1);
		Repository repo2 = getRepository(2);

		Repository fedxRepo = fedxRule.getRepository();

		fedxRule.setConfig(config -> {
			config.withBoundJoinBlockSize(10);
			config.withEnableOptionalAsBindJoin(bindLeftJoinOptimizationEnabled);
		});

		// add a person
		try (RepositoryConnection conn = repo1.getConnection()) {
			var p = Values.iri("http://ex.com/p1");
			var otherP = Values.iri("http://other.com/p1");
			conn.add(p, OWL.SAMEAS, otherP);
		}

		// add name for person 1
		try (RepositoryConnection conn = repo2.getConnection()) {
			var otherP = Values.iri("http://other.com/p1");
			conn.add(otherP, FOAF.NAME, Values.literal("Person 1"));
		}

		// mark that repo2 for some reason has foaf:age statements (e.g. old cache entry)
		Endpoint repo2Endpoint = endpoints.get(1);
		federationContext().getSourceSelectionCache()
				.updateInformation(new SubQuery(null, FOAF.AGE, null), repo2Endpoint, true);

		fedxRule.enableDebug();

		try {
			// run query which joins results from multiple repos
			// the age does not exist for any person
			try (RepositoryConnection conn = fedxRepo.getConnection()) {
				String query = "PREFIX foaf: <http://xmlns.com/foaf/0.1/> " +
						"SELECT * WHERE { "
						+ " ?person owl:sameAs ?otherPerson . "
[...]

PR #5230, opened 2025-01-15
GH-5229: fix left bind join in FedX for single binding input

aschwarte10:
GitHub issue resolved: #5229 

This change fixes a situation that can incorrectly cause empty results. It happens when the input of the left argument is a single binding set and for special source selection situations (e.g. the right argument is marked as ExclusiveStatement while the endpoint does not provide data)

To avoid the issue we also use the regular left join logic also for a single binding set input, which can handle the situation properly.

Issue is covered with a unit test.

----


hmottestad (approved):
Looks good. Thanks for the unit test.

Issue #5231, opened 2025-01-15
FedX: Poor potential query performance for RepositoryConnections#hasStatements()


aschwarte10:
### Current Behavior

The query performance of RepositoryConnections#hasStatements() is poor for queries like

- `null, rdf:type, null`
- `null, null, null`

Reason: the current implementation fetches all statements using `getStatements()` only to see if one of the federation members returns a value.

 

### Expected Behavior

Ideal solution: do not delegate to getStatements, but make use of hasStatements on individual federation members.

### Steps To Reproduce

Stacktrace:

```
Thread [main] (Suspended (breakpoint at line 565 in FederationEvalStrategy))	
	MetaphactoryFedXEvaluationStrategy(FederationEvalStrategy).getStatements(QueryInfo, Resource, IRI, Value, Resource...) line: 565	
	FedXConnection.getStatementsInternal(Resource, IRI, Value, boolean, Resource...) line: 316	
	FedXConnection(AbstractSailConnection).hasStatementInternal(Resource, IRI, Value, boolean, Resource[]) line: 445	
	FedXConnection(AbstractSailConnection).hasStatement(Resource, IRI, Value, boolean, Resource...) line: 432	
	FedXRepositoryConnection(SailRepositoryConnection).hasStatement(Resource, IRI, Value, boolean, Resource...) line: 335	
```
	

### Version

5.1.0

### Are you interested in contributing a solution yourself?

None

PR #5232, opened 2025-01-15
GH-5231: fix poor query performance for hasStatements() in FedX

aschwarte10:
GitHub issue resolved: #5231 

The previous implementation of the FedXConnection was delegating "hasStatements()" to the implementation of "getStatements()", where the latter was actually fetching data from the federation members.

For checks hasStatements() checks like {null, rdf:type, null} or even {null, null, null} the implementation is problematic as it would fetch all data matching the pattern from the federation members, only to answer if it actually exists.

We now make use of "existence" check on the federation members, and can actually rely on the source selection cache for this.

Unit test coverage has been added.

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
@@ -17,6 +17,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
+import org.eclipse.rdf4j.common.annotation.Experimental;
 import org.eclipse.rdf4j.common.iteration.CloseableIteration;
 import org.eclipse.rdf4j.common.iteration.EmptyIteration;
 import org.eclipse.rdf4j.common.iteration.SingletonIteration;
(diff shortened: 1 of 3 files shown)


tools/federation/src/main/java/org/eclipse/rdf4j/federated/evaluation/FederationEvalStrategy.java
```diff
@@ -605,6 +605,47 @@ public CloseableIteration<Statement> getStatements(QueryInfo queryInfo, Resource
...
+
+	/**
+	 * Returns the accessible federation members in the context of the query. By default this is all federation members.
+	 *
+	 * @param queryInfo
+	 * @return
+	 */
+	protected List<Endpoint> getAccessibleFederationMembers(QueryInfo queryInfo) {
+		return federationContext.getFederation().getMembers();
+	}
```
hmottestad, line 647: Can you explain more why this needs to be protected. I didn't see that it's override anywhere in the code.

hmottestad: Sound good. Can you update the docs to reflect that this is meant to be an extension point for subclasses. If you are unsure if you need to change it in future you can annotate it as experimental if you want.

aschwarte10: Good suggestion, done

hmottestad (approved):
You can merge when you are ready.

hmottestad:
I cancelled the test that was taking too long. It's an issue related to some very specific edge cases when connections and resources are not closed before trying to shutdown the ShaclSail. I have a PR open for trying to fix it, but it's a bit tricky.

Issue #5234, opened 2025-01-20
FedX: Support to influence source selection cache parameters


aschwarte10:
### Current Behavior

The source selection cache in FedX is currently implemented as in memory cache which under the hood uses Guava. There exists some means of fine-tuning the cache configuration using a cache spec, but this is too limited.

### Expected Behavior

The implementation should allow to acceppt a Guava Cache builder for configuration

### Version

5.1.0

### Are you interested in contributing a solution yourself?

Yes

hmottestad:
This is necessary for performance tuning.

PR #5235, opened 2025-01-20
GH-5234: fix limited programmatic configuration support of FedX source selection cache

aschwarte10:
GitHub issue resolved: #5234 

Add an additional constructor accepting a supplier for an initialized cache.

----

### DIFF
--- a/tools/federation/src/main/java/org/eclipse/rdf4j/federated/cache/SourceSelectionMemoryCache.java
+++ b/tools/federation/src/main/java/org/eclipse/rdf4j/federated/cache/SourceSelectionMemoryCache.java
@@ -12,6 +12,7 @@ package org.eclipse.rdf4j.federated.cache;
 
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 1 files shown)


jetztgradnet (approved):
LGTM. An additional constructor should not cause any binary incompatibilities.

hmottestad:
This is a new feature though and not really a bug fix.

aschwarte10:
> This is a new feature though and not really a bug fix.

Technically, you are absolutely right, good eyes. Let me try to explain why I reported this as bug, which in our case impacts performance / stability of a federated system. For our use-cases more control over the cache strategies (eviction, size, introspection) is required to offer good performance. 

If this change is not acceptable for a patch release, we in our case would (temporarily) need to override/customize some parts of the source selection logic to make the guava cache accessible. Please let me know.

hmottestad:
We can change it from bug to performance and then we can include it.

aschwarte10:
> We can change it from bug to performance and then we can include it.

Thanks, I see that you already did this. Good to know!

Issue #5343, opened 2025-05-31
Speed-up size() in LMDB Sail


odysa:
### Problem description

There are about 50M triples in our LMDB store. `size()` averagely takes above 60s.
The current implementation counts the statement iterator.

```java
@Override
protected long sizeInternal(Resource... contexts) throws SailException {
    try (Stream<? extends Statement> stream =
             getStatementsInternal(null, null, null, false, contexts).stream()) {
        return stream.count();
    }
}
```

### Preferred solution

`MDB_stat` exposes ms_entries, which is the exact number of data items  in a database.
Could we  leverage  this value instead of iterating over every statement to compute the store size?

If I’m reading the docs and code correctly, `ms_entries` should equal the entry count of any of our indexes (e.g., spoc)?
```c
typedef struct MDB_stat {
	unsigned int	ms_psize;			/**< Size of a database page.
											This is currently the same for all databases. */
	unsigned int	ms_depth;			/**< Depth (height) of the B-tree */
	mdb_size_t		ms_branch_pages;	/**< Number of internal (non-leaf) pages */
	mdb_size_t		ms_leaf_pages;		/**< Number of leaf pages */
	mdb_size_t		ms_overflow_pages;	/**< Number of overflow pages */
	mdb_size_t		ms_entries;			/**< Number of data items */
} MDB_stat;
```

### Are you interested in contributing a solution yourself?

Yes

odysa:
Can we have a non-materializing getStatements only for `size()`. Get rid of all `initValue`?
```java
	@Override
	protected CloseableIteration<? extends Statement> getStatementsInternal(Resource subj, IRI pred,
			Value obj,
			boolean includeInferred, Resource... contexts) throws SailException {
		return new IterationWrapper<Statement>(
				super.getStatementsInternal(subj, pred, obj, includeInferred, contexts)) {
			@Override
			public Statement next() throws SailException {
				// ensure that all elements of the statement are initialized (lazy values are resolved)
				Statement stmt = super.next();
				initValue(stmt.getSubject());
				initValue(stmt.getPredicate());
				initValue(stmt.getObject());
				initValue(stmt.getContext());
				return stmt;
			}
		};
	}
```

kenwenzel:
I think it is called there because the statements must also exist even when the underlying SAIL was closed.

PR #5345, opened 2025-05-31
GH-5343 Make LMDBSail Size() 36000x Faster 🚀🚀🚀🚀

odysa:
GitHub issue resolved: # 
#5343 

This PR introduces an optimization for the `size(...)` method in the LMDBstore implementation. Introduce a `cardinalityExact` to calcualte the exact size, leverageing`mdb_stats` when possible.

## Key Changes
- **ChangeSet**

	- **Fast-path Optimization**  
 	 	- When there are no `approved` or `deprecated` changes in the current transaction or not in a transcation, the method directly delegates to the `derivedFrom` store for fast cardinality estimation.

	- **Fallback to Iterator**  
 	 	- When changes exist in the transaction, the method falls back to streaming through matching statements using `getStatements(...).stream().count()`. This bypasses LMDB’s lazy evaluation to ensure consistency, even with uncommitted changes.

- **Low-level Size Calculation (`cardinalityExact`)**  
  - If the pattern is completely unspecified (i.e. all wildcards), the method uses LMDB's `mdb_stat` to return the total size efficiently.
  - For specific patterns, it iterates over both **explicit** and **implicit** triples and counts the results.

## Perf
I created a LMDBSail with 10M triples.
Original `size()`: 21802ms
Screenshot 2025-05-31 at 11 57 46 AM

Optimized `size()`: 
685.6 μs to get the full size by leveraging `mdb_stats`.
274.2 ms to get the size of a context of 5000000 triples.
```
Total Size: 10000000, Time taken: 685.6 μs
Size in context: 5000000, Time taken: 274.2 ms
```
----

### DIFF
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDataset.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDataset.java
@@ -132,7 +132,8 @@ public interface SailDataset extends SailClosable {
 
 	@Experimental
 	default long size(final Resource subj, final IRI pred, final Value obj, final Resource... contexts) {
-		return getStatements(subj, pred, obj, contexts).stream()
-				.count();
+		try (var iter = getStatements(subj, pred, obj, contexts)) {
+			return iter.stream().count();
+		}
 	}
 }
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
@@ -392,6 +392,8 @@ class SailDatasetImpl implements SailDataset {
 		}
 
 		// Fallback path: iterate over all matching statements
-		return getStatements(subj, pred, obj, contexts).stream().count();
+		try (var iter = getStatements(subj, pred, obj, contexts)) {
+			return iter.stream().count();
+		}
 	}
 }
--- a/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
+++ b/core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
@@ -1035,19 +1035,19 @@ public abstract class SailSourceConnection extends AbstractNotifyingSailConnecti
 	}
 
 	/**
-	 * Returns the number of statements in the snapshot, optionally including inferred statements, for the given
-	 * contexts. This method reads the size directly from the dataset within the current isolation level.
+	 * Returns the number of statements in the snapshot for the given contexts. This method reads the size directly from
+	 * the dataset within the current isolation level.
 	 *
-	 * @param includeInferred whether to include inferred statements in the count
-	 * @param contexts        the RDF contexts (named graphs) to restrict the count to; if none are provided, counts all
-	 *                        contexts
+	 * @param contexts the RDF contexts (named graphs) to restrict the count to; if none are provided, counts all
+	 *                 contexts
 	 * @return the number of statements in the dataset
 	 * @throws SailException if an error occurs while accessing the Sail store
 	 */
 	@Experimental
-	protected long getSizeFromSnapshot(final boolean includeInferred, final Resource... contexts) throws SailException {
-		try (SailSource branch = branch(IncludeInferred.fromBoolean(includeInferred))) {
-			return branch.dataset(getIsolationLevel()).size(null, null, null, contexts);
+	protected long getSizeFromSnapshot(final Resource... contexts) throws SailException {
+		try (SailSource branch = branch(IncludeInferred.explicitOnly);
+				SailDataset dataset = branch.dataset(getIsolationLevel())) {
+			return dataset.size(null, null, null, contexts);
 		}
 	}
 
--- a/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreConnection.java
+++ b/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreConnection.java
@@ -202,6 +202,6 @@ public class LmdbStoreConnection extends SailSourceConnection {
 
 	@Override
 	protected long sizeInternal(Resource... contexts) throws SailException {
-		return super.getSizeFromSnapshot(false, contexts);
+		return super.getSizeFromSnapshot(contexts);
 	}
 }


odysa:
Let me further explore the materialize-nothing solution to see how fast it can be.

kenwenzel:
> Let me further explore the materialize-nothing solution to see how fast it can be.

Maybe we can parameterize the cardinality() function with `exact` and reuse the existing code?

odysa:
Ok.
Counting 10M triples without context filter only took 685.2 μs
Counting triples within a context requires scanning all entires. It took 274.2ms to scan 5000000 triples.
```
Total Size: 10000000, Time taken: 685.6 μs
Size in context: 5000000, Time taken: 274.2 ms
```

odysa:
My exactCardinality implementation.

```java
protected long exactCardinality(long subj, long pred, long obj, long context) throws IOException {
		TripleIndex index = getBestIndex(subj, pred, obj, context);

		if(context == -1){
			return txnManager.doWith((stack, txn) -> {
				long cardinality = 0;
				for (boolean explicit : new boolean[] { true, false }) {
					int dbi = index.getDB(explicit);
					MDBStat stat = MDBStat.mallocStack(stack);
					mdb_stat(txn, dbi, stat);
					cardinality += (long) stat.ms_entries();
				}
				return cardinality;
			});
		}

		return txnManager.doWith((stack, txn) -> {
			long cardinality = 0;

			MDBVal maxKey = MDBVal.malloc(stack);
			ByteBuffer maxKeyBuf = stack.malloc(TripleStore.MAX_KEY_LENGTH);
			index.getMaxKey(maxKeyBuf, subj, pred, obj, context);
			maxKeyBuf.flip();
			maxKey.mv_data(maxKeyBuf);

			MDBVal keyData = MDBVal.mallocStack(stack);
			ByteBuffer keyBuf = stack.malloc(TripleStore.MAX_KEY_LENGTH);
			MDBVal valueData = MDBVal.mallocStack(stack);
			PointerBuffer pp = stack.mallocPointer(1);

			long[] keyComponents = new long[4];

			for (boolean explicit : new boolean[] { true, false }) {
				keyBuf.clear();
				index.getMinKey(keyBuf, subj, pred, obj, context);
				keyBuf.flip();

				int dbi = index.getDB(explicit);
				keyData.mv_data(keyBuf);

				long cursor = 0;
				try {
					E(mdb_cursor_open(txn, dbi, pp));
					cursor = pp.get(0);

					int rc = mdb_cursor_get(cursor, keyData, valueData, MDB_SET_RANGE);
					while (rc == MDB_SUCCESS && mdb_cmp(txn, dbi, keyData, maxKey) < 0) {
						// decode key to check context
						ByteBuffer keyBytes = keyData.mv_data();
						Varint.readListUnsigned(keyBytes, keyComponents);

						if (keyComponents[3] == context) {
							cardinality++;
						}

						rc = mdb_cursor_get(cursor, keyData, valueData, MDB_NEXT);
					}
				} finally {
					if (cursor != 0) {
						mdb_cursor_close(cursor);
					}
				}
			}

			return cardinality;
		});
	}
```

odysa:
For changeset:

If there are no approved or deprecated changes, it uses a fast direct lookup. Otherwise, it falls back to iterating over matching statements, which remains performant since it avoids LMDB’s lazy evaluation.
```java
@Override
	public long size(final Resource subj, final IRI pred, final Value obj, final Resource... contexts) {
		// Fast path: no approved or deprecated
		if (!changes.hasApproved() && !changes.hasDeprecated()) {
			return derivedFrom.size(subj, pred, obj, contexts);
		}

		// Fallback path: iterate over all matching statements; remains fast due to bypassing LMDB's lazy evaluation
		return getStatements(subj, pred, obj, contexts).stream().count();
	}
```

kenwenzel:
Hi @odysa,

the context size is also already tracked:
https://github.com/eclipse-rdf4j/rdf4j/blob/157627bfdaef4a363317fd3c10f3d2ca2792b32d/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TripleStore.java#L913

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -1033,4 +1035,10 @@ private boolean hasStatement(SailDataset dataset, Resource subj, IRI pred, Value
 		}
 	}
 
+	protected long calculateSize(final boolean includeInferred, final Resource... contexts) throws SailException {
+		try (SailSource branch = branch(IncludeInferred.fromBoolean(includeInferred))) {
+			return branch.dataset(getIsolationLevel()).size(null, null, null, contexts);
+		}
+	}
+
```
hmottestad: Any reason for why we need a new method here?

From a quick glance I saw that there is `protected long sizeInternal(Resource... contexts) throws SailException {` which I wonder if would be a better place for this code?

odysa: This optimization is only for LMDB, so I made 2 different pathes for size.
LMDB gets size directly from `banch.dataset`
Other sails call `getStatementsInternal` to count.

I don't want to accidently break other sails, so leave `sizeInternal` unchanged.

odysa:
Done getting size from tracked context size

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -383,4 +376,15 @@ private boolean isDeprecated(Triple triple, List<Statement> deprecatedStatements
...
 	}
+
+	@Override
+	public long size(final Resource subj, final IRI pred, final Value obj, final Resource... contexts) {
+		// Fast path: no approved or deprecated
+		if (!changes.hasApproved() && !changes.hasDeprecated()) {
+			return derivedFrom.size(subj, pred, obj, contexts);
+		}
+
+		// Fallback path: iterate over all matching statements; remains fast due to bypassing LMDB's lazy evaluation
```
kenwenzel: Comment is a bit misleading - it actually remains fast by using (some kind of) lazy evaluation.
Maybe just use:
// Fallback path: iterate over all matching statements

kenwenzel:
Besides my comment it looks good to me.
@hmottestad What do you think about the changes in the SAIL implementation?

odysa:
>Another aspect is to make sure that this new way of counting follows all the transactional isolation levels that are supported by the LMDB Store.

Lemme add more test cases to it.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TripleStore.java
```diff
@@ -656,6 +658,69 @@ protected void filterUsedIds(Collection<Long> ids) throws IOException {
...
+					dbi = index.getDB(false);
+					mdb_stat(txn.get(), dbi, stat);
+					cardinality += stat.ms_entries();
+				}
+				return cardinality;
+			}
+		}
+
+		// Fast path: if only context is specified, return the size of the given context
+		if (subj == LmdbValue.UNKNOWN_ID && pred == LmdbValue.UNKNOWN_ID
```
kenwenzel: Could be combined with previous if where context variable is checked inside. This would also avoid the slower call to stackGet().

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -10,15 +10,8 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.base;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
kenwenzel: @hmottestad Do we have any rules when using wildcard imports are allowed?

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -383,4 +376,15 @@ private boolean isDeprecated(Triple triple, List<Statement> deprecatedStatements
...
+
+	@Override
+	public long size(final Resource subj, final IRI pred, final Value obj, final Resource... contexts) {
+		// Fast path: no approved or deprecated
+		if (!changes.hasApproved() && !changes.hasDeprecated()) {
+			return derivedFrom.size(subj, pred, obj, contexts);
+		}
+
+		// Fallback path: iterate over all matching statements
+		return getStatements(subj, pred, obj, contexts).stream().count();
```
kenwenzel: @odysa Does this call materialize/resolve all values in LmdbStore?

kenwenzel (commented):
Thank you for the good work. I only have some minor comments.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
odysa: No. It calls `getStatements` here
https://github.com/eclipse-rdf4j/rdf4j/blob/d78001b92d61fc3cfb462e10dee0180f3f85ef67/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java#L927-L941
And eventually calls `getNextElement`
https://github.com/eclipse-rdf4j/rdf4j/blob/d78001b92d61fc3cfb462e10dee0180f3f85ef67/core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbStatementIterator.java#L53-L78

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
hmottestad: We usually don't use wildcard, but it's unfortunately not picked up by the formatter.

hmottestad:
This is still on my list of things to do. I've just been a bit too busy with other things, RDF4J is a side project that I work on next to my regular job.

odysa:
> This is still on my list of things to do. I've just been a bit too busy with other things, RDF4J is a side project that I work on next to my regular job.

Thank you so much. @hmottestad I really appreciate all the work that’s gone into RDF4J, and it's impressive that you're maintaining such a big project in your free time. If there’s anything I can do to help with maintaining the project, such as reviewing PRs or anything else, I’d be happy to contribute.

kenwenzel:
LGTM
@hmottestad Same with me - that's why I asked for maybe using funding opportunities.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -383,4 +383,15 @@ private boolean isDeprecated(Triple triple, List<Statement> deprecatedStatements
...
```
hmottestad: `return getStatements(subj, pred, obj, contexts).stream().count();` should probably we wrapped in a try-with-resource.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDataset.java
```diff
@@ -130,4 +130,9 @@ default Comparator<Value> getComparator() {
 		return null;
 	}
 
+	@Experimental
+	default long size(final Resource subj, final IRI pred, final Value obj, final Resource... contexts) {
+		return getStatements(subj, pred, obj, contexts).stream()
```
hmottestad: I believe that this also needs to be closed.

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceConnection.java
```diff
@@ -1033,4 +1034,21 @@ private boolean hasStatement(SailDataset dataset, Resource subj, IRI pred, Value
...
+	 * contexts. This method reads the size directly from the dataset within the current isolation level.
+	 *
+	 * @param includeInferred whether to include inferred statements in the count
+	 * @param contexts        the RDF contexts (named graphs) to restrict the count to; if none are provided, counts all
+	 *                        contexts
+	 * @return the number of statements in the dataset
+	 * @throws SailException if an error occurs while accessing the Sail store
+	 */
+	@Experimental
+	protected long getSizeFromSnapshot(final boolean includeInferred, final Resource... contexts) throws SailException {
```
hmottestad: Do we need `includeInferred`? I can't see that the other size related methods support differentiating between inferred or explicit.

hmottestad: I kinda wish we could just use the existing sizeInternal(...) method, but I get that it's better to be safe and introduce a separate method for now and then later on we can consider changing up sizeInternal(...).

core/sail/base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
hmottestad: As in:

```
try(var stream = getStatements(subj, pred, obj, contexts).stream()){
  return stream.size();
}
```

kenwenzel:
@hmottestad @odysa Is this ready for merging?

PR #5344, opened 2025-05-31
GH-5310 Make the CostantOptimizer work for aggregates, and be more co…

JervenBolleman:
…nservative in the case of failures

GitHub issue resolved: # 

Briefly describe the changes proposed in this PR:

----
PR Author Checklist (see the [contributor guidelines](https://github.com/eclipse/rdf4j/blob/main/CONTRIBUTING.md) for more details):

 - [x] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [ ] I've added tests for the changes I made
 - [ ] I've applied [code formatting](https://github.com/eclipse/rdf4j/blob/main/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)
 - [ ] I've [squashed](https://rdf4j.org/documentation/developer/squashing) my commits where necessary 
 - [ ] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change

PR #5351, opened 2025-06-12
GH-5310 Correctly evaluate constant grouping constants for empty result sets.

JervenBolleman:
GitHub issue resolved: #GH-5310

The logic to deal with group elements and empty results set was not correct for constant operator values.
This is fixed, plus minor code cleanups.

----

 - [X] I've added tests for the changes I made

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
@@ -319,8 +319,12 @@ public class GroupIterator extends AbstractCloseableIteratorIteration<BindingSet
 			} else {
 				List<AggregateCollector> collectors = makeCollectors(aggregates);
 				List<Predicate<?>> predicates = new ArrayList<>(aggregates.size());
-				for (int i = 0; i < aggregates.size(); i++) {
-					predicates.add(ALWAYS_TRUE_BINDING_SET);
+				for (var ag : aggregates) {
+					if (ag.agg instanceof WildCardCountAggregate) {
+						predicates.add(ALWAYS_TRUE_BINDING_SET);
+					} else {
+						predicates.add(ALWAYS_TRUE_VALUE);
+					}
 				}
 				final Entry entry = new Entry(null, collectors, predicates);
 				entry.addSolution(EmptyBindingSet.getInstance(), aggregates);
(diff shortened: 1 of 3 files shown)


core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/ConstantOptimizerTest.java
```diff
@@ -112,12 +115,49 @@ public void testFunctionOptimization() throws RDF4JException {
 
 	}
 
+	@Test
+	public void testAggregateOptimization() throws RDF4JException {
+		String query = "prefix ex: <ex:>" + "select (max(1) AS ?a) \n " + "where {\n" + "?x a ?z \n"
```
odysa, line 120: Let's try to test all aggregate functions here.
```java
String query = String.join("\n",
		"PREFIX ex: <ex:>",
		"SELECT",
		"  (MAX(1) AS ?a)",
		"  (MIN(1) AS ?b)",
		"  (AVG(1) AS ?c)",
		"  (COUNT(1) AS ?d)",
		"  (COUNT(DISTINCT 1) AS ?e)",
		"  (COUNT(*) AS ?f)",
		"WHERE {",
		"  ?x a ?z ;",
		"  ex:someProperty ?val .",
		"}"
);
.....

assertTrue(bindings.hasBinding("a")); 
assertTrue(bindings.hasBinding("b")); 
assertTrue(bindings.hasBinding("c"));
assertTrue(bindings.hasBinding("d")); 
assertTrue(bindings.hasBinding("e")); 
assertTrue(bindings.hasBinding("f"));
assertEquals(1, ((Literal) bindings.getBinding("a").getValue()).intValue());
assertEquals(1, ((Literal) bindings.getBinding("b").getValue()).intValue());
assertEquals(1, ((Literal) bindings.getBinding("c").getValue()).intValue());
assertEquals(1, ((Literal) bindings.getBinding("d").getValue()).intValue());
assertEquals(1, ((Literal) bindings.getBinding("e").getValue()).intValue());
assertEquals(0, ((Literal) bindings.getBinding("f").getValue()).intValue());
```

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/iterator/GroupIterator.java
```diff
@@ -319,8 +319,12 @@ private List<Entry> emptySolutionSpecialCase(List<AggregatePredicateCollectorSup
 			} else {
 				List<AggregateCollector> collectors = makeCollectors(aggregates);
 				List<Predicate<?>> predicates = new ArrayList<>(aggregates.size());
-				for (int i = 0; i < aggregates.size(); i++) {
-					predicates.add(ALWAYS_TRUE_BINDING_SET);
+				for (var ag : aggregates) {
+					if (ag.agg instanceof WildCardCountAggregate) {
```
odysa, line 323: Looks like there is no better way to test `AggregateFunction<CountCollector, BindingSet>` in Java. Just need to be careful when somebody in the future make a new BindingSet aggregator. (very unlikely?)

JervenBolleman: Yeah, in a newer java we could make that a sealed class and we would get compilation time errors.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/ConstantOptimizerTest.java
JervenBolleman: I Will include that!

JervenBolleman:
> Does this unblock the previous PR #5344 ?

No, previous PR#5344 is broken and is not the right solution.

JervenBolleman:
@odysa @hmottestad this is getting into the specification corners of how aggregates are supposed to behave in the face of the empty result set.
```sparql
SELECT
(COUNT(1) AS ?d)
WHERE {
  ?s ?p ?o
}
```
on an empty result set should I believe return '0' not '1'.

```sparql
SELECT
(MAX(1) AS ?d)
WHERE {
  ?s ?p ?o
}
```
I believe should '1', other stores return an UNDEF/null binding here.

If it should be an UNDEF the logic in the empty case can be simplified. To return a zero in the case of the count operators and a UNDEF for all other cases.

core/queryalgebra/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/ConstantOptimizerTest.java
JervenBolleman: @odysa 
I think this is not right for the empty case, but I am not sure.
"  (COUNT(1) AS ?d)" !=  assertEquals(1, ((Literal) bindings.getBinding("d").getValue()).intValue());

odysa:
@JervenBolleman I agree `count` should return 0 in this case.

odysa (approved):

Issue #5327, opened 2025-05-09
Support RDF 1.2


kenwenzel:
### Problem description

- [ ] Update Turtle parser / writer to support reifying triples and triple terms https://www.w3.org/TR/rdf12-turtle/#reifying-triples
- [ ] Update N-Triples and N-Quads parsers / writers to support triple terms https://www.w3.org/TR/rdf12-n-triples/#triple-terms
- [ ] Update SPARQL parser to support `ReifiedTriple`, `TripleTerm` and `TripleTermData` https://www.w3.org/TR/sparql12-query/#sparqlGrammar
- [ ] ...

### Are you interested in contributing a solution yourself?

None

nguyenm100:
also

https://w3c.github.io/rdf-semantics/spec/

hmottestad:
What is the status on RDF 1.2? We've already implemented a previous version of RDF star, and we don't want to implement something that is going to change in the near future. 

Bigger question is how we should design the backend to support reified statements more efficiently than just converting it to regular statements. We want to extend the data model on the sail side to bring us closer to the performance of property graphs. That's probably a completely different issue though.

nguyenm100:
According to Ora (at KGC this past week) it's largely done and it will be released for public review this summer.. he doesn't anticipate any large changes from this point.

kenwenzel:
I don't know the state of RDF 1.2 but I hope it will be quite stable.

Do you have some specific ideas on improving performance? Maybe we could borrow some ideas from QLever for improving our indexes.

Ostrzyciel:
> Bigger question is how we should design the backend to support reified statements more efficiently than just converting it to regular statements. We want to extend the data model on the sail side to bring us closer to the performance of property graphs. That's probably a completely different issue though.

There are a few ways to approach this, with different performance trade-offs. So far, I've only thought about indexing RDF-star, but RDF 1.2 has more moving parts. I'm also a bit in the dark as to what are the expected usage patterns for the new constructs, because that would inform index design a lot.

If this helps, I would be happy to help in brainstorming this.

For what it's worth, oxigraph has some nice documentation about its indexing strategy and hacks, to for example skip lookups in the value table with some literal types: https://github.com/oxigraph/oxigraph/wiki/Architecture (I think Jena TDB2 does something similar with inlining values... this has both pros and cons). But that's a separate topic :)

kenwenzel:
[...]

PR #5373, opened 2025-07-10
GH-5327 Design Doc for RDF 1.2 RFC

### LINKED ISSUE #5804: RDF4J-STF-M2: Standards
- Implement core support for in parsers, serializers, and compatible SAIL stores (MemoryStore, LmdbStore)

SarahGrand:
GH-5327
Design Doc for RDF 1.2 implementation


odysa:
RFC: RDF1.2 Design
Our team has reviewed the RDF1.2 design (Turtle & MemoryStore) internally and would greatly appreciate feedback from the RDF4J community. 
Your insights and comments are very welcome!
cc @hmottestad @kenwenzel @nguyenm100

kenwenzel:
@SarahGrand Thank you for the good compilation of main changes in RDF 1.2 and implementation aspects in RDF4J.
As far as I can see, the MemoryStore needs only minimal changes and you already have a good plan to extend the Turtle parser and writer.
I would propose to store the language direction in a separate field within `Literal` that may have a similar signature to `Optional<String> getLanguageDirection()`. You could even introduce an enumeration `enum LanguageDirection { LTR, RTL }` and use this instead of a generic `String` datatype.
@hmottestad Would it be fine to integrate everything related to RDF 1.2 directly into the existing packages? Should we drop and/or rename the RDF-star related classes?

hmottestad:
I'm still on vacation. I'll try to take a look at this on Wednesday.

odysa:
@hmottestad @kenwenzel Since this introduces breaking changes, do you think we should release it as a major version, version 6, and create a new branch for it, separate from develop?

nguyenm100:
hey folks, can we get a decision on which branch the PR should be submitted to? @kenwenzel @hmottestad

kenwenzel:
@nguyenm100 @SarahGrand You can use `rdf12` and we decide later if it should be part of RDF4J 5 or an upcoming Version 6.

hmottestad:
@kenwenzel is that branch up to date with the develop branch?

kenwenzel:
@SarahGrand I would propose to intoduce a separate method `getBaseDirection()`. I understand that concatenation might introduce a slight performance cost (although most writers are probably stream-based) but in the end splitting (especially when evaluating SPARQL) could be even more costly.

gnomadix:
now that RDF1.2 has been ratified - is there a plan / timescale to support it ?

kenwenzel:
@gnomadix Yes. We plan to implement this approx. within the next two months.

kenwenzel:
RDF 1.2 support is now fully implemented for MemoryStore and LmdbStore. Support for triple terms in NativeStore is planned.

Issue #5374, opened 2025-07-10
Timeout parameter on queries using underlying LMDB Sail cause the RDF4J to crash


arastogi39:
### Current Behavior

So, of late we started to notice that if we set a timeout parameter in an RDF4J Sparql request, using this API -> https://rdf4j.org/documentation/reference/rest-api/#tag/SPARQL/paths/~1repositories~1%7BrepositoryID%7D/get and search "timeout", this causes the rdf4j server to crash.

We are using the LMDB sail as the backend for the repository, which we created, and to fire this query. Attaching the err_pid file dump as well.

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000010133fd4c, pid=62774, tid=58371
#
# JRE version: OpenJDK Runtime Environment Temurin-17.0.13+11 (17.0.13+11) (build 17.0.13+11)
# Java VM: OpenJDK 64-Bit Server VM Temurin-17.0.13+11 (17.0.13+11, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# C  [liblwjgl_lmdb.dylib+0xbd4c]
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
#   https://github.com/adoptium/adoptium-support/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

---------------  T H R E A D  ---------------

Current thread (0x00000001292e2400):  JavaThread "repositories/anujtestrepo" daemon [_thread_in_native, id=58371, stack(0x000000017754c000,0x000000017774f000)]

Stack: [0x000000017754c000,0x000000017774f000],  sp=0x000000017774ca10,  free space=2050k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [liblwjgl_lmdb.dylib+0xbd4c]
J 13183  org.lwjgl.util.lmdb.LMDB.nmdb_cursor_get(JJJI)I (0 bytes) @ 0x00000001400b1548 [0x00000001400b14c0+0x0000000000000088]
j  org.eclipse.rdf4j.sail.lmdb.LmdbRecordIterator.next()[J+229
[...]

PR #5376, opened 2025-07-16
GH-5374 check if the LmdbRecordIterator is closed after getting the read lock

hmottestad:
GitHub issue resolved: #5374 

----


core/sail/base/src/main/java/org/eclipse/rdf4j/sail/TripleSourceIterationWrapper.java
```diff
@@ -48,9 +48,6 @@ public boolean hasNext() throws QueryEvaluationException {
 		} catch (QueryEvaluationException e) {
 			throw e;
 		} catch (Exception e) {
-			if (e instanceof InterruptedException) {
```
hmottestad, line 51: This is dead code, as InterruptedException is a checked exception that can't be thrown here.

hmottestad:
The close method looks like this:

```
private void closeInternal(boolean maybeCalledAsync) {
	if (!closed) {
		long stamp;
		if (maybeCalledAsync && ownerThread != Thread.currentThread()) {
			stamp = txnLock.writeLock();
		} else {
			stamp = 0;
		}
		try {
			if (!closed) {
				...
			}
		} finally {
			closed = true;
			if (stamp != 0) {
				txnLock.unlockWrite(stamp);
			}
		}
	}
}
```

@kenwenzel Is the cost of acquiring the write lock so large that we can't just always acquire it when closing? 

We also have the [ReadPrefReadWriteLockManager](https://github.com/jetztgradnet/rdf4j/blob/main/core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/ReadPrefReadWriteLockManager.java) and the [AbstractReadWriteLockManager](https://github.com/jetztgradnet/rdf4j/blob/main/core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/AbstractReadWriteLockManager.java#L29) that have higher throughput for read locks than StampedLock and also support lock tracking so that lost locks can be detected. Though I don't think the lock tracking is really needed in this code.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
kenwenzel, line 1: Superfluous import.

Issue #5358, opened 2025-06-26
Sparql federated services submits incorrect query with a VALUES clause


aschwarte10:
### Current Behavior

We found that the query engine produces incorrect queries for a SPARQL 1.1 federated service, when the service clause contains a VALUES clause.

Though we identified it in a different context, it can be reproduced with the following example query:

```
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX sh: <http://www.w3.org/ns/shacl#>
SELECT * WHERE {
  SERVICE <https://query.wikidata.org/sparql> {
    VALUES ?resource {
     <http://www.wikidata.org/entity/Q7455975> 
    }
    ?resource <http://www.wikidata.org/prop/direct/P856> ?website
 }
}
```

The actual submitted query to the wikidata SPARQL endpoint on HTTP level is replacing the `?resource` variable in the VALUES clause, i.e. something like

```
...
VALUES <http://www.wikidata.org/entity/Q7455975> {
   <http://www.wikidata.org/entity/Q7455975> 
  }
...
```

A workaround is to use `BIND(<http://www.wikidata.org/entity/Q7455975> AS ?resource)` in case of single value. Note that in our real use-case we have generated queries where we rely on the use of VALUES inside the service clause.

Hint: the problem is in `org.eclipse.rdf4j.query.algebra.evaluation.impl.evaluationsteps.ServiceQueryEvaluationStep#evaluate()` where `getBoundVariables` will retrieve a binding for `?resource`. This binding gets inserted into the query string at execution time from the SPARQLRepository executing it.

### Expected Behavior

I expect to be able to use VALUES (also with single bindings) in a SERVICE

### Steps To Reproduce

For example using the following snippet

```
Repository repo = new SailRepository(new MemoryStore());
  try (var conn = repo.getConnection()) {

      var tq = conn.prepareTupleQuery("""
                  PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
                  PREFIX sh: <http://www.w3.org/ns/shacl#>
                  SELECT * WHERE {
                    SERVICE <https://query.wikidata.org/sparql> {
                      VALUES ?resource {
                       <http://www.wikidata.org/entity/Q7455975>
                      }
                      ?resource <http://www.wikidata.org/prop/direct/P856> ?website
                   }
                  }
              """);

      try (var tqr = tq.evaluate()) {
[...]

PR #5408, opened 2025-08-20
GH-5358: fix query evaluation errors when using VALUES clause in SERVICE

aschwarte10:
GitHub issue resolved: #5358

The BindingSetAssingmentInlinerOptimizer attempts to inject a single value from VALUES clauses into matching variables.

For SERVICE causes this is problematic (when they later get evaluated through the SparqlFederatedService):

the SPARQLFederatedService computes the set of bound variables and applies logic. Here it is important, that binding set assignments are not inlined. Otherwise, a query evaluation exception is thrown due to a malformed query.

The actual query at evaluation time (in case such inlining happened) would contain something like

VALUES <http://www.example.com/MyResource> {
   <http://www.example.com/MyResource>
}

i.e. the variable name of the VALUES clause incorrectly replaced.

This change fixes it by not applying inlining insider SERVICE nodes.

Various unit tests on different levels have been added, which I used to get closer to the problem.

----

### DIFF
--- a/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
+++ b/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Eclipse RDF4J contributors.
+ * Copyright (c) 2025 Eclipse RDF4J contributors.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Distribution License v1.0
--- a/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
+++ b/compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
@@ -1,0 +1,43 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *******************************************************************************/
+package org.eclipse.rdf4j.repository.sparql.federation;
+
+import org.eclipse.rdf4j.repository.Repository;
+import org.eclipse.rdf4j.repository.sail.SailRepository;
+import org.eclipse.rdf4j.sail.memory.MemoryStore;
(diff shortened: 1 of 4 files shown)


compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.rdf4j.repository.Repository;
+import org.eclipse.rdf4j.repository.sail.SailRepository;
+import org.eclipse.rdf4j.sail.memory.MemoryStore;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+public class SparqlFederatedServiceIntegrationTest {
+
+	@Test
+	@Disabled("manual test to demonstrate the original issue of GH-5358")
```
aschwarte10, line 22: if prefered, I can also delete this test class. It is just one example to quickly run check whether the fix is working. This one was previously throwing an exception

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
```diff
@@ -0,0 +1,43 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Eclipse RDF4J contributors.
```
JervenBolleman: Is this year correct?

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
JervenBolleman: Disabled is fine, spinning up our own SPARQL endpoint and using that for a test would be even nicer.

JervenBolleman (changes requested):
Just the copyright year nit, otherwise looks good to me.

JervenBolleman:
@aschwarte10 the dash check failing is a spurious error by the service and is not something you need to worry about here.

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
aschwarte10: Yep, would have wished for that. Had also a quick look, but my bandwidth is too limited

compliance/repository/src/test/java/org/eclipse/rdf4j/repository/sparql/federation/SparqlFederatedServiceIntegrationTest.java
aschwarte10: no .. wrong copy and paste. Thanks 👍 

Fixed now

JervenBolleman:
LGTM I will merge if no objections by Monday

Issue #5433, opened 2025-09-16
Lock contention slows down LmdbStore for parallel read transactions


Ostrzyciel:
### Current Behavior

I set up an LmdbStore and had it execute read-only SPARQL SELECT queries from multiple clients in parallel. There were no writes.

I have noticed that as the number of clients increased, queries per second dropped dramatically, and CPU usage kept increasing. On a 12 core machine, at 1 client, I saw ~60 queries per second per core (load: ~1), while at 16 clients I saw ~2.5 queries per second per core (load =12, so 100% CPU usage). Note that this was done over HTTP with simple queries, so I would expect this to not scale so poorly...

I profiled RDF4J under this workload (16 clients) and got this:

<img width="2439" height="774" alt="Image" src="https://github.com/user-attachments/assets/99cf8bbe-f230-4425-965f-f3d36a19cc1a" />

This was done with a regular sampling profiler, which is vulnerable to CPU skid. The `setInternalID` is just a setter, so its presence in the flame graph is a skid artifact. The real culprit is `StampedLock.readLock` in `ValueStore.getLazyValue(long)`.

I had a look at this lock. It is optimized to allow for obtaining optimistic read locks without synchronizing state of the lock over all threads (via a CAS), which is a super costly operation (you have to go through all the caches, all cores, and so on...). This cost can be avoided with the `tryOptimisticRead` method, which does not require any synchronization, but instead requires us to validate the stamp at the end of the read.

See https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/StampedLock.html and https://stackoverflow.com/a/43741577

However, the current implementation uses regular read locks, not optimistic read locks. This causes a CAS per each RDF term we read from the DB.

The other issue is that we acquire this lock 3–4 times per statement, which is pretty crazy... We could instead acquire the lock once and perform all 3–4 operations in one batch. If we are lucky, it may even get auto-vectorized.

### Expected Behavior

I would expect the parallel query performance to not be so bad.

### Version

5.1.5

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

I'm working on a solution for this.

kenwenzel:
@Ostrzyciel We can also try to replace it with https://github.com/eclipse-rdf4j/rdf4j/blob/main/core/sail/api/src/main/java/org/eclipse/rdf4j/common/concurrent/locks/ReadWriteLockManager.java which is also used in other places.
@hmottestad also suggested this some time ago.

[...]

PR #5434, opened 2025-09-16
GH-5433 Fix read lock contention in LmdbStore

Ostrzyciel:
GitHub issue resolved: #5433

### (old)

Replaces the current regular read locks in LMDB ValueStore reads with optimistic read locks, which lets us avoid a CAS. This makes a huge difference for parallel queries, and I saw our internal benchmark report much better results with this change.

Alternatively, this could use `StampedLockManager`, but this would require allocating an object for each time we acquire a lock... that's a lot of overhead for very hot methods such as these. The solution with `long`s has basically no overhead.

The performance in these methods could probably still be better (see the issue for profiler charts) – but this does not seem to be related to locks anymore, so I will look into that separately.

Initially I also lumped this together with a similar change in `LmdbRecordIterator`, but that turned out to be a harder case to solve, so I'm deferring it for a future PR.

### Current

Following a discussion with @kenwenzel, it turns out the lock in `getLazyValue` was not needed after all, so I simply removed it.

----

### DIFF
--- a/core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/RecordIteratorBenchmark.java
+++ b/core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/RecordIteratorBenchmark.java
@@ -28,6 +28,7 @@ import org.openjdk.jmh.infra.Blackhole;
 @Warmup(iterations = 5)
 @BenchmarkMode({ Mode.AverageTime })
 @Fork(value = 4, jvmArgs = { "-Xms1G", "-Xmx1G" })
+//@Fork(value = 1, jvmArgs = {"-Xms1G", "-Xmx1G", "-XX:StartFlightRecording=jdk.CPUTimeSample#enabled=true,filename=profile.jfr,method-profiling=max","-XX:FlightRecorderOptions=stackdepth=1024", "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"})
 @Threads(value = 8)
 @Measurement(iterations = 10)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)


core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
```diff
@@ -126,73 +126,83 @@ class LmdbRecordIterator implements RecordIterator {
...
-			if (closed) {
-				log.debug("Calling next() on an LmdbRecordIterator that is already closed, returning null");
-				return null;
+		long stamp = txnLock.tryOptimisticRead();
+		long[] value = nextInternal();
+		if (!txnLock.validate(stamp)) {
+			stamp = txnLock.readLock();
+			try {
+				// Retry read
+				value = nextInternal();
```
hmottestad: Don't we risk that calling nextInternal() twice will skip an element if the write happened right after the first call to nextInternal() and right before validating the lock?

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
```diff
@@ -126,73 +126,83 @@ class LmdbRecordIterator implements RecordIterator {
 
 	@Override
 	public long[] next() {
-		long stamp = txnLock.readLock();
-		try {
-			if (closed) {
-				log.debug("Calling next() on an LmdbRecordIterator that is already closed, returning null");
-				return null;
+		long stamp = txnLock.tryOptimisticRead();
+		long[] value = nextInternal();
```
hmottestad: @kenwenzel Is the read/write locking needed as a protection for LMDB, or is it safe to concurrently do all the operations in `nextInternal()` while there is a write?

hmottestad:
Btw. The `ReadPrefReadWriteLockManager` implements a faster read lock by using two LongAdder objects. The implementation is inspired from https://concurrencyfreaks.blogspot.com/2013/09/combining-stampedlock-and-longadder-to.html

You might want to try something similar to compare performance against optimistic locking.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
Ostrzyciel: Agh, yeah, you are right, that will be an issue. Let me have a look at what I can do about this.

Ostrzyciel: I think I was too trigger-happy with an optimistic lock here. We should try something else here (e.g., the `ReadPrefReadWriteLockManager` you proposed), but IMO this is for a separate PR.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
kenwenzel: @hmottestad It's only required to protect txnRef itself to not accidentally close read transactions during a read operation. As far as I remember this is required to correctly support autogrow which needs to close all transactions.

kenwenzel:
@Ostrzyciel I think we can safely remove the `revisionLock` from `getLazyValue` at all. I think it was just a safety measure from my side.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
hmottestad: So if autogrow happens in the middle of a read operation that doesn't hold the lock, what will happens with the read operation?

kenwenzel:
@Ostrzyciel It is no problem at this point. LazyRevision is required to track uninitialized values of a specific Revision that might have been deleted in the meantime. For those values the ID to value mapping is preserved until the specific LazyRevision is garbage collected.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
kenwenzel: It will fail as the transaction will be closed.

kenwenzel (approved):
LGTM - thank you!

hmottestad (approved):

kenwenzel:
> Looks good to me too.
> 
> I wonder what the hit rate on the valueCache is with multiple threads. Might be worth using VarHandle and acquire/release to see values cached by other threads.

Good question - we could investigate this. Should we create an issue for this?

kenwenzel:
@hmottestad @Ostrzyciel I've collected the relevant commits here. I hope this is OK with you?

Issue #5118, opened 2024-08-27
NotifyingSailConnection should distinguish between explicit and inferred statements


kenwenzel:
### Problem description

When listening to changes on `NotifyingSailConnection` the consumer is unable to distuingish between explicit and inferred statements.
In scenarios where undo/redo or audit logging should be implemented it would be very helpful if this distinction could be drawn.

### Preferred solution

The interface `SailConnectionListener` could be extended to include an `inferred` flag as parameter for the notification methods.
The default implementation of those methods could just delegate to the existing ones.

java```
public interface SailConnectionListener {
  void statementAdded(Statement st);
  void statementRemoved(Statement st);
  
  default void statementAdded(Statement st, boolean inferred) {
    statementAdded(st);
  }

  void statementRemoved(Statement st, boolean inferred) {
    statementRemoved(st);
  }
}
```

### Are you interested in contributing a solution yourself?

None

### Alternatives you've considered

The alternative would be to directly include the `inferred` flag within the statement interface.
This is the approach that we have used for KOMMA:
https://github.com/komma/komma/blob/master/bundles/core/net.enilink.komma.core/src/main/java/net/enilink/komma/core/IStatement.java

### Anything else?

_No response_

hmottestad:
Sounds good! We’ll have to be a bit careful that we don’t make any breaking changes. 

I think the reason why Statements don’t have an inferred flag is that users can’t insert their own inferred statements. We could introduce an interface StatementWithInferredStatus (or something a bit cleaner) and then we can use instanceOf to check if the underlying store supports that feature (eg. produces statements that implement that interface).

kenwenzel:
As a first step we could just extend the listener as sketched above. This would save users from false assumptions about the support of `inferred` states.

At a later point:
Instead of `StatementWithInferredStatus` I would prefer to simply extend the `Statement` interface with `isInferred()` and add a default method that just returns `false`.
Within KOMMA I have decided to not include the `inferred` flag for `equals()` and `hashCode()` but I am not sure if this is a good/correct decision.

hmottestad:
To extend the listener we would have to add two new methods in order to keep is backwards compatible, and both of those methods need a default implementation on the interface side. Those implementations would probably just pass down to the original methods, but for users who implement the methods there would not be a guarantee that the sail actually calls the new methods.

kenwenzel:
[...]

PR #5445, opened 2025-09-22
Extend methods of SailConnectionListener with inferred argument

kenwenzel:
GitHub issue resolved: #5118 

Extend methods of SailConnectionListener with inferred argument to distinguish between explicit and inferred statements.

----

### DIFF
--- a/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractNotifyingSailConnection.java
+++ b/core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractNotifyingSailConnection.java
@@ -48,16 +48,26 @@ public abstract class AbstractNotifyingSailConnection extends AbstractSailConnec
 		return !listeners.isEmpty();
 	}
 
+	@Deprecated(since = "5.2.0", forRemoval = true)
 	protected void notifyStatementAdded(Statement st) {
+		notifyStatementAdded(st, false);
+	}
+
+	protected void notifyStatementAdded(Statement st, boolean inferred) {
 		for (SailConnectionListener listener : listeners) {
-			listener.statementAdded(st);
+			listener.statementAdded(st, inferred);
 		}
 
 	}
 
+	@Deprecated(since = "5.2.0", forRemoval = true)
 	protected void notifyStatementRemoved(Statement st) {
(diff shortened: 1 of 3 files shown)


kenwenzel:
@hmottestad This should be API-compatible and therefore can be merged into main.

core/sail/api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractNotifyingSailConnection.java
```diff
@@ -48,16 +48,16 @@ protected boolean hasConnectionListeners() {
 		return !listeners.isEmpty();
 	}
 
-	protected void notifyStatementAdded(Statement st) {
+	protected void notifyStatementAdded(Statement st, boolean inferred) {
```
hmottestad, line 56: This method is protected, so it can be overridden by any subclass. That makes this an API level change, since the existing method signature is removed. 

You can try to keep the existing method and have it call the new method, probably best to assume that all statements are explicit for backwards compatibility.

kenwenzel: Thank you - this works.
Can we also merge this directly to 'main'?

hmottestad: This is more of a feature than a bug and needs to go to develop.

kenwenzel: @hmottestad I want to add deprecation annotations to methods without the inferred parameter. Should I declare since=5.2.0 or another version?

hmottestad: Yeah. 5.2.0.

kenwenzel: OK 👍 
Do you merge the PR?

hmottestad:
Feel free to merge :)

Issue #5291, opened 2025-04-11
LuceneSail fsyncs after each TX, leading to very poor performance


Ostrzyciel:
### Current Behavior

LuceneSail runs the `commit()` method on the `IndexWriter` every time a transaction is committed:

https://github.com/eclipse-rdf4j/rdf4j/blob/782ffd84647a851607913dda7c8707963e1a5cd6/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneIndex.java#L704-L714

This actually translates to a bit more than a transaction commit on the side of Lucene – it results in a whole lot of fsyncs. This is documented as "This may be a costly operation, so you should test the cost in your application and do it only when really necessary" in Lucene: https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/index/IndexWriter.html#commit--

I came across this while investigating very poor insert performance when using `LuceneSail`. When *any* writes are made to Lucene, even super-trivial, performance of the entire app drops by more than tenfold.

Here is the flame graph of CPU time for RDF4J Server running this workload + a few other repositories in parallel:

![Image](https://github.com/user-attachments/assets/62929535-81e4-4c41-8200-c590d06a3ffe)

### Expected Behavior

`NativeStore` has a parameter called `forceSync` that decides whether the Store should fsync after each transaction commit. It's off by default. I would suggest to add an identical parameter to `LuceneSail`. My suggestion is to also make it false by default, to save the poor users some hair. I imagine there are relatively few use cases that actually require an fsync to a text index after each commit. But, I'm also fine with it being true by default, this is up to the maintainers, of course.

When `forceSync` is enabled we would call `commit()` as done currently. Otherwise, we would call `flush()` which does the same thing (move written data from the memory to the filesystem), except fsyncing the files.

### Steps To Reproduce

Set up a `LuceneSail` and do some writes to it, with triples involving literals.

### Version

latest main (5.3.0?)

### Are you interested in contributing a solution yourself?

Yes

### Anything else?

I will submit a PR to resolve it.

Ostrzyciel:
Ah... I've just discovered that `flush()` doesn't make the changes visible to readers. That's a big problem.

I'm not so sure anymore if I know how to fix this. I will investigate further.

Ostrzyciel:
See what the wise people of the internet say on this matter: https://stackoverflow.com/a/34116569

[...]

PR #5446, opened 2025-09-22
GH-5291 Add asynchronous fsync to LuceneSail

Ostrzyciel:
GitHub issue resolved: #5291

As described in the issue, the current setup with an fsync after each transaction is very safe, but also a huge bottleneck when dealing with many small transactions. This PR introduces an option that allows for asynchronous fsyncs in the background, on a fixed interval. If there is nothing to sync, it does nothing.

I tested this with the original workload with which I found the issue. When I set `fsyncInterval` to 5000 ms, it went from ~10–12 TX/s to ~100–150 TX/s, over an HTTP connection. That's basically the same as when I tried removing the fsync entirely. Great :)

----

### DIFF
--- a/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
+++ b/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.lucene.impl;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FilterDirectory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Wrapper around a Lucene Directory that batches sync and metadata sync calls to be executed at a fixed interval.
+ *
+ * @author Piotr Sowiński
+ */
+class DelayedSyncDirectoryWrapper extends FilterDirectory {
+
+	final private Logger logger = LoggerFactory.getLogger(getClass());
+
+	final private ScheduledExecutorService scheduler;
+
+	final private AtomicBoolean needsMetadataSync = new AtomicBoolean(false);
+
+	final private AtomicReference<IOException> lastSyncException = new AtomicReference<>(null);
+
+	final private HashSet<String> pendingSyncs = new HashSet<>();
+
+	final private int maxPendingSyncs;
+
+	/**
+	 * Creates a new instance of LuceneDirectoryWrapper.
+	 *
+	 * @param in              the underlying directory
+	 * @param fsyncInterval   the interval in milliseconds writes after which a fsync is performed
+	 * @param maxPendingSyncs the maximum number of pending syncs to accumulate before forcing a sync
+	 */
+	DelayedSyncDirectoryWrapper(Directory in, long fsyncInterval, int maxPendingSyncs) {
+		super(in);
+		assert fsyncInterval > 0;
+		assert maxPendingSyncs > 0;
+		this.maxPendingSyncs = maxPendingSyncs;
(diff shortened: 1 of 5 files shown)


hmottestad:
I don't think that the close() method on the directory in the LuceneIndex class is ever called.

Ostrzyciel:
> I don't think that the close() method on the directory in the LuceneIndex class is ever called.

@hmottestad oops, you are right! Fixed that and added a test to make sure it happens.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
```diff
@@ -0,0 +1,103 @@
...
+				super.sync(toSync);
+			} catch (IOException e) {
+				logger.error("IO error during a periodic sync of Lucene index files", e);
+			}
+		}
+		if (this.needsMetadataSync.getAndSet(false)) {
+			try {
+				super.syncMetaData();
+			} catch (IOException e) {
+				logger.error("IO error during a periodic sync of Lucene index metadata", e);
```
hmottestad, line 88: I'm a bit worried that if for some reason there is a persistent issue, then we may end up logging continuously but never actually throwing an exception. 

What would usually happen if an IO exception was thrown (with the original code)? Would it bring down the entire application or just a particular transaction?

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	@Override
+	public void close() throws IOException {
+		// Finish the current sync task, if in progress and then shut down
+		scheduler.shutdown();
+		// Do a final sync of any remaining files
+		doSync();
+		super.close();
+	}
```
hmottestad, line 130: Can you structure this as nested try/finally?

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneIndex.java
```diff
@@ -390,6 +399,9 @@ public void shutDown() throws IOException {
 						if (toCloseIndexWriter != null) {
 							toCloseIndexWriter.close();
 						}
+						// Close the directory -- if asynchronous fsync is used, this will clean
+						// up the scheduler thread too.
+						directory.close();
```
hmottestad: Same here, should be nested try/finally.

hmottestad:
Thanks for the good fix. I think this will be a good solution overall, just some small things I want to be sure are robust.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
```diff
@@ -0,0 +1,103 @@
...
+			} catch (IOException e) {
+				logger.error("IO error during a periodic sync of Lucene index metadata", e);
+			}
+		}
+	}
+
+	@Override
+	public void sync(Collection<String> names) throws IOException {
+		synchronized (pendingSyncs) {
+			pendingSyncs.addAll(names);
```
hmottestad, line 103: How big is this likely to grow? Should we have a hard limit (possibly configurable) so that we don't run out of memory before we sync?

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
Ostrzyciel: This would result in a transaction rollback: https://github.com/eclipse-rdf4j/rdf4j/blob/18f9a56f7e46fa8e923cd2069bd9c5799ee6c461/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSailConnection.java#L266-L269

We cannot do the same thing 1:1 with asynchronous fsyncs, because we don't wait for the result of the fsync. The next best thing we can do is to throw an exception on the next transaction.

I've added a bit of code for that, along with a test.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
Ostrzyciel: From what I can tell, there is no limit on this, it depends on Lucene index size. I added a configurable limit for this, set to 5000 files by default – should be good enough. There is also a test for this.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/DelayedSyncDirectoryWrapper.java
Ostrzyciel: Done.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneIndex.java
Ostrzyciel: Done.

Ostrzyciel:
> @Ostrzyciel if it gets too complicated I'm open for adding a system property for disabling fsync in the lucene sail, if you believe that's a viable option for you.

Yeah, it would work for my case fine. I'm wrapping up the implementation of this right now, running the end-to-end tests.

Issue #5447, opened 2025-09-22
Improve performance of queries in the LMDB Store


hmottestad:
Improve performance of queries in the LMDB Store.

kenwenzel:
The parallel processing is impressive as it shows almost no overhead now. Cool.

PR #5448, opened 2025-09-22
GH-5447 LMDB Store query performance improvements

hmottestad:
GitHub issue resolved: #5447 

Currently a collection of various performance improvements. Still WIP.

----

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,4 +1,4 @@
-AGENTS.md
+# AGENTS.md
 
 Welcome, AI Agent! Your persistence, curiosity, and craftsmanship make a difference. Take your time, work methodically, validate thoroughly, and iterate. This repository is large and tests can take time — that’s expected and supported.
 
@@ -11,7 +11,7 @@ You need to read the entire AGENTS.md file and follow all instructions exactly. 
 ## Read‑Me‑Now: Zero‑Exception Test‑First Rule (Stricter)
 
 **You may not touch production code until a smallest‑scope failing automated test exists inside this repo and you have captured its report snippet.**
-No exceptions. A user‑provided stack trace or “obvious” contract violation is **not** a substitute for a failing test in the repository.
+A user‑provided stack trace or “obvious” contract violation is **not** a substitute for a failing test in the repository.
 
 **Auto‑stop:** If you realize you patched production before creating the failing test, **stop**, revert the patch, and resume from “Reproduce first”.
 
@@ -23,6 +23,32 @@ No exceptions. A user‑provided stack trace or “obvious” contract violation
 
 It is illegal to `-am` when running tests!
 It is illegal to `-q` when running tests!
+
+> **Clarification & carve‑out:** For **strictly behavior‑neutral refactors** that are already **fully exercised by existing tests**, you may use the **Small‑Change Fast Path** defined below. In that case you must capture **pre‑change passing evidence** at the smallest scope that hits the code you’re about to edit, then show **post‑change passing evidence** from the **same test selection**.
+> **No exceptions for any behavior‑changing change** — for those, you must follow Full TDD (Routine A).
+
+---
+
+## Two Routines: Choose Your Path
+
+**Default:** Routine A — Full TDD
+**Optional:** Routine B — Small‑Change Fast Path (tiny, low‑risk changes; see gates below)
+
+**Decision quickstart**
+
+1. **Is new externally observable behavior required?**
+   → **Yes:** Routine A (Full TDD). Add the smallest failing test first.
+   → **No:** continue.
+
+2. **Does a failing test already exist in this repo that pinpoints the issue?**
+   → **Yes:** Routine B (Bugfix with existing failing test).
+   → **No:** continue.
+
+3. **Is the edit strictly behavior‑neutral and limited in scope with clear existing tests hitting the code?**
+   → **Yes:** Routine B (Refactor/micro‑perf).
+   → **No or unsure:** Routine A.
+
+**When in doubt, choose Routine A (Full TDD).** Ambiguity is risk; tests are insurance.
 
 ---
 
@@ -39,6 +65,7 @@ It is illegal to `-q` when running tests!
     * Only necessary files changed; headers correct for new files.
     * Clear final summary: what changed, why, where, how verified, next steps.
     * **Evidence present:** failing test output (pre‑fix) and passing output (post‑fix) are both shown.
+      *For refactor Fast Path, pre‑ and post‑change **passing** snippets from the same selection are required.*
 
 ### No Monkey‑Patching or Band‑Aid Fixes (Non‑Negotiable)
 
@@ -71,21 +98,24 @@ Review bar and enforcement
 
 * Treat this policy as a blocking requirement. Changes that resemble workarounds will be rejected.
 * Your final handoff must demonstrate: failing test before the fix, explanation of the root cause, minimal fix at source, and passing targeted tests after.
+  *For refactor Fast Path, show pre‑/post‑green evidence and prove the tests hit the edited code.*
 
 ---
 
-## Enforcement & Auto‑Fail Triggers 
+## Enforcement & Auto‑Fail Triggers
 
 Your run is **invalid** and must be restarted from “Reproduce first” if any of the following occur:
 
-* You modify production code before adding and running the smallest failing test in this repo.
+* You modify production code before adding and running the smallest failing test in this repo **for behavior‑changing work**.
 * You proceed without pasting a surefire/failsafe report snippet from `target/*-reports/`.
 * Your plan does not have **exactly one** `in_progress` step.
 * You run tests using `-am` or `-q`.
 * You treat a narrative failure description or external stack trace as equivalent to an in‑repo failing test.
+* **Fast Path specific:** you cannot demonstrate that existing tests exercise the edited code (see “Hit Proof” below), or you fail to capture both pre‑ and post‑change **matching** evidence snippets.
 
 **Recovery procedure:**
 Update the plan (`in_progress: create failing test`), post a preamble, create the failing test, run it, capture the report snippet, then resume.
+For Fast Path refactors: if hit proof or scope gates fail, **switch to Full TDD** and add the smallest failing test.
 
 ---
 
@@ -114,6 +144,16 @@ Report: <module>/target/surefire-reports/<file>.txt
 Snippet:
 <copy 10–30 lines capturing the failure or success summary>
 ```
+
+**Fast Path additions**
+
+* **Pre‑green (refactor only):** capture a pre‑change passing snippet from the **most specific** test selection that hits your code (ideally a class or method).
+* **Hit Proof:** show that the selected tests exercise the edited class/method. Acceptable forms:
+
+    * Reference to an existing test class/method that directly calls the edited class/method, plus a short `rg -n` snippet showing the call site; **or**
+    * A Surefire/Failsafe output line containing the edited test class/method names; **or**
+    * A stacktrace/snippet from a deliberately failing assertion added **in a new dedicated test** (not production) that proves the path is executed (remove after proof).
+* **Post‑green:** after the patch, re‑run the **same selection** and capture a passing snippet.
 
 ---
 
@@ -149,9 +189,10 @@ Plan
 
 **Rule of thumb**
 
-* ✅ Use `-am` **only** for compile/verify with tests skipped (e.g. `-Pquick`).:
+* ✅ Use `-am` **only** for compile/verify with tests skipped (e.g. `-Pquick`):
 
     * `mvn -o -pl <module> -am -Pquick install`
+
 * ❌ Do **not** use `-am` with `verify` when tests are enabled.
 
 **Two-step pattern (fast + safe)**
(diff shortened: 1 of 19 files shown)


AGENTS.md
```diff
@@ -607,7 +747,7 @@ Do **not** modify existing headers’ years.
 
     * `-Dtest=ClassName`
     * `-Dtest=ClassName#method`
-    * `-Dit.test=ITClass#method`
+    * `-Dit.test=ITClassName[#method]`
```
hmottestad: Revert.

core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/ArrayBindingSet.java
```diff
@@ -403,6 +403,23 @@ public void addAll(ArrayBindingSet other) {
 
 	}
 
+//	@Override
+//	public boolean equals(Object other){
```
hmottestad, line 407: TODO: During GRPUP BY I noticed that ArrayBindingSet could have use for it's own equals method, which can be faster. 

We should also look into if hashcode is cached in the super class.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
```diff
@@ -347,19 +354,23 @@ protected void handleClose() throws SailException {
 	 */
 	CloseableIteration<? extends Statement> createStatementIterator(
 			Txn txn, Resource subj, IRI pred, Value obj, boolean explicit, Resource... contexts) throws IOException {
+		if (!explicit && !mayHaveInferred) {
+			// there are no inferred statements and the iterator should only return inferred statements
+			return EMPTY_ITERATION;
```
hmottestad: This makes a very big difference when not using inferencing. We can very quickly return an empty iterator without any interaction with the underlying LMDB implementation. I have had a similar optimisation for the Memory Store for a while now and it's been working very well.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/IndexKeyWriters.java
```diff
@@ -0,0 +1,403 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *******************************************************************************/
+package org.eclipse.rdf4j.sail.lmdb;
+
+import java.nio.ByteBuffer;
+
+final class IndexKeyWriters {
```
hmottestad, line 15: This class provides hard coded variants of key generation and shouldMatch logic. It's faster to pick a specific method once, instead of having a single method that supports all key combinations using a loop.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
```diff
@@ -130,6 +136,15 @@ class LmdbRecordIterator implements RecordIterator {
...
 	}
 
+	public GroupMatcher getGroupMatcher() {
+		if (groupMatcher != null)
+			return groupMatcher;
+		if (matchValues) {
+			this.groupMatcher = index.createMatcher(subj, pred, obj, context);
+		}
+		return groupMatcher;
+	}
```
hmottestad: Lazy group matcher. Not sure this makes much of a difference.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbStatementIterator.java
```diff
@@ -23,7 +24,7 @@
  * A statement iterator that wraps a RecordIterator containing statement records and translates these records to
  * {@link Statement} objects.
  */
-class LmdbStatementIterator extends LookAheadIteration<Statement> {
+class LmdbStatementIterator extends AbstractCloseableIteration<Statement> {
```
hmottestad, line 27: From experience we should try to avoid having these very low level iterators use inheritance. Object orientation, inheritance and polymorphism can be unexpectedly expensive. But not sure if it makes that much of a difference here as it did for the MemoryStore when I changed it there.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/IndexKeyWriters.java
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	static MatcherFactory matcherFactory(String fieldSeq) {
+		switch (fieldSeq) {
+		case "spoc":
+			return IndexKeyWriters::spocShouldMatch;
+		case "spco":
+			return IndexKeyWriters::spcoShouldMatch;
+		case "sopc":
+			return IndexKeyWriters::sopcShouldMatch;
```
hmottestad, line 92: We might benefit from fieldSeq being an enum instead of a string. Not sure if it will make that much of a performance difference though, but might be cleaner code.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -89,19 +101,102 @@ private Varint() {
...
+			try {
+				bb.put((byte) 0xFF); // header for 8 payload bytes
+				bb.putLong(Long.MAX_VALUE); // 7F FF FF FF FF FF FF FF
+			} finally {
+				if (prev != ByteOrder.BIG_ENDIAN) {
+					bb.order(prev);
+				}
+			}
+			return;
+		}
```
hmottestad, line 115: I've attempted to optimize both the writeUnsigned and readUnsigned to reduce the usage of loops and also to read/write multiple bytes at a time. I did run into issues with endianness that I'm not sure if I'm managing correctly. The original code didn't really care about endianness, but if I don't do anything to handle it then some tests start failing.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -203,24 +372,35 @@ public static long readUnsigned(ByteBuffer bb, int pos) throws IllegalArgumentEx
...
-			return 1;
-		} else if (a0Unsigned <= 248) {
-			return 2;
-		} else if (a0Unsigned == 249) {
-			return 3;
-		} else {
-			int bytes = a0Unsigned - 250 + 3;
-			return 1 + bytes;
-		}
+		return FIRST_TO_LENGTH[a0 & 0xFF];
```
hmottestad, line 394: This makes a pretty decent performance difference actually. Essentially a pre-calculated lookup table instead of branching.

I think I might need to clean up some of the other code to see if I can use this in other places too.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -289,32 +469,47 @@ public static void readListUnsigned(ByteBuffer bb, int[] indexMap, long[] values
...
-	 */
-	private static void writeSignificantBits(ByteBuffer bb, long value, int bytes) {
-		while (bytes-- > 0) {
-			bb.put((byte) (0xFF & (value >>> (bytes * 8))));
-		}
+	public static void readQuadUnsigned(ByteBuffer bb, int[] indexMap, long[] values) {
+		values[indexMap[0]] = readUnsigned(bb);
+		values[indexMap[1]] = readUnsigned(bb);
+		values[indexMap[2]] = readUnsigned(bb);
+		values[indexMap[3]] = readUnsigned(bb);
```
hmottestad, line 475: Quads are always 4, so this helps the JVM inline and optimize the byte code.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -289,32 +466,35 @@ public static void readListUnsigned(ByteBuffer bb, int[] indexMap, long[] values
...
+		} else {
+			return readSignificantBitsHeap(bb, n);
 		}
-		return value;
+	}
+
+	private static long readSignificantBitsDirect(ByteBuffer bb, int n) {
+		return SignificantBytesBE.readDirect(bb, n);
+//		return VarUInt.readUnsigned(bb);
+	}
```
hmottestad, line 494: TODO: Find out where this is called from and if we can be certain about the number of bytes somewhere upstream. Then we might be able to use specific lambas like we use in the Bytes class.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -346,19 +526,27 @@ private static int compareRegion(ByteBuffer bb1, int startIdx1, ByteBuffer bb2,
 	 */
 	public static class GroupMatcher {
 
-		final ByteBuffer value;
 		final boolean[] shouldMatch;
 		final int[] lengths;
+		final Bytes.RegionComparator[] cmps;
 
 		public GroupMatcher(ByteBuffer value, boolean[] shouldMatch) {
```
hmottestad, line 534: TODO: Since this code is only ever called with a ByteBuffer backed by an array, then we can probably change the signature to take the array of bytes instead.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -346,19 +526,27 @@ private static int compareRegion(ByteBuffer bb1, int startIdx1, ByteBuffer bb2,
...
 			int pos = 0;
-			for (int i = 0; i < lengths.length; i++) {
-				int length = firstToLength(value.get(pos));
-				lengths[i] = length;
-				pos += length;
+			for (int i = 0; i < 4; i++) {
+				int len = firstToLength(valueArray[baseOffset + pos]);
+				lengths[i] = len;
+
+				cmps[i] = Bytes.capturedComparator(valueArray, baseOffset + pos, len);
```
hmottestad: Here we populate an array with lambda comparators where each lambda is hard coded for a specific length. This is possible because the left side in the compare (valueArray) is known already, and we also know the effective position and length,

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
```diff
@@ -367,20 +555,77 @@ public GroupMatcher(ByteBuffer value, boolean a, boolean b, boolean c, boolean d
...
+					if (cmps[1].compare(other, otherPos) != 0) {
+						return false;
+					}
+					if (!shouldMatch[2] && !shouldMatch[3]) {
+						return true;
+					}
+				}
+
+				otherPos += otherLen;
+			}
```
hmottestad: You wouldn't believe it, but this is considerably faster than a loop. Even when we hard code the loop for 4 iterations. Unrolling the loop also allows us to terminate early if no subsequent parts need to match.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/util/SignificantBytesBE.java
```diff
@@ -0,0 +1,95 @@
...
+		case 4:
+			return u32(getIntBE(bb, littleEndian));
+		case 3:
+			return (((long) u16(getShortBE(bb, littleEndian)) << 8))
+					| (bb.get() & 0xFFL);
+		// TODO: add 1 and 2 byte cases here!!!
+		default:
+			throw new AssertionError("unreachable");
+		}
+	}
```
hmottestad, line 93: This is faster than the original approach, which used a loop, but I really want to figure out how I can provide a lambda directly to the calling code.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
kenwenzel: @hmottestad I've also unrolled the writeSignificantBits method in my other PR.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/util/SignificantBytesBE.java
kenwenzel: @hmottestad This is exactly what I'm using in my other PR for writeSignificantBits. I think the main trick is to reduce the number of calls to the ByteBuffer.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/Varint.java
kenwenzel: @hmottestad I've added a working version here:
https://github.com/kenwenzel/rdf4j/commit/3f0a6e6d5b0bcb32a268cc9788ae82c5101d38ff

We should decide if we use yours or mine :-)

kenwenzel: @hmottestad I've just read this:
https://github.com/LWJGL/lwjgl3-wiki/wiki/1.3.-Memory-FAQ#what-javaniobyteorder-should-be-used

I'm not sure if this is an requirement or just a best practice.

Issue #5063, opened 2024-06-28
rdf4j-spring uncompatible with Spring Boot 3


florianchappaz:
### Current Behavior

Running a project relying on Spring Boot v3.0.4, I'm getting this error when the RDF4JConfig attempts to build the RDF4JTemplate:
```sh
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.eclipse.rdf4j.repository.Repository' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
```

rdf4j spring components are built relying on Spring Boot v2.7.16, making them uncompatible with a Spring Boot v3 project.

### Expected Behavior

I expected the @Import(RDF4JConfig::class) to automatically build the Repository from my application.properties.

### Steps To Reproduce

1. Setup a Maven project based on 
```xml
<parent>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-parent</artifactId>
  <version>3.0.4</version>
  <type>pom</type>
</parent>
```

2. Follow the basic setup presented in [rdf4j-spring](https://rdf4j.org/documentation/programming/spring/), especially the setup of the application properties.

### Version

5.0.0

### Are you interested in contributing a solution yourself?

Perhaps?

### Anything else?

It seems like the issue has already been mentioned in this [discussion](https://github.com/eclipse-rdf4j/rdf4j/discussions/4509).

hmottestad:
RDF4J is compatible with spring-boot 2.7.16.

fkleedorfer:
@florianchappaz this may or may not be a compatibility issue. It may just be a configuration issue. 

Have you tried using an inmemory repository?
The property to activate that is 

`rdf4j.spring.repository.inmemory.enabled=true`

(and if you want: `rdf4j.spring.repository.inmemory.use-shacl-sail=true` )

fkleedorfer:
FYI, I aim to upgrade rdf4j-spring to support to the latest spring boot version in the not so distant future. I haven't looked into compatibility issues with spring boot 3 so far, but I would not expect it to work out of the box.

florianchappaz:
## Issue Investigation

I've investigated the issue further and here are my findings:

1. **Using in-memory repository:** Same error persists.

2. **VPN disabled:** 
   - My GraphDB instance is only accessible through VPN.
[...]

PR #5541, opened 2025-10-30
Upgrade to Spring 6 and Spring-Boot 3

### LINKED ISSUE #5076: Upgrade to Spring 6
### Current Behavior

While version 5.0.1 fixed a lot of CVEs, there is now one major CVE left on Spring framework 5.3 which is probably non-trivial to fix.

- CVE-2016-1000027

It appears that there will not be a 5.3.x release addressing the issue.

Upgrading to Springboot 3 (see also #5063) will fix partially fix this because Springboot 3 uses Spring 6, but we may also need to upgrade the (non-springboot) spring dependencies of the workbench.

### Expected Behavior

Upgrading to a newer release of zookeeper fix the reported CVE for zookeeper dependency

### Steps To Reproduce

_No response_

### Version

5.0.1

Perhaps?

### Anything else?

See also https://github.com/spring-projects/spring-framework/issues/24434

Upgrading may not be strictly necessary, since one of the comments state

> Having said that it can be used as a reminder to check that there are no HTTP Invoker endpoints exposed to untrusted clients. If there are none, then nothing further to do.

(but automated scanners will still report this as a serious issue)

fkleedorfer:
GitHub issues resolved: #5076, #5063, 

We don't want to break rdf4j for existing users, so we have to make new artifacts.

## Approach:
* [x] Copy `spring-components` to `spring6-components`,
* [x] Rename modules: `s/spring/spring6/` 
* [x] update those modules to latest spring/spring-boot versions (major versions 6 and 3, respectively), adapt dependencies, fix whatever needs to be
* [x] Copy `tools/server-spring` to `tools/server-spring6`
* [x] update that module to latest spring/spring-boot versions (major versions 6 and 3, respectively), adapt dependencies, fix whatever needs to be

**Note:** There is substantial code duplication between the two submodules, the changes required for spring6 are minimal. However, extracting common code requires additional work and might lead to bugs, I would be happy to leave things as is, for now.

## Multiple Submodules of `tools` not changed

Some submodules of `tools` were **not** duplicated/adapted to spring6 and thus still use spring5:
* `rdf4j-http-server`
* `rdf4j-federation`
* `rdf4j-runtime-osgi`
* `rdf4j-workbench`
 
Transitively, they all depend on Spring 5, and I am not sure if they should be duplicated or updated. I am leaning toward duplicating, but it's a lot of duplication, and it would warrant some restructuring work: I'd introduce a `tools/spring5` and a `tools/spring6` intermediate pom-packaged parent for all the config changes, without changing the original artifact names for backward compatibility - thus, it will be clear how to handle the upcoming Spring 7 release. 
I would suggest leaving this aspect (handling other submodules of tools) out of this PR.

## Github Actions bumped to Java17

GitHub CI actions have to be changed to use Java 17 instead of Java 11, as spring 6 artifacts come compiled for that version. RDF4J still compiles to 11, so it should not impact clients. Necessary changes have been made (mosly including artifacts that are shipped with JDK versions < 17).

----


fkleedorfer:
... They do not fail, they just time out

hmottestad:
I just managed to merge main into develop, so if you rebase your branch onto develop your build should be a lot more stable. 

The requirement for java(11) is configured in the GitHub repo setting somewhere by the webmasters at Eclipse Foundation, once we permanently move to a newer Java version we can ask them to change it for us.

I think it's enough to just match the name, I don't know if it needs any steps or not.

hmottestad:
Maybe it's just the space before (11) that is wrong?

fkleedorfer:
with the job 'build' that has strategy/matrix configuration containing '11', it works.

fkleedorfer:
Now all that's failing is the slow-tests job. Should we give it more than 6 hours?

fkleedorfer:
The develop branch has the same problem - but only when building with java 17:

image

hmottestad:
I believe it'll have to wait for the next major release. Last we looked into this we ran into issues with Elasticsearch, Solr and Lucene needing to be upgraded too. And Elasticsearch needs a massive rewrite if we want to upgrade it since they changed their license to one that is incompatible. 

You can start with formatting the code, and then running the license tool to see what we need to do for IP work.

hmottestad:
develop branch has been upgraded to latest spring and spring-boot

PR #5549, opened 2025-11-03
LMDB Store: working on new ID based join iterator 

hmottestad:
GitHub issue resolved: # 

----


hmottestad:
@kenwenzel I’m working on an ID-based join for the LMDB store where we don’t need to create LmdbValue objects or binding sets. There is something not quite right though, causing a bunch of tests that test SNAPSHOT and SERIALIZABLE isolation to fail. I think that ID-based joins won’t work for SERIALIZABLE isolation, but I don’t quite understand what makes SNAPSHOT fail.

kenwenzel:
@hmottestad This is a great idea. Can you please explain why you would expect ID-based joins to fail for SERIALIZABLE? What is different in this case from joining on the value objects?

kenwenzel:
@hmottestad For a DUPSORT-based implementation see also https://github.com/eclipse-rdf4j/rdf4j/pull/5558
I'm sure we get this faster.

Is it possible that you remove anything related to DUPSORT from this PR and we implement it somewhere else?
I'm interested in reducing the DB size as much as we can.

PR #5572, opened 2025-11-22
GH-5291: Make it possible to disable transactions in LuceneSail

Ostrzyciel:
GitHub issue resolved: #5291

~Second~ third take at solving #5219, after #5446 and #5473. The original idea was to wrap Lucene's Directory into a class that would defer fsyncs and then execute them on a timer. This would allow us to keep full transactional support, and also reap the benefits of improved performance.

While that solution seemed to work initially, now I'm pretty sure it's impossible to implement in a 100% correct manner that would be reliable in the long term. Lucene does a lot more with files aside from fsyncing them – it also renames and deletes them. If we intervene only in the fsync process, we see inconsistent state, where, for example, we try to fsync files that no longer exist, or never fsync files that were renamed in the meantime.

Here I took a simpler approach, where we simply add the option to disable transaction/rollback support in LuceneSail. Then, during transaction commit, we call `flush()` on the index (not a full commit), and the real commit is done on a timer. Changes are made visible to readers after the commit only.

This sounds like a pretty big compromise, but for my use case this is totally fine. If we have slight inconsistencies in the text index because of a missed rollback or two, that's fine, nothing is going to collapse. Similarly, having a 10 second delay between insert and the text being visible in queries is also totally acceptable.

I think that with enough engineering effort you could make this better (e.g., have instant visibility of the changes in readers), but quite honestly... this is good enough.

I ran some end-to-end tests and verified that this indeed does result in a ~10x throughput speedup in insert transactions.

----

### DIFF
--- a/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneIndex.java
+++ b/core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneIndex.java
@@ -229,6 +229,9 @@ public class LuceneIndex extends AbstractLuceneIndex {
 			this.fsyncIntervalMillis = NumberUtils.toLong(
 					parameters.getProperty(LuceneSail.FSYNC_INTERVAL_KEY),
 					LuceneSail.DEFAULT_FSYNC_INTERVAL);
+			if (this.fsyncIntervalMillis <= 0) {
+				throw new IllegalArgumentException(LuceneSail.FSYNC_INTERVAL_KEY + " must be > 0");
+			}
 		}
 
 		postInit();
@@ -294,9 +297,11 @@ public class LuceneIndex extends AbstractLuceneIndex {
 
 	private void setUpFsyncScheduler() {
 		// If transactions are disabled, launch a background thread to fsync periodically.
-		if (this.transactionsEnabled || this.fsyncIntervalMillis <= 0) {
+		if (this.transactionsEnabled) {
 			return;
 		}
+		// Checked in initialize(), assertion here for clarity.
+		assert this.fsyncIntervalMillis > 0;
 
 		// Use a daemon thread so the scheduler does not prevent JVM shutdown.
 		this.fsyncScheduler = new ScheduledThreadPoolExecutor(1, r -> {
--- a/core/sail/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneNonTransactionalTest.java
+++ b/core/sail/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneNonTransactionalTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Eclipse RDF4J contributors.
+ * Copyright (c) 2026 Eclipse RDF4J contributors.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Distribution License v1.0
--- a/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
+++ b/core/sail/lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/LuceneSail.java
@@ -308,9 +308,8 @@ public class LuceneSail extends NotifyingSailWrapper {
 	/**
 	 * Set the key "fsyncInterval=&lt;t&gt;" as sail parameter to configure the interval in milliseconds in which fsync
 	 * is called on the Lucene index, default is defined in {@link #DEFAULT_FSYNC_INTERVAL}. Changes in the index will
-	 * become visible to readers at most after the interval is elapsed. If set to 0 or a negative value, fsync will be
-	 * disabled completely and there are no guarantees as to when your data will be persisted. This setting is only used
-	 * when {@link #TRANSACTIONAL_KEY} is set to false (rollbacks disabled).
+	 * become visible to readers at most after the interval is elapsed. This must be set to a value &gt; 0. This setting
+	 * is only used when {@link #TRANSACTIONAL_KEY} is set to false (rollbacks disabled).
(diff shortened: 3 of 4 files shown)


Ostrzyciel:
Thank you for the review @hmottestad !

> * **Background fsync task can race with close/commit, potentially reopening the writer after shutdown**

`getIndexWriter` is defined as follows:

```java
public synchronized IndexWriter getIndexWriter() throws IOException
```

So, this access is synchronized. Additionally, it has an openness check that will throw a SailException (that's fine):

```java
if (closed.get()) {
	throw new SailException("Index has been closed");
}
```

And even if we get a rare condition where the index is closed between us obtaining a reference to the index writer and calling commit(), the `commit()` method also starts with an openness check (`ensureOpen();`).

So, no, it's not possible for this to happen. Lucene explicitly disallows opening a closed IndexWriter. Worst that can happen is a very rare error message being written to the log, but that's fine, I think.

> * **`fsyncIntervalMillis <= 0` silently disables commits when transactions are off**

That's intended behavior, but indeed, it was not documented. I added a line about this.

> * **Scheduler shutdown does not force stop on timeout**

> The scheduler may continue to run and commit against a closed or closing index writer, which can reintroduce the race/leak described above.

As I pointed out above, this resource leak is impossible.

> Suggested fix: After the timeout, call shutdownNow() and/or loop with a bounded wait. Combine with the closed guard suggested above to prevent any commit attempts after shutdown.

That's fair. I added that. At least we won't have a stray thread, as long as Lucene is able to respond to thread interruptions. If it will ignore these, there is nothing more we can do.

core/sail/lucene/src/test/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneNonTransactionalTest.java
```diff
@@ -0,0 +1,67 @@
...
+	protected long waitAfterCommitMillis() {
+		return 200;
+	}
+
+	@Test
+	public void testIndexSettings() {
+		assertNotNull(index);
+		// Make sure the thread for periodic fsync is running
+		var allThreads = Thread.getAllStackTraces().keySet();
+		assertThat(allThreads.stream().filter(t -> t.getName().startsWith("rdf4j-lucene-fsync-")).count())
```
hmottestad: Maybe we should have a little loop here and check for the thread, just in case the thread hasn't started yet. To allow maybe 1 second for it to start?

Ostrzyciel: Done, I added an 0.5 s wait – should be enough, I think.

core/sail/lucene/src/main/java/org/eclipse/rdf4j/sail/lucene/impl/LuceneIndex.java
```diff
@@ -707,14 +795,21 @@ public synchronized void begin() throws IOException {
...
 	public synchronized void commit() throws IOException {
-		getIndexWriter().commit();
-		// the old IndexReaders/Searchers are not outdated
-		invalidateReaders();
+		if (this.transactionsEnabled) {
+			getIndexWriter().commit();
+			// the old IndexReaders/Searchers are not outdated
+			invalidateReaders();
+		} else {
+			getIndexWriter().flush();
```
hmottestad, line 808: If the fsync interval is 0 or negative the `invalidateReaders();` method is never called. Is it safe to call `invalidateReaders();` without calling commit?

hmottestad:
```
 * Copyright (c) 2025 Eclipse RDF4J contributors.
 *
 *
// Some portions generated by Codex
package org.eclipse.rdf4j.sail.lucene.impl;

import java.io.File;
import java.io.IOException;
import java.util.Properties;

import org.apache.lucene.analysis.standard.StandardAnalyzer;
import org.apache.lucene.store.NIOFSDirectory;
import org.eclipse.rdf4j.sail.lucene.LuceneSail;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.io.TempDir;

public class LuceneNonTransactionalNoFsyncTest extends AbstractGenericLuceneTest {

	@TempDir
	public File dataDir;

	@Override
	protected void configure(LuceneSail sail) throws IOException {
		LuceneIndex index = new LuceneIndex(new NIOFSDirectory(dataDir.toPath()), new StandardAnalyzer());
		Properties params = new Properties();
		params.setProperty(LuceneSail.TRANSACTIONAL_KEY, "false");
		params.setProperty(LuceneSail.FSYNC_INTERVAL_KEY, "0");
		params.setProperty(LuceneSail.LUCENE_DIR_KEY, dataDir.getAbsolutePath());
		try {
			index.initialize(params);
		} catch (Exception e) {
			throw new RuntimeException(e);
		}
		sail.setLuceneIndex(index);
	}

	@Test
	public void testCommitRefreshesReadersWithoutFsync() throws Exception {
		assertQueryResult("one", PREDICATE_1, SUBJECT_1);

		connection.begin();
		connection.add(SUBJECT_4, PREDICATE_1, vf.createLiteral("stale"));
		connection.commit();

		sleepAfterCommitIfNeeded();
		assertQueryResult("stale", PREDICATE_1, SUBJECT_4);
	}
}
```

Ostrzyciel: Oh... okay, yeah, there are quite a few caveats popping up :D

You know what? Maybe the easiest way out of this is to just disallow non-positive fsync interval values. I made a new commit, updating the option docs, adding some checks & assertions, and a couple of tests.

Does this make more sense?

hmottestad: I think that's a simple and understandable solution.

Issue #5573, opened 2025-11-26
Support to refine the @context of a JSON-LD document


aschwarte10:
### Problem description

We have the use-case to refine the @context of a JSON-LD document created through the RIO JSON-LD writer.

Specifically, we want to optionally in our application use a more compact context, e.g. to use "label" instead of "rdfs:label".

A full JSON-LD example with a refined context:

```
{
    "@id": "https://example.com/bob",
    "@type": "Person",
    "label": {
        "en": "Bob"
    },
    "name": "Bob",
    "@context": {
        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
        "foaf": "http://xmlns.com/foaf/0.1/",
        "@vocab": "https://example.com/",
        "label": {
            "@id": "rdfs:label",
            "@container": "@language"
        },
        "name": "foaf:name",
        "Person": "foaf:Person"
    }
}
```

### Preferred solution

Configurability on API level using a RioSetting that allows to refine the context

### Are you interested in contributing a solution yourself?

Yes

### Alternatives you've considered

Custom JSON-LD Writer with the same logic as the existing JSONLDWriter

PR #5574, opened 2025-11-26
GH-5573: support to refine the @context of a JSON-LD document

aschwarte10:
This change introduces a new setting for the JSONLDWriter, that allows
to configure a JSON-LD Context as Hasmac JsonDocument.

GitHub issue resolved: #5573 

With such JSONLDContextProvider it is possible to refine the JSON-LD @context in an application.

Example:

```
{
    "@id": "https://example.com/bob",
    "@type": "Person",
    "label": {
        "en": "Bob"
    },
    "name": "Bob",
    "@context": {
        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
        "foaf": "http://xmlns.com/foaf/0.1/",
        "@vocab": "https://example.com/",
        "label": {
            "@id": "rdfs:label",
            "@container": "@language"
        },
        "name": "foaf:name",
        "Person": "foaf:Person"
    }
}
```

----


aschwarte10:
>  would it be a good solution if we allow you to completely override the context document?

You mean something like

```
mpJsonLd.set(JSONLDSettings.CONTEXT, "{ ... ful context here ... }")
```

(Similar to how it is done in `org.eclipse.rdf4j.rio.jsonld.JSONLDWriterTest.testFraming()`

Sure, this is also fine for use. We just want control over the context. 

Great feedback (looking at consistency 👍 

@hmottestad can you confirm that the above is what you mean? Then I can update the PR by beginning of next week. Thanks

PR #5582, opened 2025-11-28
LMDB: minor code cleanup

Ostrzyciel:
GitHub issue resolved: –

Remove unused fields & fix a few other minor code issues.

Some fields could be replaced by local vars. Nested static classes don't have to hold a reference to the parent.

If we are lucky, removing a field here and there shortens the object by enough to make it fit into one less cache line (multiples of 64 bytes). I had a look at it with the Java Object Layout tool and I think that's the case only in one class here... still, a cleanup won't hurt.

----


hmottestad:
I don't really see any issues, but I think it's best if @kenwenzel can double check.

PR #5730, opened 2026-03-30
LMDB write path optimizations (+34% throughput)

odysa:
## Summary

- **MDB_WRITEMAP**: Enable writable mmap mode for LMDB environments, eliminating write() syscall overhead for direct mmap page writes
- **Inline findId fast path**: When a write transaction is active, bypass `readTransaction`/`writeTransaction` lambda wrappers in `ValueStore.findId()`, eliminating per-call lock acquisition (`txnLock.readLock()`) and extra `MemoryStack` frames for ~333K+ new value creations per bulk load
- **4-way set-associative value cache**: Replace direct-mapped cache with 4-way set-associative design (FIFO eviction per set) to reduce pathological collision eviction
- **Thread-local value cache**: Add per-thread caches (256 slots) to reduce CPU cache-line bouncing on the shared `valueCache` array under concurrent reads
- **Lock batching for iterators**: `LmdbRecordIterator` and `LmdbContextIdIterator` batch 512 rows per lock acquisition, with optimistic lock-free reads when no writer is active
- **MDB_NORDAHEAD support**: Add configurable `noReadAhead` flag (`MDB_NORDAHEAD`) to reduce page cache pollution on large datasets

## Benchmark results

1M bulk write (single transaction, 3 indexes: spoc, posc, ospc):

| State | stmts/sec | Change |
|-------|-----------|--------|
| Baseline | ~264K | — |
| + MDB_WRITEMAP | ~285K | +8% |
| + Inline findId | ~354K | +34% total |

## Test plan

- [ ] `mvn test -pl core/sail/lmdb -Dtest=TripleStoreTest` — core data integrity (2/2 passing)
- [ ] `mvn test -pl core/sail/lmdb -Dtest=LmdbStoreTest` — store-level tests
- [ ] `mvn test -pl core/sail/lmdb -Dtest=LmdbStoreConnectionTest` — connection tests
- [ ] Full LMDB test suite: `mvn test -pl core/sail/lmdb`
- [ ] Run `BulkLoadBenchmark` at 1M/10M scale to verify throughput gains

🤖 Generated with [Claude Code](https://claude.com/claude-code)


core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/config/LmdbStoreConfig.java
hmottestad, line 235: NO_READAHEAD is already implemented.

odysa: Good catch, thank you! You're right — `NO_READAHEAD` (IRI: `noReadahead`) was already present in the base branch schema. Our PR mistakenly added a second constant `NO_READ_AHEAD` (IRI: `noReadAhead`, capital A) which caused the flag to be serialized/deserialized via two different IRIs. Removed the duplicate in fcdff7c9a1 — only the canonical `NO_READAHEAD` constant is kept now.

odysa:
Closing this PR — after thorough re-benchmarking on `TransactionsPerSecondBenchmark.largerTransactionLevelNone` (10K stmts/txn), the changes do **not** show a statistically significant improvement over `develop`.

Detailed measurements (5 runs each, 2 warmup × 8s + 5 measurement × 8s, 1 fork, single thread):

| Branch | Avg (ops/s) |
|---|---:|
| `develop` baseline | 31.12 |
| Inline `findId` only | 31.11 |

**Δ ≈ 0%** — within the benchmark's ±10% per-run noise floor.

Even reducing scope to just the inline `findId` fast path (the cleanest change in the original set) yielded no measurable throughput gain on this workload. The original "+34% throughput" headline was based on a custom `BulkLoadBenchmark` that was developed alongside the optimizations and didn't survive validation against existing JMH benchmarks.

Possible reasons the optimizations don't pay off here:
- JIT already inlines the `readTransaction`/`writeTransaction` wrappers and scalar-replaces lambda allocations
- Per-call lock acquire/release on `txnLock.readLock()` is a single CAS pair, much cheaper than expected
- Real wall-clock cost per statement is dominated by LMDB's B+tree work in `mdb_put`, not Java-side wrapping

All 514 LMDB tests (`TripleStoreTest`, `LmdbStoreTest`, `LmdbStoreConnectionTest`, `LmdbStoreConcurrencyTest`) pass on the proposed change — it's correct, just not impactful on this workload.

Closing rather than landing a no-op change. If meaningful write-path improvements are pursued in the future, structural changes (sorted batch writes with `MDB_APPEND`, reducing `AddQuadOperation` GC pressure, or the producer/consumer queue) would be more promising directions.

hmottestad:
FYI. You should test out using READ_COMMITTED on develop. It is actually faster than NONE due to some batching and reordering.

Issue #5709, opened 2026-02-26
Vocabularies: add 'EU' package


barthanssens:
### Problem description

It would be nice to add a new vocabulary package with predefined constants for "EU"-vocabularies, primarily used by EU data portals and EU Dataspaces
(e.g. Geo/Mobility/StatDCAT-AP, ADMS, CPSV-AP ... from SEMIC.eu programme, CiteDCAT from EU Joint Research Center, maybe ELI from EU Publication Office...) 

While trivial to generate, IMHO it offers a nice touch to be able to use them "out-of-the-box".

### Preferred solution

A  org.eclipse.rdf4j.model.vocabulary.eu package with several new vocabulary classes

### Are you interested in contributing a solution yourself?

Yes

PR #5739, opened 2026-04-09
GH-5709: added vocabularies typically used in EU projects

barthanssens:
GitHub issue resolved: #5709

- New "vocabulary.eu" package with European DCAT-AP vocabularies, ADMS (typically used in data spaces and data portals), Core Vocabularies (procedure exchange) and ELI Ontology (legislation)
- New "vocabulary.annotation" package with OA (web annotation), DQV, DUV
- New "vocabulary.biblio" package with BIBO, FBBR, FABIO
- Added a copy of package-protected vocabularies helper class to "org.eclipse.rdf4j.model.base.model.base", but this time as a public class (to allow other packages to use it), albeit with InternalOnly annotation

----

### DIFF
--- a/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/Vocabularies.java
+++ b/core/model-api/src/main/java/org/eclipse/rdf4j/model/base/Vocabularies.java
@@ -1,0 +1,75 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Eclipse RDF4J contributors.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *******************************************************************************/
+
+package org.eclipse.rdf4j.model.base;
+
+import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.Namespace;
+
+/**
+ * Utility methods related to RDF vocabularies.
+ *
+ * @author Alessandro Bollini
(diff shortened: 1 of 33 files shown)


hmottestad:
@barthanssens for these new vocabularies, can you use the case from the iri name instead of all caps? It makes it a lot easier for users to find the correct vocabulary when our names are identical to the vocabulary.

barthanssens:
@hmottestad uhm, for the java class name you mean ? e.g. GeoDCAT-AP instead of GEODCATAP ? Sure can do.

And/or for the IRIs for the RDF properties / classes as well ?

barthanssens:
So I've changed the all-uppercase IRI for classes and properties of EU vocabularies package, would you like me to do the same for the newly added (but more generic) vocabularies like BIBO in the "main" vocabularies package ?
(I think it's indeed a good idea, but it breaks the current convention used in the very old/widely used vocabularies like DCTERMS)

hmottestad:
> So I've changed the all-uppercase IRI for classes and properties of EU vocabularies package, would you like me to do the same for the newly added (but more generic) vocabularies like BIBO in the "main" vocabularies package ?
> 
> (I think it's indeed a good idea, but it breaks the current convention used in the very old/widely used vocabularies like DCTERMS)

Thanks. 

We don't need to do it for existing vocabularies or for new IRIs to existing vocabularies. 

For new vocabularies it reduces confusion and improves discoverability. It's always challenging when a vocabulary uses the same term for a class and a predicate when we use uppercase for all Java constants.

hmottestad:
And yes, think it's a good idea to use the same case as the IRI for all new vocabulary sets we introduce.

barthanssens:
Ok, makes sense, I'll change the IRIs of the other newly added vocabularies as well

barthanssens:
Split / moved into newer java packages. 

Also moved the Vocabularies helper class to model.base and made it public to make it reusable across packages
(actually this java class is already a copy of the package-protected one in existing ..vocabularies package, so it might be good idea to deprecated/remove that one as well, and use the now-public Vocabularies helper class to create internedIRI in the existing DCTERMS etc vocabularies. Shouldn't make a difference for the end-user)

PR #5796, opened 2026-04-27
GH-5720 Sketch based optimizer new cascade style planners and optimisers

### LINKED ISSUE #5720: RDF4J-STF-M1: Query Optimization
Implement theta sketches to estimate join cardinality for better BGP (basic graph pattern) reordering. This feature allows to estimate the result cardinality of pairwise combination of graph patterns to better predict the complexity of joins. This allows us to reorder BGP based on the estimated result cardinality to reduce the number of intermediate results.

hmottestad:
GitHub issue resolved: #5720 

----


hmottestad:
Worst query is about 8x slower than on develop. Best query is 1817x faster :)

latest-vs-develop-theme-query-compare

hmottestad:
Seems that the overhead is around 0.2 ms, so any query that is faster than that on develop will be slower on this branch.

kenwenzel:
The results look a lot more balanced in comparison to the `develop` branch.
Especially the very slow queries disappeared completely.
I wonder how much additional memory (RAM and harddisk) is required.

hmottestad:
This now uses mmap, so you don't need as much memory as before. It'll just page out. But you do need memory for the actual sketch objects and maps and lists, so I think it still requires a more than it should.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
```diff
@@ -290,9 +290,17 @@ abstract static class StatefulOperation implements Operation {
 	 */
 	public LmdbSailStore(File dataDir, StoreProperties properties, LmdbStoreConfig config)
 			throws IOException, SailException {
+		this(dataDir, properties, config, true);
+	}
+
+	public LmdbSailStore(File dataDir, StoreProperties properties, LmdbStoreConfig config,
+			boolean sketchBasedJoinEstimatorEnabled)
```
kenwenzel, line 300: Can we move `sketchBasedJoinEstimatorEnabled` to `LmdbStoreConfig` and disable it by default?
Especially for embedded uses with smaller DBs the storage overhead may not pay off.
What do you think?

hmottestad:
lmdb-theme-latest-vs-develop

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbSailStore.java
hmottestad: I was thinking to maybe tie it to the Xmx, so that if you configure your jvm with 2GB of memory or more, then the sketches will be enabled by default.

hmottestad:
[results-2026-05-08-3-vs-develop.pdf](https://github.com/user-attachments/files/27534737/results-2026-05-08-3-vs-develop.pdf)

results-2026-05-08-3-vs-develop

PR #5812, opened 2026-05-05
GH-5810 fix for lmdb crash due to incorrect transaction handling

### LINKED ISSUE #5810: LMDB read transactions can reset while record iterators are still active
### Description

While investigating LMDB read transaction lifecycle, I found that LMDB read transactions can be reset while `RecordIterator`/cursor users are still active. The same area also has close/reset behavior that can wait on unrelated cursors or fail to close cursors owned by the transaction being closed.

This risks active LMDB cursors using a reset/closed transaction, close paths hanging under concurrent access, and read-only work crossing transaction/thread boundaries incorrectly.

### Local regression evidence

Work-in-progress branch: `fix-read-txn-bug`.

`TripleStoreTxnLockTest.commitResetWaitsForOpenRecordIteratorUntilClose` currently reproduces the core issue:

```text
[commit reset must wait while an LMDB record cursor is open]
Expecting value to be false but was true
```

Related failing coverage from the same investigation:

```text
[read txn close must not wait for unrelated open LMDB record cursors]
Expecting value to be true but was false
```

```text
[read txn close must close its own unclosed LMDB record cursors before reset]
Expecting value to be true but was false
```

I also added local coverage for `ValueStoreTest.readOnlyLookupFromAnotherThreadDoesNotUseUncommittedWriteTxn`, because read-only lookup from a different thread should not use another thread's in-flight write transaction.

### Expected behavior

- Writer commit/reset waits while LMDB record cursors opened on the read transaction are active.
- Closing a read transaction closes or waits only for its own open cursors.
- Closing an iterator from another thread does not deadlock.
- Read-only lookup from another thread does not observe values from an uncommitted write transaction.

### Suspected scope

`core/sail/lmdb`, especially `TxnManager`, `LmdbRecordIterator`, `LmdbContextIdIterator`, `TripleStore`, and `ValueStore` read/write transaction ownership.

hmottestad:
Fixes #5810.

## Issue

LMDB read transactions can be reset/deactivated while Java iterators still own native LMDB cursors. The earlier fix correctly made idle iterators non-blocking, but it moved too much coordination into the hot row path: every `next()` went through the global transaction read lock plus synchronized cursor start/finish bookkeeping. That preserved correctness, but regressed the 8-thread `RecordIteratorBenchmark.iterateAll` run to ~42 ms/op.

The root cause is that there are two different lifetimes that need different treatment:

- idle open cursors must stay registered so transaction close can clean them up, but they must not block commit/reset/resizes
- active native cursor calls must be protected while LMDB is reading from the transaction, but only for that short native-use window

Using one global per-`next()` read lock for both cases put lock coordination, monitor traffic, `notifyAll`, and `LongAdder` work back in the row scan loop.

## Root-cause fix

This PR keeps cursor ownership explicit while splitting idle cursor lifetime from active native cursor use:

- `TxnManager.Txn` now exposes a package-private `CursorUse` guard through `beginCursorUse()`
- the transaction has an atomic native-use gate: high bit means reset/deactivate/native close is transitioning, low bits count active cursor users
- `reset()`, `deactivate()`, and native transaction close set the transition bit, wait only for active cursor users, then reset/free the native transaction
- idle registered cursors remain registered for cleanup, but no longer block commit/reset/resize just because they are open
- `LmdbRecordIterator`, `LmdbContextIdIterator`, and `PersistentSet.ElementIterator` no longer take the global transaction read lock on every `next()`
- native cursor calls use direct enter/exit cursor-use methods to avoid allocating/closing an `AutoCloseable` guard per row
- per-cursor async close is handled by a small package-private `CursorUseState`; the hot path is volatile stores/checks, and close waits only when it races an active cursor call
- version-based cursor renewal after transaction reset is preserved

No public API changes.

## Why this stays performant

The fix pays synchronization only at the boundary that needs it:

- commit/reset/native close coordinate with active native cursor calls, not idle iterator objects
- iterator `next()` avoids the global `StampedLongAdderLockManager` read lock
- the old synchronized `startCursorUse()` / `finishCursorUse()` path and unconditional `notifyAll()` are gone from the scan loop
- cursor close coordination is per cursor, so async close does not serialize unrelated iterators

The remaining steady-state row scan cost is mostly LMDB cursor/comparator work and key decoding, with only the lightweight transaction/cursor-use guards around native calls.

## Benchmark

Docker JMH/JFR run against the LMDB store iterator benchmark with 8 threads:

```text
bash .codex/skills/docker-jfr-benchmark-loop/scripts/run-docker-jfr-loop.sh \
  org.eclipse.rdf4j.sail.lmdb.RecordIteratorBenchmark.iterateAll
```

Environment: Docker image `maven:3.9.11-eclipse-temurin-25`, JMH 1.37, JDK 25.0.1, `-Xms1G -Xmx1G`, one fork, no warmup, ten 10s measurement iterations, 8 synchronized threads. The Docker helper also enabled JFR CPU-time profiling with stack depth 1024 and `-XX:+DebugNonSafepoints`.

Regression rerun before the final fix:

```text
Benchmark                           Mode  Cnt   Score   Error  Units
RecordIteratorBenchmark.iterateAll  avgt   10  42.106 ± 2.243  ms/op
```

Final run after this PR update:

```text
Benchmark                           Mode  Cnt   Score   Error  Units
RecordIteratorBenchmark.iterateAll  avgt   10  28.740 ± 1.470  ms/op
```

Run detail: min/avg/max = 27.036 / 28.740 / 29.972 ms/op. JFR recording: `core/sail/lmdb/target/RecordIteratorBenchmark.iterateAll.jfr`.

Final JFR top methods no longer include `startCursorUse`, `finishCursorUse`, unconditional `notifyAll`, or `LongAdder.add`. The top entries are now key decoding, LMDB cursor/comparator calls, `LmdbRecordIterator.next`, and the lighter direct cursor-use guard.

## Tests

Added focused regression coverage for:

- commit/reset waits while a `Txn.CursorUse` guard is active
- idle registered cursors do not block commit/reset
- async record iterator close remains idempotent without relying on the global read lock

Re-ran focused LMDB lifecycle/concurrency coverage:

```text
mvn -o -Dmaven.repo.local=.m2_repo -pl core/sail/lmdb \
  -Dtest=TripleStoreTxnLockTest,PersistentSetTest,TxnManagerTest,TxnRecordCacheTest verify
```

Result:

```text
Tests run: 18, Failures: 0, Errors: 0, Skipped: 0
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Also ran:

```text
mvn -T 1C -o -Dmaven.repo.local=.m2_repo -Pquick clean install
mvn -o -Dmaven.repo.local=.m2_repo -T 2C process-resources
cd scripts && ./checkCopyrightPresent.sh
```

----
PR Author Checklist (see the [contributor guidelines](https://github.com/eclipse-rdf4j/blob/main/CONTRIBUTING.md) for more details):

 - [x] I've applied [code formatting](https://github.com/eclipse-rdf4j/blob/main/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)


kenwenzel:
I've implemented a simple pooling mechanism for re-using iterator state here:
https://github.com/eclipse-rdf4j/rdf4j/pull/5558/changes/c815e5d9bd472dc7907bd4c04373d7cdf551795e

On first sight this seems a lot simpler than the proposed approach. But nevertheless, I'll take a closer look tomorrow morning.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
```diff
@@ -236,36 +255,105 @@ private boolean matches() {
 	}
 
 	private void closeInternal(boolean maybeCalledAsync) {
+		closeInternal(maybeCalledAsync, false);
+	}
+
+	private void closeInternal(boolean maybeCalledAsync, boolean readLockHeld) {
```
kenwenzel: Seems like maybeCalledAsync is useless now.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/LmdbRecordIterator.java
```diff
@@ -236,36 +255,105 @@ private boolean matches() {
...
+				interrupted = true;
+			}
+		}
+		if (interrupted) {
+			currentThread.interrupt();
+		}
+	}
+
+	private void closeCursor(boolean readLockHeld) {
+		if (!beginCloseCursor()) {
```
kenwenzel: Is it possible that this is called from a thread different from the owner/creator thread?
If yes then I think it is not allowed to close cursors created from another thread.

kenwenzel (commented):
Hi @hmottestad,

thank you for the PR. I'm not completely sure that the cursor cannot be closed from a "foreign" thread.
But maybe I just don't fully understand the logic.

If `close()` is called from a foreign thread (due to timeout - we have an open issue for this) then the best reaction would be to mark the iterator as closed and let the owner thread close the cursor in the next iteration somehow.

What do you think?

Issue #5813, opened 2026-05-06
Support for a grouped source selection optimization in FedX


aschwarte10:
FedX uses ASK queries for source selection.

For queries that consist of many patterns and for cases when the federation is large, this will use many slots of the parallel scheduler for the "simple" http requests.

The idea of this optimization is to use a grouped source selection strategy (grouping source selection requests per endpoint)

aschwarte10:
Completed with github.com/eclipse-rdf4j/rdf4j/pull/5816

PR #5815, opened 2026-05-06
GH-5813: fix maven module setup for use with Eclipse + fix build

aschwarte10:
- add junit platform launcher to federation classpath
- fix spotless and surefire integration to eclipse

Commit 2 fixes the build:

GH-5813: fix build after merging main to develop (server-boot test)

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -423,11 +423,30 @@
 								<lifecycleMappingMetadata>
 									<pluginExecutions>
 										<pluginExecution>
+											<pluginExecutionFilter>
+												<groupId>org.apache.maven.plugins</groupId>
+												<artifactId>maven-surefire-plugin</artifactId>
+												<versionRange>[2.0,)</versionRange>
+												<goals>
+													<goal>test</goal>
+												</goals>
+											</pluginExecutionFilter>
 											<action>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -423,11 +423,30 @@
...
-													<runOnIncremental>false</runOnIncremental>
-												</execute>
+												<ignore />
+											</action>
+										</pluginExecution>
+										<pluginExecution>
+											<pluginExecutionFilter>
+												<groupId>com.diffplug.spotless</groupId>
+												<artifactId>spotless-maven-plugin</artifactId>
+												<versionRange>[1.0,)</versionRange>
```
hmottestad, line 442: What is the point of using a version range? I've not seen this before and not sure what it means.

aschwarte10: The versionRange attribute itself is required, I use kind of "all versions" to have this regardless of the specific spotless version (e.g. when we do updates of spotless we do not have to adjust the m2e configuration)

aschwarte10: I would like to merge this PR now to be able to continue with my functional changes that require a stable build. If you prefer I can follow up later and use exact versions

PR #5822, opened 2026-05-11
Stable order

hmottestad:
LMDB uses ordered indexes. With DISTINCT we can take advantage of this so that we don’t need to keep all results in memory. If we know that a particular variable will always be ordered, then we can clear the hash set when we see a new binding for that variable. 

We can also expose the stable order as a best-effort ordering in ORDER BY. A user can request to have results ordered by a particular variable according to the internal order in LMDB, to make it possible to support cheap pagination.

GitHub issue resolved: # 

----


core/queryparser/sparql/src/test/java/org/eclipse/rdf4j/query/parser/sparql/SPARQLParserTest.java
```diff
@@ -297,6 +297,69 @@ public void testOrderByWithAliases2() {
 
 	}
 
+	@Test
+	public void testOrderByStableIndexParses() {
+		String query = "SELECT * WHERE { ?a a ?type. } ORDER BY STABLE_INDEX(?a)";
```
hmottestad, line 302: @kenwenzel what do you think of this idea?

kenwenzel: @hmottestad I think it might be usable for specific requirements. I'm a bit unsure if this will be used by a larger user base. Most users maybe try to keep the queries in line with standard SPARQL as much as possible.

hmottestad:
@codex rebase this PR onto develop

PR #5826, opened 2026-05-14
lmdb predicate guarantees

hmottestad:
GitHub issue resolved: # 

----


kenwenzel:
@hmottestad Do you think it would be a good idea to support D-ebtailment in the value store level instead of tracking the predicate ranges? If yes,.then we could create a specification issue for this.

hmottestad:
It's still not clear to me how that will let us optimize FILTERS into VALUES in SPARQL. Do you have an example?

kenwenzel:
> It's still not clear to me how that will let us optimize FILTERS into VALUES in SPARQL. Do you have an example?

I would try to implement something like multi-datatype lookups in value store. If, for example, an integer "10" is looked up but only a "10" long is stored then the latter will be directly returned.
But I'm not completely sure about the implications.

kenwenzel:
https://www.w3.org/TR/sparql12-entailment/#DEntRegime

hmottestad:
> https://www.w3.org/TR/sparql12-entailment/#DEntRegime

I think it's quite tricky to get right. We can go all in and say that we will normalise all numbers to either decimal or integer, so if the user inserts 1.0 and 1, then we normalise these to the same literal and the same internal LMDB ID. 

If we want to meet users half way, normalise literals but retain datatype then we need some way to represent two literals with the same internal LMDB ID while still being able to retrieve the two different literals in both different datatypes. 

We could encode this in the ID as: node type + d-entailed xsd type + normalised value + actual xsd data type. 

So 1.0 and 1 would have almost the same internal ID, but with a different byte at the end to encode the actual datatype.

hmottestad:
I guess we could just reserve the last byte for identifying the users literal. Then we don't need to normalise the actual value. 

1.0 and 001 and 1 and 01.0000000 would all have the the same first 7 bytes. The 8th byte would be 0 for 1.0, 1 for 001 and 2 for 1 and 3 for 01.0000000.

hmottestad:
```

"1.0"         -> 0x0000000000002a00
"001"         -> 0x0000000000002a01
"1"           -> 0x0000000000002a02
"01.0000000"  -> 0x0000000000002a03

```

A D-entailment lookup for "1"^^xsd:integer becomes a prefix/range scan:

```
value_id = 42

scan:
  lower = 42 << 8
  upper = ((42 + 1) << 8) - 1
```

kenwenzel:
> ```
> 
> "1.0"         -> 0x0000000000002a00
> "001"         -> 0x0000000000002a01
> "1"           -> 0x0000000000002a02
> "01.0000000"  -> 0x0000000000002a03
> ```
> 
> A D-entailment lookup for "1"^^xsd:integer becomes a prefix/range scan:
> 
> ```
> value_id = 42
> 
> scan:
>   lower = 42 << 8
>   upper = ((42 + 1) << 8) - 1
> ```

This is more or less exactly what I also thought.

What happens If "1"^^xsd:byte and "1"^^xsd:int are both stored in the db?

hmottestad:
They would have two different internal IDs where the first 7 bytes are identical. So when we lookup using D-entailment we just use the first 7 bytes of the ID, when we lookup to retrieve the value from the value store we use all 8 bytes. 

When we lookup using the first 7 bytes, we would get both 1 xsd:byte and 1 xsd:int as two distinct IDs. So if we count all objects where FILTER(?o == 1) then we would translate that into something like VALUES(?o) {"1"lmdb:dEntailedXsdInt} which would trigger the special part in the code that would ignore the last byte when doing a lookup in the lmdb triple store.

Though the last bit there with the VALUES I haven't thought through completely. Usually a bound variable would only match a single literal, but in this case we want the bound variable to match multiple literals. So maybe we need to come up with something else that allows us to express that correctly. Maybe as extension of the RDF4J Var object with a new field for d-entailed internal lmdb ID.

PR #5844, opened 2026-05-29
GH-5843: fix NativeStore BTree corruption recovery

### LINKED ISSUE #5843: NativeStore BTree corruption (IllegalArgumentException: Trying to remove largest value from an empty node)
### Current Behavior

In our usages of the NativeStore we are facing instabilities rather frequently now, with BTree errors (as depicted in below stacktrace):

```
java.lang.IllegalArgumentException: Trying to remove largest value from an empty node in /runtime-data/data/repositories/metis-ci-integration/triples-spoc.dat
at org.eclipse.rdf4j.sail.nativerdf.btree.BTree.removeLargestValueFromTree(BTree.java:938)
at org.eclipse.rdf4j.sail.nativerdf.btree.BTree.removeFromTree(BTree.java:903)
at org.eclipse.rdf4j.sail.nativerdf.btree.BTree.remove(BTree.java:848)
at org.eclipse.rdf4j.sail.nativerdf.TripleStore.commit(TripleStore.java:997)
at org.eclipse.rdf4j.sail.nativerdf.TripleStore.processUncompletedTransaction(TripleStore.java:380)
at org.eclipse.rdf4j.sail.nativerdf.TripleStore.<init>(TripleStore.java:208)
at org.eclipse.rdf4j.sail.nativerdf.NativeSailStore.<init>(NativeSailStore.java:174)
at org.eclipse.rdf4j.sail.nativerdf.NativeStore.initializeInternal(NativeStore.java:489)
at org.eclipse.rdf4j.sail.helpers.AbstractSail.init(AbstractSail.java:183)
at org.eclipse.rdf4j.repository.sail.SailRepository.initializeInternal(SailRepository.java:158)
at org.eclipse.rdf4j.repository.base.AbstractRepository.init(AbstractRepository.java:36)
```

We did not observe the issues prior to RDF4J 5.2.2, i.e. for us it is only happening recently after switching to RDF4J 6.0. 

---

Summary created by AI agent after analysis

Summary

  After a JVM crash during a transaction commit, reopening a NativeStore can permanently fail with:

```
  java.lang.IllegalArgumentException: Trying to remove largest value from an empty node in .../triples-spoc.dat
      at BTree.removeLargestValueFromTree(BTree.java:938)
      at BTree.removeFromTree(BTree.java:903)
      at BTree.remove(BTree.java:848)
      at TripleStore.commit(TripleStore.java:982)
      at TripleStore.processUncompletedTransaction(TripleStore.java:363)
      at TripleStore.<init>(TripleStore.java:198)
```

  The store cannot be opened at all until the data files are manually repaired or deleted.

  ---
  Root Cause

  There are two separate problems that combine to make this fatal:

  1. BTree structural corruption after crash (pre-existing)

  With forceSync=false (the default), the kernel may write BTree node pages to disk out of order. During a BTree merge operation, the sequence is:

  1. The right sibling's valueCount is set to 0 (node becomes empty) → written to disk
  2. The parent node's separator value and child pointer are removed → not yet on disk
  3. Crash

  On recovery, the parent node still references the now-empty leaf as a child. The B-Tree invariant (every non-root leaf has ≥ minValueCount values) is violated on disk.

  2. processUncompletedTransaction does not handle RuntimeException

  During crash recovery, TripleStore.<init> calls processUncompletedTransaction(COMMITTING) which calls commit(). When commit() iterates all triples to apply removals, btree.remove() recurses into the corrupt empty node and throws
  IllegalArgumentException — a RuntimeException. Because processUncompletedTransaction only catches IOException, the exception propagates through the constructor and leaves the store permanently inaccessible. There is no fallback to
  rollback().

```
  // processUncompletedTransaction — only catches IOException
  try {
      commit();
  } catch (IOException e) {  // IllegalArgumentException is NOT caught here
      logger.error("Failed to restore from uncompleted commit", e);
      throw e;
  }
```

### Expected Behavior

The NativeStore is expected to be stable and at minimum allows to recover

### Steps To Reproduce

  ---
  Reproducer

  A regression test has been added to TripleStoreRecoveryTest.testCommitRecoveryWithBTreeCorruption() that:

  1. Creates a 2-level BTree (98 triples → root split at triple 50)
  2. Marks the root-level triple for removal inside a transaction (sets REMOVED_FLAG)
  3. Simulates the partial-flush crash by zeroing the left child's valueCount directly in the .dat file
  4. Forces txn status to COMMITTING
  5. Asserts that reopening throws IllegalArgumentException

### Version

6.0 M1

None

### Anything else?

---
  Proposed Fix

  processUncompletedTransaction should catch RuntimeException from commit() and fall back to rollback() so the store remains accessible (with the unrecoverable transaction rolled back):

```
  case COMMITTING:
      logger.info("Detected uncompleted commit, trying to complete");
      try {
          commit();
          logger.info("Uncompleted commit completed successfully");
      } catch (IOException e) {
          logger.error("Failed to restore from uncompleted commit", e);
          throw e;
      } catch (RuntimeException e) {
          logger.error("BTree corruption detected during commit recovery, attempting rollback", e);
          try {
              rollback();
              logger.warn("Rolled back corrupt commit; data from that transaction is lost");
          } catch (IOException rollbackEx) {
              e.addSuppressed(rollbackEx);
              throw new IOException("Crash recovery failed: commit and rollback both failed", e);
          }
      }
      break;
```

  A longer-term fix is to ensure BTree merge operations write parent and child nodes in a safe order, or to add a post-crash BTree consistency check that detects and repairs empty-child violations before attempting commit().

aschwarte10:
GitHub issue resolved: #5843

See #5843 for detailed write-up

**Commit 1 GH-5843: add regression test for BTree corruption crash-recovery failure**

Adds testCommitRecoveryWithBTreeCorruption to TripleStoreRecoveryTest,
reproducing the scenario where a crash leaves a non-leaf BTree node with
an empty leaf child. On reopen,
processUncompletedTransaction(COMMITTING) calls commit(), which calls
btree.remove() on the root-level triple, which recurses into the empty
child via removeLargestValueFromTree and throws IllegalArgumentException
— making the store permanently unreadable.

The test simulates the exact production stack trace by: building a
2-level BTree, marking the root's separator value for removal, zeroing
the left child's valueCount on disk to mimic a partial-flush crash, and
forcing txn status to COMMITTING. It asserts that reopening throws
IllegalArgumentException, confirming the bug is reproducible before any
fix is applied.

**Commit 2: GH-5843: recover from BTree corruption during crash recovery**

processUncompletedTransaction only caught IOException from commit(), so
an IllegalArgumentException thrown by BTree.removeLargestValueFromTree
(when a non-leaf node's child is an empty leaf due to a partial-flush
crash) propagated through the TripleStore constructor and left the store
permanently unreadable.

Add a RuntimeException catch that falls back to rollback(), keeping the
store accessible at the cost of losing the uncommitted transaction. If
rollback also fails, both exceptions are surfaced via addSuppressed.

Update the regression test to assert the store opens and is readable
after recovery instead of asserting the exception.

----

### DIFF
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TripleStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TripleStore.java
@@ -383,10 +383,10 @@ class TripleStore implements Closeable {
 				logger.error("Failed to restore from uncompleted commit", e);
 				throw e;
 			} catch (RuntimeException e) {
-				logger.error("BTree corruption detected during commit recovery, attempting rollback", e);
+				logger.error("Commit recovery failed with runtime exception, attempting rollback", e);
 				try {
 					rollback();
-					logger.warn("Rolled back corrupt commit; data from that transaction is lost");
+					logger.warn("Rolled back failed commit recovery; data from that transaction is lost");
 				} catch (IOException rollbackEx) {
 					e.addSuppressed(rollbackEx);
 					throw new IOException("Crash recovery failed: commit and rollback both failed", e);
--- a/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TripleStore.java
+++ b/core/sail/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/TripleStore.java
@@ -382,6 +382,15 @@ class TripleStore implements Closeable {
 			} catch (IOException e) {
 				logger.error("Failed to restore from uncompleted commit", e);
 				throw e;
+			} catch (RuntimeException e) {
+				logger.error("BTree corruption detected during commit recovery, attempting rollback", e);
+				try {
+					rollback();
+					logger.warn("Rolled back corrupt commit; data from that transaction is lost");
+				} catch (IOException rollbackEx) {
+					e.addSuppressed(rollbackEx);
+					throw new IOException("Crash recovery failed: commit and rollback both failed", e);
+				}
 			}
 			break;
 		case ROLLING_BACK:
--- a/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/TripleStoreRecoveryTest.java
+++ b/core/sail/nativerdf/src/test/java/org/eclipse/rdf4j/sail/nativerdf/TripleStoreRecoveryTest.java
@@ -12,8 +12,10 @@ package org.eclipse.rdf4j.sail.nativerdf;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.io.File;
+import java.io.RandomAccessFile;
 
 import org.eclipse.rdf4j.sail.nativerdf.TxnStatusFile.TxnStatus;
 import org.eclipse.rdf4j.sail.nativerdf.btree.RecordIterator;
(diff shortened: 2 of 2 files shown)


hmottestad:
@copilot review this PR

Issue #5879, opened 2026-06-18
ApacheHC5RDF4JHttpClient.buildRequest should be protected to allow request cancellation by subclasses


sava-savov-ontotext:
### Problem description

Currently, buildRequest in ApacheHC5RDF4JHttpClient is package-private, which prevents subclasses from accessing the constructed HttpUriRequestBase instance. This makes it impossible for library users to cancel in-flight HTTP requests, such as long-running SPARQL queries.

### Preferred solution

Change buildRequest visibility from private to protected.

This allows subclasses to override buildRequest, capture the constructed request, and expose it for cancellation.
HttpUriRequestBase already has cancel() / abort() wired into Apache HC5's connection lifecycle via CancellableDependency, so this is the natural cancellation handle. The change is minimal and non-breaking.

### Are you interested in contributing a solution yourself?

Yes

aschwarte10:
Resolved with https://github.com/eclipse-rdf4j/rdf4j/pull/5881

PR #5881, opened 2026-06-18
GH-5879 Make buildRequest protected in ApacheHC5RDF4JHttpClient to allow request cancellation.

sava-savov-ontotext:
Current implementation has buildRequest as package-private, which prevents subclasses from accessing the constructed HttpUriRequestBase instance. 
This makes it impossible for library users to cancel long-running HTTP requests such as SPARQL queries.

GitHub issue resolved: #5879


aschwarte10 (approved):
LGTM if this is needed for extensions

PR #5913, opened 2026-06-29
GH-5909 Use parameter of BNODE() function as prefix of BNode ID.

### LINKED ISSUE #5909: SPARQL BNODE function should use argument as prefix for generated blank node identifiers
### Problem description

The SPARQL BNODE function currently generates blank node identifiers without using a provided argument as a prefix. When using BNODE with an argument, the argument should be used as a prefix for the generated blank node identifiers to allow for more predictable and controlled blank node naming.

### Preferred solution

When the BNODE function is called with an argument (e.g., BNODE("myprefix")), the generated blank node identifiers should use the provided argument as a prefix. For example, BNODE("ex") should generate identifiers like "ex0", "ex1", "ex2", etc., rather than generating arbitrary identifiers.

_No response_

_No response_

### Anything else?

This enhancement would align the BNODE function behavior more closely with SPARQL specifications and provide more control over blank node identifier generation in query results.

kenwenzel:
GitHub issue resolved: #5909

Use parameter of BNODE() function as prefix for generated BNode IDs within SPARQL queries.

----

### DIFF
--- a/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
+++ b/core/queryalgebra/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/QueryEvaluationContext.java
@@ -167,7 +167,7 @@ public interface QueryEvaluationContext {
 			}
 
 			return bnodesByLabel.computeIfAbsent(nodeLabel,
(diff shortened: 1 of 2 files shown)


aschwarte10 (approved):
Thanks for the fix! Changes the behavior for our use-case back to what it was before - verified with our unit test.

The pattern makes sense to me 👍

Issue #5949, opened 2026-07-12
Support serialization to RDF 1.2 Basic and RDF 1.1


ivairali:
### Problem description

The RDF1.2 [server considerations] https://www.w3.org/TR/rdf12-concepts/#server-considerations section propose approaches for serializing RDF 1.2 to RD F1.2 basic and further encoding directional literals when serializing a graph or dataset which uses features incompatible with a requested version.

### Preferred solution

This issue proposes adding a writer setting to Rio writers to perform these downgrade transformations during statements serialization. 

The settings would allow writers to serialize output as RDF 1.2 Basic by encoding triple terms as proposition forms, or as RDF 1.1, where triple terms are also encoded as proposition forms and directional language-tagged strings are encoded using the [i18n namespace](https://www.w3.org/TR/json-ld11/#the-i18n-namespace).

### Are you interested in contributing a solution yourself?

Yes

PR #5950, opened 2026-07-13
GH-5949 Implement serialization to rdf 1.2 basic and rdf 1.1

ivairali:
GitHub issue resolved: #5949

Implementation of basic encode algorithm (RDF 1.2 full -> RDF 1.2 - basic) and conversion of direction literals using the I18N namespace ("مكتبة"^^i18n:ar-eg_rtl). The conversion to RDF 1.1 uses the proposition forms as in the basic encode algorithm, but in addition also converts the direction literals. 

Introduced BasicWriterSetting.RDF_OUTPUT_VERSION property and VersionLabel enum: "1.2", "1.2-basic" and "1.1" where "1.2" is the default. 
Enhanced the AbstractRDFWriter class to handle the downgrade of the output when the property is set via dedicated methods in the Statements class.  Created a helper class to hold the conversion state, that is the blank nodes assigned for each triple term. Created DatatypeHandler I18N so RDF Parsers can deserialize successfully the RDF downgraded to 1.1. 

----

 - [v ] my pull request is [self-contained](https://rdf4j.org/documentation/developer/merge-strategy/#self-contained-changes-pull-requests-and-commits)
 - [ v] I've added tests for the changes I made
 - [v ] I've applied [code formatting](https://github.com/eclipse/rdf4j/blob/main/CONTRIBUTING.md#code-formatting) (you can use `mvn process-resources` to format from the command line)
 - [ v] I've [squashed](https://rdf4j.org/documentation/developer/squashing) my commits where necessary 
 - [ v] every commit message starts with the issue number (GH-xxxx) followed by a meaningful description of the change


hmottestad:
@ivairali please merge develop into this branch

core/model/src/test/java/org/eclipse/rdf4j/model/util/StatementsTest.java
```diff
@@ -158,4 +164,201 @@ public void testStatement_NullContext() {
...
+		Model expectedModel = new LinkedHashModel();
+		expectedModel.add(reifier1, RDF.REIFIES, bnode, context);
+		expectedModel.add(bnode, RDF.TYPE, RDF.PROPOSITION_FORM, context);
+		expectedModel.add(bnode, RDF.PROPOSITION_FORM_SUBJECT, subject, context);
+		expectedModel.add(bnode, RDF.PROPOSITION_FORM_PREDICATE, predicate, context);
+		expectedModel.add(bnode, RDF.PROPOSITION_FORM_OBJECT, object, context);
+
+		Model preservedModel = new LinkedHashModel();
+		RDFVersionsConversionContext rdf12ConversionContext = new RDFVersionsConversionContext();
+		Statements.convertRDFTo12Basic(vf, st1, preservedModel::add, rdf12ConversionContext);
```
desislava-hristova-ontotext: The tests says: testStatementRDF12ConvertTo11IsIdempotent, but you call convertRDFTo12Basic inside, is that intentional?

core/rio/api/src/test/java/org/eclipse/rdf4j/rio/RDFWriterTest.java
```diff
@@ -1856,6 +1857,111 @@ public void testReificationConversion() throws IOException {
...
+				.anyMatch(st -> st.getObject().isTripleTerm()));
+
+		assertTrue(parsedOutput.contains(null, RDF.TYPE, RDF.PROPOSITION_FORM));
+		IRI datatype = vf.createIRI("https://www.w3.org/ns/i18n#he_rtl");
+		assertTrue(parsedOutput.contains(null, RDF.PROPOSITION_FORM_OBJECT, vf.createLiteral("שלום", datatype)));
+	}
+
+	@Test
+	public void testRDF12FullTo12BasicConversion() throws IOException {
+		Model model = new LinkedHashModel();
```
desislava-hristova-ontotext, line 1916: We can add the directional literal here and verify that with basic conversion it is untouched.

hmottestad:
I had Codex review this PR and is came up with the following list. Do any of them make sense?

P1 — Valid triple-term objects with non-rdf:reifies predicates fail. [Statements.java:407-416](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/model/src/main/java/org/eclipse/rdf4j/model/util/Statements.java#L407-L416) rejects them and [:444](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/model/src/main/java/org/eclipse/rdf4j/model/util/Statements.java#L444) hard-codes rdf:reifies. Thus :s :p <<(:a :b :c)>> throws instead of becoming :s :p _:b. The [W3C basic-encoding algorithm](https://w3c.github.io/rdf-interop/spec/) applies to every triple-term object, not only rdf:reifies. Preserve the original predicate and remove this guard.

P1 — The conversion map is incorrectly shared across named graphs. [AbstractRDFWriter.java:114-121](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/AbstractRDFWriter.java#L114-L121) creates one context for the whole document, while [RDFVersionsConversionContext.java:28-34](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/model/src/main/java/org/eclipse/rdf4j/model/util/RDFVersionsConversionContext.java#L28-L34) keys only by triple term. The same term in graphs g1 and g2 gets one blank node, with proposition-form statements emitted only in the first graph. The W3C spec requires encoding each graph independently and says generated blank nodes must not be reused across dataset graphs.

P1 — NDJSON-LD resets the conversion map every 1024 statements. [NDJSONLDWriter.java:63-79](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/jsonld/src/main/java/org/eclipse/rdf4j/rio/ndjsonld/NDJSONLDWriter.java#L63-L79) creates a new context per buffer; the default flush size is 1024 ([BufferedGroupingRDFHandler.java:33-51](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/api/src/main/java/org/eclipse/rdf4j/rio/helpers/BufferedGroupingRDFHandler.java#L33-L51)). However, the parser treats the whole stream as one document with one BNode context ([NDJSONLDParser.java:42-45](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/jsonld/src/main/java/org/eclipse/rdf4j/rio/ndjsonld/NDJSONLDParser.java#L42-L45)). Repeated triple terms across a flush boundary receive different blank nodes and duplicate proposition forms.

P2 — Valid no-language i18n datatypes are mishandled. i18n#_rtl is valid, but [I18NDatatypeHandler.java:59-75](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/datatypes/src/main/java/org/eclipse/rdf4j/rio/datatypes/I18NDatatypeHandler.java#L59-L75) passes an empty language tag to BCP47 validation. Conversely, i18n#rtl is accepted even though the _ separator is required. The [JSON-LD i18n datatype definition](https://w3c.github.io/json-ld-syntax/#the-i18n-namespace) covers the optional-language form.

P2 — Directional-language downgrade does not lowercase language tags. [Statements.java:447-455](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/model/src/main/java/org/eclipse/rdf4j/model/util/Statements.java#L447-L455) inserts the original language string into the datatype IRI. AR-EG becomes i18n#AR-EG_rtl, contrary to the JSON-LD normalization rule.

P2 — RDF_OUTPUT_VERSION is not advertised consistently. The conversion is implemented centrally, but supported-settings lists omit it in at least [N-Triples/N-Quads](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/ntriples/src/main/java/org/eclipse/rdf4j/rio/ntriples/NTriplesWriter.java#L152-L157), [Binary](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/binary/src/main/java/org/eclipse/rdf4j/rio/binary/BinaryRDFWriter.java#L103-L111), and legacy JSON-LD ([JSONLDWriter.java:191-202](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/jsonld-legacy/src/main/java/org/eclipse/rdf4j/rio/jsonld/legacy/JSONLDWriter.java#L191-L202)). Direct set() works, but clients cannot discover support through the Rio API.

ivairali:
@hmottestad I have checked the comments and they make sense. I have introduced fixes in the code and changes in tests. The only thing I'd like to point out is the last comment:

2. "P2 — RDF_OUTPUT_VERSION is not advertised consistently. The conversion is implemented centrally, but supported-settings lists omit it in at least [N-Triples/N-Quads](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/ntriples/src/main/java/org/eclipse/rdf4j/rio/ntriples/NTriplesWriter.java#L152-L157), [Binary](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/binary/src/main/java/org/eclipse/rdf4j/rio/binary/BinaryRDFWriter.java#L103-L111), and legacy JSON-LD ([JSONLDWriter.java:191-202](https://github.com/eclipse-rdf4j/rdf4j/blob/ba38d35b89dac80ba741e86a97a1439960b6a28f/core/rio/jsonld-legacy/src/main/java/org/eclipse/rdf4j/rio/jsonld/legacy/JSONLDWriter.java#L191-L202)). Direct set() works, but clients cannot discover support through the Rio API."

The legacy JSONLD does not support the property at this point, because this writer calls `JSONLDInternalRDFParser serialiser` which uses `jsonldjava` library, and the later does not support RDF 1.2 direction-aware language-tagged literals. So the writer can support 1.2 to 1.1 conversion, because the transformed statements will have neither 1.2 triple terms nor directional literals, thus are processed successfully by the external library. But this is not a full downgrade support, because 1.2 to 1.2 basic conversion preserves directional literals which cannot be processed.

For the rest of the writers I have made the Rio API consistent for the users.

hmottestad:
Hi Ralitsa,

I took a closer look at this and put the regression tests and some proposed fixes on a separate branch so I would not rewrite your PR branch:

https://github.com/eclipse-rdf4j/rdf4j/tree/GH-5949-review-tests-and-fixes

I kept all of the regression tests in the first new commit after the existing branch history, followed by one focused commit per proposed fix. The tests cover deeply nested triple-term conversion, ambiguous mixtures of RDF 1.2 Full and Basic encodings, output-version configuration and announcements, and directional literals in the legacy JSON-LD writers.

Commits:

- Tests: [`45140b8ba8` — Add serialization regression tests](https://github.com/eclipse-rdf4j/rdf4j/commit/45140b8ba821d80140f13120a5fca7b291e326fc)
- Proposed fix: [`d3a4749444` — Convert nested triple terms recursively](https://github.com/eclipse-rdf4j/rdf4j/commit/d3a4749444e43b2fb454f06fa76535496e1976e4)
- Proposed fix: [`f4e4ffe1bb` — Reject ambiguous RDF graph encodings](https://github.com/eclipse-rdf4j/rdf4j/commit/f4e4ffe1bbf7d1430a20bb0fcac8c8a5bfabf5d2)
- Proposed fix: [`cf37b2a2ce` — Parse configured RDF output versions](https://github.com/eclipse-rdf4j/rdf4j/commit/cf37b2a2ce4eadd30dd74e6f7dcc8a6339f5efdd)
- Proposed fix: [`d7ebca3474` — Announce configured RDF output versions](https://github.com/eclipse-rdf4j/rdf4j/commit/d7ebca3474bd8467c153254228c87e6a6c0fb5ba)
- Proposed fix: [`f6edf12f24` — Preserve legacy JSON-LD direction](https://github.com/eclipse-rdf4j/rdf4j/commit/f6edf12f245ab21096b61598015aeaadfa71fa5a)
- Proposed fix: [`dfcfa0b2ec` — Advertise legacy RDF output versions](https://github.com/eclipse-rdf4j/rdf4j/commit/dfcfa0b2ec9d18acc49c54d92f174d7c13a8c50a)

With these changes, 3,150 tests pass across the affected and neighboring modules. I also found one Binary parser test failure, but it reproduces unchanged on current `develop`, so I have not included an unrelated fix for it here.

Could you take a look and see whether you agree with the tests and the proposed behavior/fixes? I kept each fix separate so they should be straightforward to review, adjust, or cherry-pick individually.

ivairali:
Hi Håvard,

Thank you, I checked the code and I think all commits make sense, so I cherry picked them.

Issue #5888, opened 2026-06-22
Backup functionality for LMDB


m-alreeni:
### Problem description

LMDB provides already  a set of [command line tools](http://www.lmdb.tech/doc/tools.html) that can be used to backup and restore the value and triple databases. A  global and store-agnostic backup system is to be implemented that provides addtional functionalities and works for other RDF4J SAILs.

### Preferred solution

`BackupCapableSail` API and an integrated store-agnostic backup service `SailBackupService` with:
- online full snapshots of LMDB value/triple data and auxiliary store files
- incremental backup packaging based on transaction delta logs
  -  copies per-transaction delta files 
  - it does use a WAL-like log for backup, but it is a custom backup delta log,
  - the deltas are written after commit via a commit listener and are used for incremental backup/restore replay 
- point-in-time restore by replaying transaction deltas from a base snapshot
  - loads the most recent full backup at/before the target txn ID, then replays those delta files in order.
- full backup takes an online snapshot of the LMDB store files (triples/, values/, plus extra store files), optionally zips it, and records metadata/checksum.
- retention and periodic backup scheduling support
 - Distributed replication using Ratis/Raft 
- ...

### Are you interested in contributing a solution yourself?

Yes

kenwenzel:
@m-alreeni Can you give a short explanation how the backup functionality will work?
Does it use some kind of WAL for operations logging?

PR #5951, opened 2026-07-15
GH-5888 backup functionality for lmdb

m-alreeni:
GitHub issue resolved: #5888

adds a local LMDB backup system that can create online full snapshots and incremental backups from committed transaction logs while the store stays live. Full backups capture the current LMDB state; incremental backups store the transaction deltas since a chosen transaction ID, which enables point-in-time restore by replaying deltas on top of the latest full backup.
It also supports ZIP compression, scheduled backups, retention cleanup, and integrity/status reporting so operators can verify backups and see recent failures or successful runs.

Still  missing or only partially covered:
- Deduplication: the current code has retention cleanup, but not true backup dedupe.
- A distributed backup layer (such  Ratis/Raft)  built on top of the local backup work.

----


kenwenzel:
@m-alreeni What happens if an auto-grow happens during backup? Is this already correctly handled?

m-alreeni:
> What happens if an auto-grow happens during backup? Is this already correctly handled?

@kenwenzel  
- If backup is copying the LMDB environment and an auto-grow is triggered, the auto-grow cannot proceed concurrently with the copy.
- If auto-grow is in progress, the backup copy waits until the resize completes.
- Likewise, if the backup starts first, the resize blocks until the snapshot copy finishes.

There is also a higher-level guard in LmdbSailStore.createOnlineSnapshot(...): it takes sinkStoreAccessLock before copying both the triple store and value store environments, so no store mutation path is active during the snapshot.
So: the locks serialize the two operations. The snapshot is consistent as a point-in-time view, and the resize cannot happen mid-copy.

 Is this correctly handled or is there a better way to do this?
I'll check it again for errors or another approach. if you have some ideas or find bugs, please tell me

kenwenzel:
@m-alreeni Can you write a tailored test that concurrently inserts and deletes data while backup is in progress?

PR #5974, opened 2026-08-13
Fix transaction and value pooling in LmdbStore + fix initialization of nextId in value store.

### LINKED ISSUE #5970: LmdbStore: native heap corruption aborting the JVM; committed and readable data lost without restart; `ValueStore$ReadTxn` ThreadLocal leaks under aborted-query load (5.3.1 / 5.3.2)
We were experiencing some issues and together with Claude Code came up with a workaround and the below analysis and bug report. I wasn't able to manually double-check all these points, I hope they make sense, but happy to do follow-up work to better pinpoint / reproduce this.

*(Edited after filing to follow the bug report template — content unchanged apart from being reorganised into these sections.)*

### Current Behavior

Operating a fleet of RDF4J servers (`eclipse/rdf4j-workbench:5.3.1-tomcat`, later `5.3.2-tomcat`) with ~100+ LmdbStore repositories per instance, we observed three related defects under sustained aborted-query load. We have detailed, timestamped evidence from production for all of them, including one capture that rules out server restarts as a necessary condition for data loss.

1. **LMDB read transactions leak onto Tomcat worker threads** and are still present in 5.3.2 (i.e. after #5807).
2. **A committed write transaction — acknowledged with 2xx over the HTTP repository protocol and subsequently readable — can vanish**, while earlier and later transactions on the same store persist. In one captured case this happened on a continuously running server, with no restart between the verified read and the loss.
3. **The JVM is aborted by glibc with native heap corruption** (`malloc_consolidate(): invalid chunk size` / `unaligned fastbin chunk detected`), repeatedly and under normal operation. This is new evidence from 2026-08-05 and, we think, the most significant of the three: it is direct proof of native memory corruption inside the server process, which would plausibly explain defect 2 rather than merely coexisting with it.

#### Environment

- `eclipse/rdf4j-workbench:5.3.1-tomcat`, upgraded to `5.3.2-tomcat` mid-investigation (both affected; the leak evidence below is from 5.3.2)
- Apache Tomcat 9.0.119 on Eclipse Adoptium JVM 25.0.3+9-LTS, Linux 5.15 x86_64, stock image glibc allocator (Ubuntu GLIBC 2.35-0ubuntu3.13)
- LmdbStore repositories created via the REST config API:
  `sail:sailType "rdf4j:LmdbStore"`, `lmdb:tripleIndexes "spoc,posc,ospc,cspo,cpos,cosp"`, `sail:iterationCacheSyncThreshold "10000"`, `sb:defaultQueryEvaluationMode "STANDARD"` (LMDB durability options at defaults)
- ~100+ repositories per server; a few of them very hot (both reads and writes)
- Writers: RDF4J `HTTPRepository` clients using `IsolationLevels.SERIALIZABLE` transactions
- Readers: SPARQL over the REST API, plus federated `SERVICE` clauses that the server resolves against its own repositories via an HTTP loopback
- Client HttpClient settings on the server (for SERVICE evaluation): `socketTimeout=10000`, `connectionTimeout=10000`, `maxConnPerRoute=100`, `maxConnTotal=120`
- Docker; store data on a local volume

#### Defect 1: `ValueStore$ReadTxn` ThreadLocal leaks (observed on 5.3.2)

After a period of sustained client-aborted queries (see "Steps To Reproduce"), a Tomcat webapp stop reports dozens of leaked ThreadLocals on `http-nio` worker threads:

```
29-Jul-2026 13:51:19.868 SEVERE [Thread-1] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks
  The web application [rdf4j-server] created a ThreadLocal with key of type [java.lang.ThreadLocal.SuppliedThreadLocal]
  ... and a value of type [org.eclipse.rdf4j.sail.lmdb.ValueStore$ReadTxn] ... but failed to remove it when the web
  application was stopped.
```

— repeated ~50× for `ValueStore$ReadTxn`, interleaved with the same message for `org.lwjgl.system.MemoryStack` (native allocations). This looks like the same leak class as #5807 (fixed in 5.3.2 for the reported path), surviving via a different path: read transactions held in ThreadLocals on request-processing threads that are never cleaned when the request is aborted.

Operational consequence we repeatedly hit before 5.3.2 (and suspect can still be reached): individual hot repositories become unusable ("wedged") while sibling repositories on the same server keep working; recovery required a Tomcat restart.

#### Defect 2: acknowledged-and-readable commits lost

Context: our application performs a per-record write into several repositories, each write being one SERIALIZABLE transaction over the HTTP repository protocol. The same transaction also maintains a monotonically increasing count and an order-independent XOR checksum in the repository, so the internal consistency of a repository can be verified independently of our client. A separate reconciliation process re-reads written records through fresh connections and independent query patterns.

**Case A — loss bracketed by restarts (multiple occurrences, 5.3.1 and 5.3.2)**

- 12:07:06 — Tomcat restarted (external healthcheck)
- 12:08:28 — write transaction committed (2xx); the data was subsequently **readable**: an independent verification pass (fresh connection, different query pattern) confirmed the record present and advanced past it
- 12:11:56 — Tomcat killed again (SIGTERM, then SIGKILL 10 s later)
- afterwards — the record is absent from that repository: no data triples, no bookkeeping triples; count and checksum are internally consistent **as if the transaction never happened**, while transactions committed later than it persist

The same record was independently lost on two separate server instances that ingested it at the same wall-clock time (both under comparable load), and on a third instance under calmer conditions the same write survived — so the record's content is not the variable.

**Case B — loss with no restart (captured once, 5.3.2)**

Application log (timestamps UTC), talking to a continuously running 5.3.2 server:

```
13:43:58.001  detected record missing from repository X; re-writing
13:43:58.049  re-write committed (2xx) and verification read succeeded
```

By ~13:52 the record was **absent again**. The server's Tomcat logs show **no webapp stop/start events in the surrounding two hours** (log stream continuous), i.e. the committed and read-back-verified transaction reverted on a running server. As in case A, the loss is surgical: the whole transaction vanished as a unit (data + bookkeeping), the repository's count/checksum remained self-consistent without it, and both earlier and later commits persist.

After this state is reached once, the repository appears permanently prone to it: repeated re-writes of the missing record kept reverting the same way, while the identical procedure on a freshly created store held durably. Wiping the data directory and re-loading fully cures the repository.

**Working hypothesis (ours, unverified)**

The observations fit a scenario in which a damaged environment (e.g. stale/corrupt reader-lock table state accumulated under the aborted-query storms and hard kills of defect 1's conditions) allows two writers to proceed from stale roots, with the last committer's meta page orphaning the earlier commit — which would produce exactly the "one transaction vanishes mid-lineage, later ones persist, bookkeeping consistent" signature. We have not verified this at the LMDB level and would welcome guidance on what to capture if it recurs (we now detect recurrence within ~5 minutes with precise record identity and timestamps, and can preserve `docker logs` and the store directory before any restart).

#### Defect 3: native heap corruption aborts the JVM (observed on 5.3.2)

Captured 2026-08-05 from `docker logs` of a continuously-running production instance. The JVM is killed by glibc's allocator, twice within four minutes, each time seconds after a clean start:

```
2026-08-05T03:49:29.795Z  malloc_consolidate(): invalid chunk size
2026-08-05T03:49:30.247Z  Aborted (core dumped)
   ... container restarts, "Server startup in [6667] milliseconds" at 03:49:38 ...
2026-08-05T03:53:45.118Z  malloc_consolidate(): unaligned fastbin chunk detected
2026-08-05T03:53:45.761Z  Aborted (core dumped)
```

Both messages are glibc detecting that its own heap metadata has been overwritten — an out-of-bounds write or double free in native code, not a Java-level fault. No JVM crash log (`hs_err_pid*`) is produced, because the abort happens in the allocator rather than via the JVM's own signal handling.

Frequency on this instance: **97 container restarts in 4 days**, and in a 36-hour window the Tomcat log shows **61 startups against only 14 graceful shutdowns** — i.e. ~47 aborts of this kind. The instance was not under unusual load; this is its steady state.

We consider this strong support for defect 2 being a genuine store-corruption bug rather than a client-side illusion: a process whose native heap is already corrupt can plausibly write a malformed LMDB meta page. It also fits the `org.lwjgl.system.MemoryStack` ThreadLocal leaks reported under defect 1 — lwjgl is the native-memory layer under the LMDB binding, and both symptoms point at the same area.

**Possibly related: heap exhaustion killing Tomcat's acceptor**

On the same instance the JVM also ran out of *Java* heap:

```
04:17:08  Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "http-nio-8080-Acceptor"
04:17:30  Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "http-nio-8080-Poller"
04:17:59  SEVERE [Catalina-utility-1] ... java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Java heap space
```

We flag this only as an observation, not as a claim against RDF4J: the heap was left at the JVM default and we do not yet have a heap dump, so we cannot say whether the growth was a leak in the server or simply an undersized heap. We have since enabled `-XX:+HeapDumpOnOutOfMemoryError` and will attach a dump if it recurs.

(Operationally this was the worst of the failure modes: with the acceptor thread dead the listening socket stayed bound but nothing accepted, so every client saw a connect timeout while the process stayed alive — no restart policy or healthcheck could observe an exit. That is our problem to handle, and we have, via `-XX:+ExitOnOutOfMemoryError`; we mention it only so the timestamps in the log excerpts above are interpretable.)

### Expected Behavior

1. Client-aborted or timed-out queries release their LMDB read transactions and native allocations; no `ValueStore$ReadTxn` / `MemoryStack` ThreadLocals survive on worker threads, and repositories do not become permanently unusable under this load.
2. A transaction acknowledged as committed over the HTTP repository protocol — and subsequently read back successfully — stays committed. It is never orphaned while transactions committed after it persist.
3. The server process is never aborted by the allocator with native heap corruption.

### Steps To Reproduce

We cannot yet provide a minimal reproducer — this is production evidence. The conditions that preceded the damage were:

1. High-rate SPARQL with many **client-aborted / timed-out requests** (the server log shows continuous `ClientAbortException: Broken pipe` / "Cannot call sendError() after the response has been committed" storms).
2. Federated `SERVICE` clauses resolved by the server against itself over an HTTP loopback (10 s socket timeout on that client — so each timeout aborts a server-side query).
3. Repeated hard restarts of Tomcat (SIGKILL after a 10 s grace) by an external healthcheck reacting to wedged repositories.
4. Concurrent writers into the affected repositories (4 client threads, SERIALIZABLE transactions) during those windows.
5. ~100+ LmdbStore repositories in one server, a few of them very hot.

What we ruled out on our side:

- **Client-side atomicity or bookkeeping bugs**: count/checksum are written in the same transaction as the data and stayed exactly consistent with the record absent; verified by independent recomputation.
- **Read-side illusions**: presence-then-absence was confirmed through separate connections and four distinct SPARQL access patterns (VALUES lookup, subject enumeration, FILTER match, direct triple pattern), all agreeing at each point in time.
- **Restart-required loss** (for case B): log-verified absence of webapp stop events in the window.
- **Record-content pathology**: identical records survive on other instances and on rebuilt stores.

### Version

5.3.2 (previously 5.3.1; both affected). Via `eclipse/rdf4j-workbench:5.3.2-tomcat`, Tomcat 9.0.119, Adoptium JVM 25.0.3+9-LTS.

Perhaps? — we are not in a position to fix LMDB internals ourselves, but we can do substantial legwork: run instrumented or patched builds in production (the abort reproduces on its own within hours), validate a fix branch such as #5812 under a real workload, and capture whatever diagnostics you ask for at the moment of recurrence.

### Anything else?

Full investigation timeline with queries, checksums, and logs: knowledgepixels/nanopub-query#142.

**Existing issues this may confirm.** While preparing this report we found open issues that appear to describe the same mechanisms. We are not claiming these are the root cause — but if they are, this report may be useful as production confirmation, since the conditions here (query timeouts concurrent with writes) match all of them:

- **#4806** — `TimeLimitIteration` calls `close()` asynchronously from another thread; explicitly notes that this "affects the `LmdbRecordIterator` that causes a *A heap has been corrupted* exception". Our defect 3 is exactly a native heap corruption arising in a workload dominated by timed-out queries.
- **#5810** / PR #5812 — LMDB read transactions can reset while `RecordIterator` cursors are still active.
- **#5965** — `org.eclipse.rdf4j.sail.lmdb.Pool` is a ThreadLocal holding native `MDBVal[]`, so a connection closed on a different thread can double-free and segfault. This would also explain the `MemoryStack` ThreadLocal leaks in our defect 1.
- **#5960** — NativeStore corruption after `QueryInterruptedException` on timeout, affecting *earlier, unrelated* records committed in successful transactions. That is a very close description of our defect 2, on a different store.

We do not believe 6.0.0 changes any of the above (the "recover LMDB readers" work in 6.0.0 is #5807, which we already have), but please correct us if a 6.0.0 change makes any of these obsolete — we can plan an upgrade if it is expected to help.

**Our questions:**

1. Is the `ValueStore$ReadTxn` ThreadLocal leak on aborted requests a known remaining path after #5807?
2. Is there a known mechanism by which a committed LMDB transaction can be orphaned as described — and would the corruption path in #5960 have an LMDB equivalent?
3. For defect 3: do #4806 / #5810 / #5965 plausibly account for glibc-level heap corruption in the LMDB/lwjgl layer under aborted read transactions, and is a fix for any of them in progress? If you can say which instrumentation would be most informative — a run under `MALLOC_CHECK_`/`glibc.malloc.check`, or a build with the lwjgl debug allocator enabled — we can run it in production. We can also preserve a core dump; the ones so far were discarded by the container's default `core_pattern`.
4. The damaged stores have since been rebuilt (a wipe-and-reload fully cures the behavior), so we can no longer provide post-mortem diagnostics from them. However, our reconciliation tooling now detects a recurrence within ~5 minutes, with the affected record's identity and timestamps — if it happens again we will preserve the store directory and full server logs before any restart. What diagnostics (e.g. `mdb_stat` / reader-table inspection, specific loggers at debug) would be most useful to capture at that moment?

kenwenzel:
GitHub issues resolved: #5970 #6022 

Briefly describe the changes proposed in this PR: Instead of using a thread-local pool that might not be correctly freed this PR associates transaction references with value pools.
It also fixes initialization of `nextId` in ValueStore.

----


core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -25,290 +24,450 @@
...
+	 */
+	private void closePooledReaders() {
+		if (txnPool == null) {
+			return;
+		}
+		Txn txn;
+		boolean aborted = false;
+		while ((txn = txnPool.poll()) != null) {
+			if (open.remove(txn)) {
+				txn.abortAndMarkClosed();
```
hmottestad, line 418: **[P1] Synchronize native transaction disposal**

During `MDB_READERS_FULL` recovery, this can abort a transaction while its releasing thread is still inside `resetNative()`: `release()` publishes the transaction before finishing that reset, and `abortAndMarkClosed()` does not acquire its monitor. Since [LMDB abort frees the handle](https://raw.githubusercontent.com/LMDB/lmdb/LMDB_0.9.33/libraries/liblmdb/mdb.c), this permits native use-after-free and JVM crashes. Serialize disposal with the transaction’s other lifecycle operations.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -25,290 +24,450 @@
...
-	Txn createReadTxnUntracked() throws IOException {
-		Txn txnRef = new Txn(createReadTxnInternal());
-		synchronized (active) {
-			active.put(txnRef, Boolean.FALSE);
-		}
-		return txnRef;
-	}
+	Txn createReadTxnInternal(boolean resetOnWrite) throws IOException {
+		checkNotClosed();
+		acquireReaderPermit();
```
hmottestad: **[P1] Account for nested reader acquisitions**

`SailSourceConnection` acquires dataset transactions before optimization, and cardinality estimation subsequently borrows another transaction through `doWith()`. When datasets occupy all 128 permits, those queries cannot complete optimization without waiting for the 30-second admission timeouts. `doWith()` also holds the manager’s read lock while waiting, delaying commits. Reserve capacity for these nested operations or reuse the query’s existing transaction.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -25,290 +24,450 @@
...
+			}
+		}
+
+		// Poison the semaphore: wake every admission waiter, which then fails fast in acquireReaderPermit().
+		// Permit accounting is irrelevant from here on, the manager is closed.
+		readerSlots.release(POOL_SIZE);
+		signalReaderInactive();
+
+		for (Pool pool : pools) {
+			pool.close();
```
hmottestad, line 282: **[P2] Close the triple store’s new buffer pools**

This is now the cleanup path for manager-owned `MDBVal` and key-buffer allocations, but `TripleStore.close()` never calls its transaction manager’s `close()`. Only `ValueStore` does. With `Pool.release()` removed from connection cleanup, buffers returned by ordinary statement/context iterators now leak across repository shutdowns. Call the triple store manager’s `close()` before closing its LMDB environment.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
```diff
@@ -1290,7 +1290,17 @@ long findTripleTermId(long subj, long pred, long obj, boolean create) throws IOE
 	public RecordIterator getTripleTerms(long subj, long pred, long obj) throws IOException {
 		TripleIndex index = TripleIndex.getBestIndex(tripleTermIndexes, subj, pred, obj, -1);
 		boolean doRangeSearch = index.getPatternScore(subj, pred, obj, -1) > 0;
-		return new LmdbRecordIterator(index, doRangeSearch, subj, pred, obj, -1, true, txnManager.getReadTxn());
+		Txn txn = txnManager.createReadTxn();
+		return new LmdbRecordIterator(index, doRangeSearch, subj, pred, obj, -1, true, txn) {
```
hmottestad: **[P2] Return the transaction when iterator construction fails**

If `LmdbRecordIterator` construction throws, for example when interrupted while waiting for its transaction lock, the newly borrowed transaction is never closed. The override below only helps after construction succeeds. Each failure permanently consumes a reader permit until store shutdown, eventually making value lookups time out. Close the transaction on construction failure.

core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreMultiRepoTest.java
```diff
@@ -0,0 +1,159 @@
...
+									.evaluate()) {
+								var count = result.stream().toList().size();
+								System.out.println(
+										"Repository " + repositoryIndex + ": Current size: " + connection.size()
+												+ ", distinct subjects: " + count);
+							}
+						} catch (Exception e) {
+							System.err.println(
+									"Repository " + repositoryIndex + ": Error checking size: " + e.getMessage());
+							e.printStackTrace(System.err);
```
hmottestad: **[P2] Propagate concurrent read failures to the test**

The scheduled query catches and prints every `Exception`, so the test can pass despite repeated reader failures while writes and the final size check succeed. Capture these failures and assert them on the test thread after the executor terminates.

hmottestad (commented):
I found five issues in the transaction and value pooling changes; details are attached inline. This was a source review against `main` at `b4379dcaabc5f25f0ec54a037d09555504be5087`. I did not run builds or tests.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: @hmottestad Do you understand what is meant here?

hmottestad: I think it is more related to the cardinality estimator. A query can take a permit for its dataset, then block because it needs another for the cardinality estimator when optimizing the query. I guess maybe the cardinality estimator shouldn't need to occupy a slot since it doesn't really care if the cardinalities are slightly off due to being either ahead or behind the current transaction.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
hmottestad: This should just be try with catch Throwable to close the txn if there is an issue

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: Addressed by making abortAndMarkClosed synchronized.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: close() is now called in TripleStore.close()

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
kenwenzel: Wrapped it with a try-catch block

core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreMultiRepoTest.java
kenwenzel: Exceptions are now collected and thrown if anything happens

kenwenzel (commented):
I reviewed the PR and the branch. Summary:
- The PR refactors LMDB transaction pooling, replaces thread-local pools with pooled buffers, improves reader admission and readers-full retry, and fixes nextId initialization in ValueStore. Several comments were implemented and tests added.

What is done
- Synchronized native transaction disposal to avoid native use-after-free.
- TripleStore now closes the TxnManager so pools/buffers get freed.
- Tests collect and assert scheduled reader failures.
- Pool now uses a lock-free ring buffer and frees native memory in close.
- nextId initialization in ValueStore fixed.

Changes still required (please address before merging)
1) Nested reader acquisitions / lock-order:
- doWith() holds the manager read lock and then waits for a reader permit; this can block while holding the read lock and delay commits. Fix by acquiring the reader (createReadTxn/acquire permit) before taking the manager read lock, or reuse a caller’s existing Txn for nested acquisitions, or reserve capacity for nested ops. See TxnManager.doWith() and createReadTxnInternal().

2) Ensure txn is always returned when iterator construction fails:
- ValueStore.getTripleTerms closes txn on IOException, but not on unchecked exceptions. Wrap iterator construction with catch(Throwable) and always close txn before rethrowing, to avoid permanently leaking a reader permit. Apply this pattern to all similar sites.

Other: run a search for createReadTxn usages followed by object construction and ensure all are safe.

Overall: this PR is close and fixes many issues, but the two outstanding items are resource/deadlock risks and should be addressed before merge.

hmottestad:
I added a new priority mode so that we can guarantee that an RDF4J transaction that opens a dataset (which acquires a read permit) and then runs a query (which requires a second read permit for the cardinality estimator) is guaranteed to be able to get the second reader permit. 

Previously we could have 127 transactions in use already, then if we started one more transaction we would have filled the last slot and if we ran a SPARQL query we would then stall waiting for one of the 127 other transactions to finish. Now we always leave one free for a quick cardinality estimation. 

The difference is that the read permit that is used in a dataset won't be released until the user commits or rollbacks their rdf4j transaction. While the read permit used in the cardinality estimator is returned back by our code, without the user being able to stop it.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
```diff
@@ -290,8 +291,7 @@ class ValueStore extends AbstractValueFactory {
 						rc = mdb_cursor_get(cursor, keyData, valueData, MDB_PREV);
 					}
 					if (rc == MDB_SUCCESS && keyData.mv_data().get(0) == ID_KEY) {
-						// remove lower 2 type bits
-						nextId = Math.max(nextId, (data2id(keyData.mv_data()) >> 2) + 1);
+						nextId = Math.max(nextId, ValueIds.getValue(data2id(keyData.mv_data())) + 1);
```
hmottestad, line 298: **[P1] Include triple-term IDs when restoring `nextId`**

This scans only `dbi` and `freeDbi`, while `findTripleTermId()` allocates from the same counter and stores those IDs exclusively in the term indexes. A public reproduction is to begin a transaction, add `(a, p, createTripleTerm(a, p, b))`, commit, shut down/reopen the repository, then begin another transaction and add `(b, p, createTripleTerm(b, p, a))`. All scalar components already exist, so the new term receives the existing term's exact ID. After another reopen, `id2tripleTerm()` selects the same first matching term for both statements, making at least one object incorrect.

Please include the term-index maximum when restoring the counter, or persist the allocation watermark. The corrected shift needs that additional source of allocated IDs.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
```diff
@@ -25,290 +25,479 @@
...
+	Txn createReadTxnInternal(boolean resetOnWrite) throws IOException {
+		return createReadTxnInternal(resetOnWrite, false);
 	}
 
-	private void waitForTrackedReaderToClose(Txn excludedTxn) throws IOException {
-		synchronized (active) {
-			if (hasTrackedReaders(excludedTxn)) {
+	private Txn createReadTxnInternal(boolean resetOnWrite, boolean priority) throws IOException {
+		checkNotClosed();
+		Semaphore readerPermit = acquireReaderPermit(priority);
```
hmottestad, line 173: **[P2] Account for both dataset readers during admission**

There is another nested acquisition before cardinality estimation: `UnionSailSource.dataset()` obtains the inferred dataset and then the explicit dataset, and both `LmdbSailDataset` constructors request ordinary permits. Concurrent callers can consume all 127 permits with their first dataset and then stall until the 30-second admission timeout while trying to acquire the second. The reserved callback permit is unavailable to this path. This also occurs with no inferred statements: `mayHaveInferred` skips statement iteration only after both datasets have been acquired.

Please make admission account for the complete dataset acquisition, for example by sharing one transaction between the two views or acquiring the required readers together. Giving a long-lived dataset the cardinality reserve would defeat the reserve's purpose.

core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreMultiRepoTest.java
```diff
@@ -0,0 +1,173 @@
...
+		ScheduledExecutorService scheduledExecutor = Executors.newScheduledThreadPool(200);
+		ExecutorService executor = Executors.newCachedThreadPool();
+		Throwable testFailure = null;
+		List<Callable<Void>> tasks = IntStream.range(0, repositoryCount)
+				.mapToObj(repositoryIndex -> (Callable<Void>) () -> {
+					File repositoryDir = new File(dataDir, "repo-" + repositoryIndex);
+					Repository repository = new SailRepository(
+							new LmdbStore(repositoryDir, new LmdbStoreConfig("spoc,posc")));
+					repository.init();
+					repositories.add(repository);
```
hmottestad, line 81: **[P2] Collect initialized repositories thread-safely**

All 200 task lambdas add to the same `ArrayList` declared above. Two concurrent `add()` calls can observe the same size and overwrite the same slot, losing one initialized repository. Joining the futures makes completed writes visible but does not prevent this race. The final `repositories.forEach(Repository::shutDown)` then omits that repository and leaks its LMDB resources. Please use a concurrent collection, synchronize the additions, or collect repositories through task results.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/util/MpmcRingBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+	private final long[] sequences;
+	private final Object[] values;
+
+	private final Sequence enqueuePos = new Sequence();
+	private final Sequence dequeuePos = new Sequence();
+
+	public MpmcRingBuffer(int capacity) {
+		if (Integer.bitCount(capacity) != 1) {
+			throw new IllegalArgumentException("Capacity must be a power of two: " + capacity);
+		}
```
hmottestad, line 72: **[P3] Reject capacity one in the public constructor**

`1` passes the power-of-two check, but this sequence-number scheme requires capacity at least `2`. With capacity `1`, `offer(A)` sets the slot sequence to `1`; `offer(B)` sees that as the next writable position and succeeds, overwriting A and setting the sequence to `2`. `poll()` still expects sequence `1`, so it spins indefinitely.

Please require a positive power-of-two capacity of at least `2`. Existing callers use 128 or 2048, so this is an exposed-constructor edge case.

hmottestad (commented):
I rechecked the current branch against main, including the allocation/lookup paths, nested dataset acquisition, test cleanup, and queue sequence transitions. Four remaining issues are detailed inline. This verification is based on source and call-path analysis; I did not run new tests.

hmottestad:
And yeah, I got an AI to do the analysis and post the comments.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
kenwenzel: Good find - we need to fix this.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: Sharing the transaction would be the best. I'll have to look into it.

core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreMultiRepoTest.java
kenwenzel: I'll fix this.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: @hmottestad I'm not sure if this is easily to fix. Is the lifecycle of SailSource the same as the SailDataset?

To share a transaction between two `SailSource`s the method `SailSourceConnection.branch(...)` needs to be modified. The method is currently private.

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/ValueStore.java
kenwenzel: Fixed

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/TxnManager.java
kenwenzel: Let's fix this with another issue at a later point.

core/sail/lmdb/src/test/java/org/eclipse/rdf4j/sail/lmdb/LmdbStoreMultiRepoTest.java
kenwenzel: Fixed

core/sail/lmdb/src/main/java/org/eclipse/rdf4j/sail/lmdb/util/MpmcRingBuffer.java
kenwenzel: Fixed

PR #5981, opened 2026-08-18
Add remote query cancellation

### LINKED ISSUE #5904: Implement query cancellation via HTTP
### Problem description

When a client aborts an ongoing SPARQL query via a client-side HTTP request (e.g., cancelling a fetch or dropping the TCP connection), the RDF4J server-side worker thread continues to execute the intensive graph traversal to completion because HTTP lacks an inherent mechanism to pass a "cancel" signal down to the JVM thread layer. Currently, there is no standardized, client-accessible HTTP API endpoint in RDF4J to uniquely identify and explicitly terminate a running query session from the client side. This results in unnecessary resource consumption, unkillable runaway queries, and potential server instability under heavy read loads.

### Preferred solution

To solve this, RDF4J can introduce a mechanism that generates a unique Query ID for every incoming execution, exposing a dedicated DELETE /rdf4j-server/queries/{queryId} endpoint to the client. When a client initiates a request, they can either provide or immediately receive this identifier, allowing them to fire a subsequent asynchronous API call to explicitly terminate the operation. Upon receiving this cancellation request, the server would map the ID to the active worker thread and safely trigger an internal interrupt to stop the graph traversal prematurely.

_No response_

_No response_

### Anything else?

_No response_

hmottestad:
Closes #5904.

## Problem

Abandoning a Workbench result or an RDF4J HTTP response does not reliably stop the query that owns the server-side evaluation. Query explanations already had a request-id coordinator, but regular tuple, graph, and boolean queries had no equivalent end-to-end cancellation path. Replicated Workbench or Server deployments also need to distinguish “this replica does not own the query” from successful cancellation so a caller can retry through round-robin routing.

## Implementation

This change extracts the explanation lifecycle into a reusable `CancellableOperationCoordinator` while preserving `AsyncExplainCoordinator` as a compatibility facade. A nestable `QueryRequestContext` propagates a client-generated `query-request-id` through HTTPRepository calls. `RDF4JProtocolSession` sends that id on non-transactional regular queries and exposes the matching cancellation POST, while `HTTPRepository` tracks active protocol sessions and exposes `cancelQuery(String)`.

Server registers regular queries before evaluation and keeps their cancellation handles alive through lazy tuple and graph result serialization. Cancellation interrupts the attached worker, closes its repository connection, and forwards the same id when the repository is HTTP-backed. Controller shutdown drains both query and explanation coordinators. Transaction endpoints explicitly reject the tracking parameters because their connection lifecycle is shared by the transaction.

Workbench retains the query editor as the cancellation controller and opens regular results in a UUID-specific secondary window. Both GET and long-query POST execution carry a fresh UUID. Explanation requests, including compare requests, also receive a fresh UUID. Deriving the result-window name from the query UUID prevents multiple browser tabs from targeting the same named window.

Regular-query and explanation cancellation return HTTP 204 only when the active request id was found. Workbench and Server return HTTP 404 for unknown or completed ids. The Workbench frontend treats HTTP and network failures as retryable and makes up to twenty retries after the initial cancellation request, allowing round-robin routing repeated opportunities to reach the replica that owns the query.

## Verification

The implementation was developed test-first; the failing and passing report snippets are retained in `initial-evidence-GH-5904.txt`.

- `python3 .codex/skills/mvnf/scripts/mvnf.py core/http/client --retain-logs`: 110 tests, zero failures/errors, 8 skipped
- `python3 .codex/skills/mvnf/scripts/mvnf.py core/repository/http --retain-logs`: 12 tests, zero failures/errors
- `python3 .codex/skills/mvnf/scripts/mvnf.py tools/server-spring --retain-logs`: 113 tests, zero failures/errors
- `python3 .codex/skills/mvnf/scripts/mvnf.py tools/workbench --retain-logs`: 389 tests, zero failures/errors
- `tools/workbench/compileTypescript.sh`: passed and regenerated `query.js` plus its source map
- `scripts/checkCopyrightPresent.sh`: passed
- `mvn -o -Dmaven.repo.local=.m2_repo -q -T 2C process-resources`: passed
- `mvn -B -ntp -Dmaven.compiler.showWarnings=false -T 1C -o -Dmaven.repo.local=.m2_repo -Pquick clean install`: BUILD SUCCESS
- `git diff --check`: passed

The implementation plan and the deliberate transaction/JVM-local registry boundaries are documented in `.agent/plans/GH-5904-query-cancellation.md`.


hmottestad:
TODO: We should also implement periodic writes to the socket when the returned format allows for writing characters that will be ignored (comments or whitespace). This would probably work for json and xml, but not for CSV since there are no characters that can be prefixed without changing the meaning of the CSV. Be wary of only writing if we are sure that noting has been written yet. The writing will trigger the connection to be closed if the client does not respond.

Issue #5977, opened 2026-08-16
Turtle parser does not resolve relative references with colons after first slash against base


QuarksToQuasars:
### Current Behavior

Hi,

I think we ran into a turtle parser bug w.r.t. relative URIs such as `<system/x:ABC67>` (please note the `:` **after** the first slash), which is not properly resolved against the base.
After parsing, it is still `<system/x:ABC67>` ... which is valid turtle, but not valid RDF w.r.t. to the abstract syntax, which requires absolute IRIs.

[RFC 3986 4.2 ](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2) only states that colons are not allowed in the "first segment of a relative-path reference", however, a colon in a later segment should be valid. So `<system/x:ABC67>` should be correctly detected as relative reference and resolved during parsing against the base.

> A path segment that contains a colon character (e.g., "this:that")
   cannot be used as the first segment of a relative-path reference, as
   it would be mistaken for a scheme name.  Such a segment must be
   preceded by a dot-segment (e.g., "./this:that") to make a relative-
   path reference.

### Expected Behavior

With, for example, a `@base <http://example.com/>` a relative reference such as `<system/x:ABC67>` in a turtle document should resolve to an absolute IRI `<http://example.com/system/x:ABC67>` during parsing.

### Steps To Reproduce

Failing test case:

`core/rio/turtle/src/test/java/org/eclipse/rdf4j/rio/turtle/TurtleParserTest.java` 

```
	@Test
	public void testRelativeIRIWithColonInLaterPathSegmentIsResolvedAgainstBase() throws IOException {
		// Per RFC 3986 section 4.2, only the *first* path segment of a relative reference must be
		// colon-free; a colon in a later segment is legal and does not make the reference absolute.
		// Such a reference must still be resolved against @base, not stored as-is.
		String data = "@base <http://example.com/> .\n" + "<system/x:ABC67> <urn:pred> <urn:obj> .";

		parser.parse(new StringReader(data), baseURI);

		assertTrue(errorCollector.getFatalErrors().isEmpty());
		assertEquals(1, statementCollector.getStatements().size());
		Statement st = statementCollector.getStatements().iterator().next();
		assertEquals(vf.createIRI("http://example.com/system/x:ABC67"), st.getSubject());
	}
```

### Version

6.0.1-SNAPSHOT

### Are you interested in contributing a solution yourself?

Perhaps?

### Anything else?

[...]

PR #5982, opened 2026-08-19
GH-5977 fix resolveURI logic for relative IRIs

DanchoIvanov:
GitHub issue resolved: # GH-5977

Fixes an issue with the AbstractRDFParser.resolveURI where a relative IRI is threated as absolute.

Introduce logic that check if string is  a valid RFC 3986 scheme followed by ':'. A colon appearing after '/', '?' or '#' does not make the reference absolute.

----

### DIFF
--- a/core/model/src/main/java/org/eclipse/rdf4j/model/util/URIUtil.java
+++ b/core/model/src/main/java/org/eclipse/rdf4j/model/util/URIUtil.java
@@ -371,4 +371,44 @@ public class URIUtil {
 				|| codePoint >= 0x0300 && codePoint <= 0x036F || codePoint >= 0x203F && codePoint <= 0x2040;
 	}
 
+	/**
+	 * True if the string starts with a valid RFC 3986 scheme followed by ':'. A colon appearing after '/', '?' or '#'
+	 * does not make the reference absolute.
+	 *
+	 * @param iri a string representing an RDF IRI reference to check.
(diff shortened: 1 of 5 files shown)


aschwarte10 (commented):
Thanks for looking into this and providing the fix 👍 

From my side the logic looks fine.

As a quick addition: could you add a unit test to the existing `TurtleParserTest` class? Either the reproducer from the issue description or alternatively something like

```
@Test
	public void testParseBaseIriParsing() throws Exception {
		
		String data = """
				@base <http://example.com/> .
				@prefix my: <https://mydomain.com/> .
				
				<system/x:ABC> a my:Item .
				
				<http://example.com/system/x:DEF> a my:Item .
				
				my:abc a my:Item .
				
				""";
		
		Reader r = new StringReader(data);

		try {
			parser.parse(r);
			var stmts = statementCollector.getStatements();
			
			assertEquals(3, stmts.size());
			
			var model = new TreeModel(stmts);
			assertEquals(Set.of(Values.iri("http://example.com/system/x:ABC"), Values.iri("http://example.com/system/x:DEF"), Values.iri("https://mydomain.com/abc")), model.subjects());
		} catch (RDFParseException e) {
			fail("parse error on correct data: " + e.getMessage());
		}
	}
```

hmottestad:
Here is a new test case.

[ResolveUriTest.java](https://github.com/user-attachments/files/31643703/ResolveUriTest.java)

PR #6007, opened 2026-08-31
GH-6006 Bound LMDB page cardinality estimation

### LINKED ISSUE #6006: LMDB page cardinality estimator performs an unbounded linear leaf walk
### Current behavior

`LmdbBtreeRangeCounter.countRange()` seeks to the first matching key and then walks every subsequent LMDB leaf page and key until the inclusive maximum key is reached. The estimator therefore has linear cost in the physical size of the selected range.

This becomes especially expensive for large RDF statement ranges because each `LmdbDataFile.readPage()` call allocates a new page-sized `ByteBuffer` and performs a positional `FileChannel.read()`. A broad cardinality estimate can consequently read and allocate for millions of pages before query planning can continue.

### Expected behavior

Page-based cardinality estimation should have bounded work for large ranges while preserving exact counts for small ranges.

A focused implementation can:

* walk and count exactly for at most 32 leaf pages;
* return the exact result when the upper bound is reached within that ceiling;
* otherwise count the first and last partial leaves exactly;
* deterministically sample 16 complete middle leaves, or 32 when a non-prefix residual filter must be applied;
* weight each sampled leaf by its virtual root-to-leaf selection probability;
* use a per-estimate page cache so shared root and branch pages are read only once;
* apply `GroupMatcher` only when bound fields remain after the contiguous index prefix;
* avoid returning a false zero for a non-empty range with a residual filter.

This changes large-range work from `O(all pages and entries in the range)` to approximately `O(32 exact leaf pages + samples * tree depth)` and keeps estimates deterministic for a fixed LMDB snapshot and range.

### Acceptance criteria

* Ranges spanning at most 32 leaf pages retain exact current behavior.
* Larger ranges have a hard upper bound on page reads independent of range cardinality.
* First and last partial leaves are counted exactly; only complete middle leaves are sampled.
* Bound non-prefix statement fields are evaluated on sampled leaves.
* Repeated estimates for the same snapshot and range return the same value.
* Existing page-cardinality regression tests remain green, with focused tests covering bounded page reads, boundary inclusivity, same-leaf ranges, residual filtering, and non-zero selective estimates.

### Scope

The fix should remain local to `LmdbBtreeRangeCounter`, `LmdbPageCardinalityEstimator`, and the `TripleStore` call site. It does not require a generic estimator abstraction, confidence intervals, adaptive allocation, learned calibration, persistent statistics, or a persistent page cache.

hmottestad:
## Summary

Fixes #6006.

This replaces the LMDB query planner's potentially unbounded leaf walk with a bounded, snapshot-consistent page
cardinality estimator. It also documents the estimator's architecture and invariants and adds an operational switch
that restores the complete RDF4J 5.3.2 estimation path when compatibility or diagnosis requires it.

```mermaid
flowchart LR
    P[Statement pattern] --> I[Binding-shape and index selection]
    I --> T[Pinned LMDB read transaction]
    T --> W[Bounded B-tree page walk]
    W --> Q[Estimate, hard bounds, and quality]
    Q --> O[Query optimizer]
    W -. decode or estimation failure .-> L[RDF4J 5.3.2 index selection and cursor sampler]
    S[Disable system property] --> L
```

The LMDB database format and query-result semantics are unchanged. This affects planner statistics and planning cost.

## Page-walking estimator

`TripleStore` maps the logical subject/predicate/object/context binding shape onto the configured physical indexes. It
chooses the longest contiguous bound prefix, then uses residual field layout as a tie-break for the page estimator.
Only that contiguous prefix restricts the B-tree range; bound fields after the first gap are sampled as residual
predicates.

`LmdbPageCardinalityEstimator` pins all work to the transaction's matching LMDB meta page and native mapping. Page,
header, and named-database caches live only for that snapshot, because LMDB can reuse page numbers after copy-on-write
updates. The file-channel path and native-map path use the same decoder.

`LmdbBtreeRangeCounter` estimates an inclusive key range as follows:

- whole-database metadata, empty ranges, same-leaf ranges, and small ranges are exact;
- partial boundary leaves are counted exactly;
- the interior is decomposed into complete sibling subtrees;
- large interior mass is sampled deterministically with inverse-probability weighting;
- residual predicates use bounded pilot and final evidence streams;
- a materially cheaper complement can be estimated and subtracted from the exact database total;
- hard lower/upper bounds, effective sample size, sampled mass, error, and disagreement accompany the point estimate;
- a second physical layout is consulted only for weak no-prefix evidence.

Explicit and inferred databases are disjoint, so their estimates and bounds are added. Estimates from two physical
layouts describe the same logical result, so they are quality-weighted and geometrically blended within the
intersection of their hard bounds instead of being added.

### LMDB format invariants

The raw decoder follows LMDB's page, node, and named-database layouts and comparator flags. `MDB_REVERSEKEY` takes
precedence over `MDB_INTEGERKEY`, matching LMDB. Inclusive RDF4J range bounds are converted to the half-open intervals
used internally by the page walk.

DUPSORT databases count logical records, not only outer keys:

- a direct value contributes one record;
- an embedded duplicate subpage contributes its logical key count (`P_SUBP`/`P_LEAF2`);
- a promoted duplicate sub-database contributes `MDB_db.md_entries`.

Contradictory or unsupported layouts fail soft rather than silently producing planner statistics from a misdecoded
page.

## RDF4J 5.3.2 compatibility fallback

Start the JVM with this property before opening the LMDB store:

```text
-Dorg.eclipse.rdf4j.sail.lmdb.disablePageWalkingEstimator=true
```

The property bypasses page walking and avoids constructing its page reader/cache. It restores both pieces of the
RDF4J 5.3.2 estimator:

1. the first configured index having the longest bound prefix is selected, preserving the old tie behavior;
2. the original three-bucket, at-most-100-rows-per-bucket cursor sampler is run unchanged.

Unexpected page-walker failures use the same complete fallback. The property is captured when each store opens;
reopen or restart the store after changing it. Missing or false leaves page walking enabled.

This is intentionally separate from `LmdbStoreConfig.setPageCardinalityEstimator(false)`. That existing repository
setting disables approximate estimation and counts matching statements exactly; the new system property selects the
5.3.2 approximate estimator.

The compatibility source was checked against tag `5.3.2`, commit
`e0bbd99d3b5d1ed1ee9e48e4394006768581e2c0`:

- the sampler body is source-identical after only normalizing the relocated `MAX_KEY_LENGTH` owner;
- the index-selector body is source-identical;
- `Statistics` and `bucketStart` are source-identical.

An independently run 5.3.2 fixture supplies exact golden outputs for unbound, single-field, two-field, prefix, and
no-prefix patterns. Before the property was implemented, six of nine golden assertions disagreed with the page walker;
the unchanged test now passes all nine.

## Maintainer documentation

The implementation now documents:

- logical binding masks versus physical key-position masks;
- prefix-range and residual-matcher responsibilities;
- transaction/meta-page/native-map snapshot boundaries;
- LMDB on-disk structure offsets and comparator precedence;
- branch-path decomposition, boundary slicing, complement estimation, and exactification;
- inverse-probability sampling, deterministic seeds, budgets, quality signals, and hard bounds;
- DUPSORT logical multiplicity;
- request-local and snapshot-local cache lifetimes, including the page cache's lock-free soft capacity;
- how explicit/inferred and independent-layout estimates differ;
- where to add tests when changing key encoding, traversal, sampling, or quality policy;
- why the 5.3.2 compatibility sampler and selector must not be retuned in place.

The user-facing LMDB Store documentation explains the switch, when it is read, automatic fail-soft fallback, and the
difference from exact counting.

## Verification

Local verification on the final working tree:

- pre-fix focused regression: 1 test failed with 6 of 9 5.3.2 estimate assertions mismatching;
- post-fix compatibility class: 3 tests, 0 failures, 0 errors, 0 skipped;
- full `core/sail/lmdb` verify: 1,181 discovered, 1,079 executed, 0 failures, 0 errors, 102 skipped;
- root `-Pquick clean install`: all 123 reactor projects passed;
- copyright check, formatter, and `git diff --check`: clean;
- executable sampler/index-selector equivalence checks against the 5.3.2 tag: clean.

The compatibility class covers the true switch, an explicit false value, and the pre-existing exact-count repository
setting. The module run includes the regular and DUPSORT range-counter tests, native DUPSORT integration, cardinality
integration, binding-shape/index-layout tests, and the new compatibility matrix.

The 102 skipped tests are existing workload-gated cases. `PageCardinalityEstimatorThemeAccuracyIT` remains
intentionally class-disabled because it loads all eight Theme datasets and emits dataset-scale CSV/Markdown accuracy
artifacts; it was not run in this verification.

Remote CI is green for current PR head `709796e2be8838269565abd6ea5c84bb511c967d`. The documentation, compatibility
switch, and its local verification will require a new CI run after they are committed and pushed.

## Performance evidence and scope

The committed Medical Records JMH comparison covers queries q0-q10. Nine of eleven mean scores improve; the largest
measured change is q2, from `51.169 +/- 50.042 ms/op` to `7.853 +/- 0.981 ms/op` (about 84.7% lower). The two small mean
regressions are within broad error intervals. These numbers are workload-scoped evidence, not a claim about every Theme
dataset.

## Compatibility notes

- No public RDF4J API or LMDB on-disk format changes.
- No new dependency.
- Existing repository configuration remains valid.
- The new JVM property is absent/false by default, so the bounded page walker remains the normal path.
- Query-result semantics are unchanged; only optimizer cardinality estimates and planning work are affected.


kenwenzel:
@hmottestad Can you also extend this for MDB_DUPSORT?

hmottestad:
I've avoided it here, because I want to merge this to main so DUPSORT isn't needed. I can ask claude to see if it can be extended for DUPSORT, but a little bit busy this week.

hmottestad:
## Accuracy summary

| Measurement | Initial | After 500% churn |
|---|---:|---:|
| Mean absolute percentage error | 0.38% | 0.38% |
| Weighted absolute percentage error | 1.31% | 1.31% |
| Median absolute percentage error | 0.00% | 0.00% |
| P95 absolute percentage error | 2.75% | 2.75% |
| Maximum absolute percentage error | 8.42% | 8.42% |
| Exact-zero / estimated-nonzero rows | 0 | 0 |

Estimator values changed for 0 of 107 pattern occurrences after churn; maximum absolute estimate drift was 0.00%.

kenwenzel (commented):
Thank you for also supporting dup sort. LGTM

@hmottestad Are you aware of changes in LMDB 1.0. I've seen that LWJGL also updated to LMDB 1.0 and we should prepare to support the new file format.

